Issue #29, opened 2015-07-29
Layers support for Leaflet Map Component


mbarto:
Implement layer support and configuration for the following layer types:
 * background layers:
  -  OSM / MapQuest
  - Google / Bing
  - WMS(T) - TMS
  - None

offtherailz:
Many issues using leaflet plugins dependencies:
* the leaflet-plugins module on npm doesn't supports modules. 
* tried to use a module friendly tarball from a pull request. Now google doesn't seems to work. 

Check this Work in progress for details : https://github.com/offtherailz/MapStore2/tree/googlebing_support

PR #37, opened 2015-07-30
#29 support for osm layer

offtherailz:
Add the possibility to add layers in a declarative way.
Supports osm layer legacy configuration .

### DIFF
--- a/web/client/components/leaflet/__tests__/map-test.jsx
+++ b/web/client/components/leaflet/__tests__/map-test.jsx
@@ -50,7 +50,7 @@ describe('LeafletMap', () => {
         expect(document.getElementsByClassName('leaflet-tile-pane').length).toBe(1);
         expect(document.getElementsByClassName('leaflet-objects-pane').length).toBe(1);
         expect(document.getElementsByClassName('leaflet-control-container').length).toBe(1);
-    });
+    })
 
     it('enables leaflet controls', () => {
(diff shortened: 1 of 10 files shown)


web/client/components/leaflet/__tests__/map-test.jsx
```diff
@@ -67,4 +68,10 @@ describe('LeafletMap', () => {
         zoomOut.click();
         expect(leafletMap.getZoom()).toBe(11);
     });
+    it('check layers init', () => {
+        const map = React.render(<LeafletMap center={{lat: 43.9, lng: 10.3}} zoom={11}>
+            <LeafLetLayer source={{ptype: 'gxp_osmsource'}}/>
+        </LeafletMap>, document.body);
+        expect(map).toExist();
```
mbarto: Can we test for layer dom existance?

web/client/components/leaflet/__tests__/map-test.jsx
```diff
@@ -49,7 +50,7 @@ describe('LeafletMap', () => {
         expect(document.getElementsByClassName('leaflet-tile-pane').length).toBe(1);
         expect(document.getElementsByClassName('leaflet-objects-pane').length).toBe(1);
         expect(document.getElementsByClassName('leaflet-control-container').length).toBe(1);
-    });
+    })
```
mbarto: Please, add back the semicolon

PR #55, opened 2015-08-14
#29 Google and Bing Support

offtherailz:
Add support for Google and Bing background.
* Uses a fork of leaflet-plugins module that supports modules (not merged yet, pending pull request).
* Adds also google api to the test depedencies.

### DIFF
--- a/web/client/components/leaflet/Layer.jsx
+++ b/web/client/components/leaflet/Layer.jsx
@@ -9,13 +9,15 @@ var React = require('react');
 var L = require('leaflet');
 var LeafletUtils = require('./WMSUtils');
 
-var LeafletLayer = React.createClass({
+var LPlugins = require('leaflet-plugins');
+
+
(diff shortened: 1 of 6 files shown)


web/client/components/leaflet/Layer.jsx
```diff
@@ -42,12 +43,37 @@ var LeafletLayer = React.createClass({
...
+                    break;
+                case "gxp_googlesource":
+                    this.layer = this.createGoogleLayer(options);
+                    break;
                 default:
             }
             if (this.layer) {
-                this.layer.addTo(this.props.map);
+                // some plugins doesn't have addTo method
+                if (source.type === "gxp_wmssource" || source.type === "gxp_osmsource" ) {
```
mircobe87: Is the property `type` right? It should be `ptype`?

mbarto:
Replaced by https://github.com/geosolutions-it/MapStore2/pull/80

Issue #118, opened 2015-09-13
Add "Fork Me on GitHub" Red ribbon on homepage


simboss:
See old mapstore page [here](http://mapstore.geo-solutions.it/mapstore/)

PR #120, opened 2015-09-14
#118 Add github ribbon

offtherailz:
Add "fork me on Github" ribbon in the home page.

### DIFF
--- a/web/client/index.html
+++ b/web/client/index.html
@@ -25,9 +25,13 @@
             margin: 0;
             padding: 0;
         }
-        
+
         .panel{
             background-color: rgba(255,255,255,0.3);
+        }
+
+        #langSelContainer {
+            margin: 8px;
         }
 
         @media (min-width: 992px) {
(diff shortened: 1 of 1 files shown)


web/client/index.html
```diff
@@ -74,8 +77,10 @@
         </style>
     </head>
     <body>
+        <a href="https://github.com/geosolutions-it/MapStore2"><img style="position: absolute; top: 0; right: 0; border: 0; z-index:100; " src="https://camo.githubusercontent.com/e7bbb0521b397edbd5fe43e7f760759336b5e05f/68747470733a2f2f73332e616d617a6f6e6177732e636f6d2f6769746875622f726962626f6e732f666f726b6d655f72696768745f677265656e5f3030373230302e706e67" alt="Fork me on GitHub" data-canonical-src="https://s3.amazonaws.com/github/ribbons/forkme_right_green_007200.png"></a>
```
mbarto: Can't we just define a new Component for this?
Not mandatory, just an idea.

offtherailz: It is possible, but it require a little more work with no gain (only a usage sample IMHO). I suggest to use this now and open a minor priority issue to fix. Let me know if it is ok for you.

mbarto: Yes, it's ok, move on.

PR #176, opened 2015-09-30
Mouse position#88

kappu72:
Close#88
CrsSelector and panle printing coordinate are configurable via css 
The component get available css from proj4js.
How it printrdinates is configurable adding more react MousePositionLable and passing these as props.
See MousePositionLableDMS and MousePositionLableYX


simboss:
Generally speaking it would be better to have commits squashed before sending a PR.

Issue #108, opened 2015-09-10
Full Screen Map Loading Spinner


simboss:
Introduce a global spinner to show when the map is loading. The spinner should:
* not be intrusive (small button / div to show over the map)
* work with both leaflet and openlayers Map components
* have a configurable delay before showing the spinner (so that is not shown at all if the map loading is very fast)

For a nice example [see this](http://theconversation.com/big-data-reveals-the-business-risk-posed-by-unreported-ukraine-protests-24232) or [this one](www.kayak.it).

PR #177, opened 2015-09-30
Full Screen Map Loading Spinner #108.

nmco:


web/client/reducers/config.js
```diff
@@ -47,6 +47,34 @@ function mapConfig(state = null, action) {
             return assign({}, state, {
                 mousePointer: action.pointer
             });
+        case LAYER_LOADING: {
+            let loadingLayers = assign((state && state.loadingLayers) || {});
+            loadingLayers[action.layerId] = true;
```
mbarto: I think you are mutating the original state.loadingLayers object here, you should avoid that: assign({}, (state && state.loadingLayers) || {}) should do the trick

web/client/reducers/config.js
```diff
@@ -47,6 +47,34 @@ function mapConfig(state = null, action) {
...
+            return assign({}, state, {
+                loadingLayers: loadingLayers
+            });
+        }
+        case LAYER_LOAD: {
+            let loadingLayers = assign((state && state.loadingLayers) || {});
+            loadingLayers[action.layerId] = false;
+            return assign({}, state, {
```
mbarto: Same as above

web/client/reducers/config.js
```diff
@@ -47,6 +47,34 @@ function mapConfig(state = null, action) {
...
+                loadingLayers: loadingLayers
+            });
+        }
+        case SHOW_SPINNER: {
+            let spinnersInfo = assign((state && state.spinnersInfo) || {});
+            spinnersInfo[action.spinnerId] = true;
```
mbarto: Same as above

web/client/reducers/config.js
```diff
@@ -47,6 +47,34 @@ function mapConfig(state = null, action) {
...
+            return assign({}, state, {
+                spinnersInfo: spinnersInfo
+            });
+        }
+        case HIDE_SPINNER: {
+            let spinnersInfo = assign((state && state.spinnersInfo) || {});
+            spinnersInfo[action.spinnerId] = false;
```
mbarto: Same as above

PR #179, opened 2015-09-30
fix #164 Leaflet mobile freeze resolution

### LINKED ISSUE #164: Leaflet map freeze after pinch to zoom
Tested with chrome for android on a galaxy note 2. After a pinch-in/out the map freeze. 
This can ne related to this issue: https://github.com/Leaflet/Leaflet/issues/1759 or similar.

offtherailz:
For leaflet maps compare the coordinates with the maps one and enable zoomAnimation should fix the problem.
We check the coordinates with 12 decimals max.


offtherailz:
Closing because it still not solve some problems

Issue #102, opened 2015-09-09
Implement zoom all control


mbarto:
Button to zoom to the user defined maxExtent (or to zoom 1, if not defined).
Should work with both leaflet and openlayers map components

chrismayer:
I suggest a button component handling the zoom logic, which is customizable via CSS. The default position could be below the zoom-in and zoom-out buttons, like in this little mockup (ignore styling of the button):

![mockup-102-1](https://cloud.githubusercontent.com/assets/1185547/10192570/979b5560-677d-11e5-9a74-a95e7985e03e.png)

What do you think?

mbarto:
Looks good to me.

Mauro


2015-09-30 14:15 GMT+02:00 Christian Mayer <notifications@github.com>:

simboss:
I believe it should be better placed within the toolbar for consistency, no?

Feedback welcome.

chrismayer:
I though it could be placed there to have all the "zoom-tools" on the same side of the map. But I agree for consistency it might be better having the new button in the toolbar. Seems we have then the advantage to have all tools in a common place. So toolbar?

mbarto:
I think we can put "pure navigation tools" like zoom+, zoom-, zoom all on
the left (the first two are already there). I would leave the right toolbar
for other functionalities.

Mauro


2015-09-30 14:39 GMT+02:00 Simone Giannecchini <notifications@github.com>:

PR #190, opened 2015-10-01
Add ZoomToMaxExtentButton component

chrismayer:
As described and discussed in #102 this PR adds a ZoomToMaxExtentButton component.

### DIFF
--- a/web/client/components/buttons/ZoomToMaxExtentButton.jsx
+++ b/web/client/components/buttons/ZoomToMaxExtentButton.jsx
@@ -11,6 +11,9 @@ var BootstrapReact = require('react-bootstrap');
 var Button = BootstrapReact.Button;
 var Glyphicon = BootstrapReact.Glyphicon;
 
+const mapUtils = require('../../utils/MapUtils');
+const configUtils = require('../../utils/ConfigUtils');
+
 
 /**
  * A button to zoom to max. extent of the map or zoom level one.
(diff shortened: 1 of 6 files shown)


mbarto:
Please, check Travis build: it's failing with this error:

web/client/examples/viewer/plugins/index.jsx
  90:12  error  Line X: Adjacent JSX elements must be wrapped in an enclosing tag

mbarto:
From an initial look at the code, it cannot work. Did you test that it is working in either leaflet or openlayers viewer?

chrismayer:
> Did you test that it is working in either leaflet or openlayers viewer?

Sure I did. All looked fine to me. What is wrong with the code?

mbarto:
I will checkout the code and do a direct test. I will let you know.

web/client/components/buttons/ZoomToMaxExtentButton.jsx
```diff
@@ -0,0 +1,100 @@
...
+            this.props.dispatch(changeMapView(this.props.mapConfig.center, 1,
+                this.props.mapConfig.maxExtent, this.props.mapConfig.size));
+        }
+
+    }
+});
+
+// In addition to the state, 'connect' puts 'dispatch' in our props.
+// connect Redux store slice with map configuration
+module.exports = connect((state) => {
```
mbarto: It would be preferrable to bind properties in the plugins index.js as we do for other plugins, instead of connecting directly to the store

web/client/components/buttons/ZoomToMaxExtentButton.jsx
```diff
@@ -0,0 +1,100 @@
...
+                    minx: maxExtent[0],
+                    miny: maxExtent[1],
+                    maxx: maxExtent[2],
+                    maxy: maxExtent[3]
+                },
+                crs: mapConfig.projection,
+                rotation: 0
+            };
+            // adapt the map view by dispatching to the corresponding action
+            this.props.dispatch(changeMapView(this.props.mapConfig.center, -1,
```
mbarto: You should get the action to call as a property without directly calling dispatch

mbarto:
I think it works only when maxExtent is not defined.
When there is a maxExtent you invoke the action with maxExtent and a -1 zoom, but if you look at both OpenLayers and Leaflet Map.jsx components, you will see that:
 * bbox is not used to change current map zoom / center
 * only center and zoom are used for that purpose and -1 is not a valid zoom value

chrismayer:
Okay I'll bind the actions the other way you described.

What would be your favourite strategy to zoom the map by center, since  

 > - bbox is not used to change current map zoom / center
 > - only center and zoom are used for that purpose and -1 is not a valid zoom value

Would extending ``_updateMapPositionFromNewProps`` in the Map.jsx files (OpenLayers and Leaflet) be an option so they check if only the extent object ``bbox`` has been changed and zoom to it?

mbarto:
Hi, Christian.
The most important thing is that the application state MUST be consistent.  So: 
 * in the reducer you need to set all the properties (zoom, center, bbox) to represent the same map; since you have a bbox, you need to calculate the related center and zoom, and set all the 3 properties on the new state
 * then you don't need to change _updateMapPositionFromNewProps

Consider that my desire is to remove the bbox from the state and make it a calculated property of center and zoom, so those 2 are the important properties for map navigation.

chrismayer:
Hi @mbarto,
I tried to address your desire. So now the max. extent will be set by center and zoom, which are calculated from ``maxExtent``.

mbarto:
Looks good. I will try to test it on monday, and then merge it. Thank you.

mbarto:
Hi Christian, I did a couple of minor adjustements and merged your work. Thank you.

PR #216, opened 2015-10-09
Fixes #209, bing apiKey

### LINKED ISSUE #209: Bing layers not running causes impossible background switch
Bing layers don't work (maybe is an API key problem). A leaflet-plugin's error stops of layer remove and do not allow to remove layer on background switcher. 

Bing.js line 128

mircobe87:
when a map configuration is loaded, all Bing layers get the proper apiKey
from localConfig.json file.


web/client/utils/ConfigUtils.js
```diff
@@ -255,6 +257,12 @@ var ConfigUtils = {
 
         retina: retina
         };
+    },
+    setBingKey: function(layer) {
```
offtherailz: We should use more generic function that assign defaults values to the layers, or something like that

PR #221, opened 2015-10-13
Close #202. Add Settings panel

### LINKED ISSUE #202: Create a settings panel with all the tools not required during navigation
Mouse position enabling button, crs selector, language selector and other controls are not needed any time. We can move all these controls in a panel visible with a settings button, like this:
image

RFC

offtherailz:
New settings panel component in the map viewer app. 
* The settings panel contains language selector and mouse position tool.
* The language selector now uses flags.
*  Removed also some div not needed in the components inside the settings panel.
* Small fixes for the about button to stay above attribution. 

Here a preview
image


web/client/components/mapcontrols/mouseposition/__test__/CRSSelector-test.js
```diff
@@ -22,9 +22,8 @@ describe('CRSSelector', () => {
         const cmp = React.render(<CRSSelector enabled={true}/>, document.body);
         expect(cmp).toExist();
 
-        const cmpDom = React.findDOMNode(cmp);
+        const cmpDom = cmp.getDOMNode();
```
mbarto: This is deprecated. Please use React.findDOMNode

PR #271, opened 2015-11-03
Add support for ImageButton in ZoomToMaxExtentButton component

nmco:
Remove div around the image in ImageButton component

### DIFF
--- a/web/client/components/buttons/ImageButton.jsx
+++ b/web/client/components/buttons/ImageButton.jsx
@@ -12,21 +12,22 @@ var ImageButton = React.createClass({
     propTypes: {
         id: React.PropTypes.string,
         image: React.PropTypes.string,
-        onClick: React.PropTypes.func
+        onClick: React.PropTypes.func,
+        style: React.PropTypes.object
     },
     getStyle() {
-        var style = {
+        var finalStyle = {
             cursor: "pointer",
             margin: 0,
             padding: 0,
             display: "inline-block"
         };
-        if (this.props.image && this.props.image !== "") {
-            assign(style, {
+        if (this.props.image) {
+            assign(finalStyle, {
                 overflow: "hidden"
             });
         } else {
-            assign(style, {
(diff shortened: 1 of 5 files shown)


nmco:
I improve a little bit (I hope) the ImageButton component: remove the div around the image tag and give the possibility to pass a custom style. I upgraded and updated all the need tests and code. I also give the possibility to the ZoomExtendButton component to support an ImageButton (the same beahvior as the InfoButton).

web/client/components/buttons/ImageButton.jsx
```diff
@@ -12,21 +12,22 @@ var ImageButton = React.createClass({
...
     },
     getStyle() {
-        var style = {
+        var finalStyle = {
             cursor: "pointer",
             margin: 0,
             padding: 0,
             display: "inline-block"
         };
         if (this.props.image && this.props.image !== "") {
```
mbarto: this is equivalent to: if(this.props.image) {

nmco:
By the way, a pattern is emerging for the buttons: being a normal (bootstrap) button or an image button. If this is a correct pattern\behavior we should probably find a way to put this in a reusable component, the last time I looked around mixins are the cools for this stuff.

web/client/components/buttons/ImageButton.jsx
nmco: I will change that :)

mbarto:
Unfortunately, mixins are deprecated, since they cannot be used with ES6 classes.
In general, my preferred approach is to split UI from behaviour, so that we have agnostic UI controls (buttons, image buttons, etc.), triggering desired behaviour by configuration (onClick). I don't like too much the current approach of having the UI and behaviour in the same place, or at least I don't like having this at the framework level. It's ok for now, but I will refactor it some day :)

nmco:
Good point, I was looking at the problem in a wrong way. Behavior and UI in the same place is not a good idea. Thank you.

PR #283, opened 2015-11-05
Implement a basic dynamic help mechanism

### LINKED ISSUE #155: Implement dynamic help system
Help should be enabled by a toggle button.
When enabled, all the "help-augmented" components should be highlighted on the UI, so that the user can ask help on a specific UI part. When an highlighted part is selected, the related help should be shown to the user.
We need to define how to help-augment any component and how to recognize them.

chrismayer:
This adds a dynamic help system to highlight 'help-augmented' elements
by a badge and show the corresponding help text by clicking the badge. The highlighting can be activated / deactivated by a button in the toolbar.

To show the mechanism the ZoomToMaxExtentButton is extended by such a "help-badge" 

Fixes #155

### DIFF
--- a/web/client/components/Help/HelpWrapper.jsx
+++ b/web/client/components/Help/HelpWrapper.jsx
@@ -1,0 +1,43 @@
+/**
+ * Copyright 2015, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+
+var React = require('react');
+var HelpBadge = require('./HelpBadge');
+require("./help.css");
+
+/**
+ * A wrapper to add a help badge to an element.
+ *
+ * Component's properies:
+ *  - helpText: {string}      the text to be displayed when this badge is clicked
+ *  - helpEnabled: {bool}     flag to steer visibility of the badge
+ *  - changeHelpText (func)   action to change the current help text
+ */
+let HelpWrapper = React.createClass({
+    propTypes: {
+        helpText: React.PropTypes.string,
+        helpEnabled: React.PropTypes.bool,
+        changeHelpText: React.PropTypes.func
+    },
+
+    render: function() {
+        return (
+            <div style={{position: "absolute"}}>
+                <HelpBadge
+                    id={"helpbadge-" + this.props.children.key}
+                    isVisible = {this.props.helpEnabled}
+                    helpText = {this.props.helpText}
+                    changeHelpText = {this.props.changeHelpText}
+                />
+                {this.props.children}
+            </div>);
+    }
+});
+
+module.exports = HelpWrapper;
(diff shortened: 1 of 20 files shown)


mbarto:
Looks good. I was wondering if we can extend existing tools with help badge functionality using an Higher Order Component that wraps existing ones, instead of adding help functionality to every component one by one.
Something like:
<Help text="..."><ZoomToMaxExtent/></Help>

chrismayer:
Thanks for you review @mbarto. I agree that a high-level wrapper would be good. I also discovered that it is not that good to extend every component and already tried some things to solve that, but no good approach so far.  I will have a deeper look on that.

chrismayer:
I did some more investigation and added an approach for a high level component, which can be  wrapped around an existing element in order to show a help badge.
The positioning is a bit tricky because the ``HelpBadge`` is no more a direct child of the component itself. I solved it with a dynamic ID and appropriate CSS.
Any thoughts welcome.

web/client/components/Help/HelpWrapper.jsx
```diff
@@ -0,0 +1,43 @@
...
+        helpText: React.PropTypes.string,
+        helpEnabled: React.PropTypes.bool,
+        changeHelpText: React.PropTypes.func
+    },
+
+    render: function() {
+        return (
+            <div style={{position: "absolute"}}>
+                <HelpBadge
+                    id={"helpbadge-" + this.props.children.key}
```
mbarto: We should add some check that one single child exist and has a key,

mbarto:
Looks good to me, please proceed with other components.

offtherailz:
Hi, It would be nice if the help text appears on mouse hover (on the ?) .
The interaction should be: the Help div appears on first hover event. The text changes on the next hover events, the help div disappear when the user disable the help. 
If would be cool if  the help div was draggable, maybe this can be a future enhancement, what do you think @mbarto ?

chrismayer:
I added some more help badges to the app's components. To add them to the MapToolbar's buttons seems to be a bit more complicated since they are created dynamically and the ``<HelpWrapper>``-tag doesn't seem to be the right component, which is expected for in the MapToolbar class. Any hints welcome how to solve this in an elegant way.

The behaviour described by @offtherailz shouldn't be that hard to implement once a decision has been made.

offtherailz:
I think we can do it passing some property helpMessage to each children and the other properties to the map toolbar directly. When the toolbar creates the buttons, it can wrap the component ( if the helpMessage property is present). This should solve the problem.

mbarto:
Yes, I think it's a good idea. I also think  hover is the way to go.

chrismayer:
I implemented the hover mechanism (f72225c) as described by @offtherailz:
>The interaction should be: the Help div appears on first hover event. The text changes on the next hover events, the help div disappear when the user disable the help.

mbarto:
Hi Christian. Do you think you have still time to add the help badge also to the MapToolbar tools?
Can you also rebase your work on master, so that we don't have conflicts?
Thanks

chrismayer:
Hi Mauro,
I already started, will provide it at the asap (no later than end of the day).

chrismayer:
>Do you think you have still time to add the help badge also to the MapToolbar tools?
Can you also rebase your work on master, so that we don't have conflicts?

Done...

PR #396, opened 2016-01-20
fixes #370

### LINKED ISSUE #370: duplicated layers not allowed
Actually the identifier of a layer in each action and tool is the layer name. This causes problems if we have duplicated layers (e.g. with different styles). 

We should allow to pass an `id` property from the configuration, by default initialized with the layer name (@source if possible).
Then all the actions, reducers and internal calls to should use the id to identify the layer

kappu72:


web/client/examples/layertree/app.jsx
```diff
@@ -55,16 +56,18 @@ const getGroup = (group, allLayers) => {
         }, []);
 
     return assign({}, {
+        id: groupName,
```
offtherailz: Clarify why also groups need an id

web/client/utils/ConfigUtils.js
```diff
@@ -72,7 +72,7 @@ var ConfigUtils = {
         other.center = ConfigUtils.getCenter(other.center);
         return {
             map: other,
-            layers: layers.map(ConfigUtils.setBingKey),
+            layers: layers.map(ConfigUtils.setBingKey).map(ConfigUtils.setLayerId),
```
offtherailz: So the id is mandatory and this function is an helper to assign the name if the id is not present.

kappu72: Yes It'a mandatory and setLayerId is an helper function that set id equal to layer's name if id is missing

web/client/examples/layertree/app.jsx
kappu72: In layers reducer deepChange function is used both for layers and group, now it checks for a generic node id not name

web/client/utils/ConfigUtils.js
```diff
@@ -268,6 +268,12 @@ var ConfigUtils = {
             layer.apiKey = defaultConfig.bingApiKey;
         }
         return layer;
+    },
+    setLayerId: function(layer) {
+        if (!layer.id) {
+            layer.id = layer.name;
```
offtherailz: We should use a progressive number to set unique id anyway, for duplicates too.

PR #419, opened 2016-02-03
Query tool basic implementation

tdipisa:
This pull request provides a basic implementation of the following components:

- QueryBuilder (root component)
- FilterField
- ComboField
- DateField


web/client/components/QueryForm/ComboField.jsx
```diff
@@ -0,0 +1,51 @@
...
+            fieldException: null,
+            onUpdateField: () => {},
+            onUpdateExceptionField: () => {}
+        };
+    },
+    renderfieldOptions(fieldOpt) {
+        return (<option key={fieldOpt} value={fieldOpt}>{fieldOpt}</option>);
+    },
+    render() {
+        const style = {width: "250px"};
```
mbarto: make this configurable, as a style property of the component with width: 250px as a default value

web/client/reducers/queryform.js
```diff
@@ -0,0 +1,144 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const {ADD_FILTER_FIELD, REMOVE_FILTER_FIELD, UPDATE_FILTER_FIELD, UPDATE_EXCEPTION_FIELD} = require('../actions/queryform');
+const assign = require('object-assign');
+
+const initialState = {
```
mbarto: This is the initial state for the example, non of a generic QueryBuilder, the initialState of a generic QueryBuilder should be empty

web/client/reducers/queryform.js
```diff
@@ -0,0 +1,144 @@
...
+                attribute: "",
+                operator: null,
+                value: null,
+                exception: null
+            };
+
+            return assign({}, state, {filterFields: [...state.filterFields, newElement]});
+        }
+        case REMOVE_FILTER_FIELD: {
+            return assign({}, state, {filterFields: state.filterFields.filter((field) => {
```
mbarto: Can be simplified: filter((field) => field.rowId !== action.rowId)

web/client/reducers/queryform.js
```diff
@@ -0,0 +1,144 @@
...
+                if (field.rowId !== action.rowId) {
+                    return field;
+                }
+            })});
+        }
+        case UPDATE_FILTER_FIELD: {
+            return assign({}, state, {filterFields: state.filterFields.map((field) => {
+                if (field.rowId === action.rowId) {
```
mbarto: if we name action.fieldName as the real name of the attribute to change, we can get rid of the switch and make it generic: field[action.fieldName] = action.fieldValue
Moreover you cannot assign new values to field, you have to create a new object, copying old values as usual assign({}, field, {[action.fieldName]: action.fieldValue})

web/client/reducers/queryform.js
```diff
@@ -0,0 +1,144 @@
...
+                        }
+                        default: break;
+                    }
+                }
+                return field;
+            })});
+        }
+        case UPDATE_EXCEPTION_FIELD: {
+            return assign({}, state, {filterFields: state.filterFields.map((field) => {
+                if (field.rowId === action.rowId) {
```
mbarto: Same as above, you cannot mutate field

PR #429, opened 2016-02-08
Fix #390 Issue Snapshot Plugin

### LINKED ISSUE #390: Snapshot tool
Create a tool to snapshot the map. 
As a user, when I click on the snapshot tool a window should show a preview of the snapshot with some information (date, size...) and allow me to download the image. 

The tool should work of course with OpenLayers and Leaflet maps. 

Here a mockup: 
snapshot tool

Here some useful link:
* Original Plugin in MapStore (1) : https://github.com/geosolutions-it/MapStore/blob/master/mapcomposer/app/static/externals/gxp/src/script/plugins/PrintSnapshot.js
* Server side part of the original Plugin: https://github.com/geosolutions-it/servicebox/blob/master/src/main/java/it/geosolutions/servicebox/UploadCanvas.java

kappu72:
Base implementation of snapshot components for ol and leaflet.
With ie or when image size is bigger then 2Mb, servicebox service is needed to allow the image saving.
In Leaflet html2canvas lib needed.
In Openlayers, it clone the map status, so any layer introduce form other component but not in map state will not be rendere.


web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -10,7 +10,9 @@ var Layers = require('../../../../utils/openlayers/Layers');
 var ol = require('openlayers');
 var objectAssign = require('object-assign');
 var CoordinatesUtils = require('../../../../utils/CoordinatesUtils');
-const {isArray} = require('lodash');
+var ConfigUtils = require('../../../../utils/ConfigUtils');
```
offtherailz: use const

web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -29,16 +31,63 @@ function getWMSURLs( urls ) {
     return urls.map((url) => url.split("\?")[0]);
 }
 
+function needProxy(uri) {
```
offtherailz: we should extenalize in a utils function. axios already have a function like this, isn't it?

web/client/examples/viewer/index.html
```diff
@@ -25,6 +25,16 @@
             padding: 0;
             overflow: hidden;
         }
+        #snapshot_map{
```
offtherailz: I think this should stay in the css of the component that initialize it, included with require. 
we should also use something like "snapshot_hidden_map"

web/client/components/Snapshot/SnapshotPanel.jsx
```diff
@@ -0,0 +1,104 @@
...
+            <Grid header={this.props.name} className="snapshot-panel" fluid={true}>
+                <Row>
+                    <Col xs={7} sm={7} md={7}><Image src={this.props.snapshot.img.data} responsive thumbnail/></Col>
+                    <Col xs={5} sm={5} md={5}>
+                       <Table responsive>
+                            <tbody>
+                               <tr>
+                                <td>Date</td><td> <DateFormat dateParams={this.props.dateFormat}/></td>
+                                </tr>
+                                <tr><td>Layers</td><td>{this.renderLayers()}</td></tr>
```
offtherailz: comma and space between layers, to allow automatic new lines

offtherailz:
The container window should be bigger

web/client/components/Snapshot/leaflet/GrabMap.jsx
```diff
@@ -0,0 +1,124 @@
...
+            this.props.onSnapshotReady(imgData, canvas.width, canvas.height, size);
+            this.props.onStatusChange("READY");
+        }
+    },
+    mapIsLoading(layers) {
+        return layers.some((layer) => layer.loading);
+    }
+});
+
+module.exports = GrabLMap;
```
offtherailz: Leaflet preview has many issues. 
* Sometimes the map is not complete, layer missing or not well stacked 
image

* When swich layer or background, the preview do not update. 
image
* it is slow

web/client/components/Snapshot/SnapshotPanel.jsx
```diff
@@ -0,0 +1,104 @@
...
+            return (<Row style={{marginTop: "5px"}}>
+                    {this.props.snapshot.error}
+                    </Row>);
+        }
+    },
+    render() {
+        return ( (this.props.snapshot.img && this.props.snapshot.img.data) || this.props.snapshot.state === "READY") ? (
+            <Grid header={this.props.name} className="snapshot-panel" fluid={true}>
+                <Row>
+                    <Col xs={7} sm={7} md={7}><Image src={this.props.snapshot.img.data} responsive thumbnail/></Col>
```
offtherailz: The thumbnail should show loading when the snapshot is updating.

web/client/components/Snapshot/SnapshotPanel.jsx
```diff
@@ -0,0 +1,104 @@
...
+                        </Table>
+                    </Col>
+                </Row>
+
+                <Row className="pull-right" style={{marginTop: "5px"}}>
+                    {(this.props.browser.ie || (this.props.snapshot.img.size > 2000 && this.props.browser.chrome)) ? this.remoteDownload() : this.localDownload()}
+                </Row>
+                {this.renderError()}
+            </Grid>
+        ) : null;
```
offtherailz: On first render, the window should show loading messages, not a blank window
image

web/client/components/map/openlayers/plugins/WMSLayer.js
kappu72: Axios has almost identical function but implemented in request interceptor.
Where you want to implement this function?

web/client/examples/viewer/index.html
kappu72: I agree

offtherailz:
Summarizing:
* The Openlayers Snapshot is good
* The leaflet snapshot is buggy
* The snapshot preview need some improvement.

web/client/components/Snapshot/SnapshotPanel.jsx
kappu72: Changed to use ul li elements

offtherailz:
Some error on preview creation. 
image

offtherailz:
Cross origin is not used anymore for leaflet. 
image

offtherailz:
During implementation the preview is very slow when using proxy.
Can we display a tainted canvas for the preview while generating the usable snapshot?

offtherailz:
There is no warning about the background not supported (Google, Bing). 
When the snapshot is open with one of theese enabled, you should force one allowed background (e.g. openstreetmap) and write an info row on the snapshot preview window about it.

offtherailz:
The snapshot window should be a little bigger, like in the previous comment

offtherailz:
The loading is cool, but the window still remains blank for too much time, and it is confusing.

PR #460, opened 2016-02-18
Fix #195 fullscreen support for mobile

### LINKED ISSUE #195: Provide fullscreen support for mobile devices
On mobile devices the navigation bar should be hidden when possible. 
See this : http://www.html5rocks.com/en/mobile/fullscreen/?redirect_from_locale=pt
See also how to add the application to the home screen of the device (using it as an app).

Calbalacrab:
Add new label into settings menu "Fullscreen";
Add new toggle button into section:
 -- ON, enter fullscreen; 
-- OFF, exit fullscreen;
Add metatag and icon for HomeScreen support
Lock screen rotation


web/client/actions/__tests__/fullScreen-test.js
```diff
@@ -0,0 +1,29 @@
...
+var expect = require('expect');
+var {TOGGLE_FULLSCREEN, toggleFullScreen} = require('../fullScreen');
+
+describe('Test correctness of the fullScreen actions', () => {
+
+    it('test fullscreen true', () => {
+        const testVal = true;
+        const retval = toggleFullScreen(testVal);
+        expect(retval.type).toBe(TOGGLE_FULLSCREEN);
+        expect(retval.state).toExist();
```
mbarto: the toggleFullScreen returned object does not contain a state property, it only contains type and fullscreen

web/client/examples/viewer/components/FullScreen.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+var full = React.createClass({
+    propTypes: {
+        fullscreen: React.PropTypes.bool,
+        glyphicon: React.PropTypes.string,
+        buttonText: React.PropTypes.string,
+        propertiesToggle: React.PropTypes.func
+    },
+    componentWillReceiveProps: function(nextProps) {
+        if (nextProps.fullscreen !== this.props.fullscreen) {
```
mbarto: This is not correct: if the property changes, we need to check the new value to decide if entering or exiting fullscreen

web/client/reducers/__tests__/fullScreen-test.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+var expect = require('expect');
+var fullScreen = require('../fullScreen');
+
+
+describe('Test the locale reducer', () => {
+    it('Change MousePosition State', () => {
```
mbarto: Uhm... mouse position state ... aren't we testing TOGGLE_FULLSCREEN ?

web/client/examples/viewer/plugins/mobile.jsx
```diff
@@ -105,6 +109,8 @@ module.exports = {
                             <LangBar key="langSelector"
                             currentLocale={props.locale}
                             onLanguageChange={props.loadLocale}/>
+                            <h5>Fullscreen</h5>
+                            <FullScreen fullscreen={false} propertiesToggle={props.toggleFullScreen} />
```
mbarto: If fullscreen is always false, the state of the button will never change. It should be tied to the current state

PR #467, opened 2016-02-25
Improvements for the query form tool

tdipisa:
With this pull request:

- When selecting a different method from the spatial selection dd, remove the previously created feature in the map.
- Verified the possibility to draw Circle and BBOX with mouse dragging
- Added the possibility to edit the center of the circle.
- Added a configurable way to set coordinates visualization in Lon/Lat if needed in BBOX and Circle details form.
- Updated the queryform example and test cases accordingly

### DIFF
--- a/web/client/components/QueryForm/GeometryDetails.jsx
+++ b/web/client/components/QueryForm/GeometryDetails.jsx
@@ -18,6 +18,8 @@ const assign = require('object-assign');
 
 const GeometryDetails = React.createClass({
     propTypes: {
+        map: React.PropTypes.object,
+        useMapProjection: React.PropTypes.bool,
         geometry: React.PropTypes.string,
         type: React.PropTypes.string,
         onShowPanel: React.PropTypes.func,
@@ -26,6 +28,8 @@ const GeometryDetails = React.createClass({
     },
     getDefaultProps() {
         return {
+            map: {},
+            useMapProjection: true,
             geometry: null,
             type: null,
             onShowPanel: () => {},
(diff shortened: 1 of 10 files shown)


web/client/components/QueryForm/GeometryDetails.jsx
```diff
@@ -43,7 +47,9 @@ const GeometryDetails = React.createClass({
             }
         }
 
+        let mapProj = this.props.map.projection;
         let bbox = ol.geom.Polygon.fromExtent(coordinates);
```
mbarto: We should not use openlayers objects here

web/client/components/QueryForm/GeometryDetails.jsx
```diff
@@ -51,10 +57,15 @@ const GeometryDetails = React.createClass({
...
         this.tempCircle[name] = parseFloat(value);
 
-        let circle = new ol.geom.Circle(this.tempCircle.center, this.tempCircle[name]);
+        let mapProj = this.props.map.projection;
+        let center = new ol.geom.Point([this.tempCircle.x, this.tempCircle.y]);
+        center = !this.props.useMapProjection ? center.clone().transform('EPSG:4326', mapProj) : center;
+
+        let coordinates = center.getCoordinates();
+        let circle = new ol.geom.Circle(coordinates, this.tempCircle.radius);
         circle = ol.geom.Polygon.fromCircle(circle, 100);
```
mbarto: Same as above. In alternative, GeometryDetails can become a mapping library dependent component

mbarto:
In general, OpenLayers object cannot be used outside of web/client/components/map/openlayers/*.jsx and Leaflet objects cannot be used outside of web/client/components/map/leaflet/*.jsx.
Fix this before merging

PR #478, opened 2016-03-01
Feature grid with example

kappu72:


web/client/examples/featuregrid/components/FeatureGridActionsSupport.jsx
```diff
@@ -0,0 +1,79 @@
...
+        actions: {}
+    },
+    componentDidMount() {
+        this.geoJSON = new ol.format.GeoJSON();
+    },
+    shouldComponentUpdate(nextProps) {
+        return (!isUndefined(nextProps.featuregrid.zoomto) && !isEqual(nextProps.featuregrid.zoomto, this.props.featuregrid.zoomto));
+    },
+    componentDidUpdate() {
+        if (this.props.featuregrid && this.props.featuregrid.zoomto && this.props.featuregrid.zoomto.length > 0) {
```
mbarto: I would move this logic in FeatureGrid, and remove ol explicit dependencies, let's use CoordinateUtils

PR #479, opened 2016-03-01
Added selection to featuregrid

kappu72:


web/client/examples/featuregrid/components/FeatureGridMap.jsx
```diff
@@ -28,26 +29,39 @@ const FeatureGridMap = (props) => {
                     <Layer key={layer.name} position={index} type={layer.type}
                         options={assign({}, layer, {srs: props.map.projection})}/>
                 )}
-                <Layer type="vector" position={1} options={{name: "States", features: props.features}} />
+                <Layer type="vector" position={1} options={{name: "States"}}>
+                {
+                    features.map( (feature) => {
+                        return React.createElement(Feature, {
```
mbarto: Please use <Feature .../> instead of React.createElement(Feature, ...)

PR #540, opened 2016-05-03
Added styler to shapefile import component

kappu72:
ShapeFile plugin, allow users to import local (shape  and geoJson format) zipped files.
Files, can be dragged or selected. Allow importing multiple files and zip file with multiple shapefiles.
After selecting the shapefile to be imported, users can choose the style for the shape geometry type.
In case of multiple import, the first file determine the geometry type for the styler.
screen shot 2016-05-03 at 10 52 08
screen shot 2016-05-03 at 10 53 31
screen shot 2016-05-03 at 10 53 15

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "proj4": "~2.3.6",
     "react": "0.14.8",
     "react-bootstrap": "0.28.1",
+    "react-color": "^2.0.0",
     "react-dom": "0.14.8",
     "react-draggable": "1.3.4",
     "react-dropzone": "^3.4.0",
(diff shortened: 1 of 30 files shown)


offtherailz:
When the user have a file selected (e.g. he has already dragged the file into the dragging area) the dragging area should disappear and instead the UI should show the file name.
After the Add all the tool should be reset to the initial state (now you can add the same shape many times)

kappu72:
Added combo box that shows all Layers to be imported. Choose a style for selected layer and add to map.

PR #563, opened 2016-05-12
Fixes #552 Handle properly the new props

### LINKED ISSUE #552: Clicking two times in the zoom to max extent button breaks map naviagtion
After clicking two times in the zoom to max extent button, the first time we try to zoom in the map we will get an empty map:
anim
Note that this only happens when using LeafLetJS with Ol3 this works fine.

nmco:
We were testing new props against the map values instead of the current props. I update the way we test the new props by checking against the current props and the map values. This pull request also ads a proper test for this.

### DIFF
--- a/web/client/components/map/leaflet/Map.jsx
+++ b/web/client/components/map/leaflet/Map.jsx
@@ -155,8 +155,7 @@ let LeafletMap = React.createClass({
         );
     },
     _updateMapPositionFromNewProps(newProps) {
-        // Do the change at the same time, to avoid glitches
-        const currentCenter = this.map.getCenter();
+
         // current implementation will update the map only if the movement
         // between 12 decimals in the reference system to avoid rounded value
         // changes due to float mathematic operations.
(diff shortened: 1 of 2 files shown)


web/client/components/map/leaflet/Map.jsx
```diff
@@ -166,14 +165,34 @@ let LeafletMap = React.createClass({
...
+        // getting all centers we need to check
+        const newCenter = newProps.center;
+        const currentCenter = this.props.center;
+        const mapCenter = this.map.getCenter();
+        // checking if the current props are the same
+        const propsCentersEqual = isNearlyEqual(newCenter.x, currentCenter.x) &&
+                                  isNearlyEqual(newCenter.y, currentCenter.y);
+        // if props are the same nothing to do, otherwise
+        // we need to check if the new center is equal to map center
+        const centerIsUpdate = propsCentersEqual ||
```
mbarto: constant/variable names are important: this should be called centerIsNotUpdated

web/client/components/map/leaflet/Map.jsx
```diff
@@ -166,14 +165,34 @@ let LeafletMap = React.createClass({
...
+
+        // getting all zoom values we need to check
+        const newZoom = newProps.zoom;
+        const currentZoom = this.props.zoom;
+        const mapZoom = this.map.getZoom();
+        // checking if the current props are the same
+        const propsZoomEqual = newZoom === currentZoom;
+        // if props are the same nothing to do, otherwise
+        // we need to check if the new zoom is equal to map zoom
+        const zoomIsUpdate = propsZoomEqual || newZoom === mapZoom;
```
mbarto: same here: this should be called zoomIsNotUpdated

nmco:
Updated. Thanks for the review.

PR #678, opened 2016-06-14
Fixes #676: expandable Toolbar

### LINKED ISSUE #676: Expandable Toolbar
We should allow an expandable Toolbar, with:
 * tool to expand / collapse
 * configuration for always visible / collapsable tools

mbarto:


web/client/components/app/StandardApp.jsx
```diff
@@ -61,7 +64,9 @@ const StandardApp = React.createClass({
     },
     init() {
         this.store.dispatch(changeBrowserProperties(ConfigUtils.getBrowserProperties()));
-
+        if (urlQuery.appConfig) {
```
offtherailz: appConfig.js is a name used to load the StandardApplication config, so it can confuse. Let's use localConfig.

PR #685, opened 2016-06-16
Move Measure tool in DrawerMenu

Gnafu:
To display the measure results in a floating panel there is a new plugin "MeasureResults".  
The new flag "showResults" in the MeasurePlugin is set to false, the results are not visibile in the panel (to be used with the MeasureResults).  

The Reset button has been removed, to disable the tools just toggle the active button.

### DIFF
--- a/web/client/components/mapcontrols/measure/MeasureResults.jsx
+++ b/web/client/components/mapcontrols/measure/MeasureResults.jsx
@@ -1,0 +1,112 @@
+/**
+ * Copyright 2016, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const React = require('react');
+const {Panel} = require('react-bootstrap');
+const Draggable = require('react-draggable');
+
+const Message = require('../../../plugins/locale/Message');
+const ReactIntl = require('react-intl');
+const FormattedNumber = ReactIntl.FormattedNumber;
+const measureUtils = require('../../../utils/MeasureUtils');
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -91,13 +91,13 @@
...
     "object-assign": "3.0.0",
     "ogc-schemas": "2.6.1",
-    "w3c-schemas": "1.3.1",
     "openlayers": "3.15.1",
     "pdfviewer": "0.3.2",
     "proj4": "2.3.14",
     "react": "0.14.8",
     "react-bootstrap": "0.28.1",
     "react-color": "2.0.0",
+    "react-contextmenu": "^1.6.0",
```
mbarto: Did you use it at the end? If not, please remove

web/client/components/mapcontrols/measure/MeasureResults.jsx
```diff
@@ -0,0 +1,181 @@
...
+            bearingMeasureEnabled: false,
+            geomType: null,
+            // reset old measurements
+            len: 0,
+            area: 0,
+            bearing: 0
+        };
+        this.props.toggleMeasure(newMeasureState);
+    },
+    getFormattedBearingValue(azimuth = 0) {
```
mbarto: Can we move common functions to a utility module and use them in both MeasureComponent and MeasureResults?

web/client/plugins/MeasureResults.jsx
```diff
@@ -0,0 +1,42 @@
+/**
```
mbarto: Can we have just one plugin that uses different components for the Toolbar and DrawerMenu?

PR #747, opened 2016-06-27
Add support to map cards for the home

offtherailz:
* thumbnail management with the `thumbnail` attribute from GeoStore. 
image


offtherailz:
this is the themed version: 
image

PR #788, opened 2016-07-01
Number text components

kappu72:


web/client/components/data/query/SimpleFilterField.jsx
```diff
@@ -9,21 +9,24 @@
 const React = require('react');
 const {Panel, ButtonToolbar, Button, OverlayTrigger, Tooltip} = require('react-bootstrap');
 const ComboField = require('./ComboField');
+const NumberField = require('./NumberField');
+const TextField = require('./TextField');
 const {isEqual} = require('lodash');
 
 const SimpleFilterFiled = React.createClass({
```
mbarto: SimpleFilterField

web/client/components/data/query/SimpleFilterField.jsx
```diff
@@ -9,21 +9,24 @@
...
         optionsValues: React.PropTypes.array,
         defaultOptions: React.PropTypes.array,
         optionsLabels: React.PropTypes.object,
-        values: React.PropTypes.array,
+        values: React.PropTypes.oneOfType([React.PropTypes.array, React.PropTypes.object, React.PropTypes.number, React.PropTypes.string]),
         multivalue: React.PropTypes.bool,
         combo: React.PropTypes.bool,
         toolbar: React.PropTypes.bool,
-        type: React.PropTypes.oneOf(['list', 'date', 'bool']),
+        type: React.PropTypes.oneOf(['list', 'number', 'string']).isRequired,
```
mbarto: Why did you remove 'date' and 'bool'?

web/client/components/data/query/SimpleFilterField.jsx
kappu72: Sorry for the typo.

web/client/components/data/query/SimpleFilterField.jsx
kappu72: I'll remove date and bool because they aren't implemented and removing them will result in a warning from React-PropTypes validator if passed as option.

PR #824, opened 2016-07-06
Fix #823, Fixed the SharePlugin tests

### LINKED ISSUE #823: Tests for the share plugins fail randomly
The SharePlugins' social network counters dispatch ajax request that sometimes make the test fail.
We should override the function for the tests.

MV88:
Override the ajax calls that sometimes cause a test failure "x After each hook"


offtherailz:
We should check if the regular component functionality works with this fix (I am not sure that the default `getCount` function is replaced with null also in the normal mode)

PR #825, opened 2016-07-06
Resolves #812, Add authentication when accessing external resources

### LINKED ISSUE #812: Use authentication when accessing external resources
The proposal for this work is available here: https://github.com/geosolutions-it/MapStore2/wiki/Security-System-Proposal

Authentication should be added to all axios request transparently using an interceptor. GeoServer sharing links should use authkey parameter if configured. OL3 and Leaflet should also use the configured authentication.

nmco:
Associated issue: https://github.com/geosolutions-it/MapStore2/issues/812

### DIFF
--- a/web/client/actions/security.js
+++ b/web/client/actions/security.js
@@ -37,7 +37,6 @@ function loginFail(e) {
 }
 
 function logout(redirectUrl) {
-    GeoStoreAPI.logout(); // this resets the credentials
     return {
         type: LOGOUT,
         redirectUrl: redirectUrl
@@ -69,9 +68,8 @@ function changePasswordFail(e) {
     };
 }
 function geoStoreChangePassword(user, newPassword) {
-    return (dispatch, getState) => {
-        let opts = GeoStoreAPI.getAuthOptionsFromState(getState());
-        GeoStoreAPI.changePassword(user, newPassword, opts).then(() => {
+    return (dispatch) => {
+        GeoStoreAPI.changePassword(user, newPassword).then(() => {
             dispatch(changePasswordSuccess(user, newPassword));
         }).catch((e) => {
             dispatch(changePasswordFail(e));
--- a/web/client/components/catalog/SharingLink.jsx
+++ b/web/client/components/catalog/SharingLink.jsx
@@ -9,6 +9,7 @@ const React = require('react');
 const Message = require('../I18N/Message');
 const {Input, Button, Glyphicon, OverlayTrigger, Tooltip} = require('react-bootstrap');
 const CopyToClipboard = require('react-copy-to-clipboard');
+const SecurityUtils = require('../../utils/SecurityUtils');
 
 const SharingLink = React.createClass({
     propTypes: {
@@ -17,12 +18,14 @@ const SharingLink = React.createClass({
         onCopy: React.PropTypes.func,
         messages: React.PropTypes.object,
         locale: React.PropTypes.string,
-        bsSize: React.PropTypes.string
+        bsSize: React.PropTypes.string,
+        addAuthentication: React.PropTypes.bool
     },
     getDefaultProps() {
         return {
             onCopy: () => {},
-            bsSize: 'small'
+            bsSize: 'small',
+            addAuthentication: false
         };
     },
     getInitialState() {
@@ -32,6 +35,8 @@ const SharingLink = React.createClass({
         if (!this.props.url) {
             return null;
         }
+        // add authentication to the url if possible
+        const url = this.props.addAuthentication ? SecurityUtils.addAuthenticationToUrl(this.props.url) : this.props.url;
         const messageId = this.state.showCopiedToolTip ? "catalog.copied" : "catalog.copyToClipboard";
         const tooltip = (
             <Tooltip id="tooltip">
@@ -39,7 +44,7 @@ const SharingLink = React.createClass({
             </Tooltip>
         );
         const copyButton = (
-            <CopyToClipboard text={this.props.url} onCopy={this.props.onCopy}>
+            <CopyToClipboard text={url} onCopy={this.props.onCopy}>
                 <OverlayTrigger placement="right" overlay={tooltip} onExited={() => this.setState({showCopiedToolTip: false})}>
                     <Button bsSize={this.props.bsSize} bsStyle="primary" className="link-button" onClick={() => this.setState({showCopiedToolTip: true})}>
                         <Glyphicon glyph="paperclip"/>&nbsp;{this.props.labelId
@@ -51,7 +56,7 @@ const SharingLink = React.createClass({
         return (
             <div className="link-sharing">
                 <Input bsSize={this.props.bsSize} className="link-input"
-                    type="text" value={this.props.url} onChange={() => {}} buttonAfter={copyButton}/>
+                    type="text" value={url} onChange={() => {}} buttonAfter={copyButton}/>
             </div>
         );
     }
--- a/web/client/libs/__tests__/ajax-test.js
+++ b/web/client/libs/__tests__/ajax-test.js
@@ -8,6 +8,54 @@
 
 var expect = require('expect');
 var axios = require('../ajax');
+const SecurityUtils = require('../../utils/SecurityUtils');
+const assign = require('object-assign');
+
+const userA = {
+    User: {
+        enabled: true,
+        groups: "",
+        id: 6,
+        name: "adminA",
+        role: "ADMIN"
+    }
+};
+
+const securityInfoA = {
+    user: userA
+};
+
+const userB = assign({}, userA, {
+    name: "adminB",
+    attribute: [{
+        name: "UUID",
+        value: "263c6917-543f-43e3-8e1a-6a0d29952f72"
+    }, {
+        name: "description",
+        value: "admin user"
+    }
+]});
+
+const securityInfoB = {
+    user: userB,
+    token: "263c6917-543f-43e3-8e1a-6a0d29952f72",
+    authHeader: 'Basic 263c6917-543f-43e3-8e1a-6a0d29952f72'
+};
+
+const authenticationRules = [
+    {
+      "urlPattern": ".*geoserver.*",
+      "method": "authkey"
+    },
+    {
+      "urlPattern": ".*not-supported.*",
+      "method": "not-supported"
+    },
+    {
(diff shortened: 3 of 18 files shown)


web/client/actions/security.js
```diff
@@ -18,6 +19,12 @@ const LOGOUT = 'LOGOUT';
 
 
 function loginSuccess(userDetails, username, password, authProvider) {
+    // letting security utils know about a successfull login
+    SecurityUtils.provideSecurityInfo({
+        user: userDetails && userDetails.User,
+        username: username,
```
offtherailz: this should stay in the reducer, and the password should not be saved (basic header instead), follow how we do it with GeoStore utilities.

web/client/components/catalog/SharingLink.jsx
```diff
@@ -32,14 +33,16 @@ const SharingLink = React.createClass({
         if (!this.props.url) {
             return null;
         }
+        // add authentication to the url if possible
+        const url = SecurityUtils.addAuthenticationToUrl(this.props.url);
```
offtherailz: the add authkey functionality have to be optional, with a configuration prop.

web/client/libs/ajax.js
```diff
@@ -92,6 +93,10 @@ axios.interceptors.request.use(config => {
     var uri = config.url || '';
     var sameOrigin = !(uri.indexOf("http") === 0);
     var urlParts = !sameOrigin && uri.match(/([^:]*:)\/\/([^:]*:?[^@]*@)?([^:\/\?]*):?([^\/\?]*)/);
+    if (uri.indexOf("http") === 0) {
+        // we are only interested in http and https requests
+        SecurityUtils.addAuthenticationToAxios(config);
```
offtherailz: Axios should be aware of Security utils, not the opposite.

web/client/reducers/security.js
```diff
@@ -11,9 +11,23 @@ const SecurityUtils = require('../utils/SecurityUtils');
...
+const {REHYDRATE} = require('redux-persist/constants');
+
 function security(state = {user: null, errorCause: null}, action) {
     switch (action.type) {
 
+        case REHYDRATE:
+            let incoming = action.payload.security;
+            if (incoming) {
+                // letting security utils know about the security info
+                SecurityUtils.provideSecurityInfo({
```
offtherailz: As before, store the security object, not username and password.

web/client/utils/SecurityUtils.js
```diff
@@ -6,18 +6,196 @@
...
+     */
+    getUserName() {
+        return this.securityInfo && this.securityInfo.username;
+    },
+
+    /**
+     * Returns the current user password or undefined if not available.
+     */
+    getUserPassword() {
+        return this.securityInfo && this.securityInfo.password;
```
offtherailz: no password to store

web/client/utils/SecurityUtils.js
```diff
@@ -6,18 +6,196 @@
...
+                addParameterToAxiosConfig(axiosConfig, 'authkey', userUuid);
+                return axiosConfig;
+            case 'basic':
+                const userName = this.getUserName();
+                const userPassword = this.getUserPassword();
+                if (!userName || !userPassword) {
+                    return axiosConfig;
+                }
+                // axios support basic authentication natively
+                axiosConfig.auth = {
```
offtherailz: Ignore auth object and inject the header instead. Something like:  
https://github.com/geosolutions-it/MapStore2/blob/master/web/client/api/GeoStoreDAO.js#L65

web/client/libs/ajax.js
```diff
@@ -92,6 +93,10 @@ axios.interceptors.request.use(config => {
     var uri = config.url || '';
     var sameOrigin = !(uri.indexOf("http") === 0);
     var urlParts = !sameOrigin && uri.match(/([^:]*:)\/\/([^:]*:?[^@]*@)?([^:\/\?]*):?([^\/\?]*)/);
+    if (uri.indexOf("http") === 0) {
+        // we are only interested in http and https requests
```
offtherailz: This is wrong, you are not checking the request type, but looking for absolute URL! also relative URL request must be intercepted.

web/client/utils/SecurityUtils.js
```diff
@@ -6,18 +6,196 @@
...
+        const foundRule = this.getAuthenticationRules().find(
+            rule => rule && rule.urlPattern && url.match(new RegExp(rule.urlPattern, "i")));
+        return foundRule ? foundRule.method : undefined;
+    },
+
+    /**
+     * This method will add to the axios config object the correct authentication
+     * method based on the request URL.
+     */
+    addAuthenticationToAxios: function(axiosConfig) {
```
offtherailz: as above, move this to axios "interceptor"

offtherailz:
Do not store anything in the SecurityUtils class. provide a reference to the store in the DefaultStore and then use SecurityUtils to get rules and apply them in axios. This should be simpler and clearer.

web/client/components/catalog/SharingLink.jsx
nmco: Method SecurityUtils.addAuthenticationToUrl will check if authentication is activated before adding anything.

offtherailz: This functionality have 2 purposes:
1. Get The link for the WMS for your personal GIS
2. Share with another user 
In the first case it is ok to skip authentication problems, but for the 2nd case is very bad to share the authentication key. 
So make it optional.

web/client/libs/__tests__/ajax-test.js
```diff
@@ -8,6 +8,53 @@
...
+const SecurityUtils = require('../../utils/SecurityUtils');
+const assign = require('object-assign');
+
+const userA = {
+    User: {
+        enabled: true,
+        groups: "",
+        id: 6,
+        name: "adminA",
+        rol: "ADMIN"
```
offtherailz: role

web/client/libs/ajax.js
```diff
@@ -10,6 +10,52 @@ var axios = require('axios');
...
+function addParameterToAxiosConfig(axiosConfig, parameterName, parameterValue) {
+    // FIXME: the parameters can also be a URLSearchParams
+    axiosConfig.params = Object.assign({}, axiosConfig.params, {[parameterName]: parameterValue});
+}
+
+/**
+ * Internal helper that adds or overrides an http header in a axios configuration.
+ */
+function addHeaderToAxiosConfig(axiosConfig, headerName, headerValue) {
+    axiosConfig.headers = Object.assign({}, axiosConfig.headers, {[headerName]: headerValue});
```
offtherailz: Object.assign is not compatible with all browsers, please use `require('object-assign)`

web/client/libs/ajax.js
```diff
@@ -10,6 +10,52 @@ var axios = require('axios');
...
 var toString = Object.prototype.toString;
 
+const SecurityUtils = require('../utils/SecurityUtils');
+
+/**
+ * Internal helper that adds an extra paramater to an axios configuration.
+ */
+function addParameterToAxiosConfig(axiosConfig, parameterName, parameterValue) {
+    // FIXME: the parameters can also be a URLSearchParams
+    axiosConfig.params = Object.assign({}, axiosConfig.params, {[parameterName]: parameterValue});
```
offtherailz: Object.assign is not compatible with all browsers, please use `require('object-assign)`

web/client/utils/SecurityUtils.js
```diff
@@ -6,18 +6,149 @@
...
+        if (!url || !this.isAuthenticationActivated()) {
+            return parameters;
+        }
+        switch (this.getAuthenticationMethod(url)) {
+            case 'authkey':
+                const userUuid = this.findUserAttributeValue('uuid');
+                if (!userUuid) {
+                    return parameters;
+                }
+                return Object.assign(parameters || {}, {'authkey': userUuid});
```
offtherailz: Object.assign is not compatible with all browsers, please use `require('object-assign)`

web/client/utils/SecurityUtils.js
```diff
@@ -6,18 +6,149 @@
...
+     * This method will add query parameter based authentications to an object
+     * containing query paramaters.
+     */
+    addAuthenticationParameter: function(url, parameters) {
+        if (!url || !this.isAuthenticationActivated()) {
+            return parameters;
+        }
+        switch (this.getAuthenticationMethod(url)) {
+            case 'authkey':
+                const userUuid = this.findUserAttributeValue('uuid');
```
offtherailz: token is already present in the security object and is more general (another login service should provide a temporary token that have to be used during the session instead of the static authkey. 
Use security.token.

web/client/libs/ajax.js
```diff
@@ -10,6 +10,52 @@ var axios = require('axios');
...
+ * Internal helper that will add to the axios config object the correct
+ * authentication method based on the request URL.
+ */
+function addAuthenticationToAxios(axiosConfig) {
+    if (!axiosConfig || !axiosConfig.url || !SecurityUtils.isAuthenticationActivated()) {
+        return axiosConfig;
+    }
+    switch (SecurityUtils.getAuthenticationMethod(axiosConfig.url)) {
+        case 'authkey':
+            const userUuid = SecurityUtils.findUserAttributeValue('uuid');
```
offtherailz: use token

web/client/localConfig.json
```diff
@@ -5,6 +5,13 @@
 	"bingApiKey": "AhuXBu7ipR1gNbBfXhtUAyCZ6rkC5PkWpxs2MnMRZ1ZupxQfivjLCch22ozKSCAn",
 	"initialMapFilter": "MS2",
   "ignoreMobileCss" : true,
+	"useAuthenticationRules": false,
+	"authenticationRules": [
+		{
+			"urlPattern": ".*geoserver.*",
```
offtherailz: No rules by default

offtherailz:
:+1: Seems great !! Let's remove the default rules and we can merge it

nmco:
Add default security rule for GeoStore. Add fix that makes GeoStoreDAO use the global axios instance.

offtherailz:
GeoStoreDAO logout() method was forcing re-creation of the custom axios instance in an initial and raw implementation of authentication persistance system. It is obsolete, You did the right thing removing it.

PR #831, opened 2016-07-07
Add Marker on GeoCoder result. Fixes #275

### LINKED ISSUE #275: Improve geocoder Feedback
* [x] A localized text should indicate something like "Search by location name or coordinates ..."
* [ ] When the item is selected, we should place a marker (configurable, disappearing if possible) centered in the location selected.

Gnafu:


Gnafu:
Closes #275

PR #872, opened 2016-07-21
Fixed #546 issue, zoom on the shapefile after insertion

### LINKED ISSUE #546: Upload shapefile should zoom on imported shapefile
Currently when you import a shapefile if this is "out of screen" it's not immediately clear that the operation was succesful. We can either:
 * zoom automatically on shapefile extents
 * show an option to the user to zoom or not (The shapefile has been succesfully imported. Would you like to zoom on it (Yes/No) )

MV88:
I have fixed the issue #546.

I noticed that when i load the shapefiles of Afganistan and NorthAmerica the zoom level is too low like 3 level behind on a normal desktop screen. On mobile screens the zoom level seems to be good.

### DIFF
--- a/web/client/components/shapefile/ShapefileUploadAndStyle.jsx
+++ b/web/client/components/shapefile/ShapefileUploadAndStyle.jsx
@@ -146,19 +146,19 @@ const ShapeFileUploadAndStyle = React.createClass({
         if (!this.state.useDefaultStyle) {
             styledLayer = StyleUtils.toVectorStyle(styledLayer, this.props.shapeStyle);
         }
-        let bbox = this.props.bbox;
         Promise.resolve(this.props.addShapeLayer( styledLayer )).then(() => {
             this.props.shapeLoading(false);
 
-            // calculate the max bbox that contain all shapefiles
-            let layers = this.props.layers;
-            layers[0].features.reduce((total, feature) => {
-                let actualBBox = feature.geometry.bbox;
-                bbox[0] = Math.min(bbox[0], actualBBox[0]); // minx
-                bbox[1] = Math.min(bbox[1], actualBBox[1]); // miny
-                bbox[2] = Math.max(bbox[2], actualBBox[2]); // maxx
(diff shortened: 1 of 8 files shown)


web/client/components/shapefile/ShapefileUploadAndStyle.jsx
```diff
@@ -127,14 +140,29 @@ const ShapeFileUploadAndStyle = React.createClass({
...
         if (!this.state.useDefaultStyle) {
             styledLayer = StyleUtils.toVectorStyle(styledLayer, this.props.shapeStyle);
         }
+        let bbox = this.props.bbox;
         Promise.resolve(this.props.addShapeLayer( styledLayer )).then(() => {
             this.props.shapeLoading(false);
+
+            // calculate the max bbox that contain all shapefiles
+            let layers = this.props.layers;
+            layers[0].features.reduce((total, feature) => {
```
mbarto: You have to assign reduce result to a new const

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
```diff
@@ -127,14 +140,29 @@ const ShapeFileUploadAndStyle = React.createClass({
...
+                let actualBBox = feature.geometry.bbox;
+                bbox[0] = Math.min(bbox[0], actualBBox[0]); // minx
```
mbarto: you have to return the new bbox, do not do side effects in reduce

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
MV88: Is it correct like this?
const bbox = layers[0].features.reduce((bboxtotal, feature) => {
                 let actualBBox = feature.geometry.bbox;
                 bboxtotal[0] = Math.min(bboxtotal[0], actualBBox[0]); // minx 
                 ....
                 bboxtotal[3] = Math.max(bboxtotal[3], actualBBox[3]); // maxy
}, this.props.bbox);

PR #898, opened 2016-07-28
Save maps in new format

Gnafu:
The save tool is not active by default, the following configuration must be used:
```
{
    "name" : "Save",
    "cfg": {
        "closeGlyph" : "1-close"
     }
},
```

### DIFF
--- a/web/client/actions/config.js
+++ b/web/client/actions/config.js
@@ -11,11 +11,12 @@ var axios = require('../libs/ajax');
 const MAP_CONFIG_LOADED = 'MAP_CONFIG_LOADED';
 const MAP_CONFIG_LOAD_ERROR = 'MAP_CONFIG_LOAD_ERROR';
 
-function configureMap(conf, legacy) {
+function configureMap(conf, mapId) {
     return {
         type: MAP_CONFIG_LOADED,
         config: conf,
-        legacy: legacy || false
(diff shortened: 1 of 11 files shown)


web/client/components/maps/SavePanel.jsx
```diff
@@ -0,0 +1,50 @@
...
+            <Dialog id="share-panel-dialog" className="modal-dialog modal-content share-win">
+                <span role="header">
+                    <span className="share-panel-title">
+                        <Message msgId="map.saveTitle"/>
+                    </span>
+                    <button onClick={this.props.onClose} className="print-panel-close close">
+                        {this.props.closeGlyph ? <Glyphicon glyph={this.props.closeGlyph}/> : <span>×</span>}
+                    </button>
+                </span>
+                <div role="body" className="share-panels">
```
offtherailz: share-panels seems to be copied from another dialog. Do not reuse same class, or give it a more general name, if you think it can be reused

web/client/components/maps/SavePanel.jsx
```diff
@@ -0,0 +1,50 @@
...
+
```
offtherailz: Body is empty?

Gnafu:
Removed the unused file

PR #906, opened 2016-08-01
Enable click on map card

### LINKED ISSUE #844: When clicking on a map thumbnail the map itself should open
Currently, to open a map from the home page you need to clik on the icon at the LRC.

This is fine but on the mobile it would be much better if we could do the same when clicking anywhere on the map thumbnail.

Gnafu:
Fixes #844

### DIFF
--- a/web/client/components/misc/GridCard.jsx
+++ b/web/client/components/misc/GridCard.jsx
@@ -24,6 +24,16 @@ const GridCard = React.createClass({
             header: ""
         };
     },
+    // click on the card triggers the first available action
+    onClick() {
+        // Users can select Title and Description without triggering the click
+        var selection = window.getSelection();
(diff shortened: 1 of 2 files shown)


web/client/components/misc/GridCard.jsx
```diff
@@ -24,6 +24,16 @@ const GridCard = React.createClass({
             header: ""
         };
     },
+    // click on the card triggers the first available action
+    onClick() {
+        // Users can select Title and Description without triggering the click
```
offtherailz: So only an area without any text is clickable?

Gnafu: No, all the card area is clickable and will open the map.
Only if the user highlights something the map will not open.

Issue #855, opened 2016-07-13
Create new LayerGroup


Gnafu:
Users should be able to create new layergroups and add layers to them

PR #911, opened 2016-08-04
Move layer from a group to another. Closes #854 #855

### LINKED ISSUE #854: Enable users to move Layers in different Group
The layers order can be changed within the same TOC layergroup but layers cannot be moved to another group.

Gnafu:
And create the group if it does not exists


Gnafu:
Updated with empty group removal and fixed tests

PR #932, opened 2016-08-30
Close #931. Add manager page and menu

### LINKED ISSUE #931: Include a manager page
The manager page can be filled with some administration tools (user management, security settings and so on...).
Also a plugin to access to the manager page.

offtherailz:
This adds to MapStore 2 a manager page. 
This page is a generic page that contains tools for administration. A plugin to access to that menu is also included ( only administrators can see the menu )

### DIFF
--- a/web/client/plugins/manager/Manager.jsx
+++ b/web/client/plugins/manager/Manager.jsx
@@ -1,0 +1,74 @@
+/**
+ * Copyright 2016, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const React = require('react');
+
+const {Nav, NavItem, Glyphicon} = require('react-bootstrap');
+const {connect} = require('react-redux');
+const {Message} = require('../../components/I18N/I18N');
+require('./style/manager.css');
+
+const Manager = React.createClass({
+    propTypes: {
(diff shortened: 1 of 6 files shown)


web/client/plugins/manager/Manager.jsx
```diff
@@ -0,0 +1,74 @@
+/**
```
mbarto: We usually put the jsx of plugins in the plugins folder, not in a subfolder. We should not change this, unless we do the same for all the other plugins

web/client/plugins/manager/ManagerMenu.jsx
```diff
@@ -0,0 +1,105 @@
+/**
```
mbarto: Is this one so different from the BurgerMenu plugin?

web/client/plugins/manager/Manager.jsx
offtherailz: This has been done because we have different plugins specific for manager, so we should group all them together in that folder.
There are many plugins now, and I suggest to group them somehow.

web/client/plugins/manager/ManagerMenu.jsx
offtherailz: Not too much, only the visibility (admin only) and the possibility to configure it manually.
In the future we could consider to make 2 instances of the burger menu with different tools wired inside it.

PR #954, opened 2016-09-09
Reload maps on successful login

### LINKED ISSUE #953: When logging in / out in HomePage Maps List is not reloaded
We should reload the maps list when the security context changes.

Gnafu:
Fixes #953


web/client/actions/security.js
```diff
@@ -43,10 +44,18 @@ function logout(redirectUrl) {
     };
 }
 
+function logoutWithReload(redirectUrl) {
```
mbarto: Let's move it to maps

Gnafu:
Rebased

PR #959, opened 2016-09-09
Fixed #947 : updated SaveAsPlugin with the changes applied to the MetaDataModal

### LINKED ISSUE #947: SaveAs functionality should be added and integrated with new changes to the metadata modal
SaveAs Plugin need to be updated after the changes applied to the thumbnail and metadata modal.

MV88:
I updated the SaveAsPlugin in order to work correctly with the changes made on the Thumbnail and MetaDataModal components. Reviewed also the zone for displaying any error relative to the map, the thumbnail and the image dropped.


Gnafu:
Adding a layer from the Catalog generates a bad map that cannot be loaded again.

MV88:
There were a bug when creating the layers of the new map. It was missing the provider parameter.

MV88:
Messed up. I will redo the PR.

PR #979, opened 2016-09-15
Fix #980 Minor ScaleBox Plugin Improvement in order to support custom projections

### LINKED ISSUE #980: Scalebox Plugin does not handle custom projections
Hi,

I am currently using the scaleboxplugin (master@mapstore2) with a map configured with the belgium coordinate system (EPSG:31370).

By default the scales shown are not correct. I think the cause is that the plugin takes, as default parameter, the scales for Google Mercator. And i do not see a way to override this default behavior.

Regards,

ndufrane:
Hi,

Proposal for the issue #980.

The modification allow the scalebox plugin to take into account the map projection for computing scales.

Regards,


simboss:
Hey Nicolas, thx for this contribution.

Would you be so kind to open an issue and reference it from the title of the PR?

PR #984, opened 2016-09-16
Fix #983 add workspace creation and selection in importer tool.

### LINKED ISSUE #983: add the possibility to create and select workspaces before import
The importer should provide a built-in functionality to create workspaces with pre-selected datastore (e.g. a jndi datastore where vector data should be importer).

offtherailz:
image

### DIFF
--- a/web/client/plugins/VectorStyler.jsx
+++ b/web/client/plugins/VectorStyler.jsx
@@ -122,7 +122,7 @@ const VectorStyler = React.createClass({
         switch (this.props.layer.describeLayer.geometryType) {
             case 'Polygon':
             case 'MultiPolygon':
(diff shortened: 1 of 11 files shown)


web/client/plugins/VectorStyler.jsx
```diff
@@ -122,20 +122,16 @@ const VectorStyler = React.createClass({
         switch (this.props.layer.describeLayer.geometryType) {
             case 'Polygon':
             case 'MultiPolygon':
-            case 'MultiSurfacePropertyType': { // TODO remove propertytype
+            case 'MultiSurface': { // TODO remove propertytype
```
mbarto: You can remove the comment

PR #1003, opened 2016-09-21
Add tests to geoserver api

offtherailz:
Add tests and test resources for the GeoServer ReST API.


Gnafu:
Again, coveralls is reporting wrong percentages.
merging.

PR #1009, opened 2016-09-22
Translation json test

saidaipparla:


web/client/actions/__tests__/locale-test.js
```diff
@@ -5,7 +5,6 @@
  * This source code is licensed under the BSD-style license found in the
```
MV88: Every file you modify you have to change the date, in the copyright section, from 2015 to 2015-2016

PR #1023, opened 2016-09-23
Fixed #1011 : fixed several bugs on permission editor

### LINKED ISSUE #1011: The permission editor need some updates
- Some translations are missing and they also should be internationalized.
- If the user isn't the owner of the map --> hide permission editor
- The groups spinner should turn off if no available groups is left

MV88:
- [x] Some translations are missing and they also should be internationalized.
- [x] If the user isn't the owner of the map --> hide permission editor
- [x] The groups spinner should turn off if no available groups is left

### DIFF
--- a/web/client/components/maps/modals/MetadataModal.jsx
+++ b/web/client/components/maps/modals/MetadataModal.jsx
@@ -123,7 +123,7 @@ const MetadataModal = React.createClass({
     },
     componentDidUpdate(prevProps) {
         if (this.props.show && !prevProps.show) {
-            if (this.props.displayPermissionEditor) {
+            if (this.props.displayPermissionEditor && (this.props.user.name === this.props.map.owner || this.props.user.role === "ADMIN" )) {
                 this.loadPermissions();
                 this.loadAvailableGroups();
             }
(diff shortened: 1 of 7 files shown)


web/client/components/maps/modals/MetadataModal.jsx
```diff
@@ -123,7 +123,7 @@ const MetadataModal = React.createClass({
     },
     componentDidUpdate(prevProps) {
         if (this.props.show && !prevProps.show) {
-            if (this.props.displayPermissionEditor) {
+            if (this.props.displayPermissionEditor && this.props.user.name === this.props.map.owner) {
```
Gnafu: ADMIN role users should be able to edit map permissions even if they are not owners

web/client/components/maps/modals/MetadataModal.jsx
```diff
@@ -157,7 +157,7 @@ const MetadataModal = React.createClass({
         this.refs.thumbnail.updateThumbnail(this.props.map, metadata);
     },
     renderPermissionEditor() {
-        if (this.props.displayPermissionEditor) {
+        if (this.props.displayPermissionEditor && this.props.user.name === this.props.map.owner) {
```
Gnafu: Same as above, ADMIN users must be able to edit permissions

web/client/components/maps/modals/__tests__/MetaDataModal-test.jsx
```diff
@@ -30,18 +30,26 @@ describe('This test for MetadataModal', () => {
         expect(metadataModalItem).toExist();
 
         const metadataModalItemDom = ReactDOM.findDOMNode(metadataModalItem);
-        expect(metadataModalItemDom).toNotExist();
+        expect(metadataModalItemDom).toExist();
```
Gnafu: please test that it is not visible

PR #1031, opened 2016-09-26
Make onMapViewChanges configurable as property of MapPlugin

manisandro:
In QWC2 we need to override the changeMapView action to store the current map center coordinates in the URL.

AFAICS the simplest solution allowing for maximum code-reuse is to make the `onMapViewChanges` callback configurable at MapPlugin level.

CC @mbarto


mbarto:
Hi Sandro, I created a more generic version here: https://github.com/geosolutions-it/MapStore2/pull/1032
The new one allows overriding every action. Let's wait for the build

PR #1036, opened 2016-09-27
Ensure initial mousePosition state is set

manisandro:
Currently no initial state is set for `mousePosition`, and adding i.e.

    mousePosition: {
        enabled: true
    },

to `defaultState` in `appConfig.js` has no effect.


mbarto:
Hi Alessandro, the purpose of this PR is not clear to me. The reducer for mousePosition is already included by the plugin, when you include it in a page. Can you explain me the need for direct inclusion?

manisandro:
Hmpf, never mind, was looking for the problem in the wrong spot. Sorry for the noise.

PR #1040, opened 2016-09-28
Add jpeg-png on available formats. Closes #816

### LINKED ISSUE #816: Implement support for image/vnd.jpeg-png format
We have developed a new output format for WMS in GeoServer which is called image/vnd.jpeg-png. See this page for [basic info](image/vnd.jpeg-png).

It dynamically select png VS jpeg depending on the content of the image that is produced to account for the ability to have transparent tiles when needed and JPEG compressed tiles otherwise.

Gnafu:


web/client/components/TOC/fragments/settings/Display.jsx
```diff
@@ -32,7 +32,7 @@ module.exports = React.createClass({
             [(<label key="format-label" className="control-label"><Message msgId="layerProperties.format" /></label>),
             (<DropdownList
                 key="format-dropdown"
-                data={this.props.formats || ["image/png", "image/png8", "image/jpeg", "image/gif"]}
+                data={this.props.formats || ["image/png", "image/png8", "image/jpeg", "image/vnd.jpeg-png", "image/gif"]}
```
mbarto: Can we write a quick test for this?

Gnafu: Done on the layers reducer. The Display component already has the required tests

PR #1046, opened 2016-09-28
fix#1039 changed from MS2 to all

saidaipparla:


web/client/api/GeoStoreDAO.js
```diff
@@ -40,7 +40,7 @@ var Api = {
             this.addBaseUrl(parseOptions(options))).then(function(response) {return response.data; });
     },
     getResourcesByCategory: function(category, query, options) {
-        let q = query || "*";
+        let q = query && "*";
```
mbarto: This is not the required fix. We should remove the MS2 filter from configuration. This is asking for * for each configured filter.

saidaipparla: ok i will change

PR #1048, opened 2016-09-29
Proposed changes for QWC2

manisandro:
Some proposed changes / RFC:

- *Store length and area unit in store*: We'd like to store the selected measurement unit in our Measure dialog [1], so it would make sense to store it in the store.
- *Fix CRSSelector not rendering*, *Fix ScaleBox not rendering*: For some reason the controls rendered as `<Input>` aren't rendered (used i.e. in [2]). I wonder if it is related to the change

        Feature/Deprecation: Rewrite form and form control API (<FormControl>, &c.), and deprecate the old API (<Input>, &c.) (#1765)

    mentioned for v0.29 in [3]
- *Sync default maxLevel with default ZoomButton.maxZoom*: With the default settings, the zoom button can zoom further in than the available scales in the scalebox
- *Only pass zoomLevel to onChange function, the changeZoomLevel action doesn't expect a scale number as second parameter*: `changeZoomLevel` from `actions/map` is connected to `ScaleBox.props.onChange` in various examples. However, the parameters passed in `ScaleBox.onComboChange` (zoom level and scale) don't match what `changeZoomLevel` expects (zoom level and mapStateSource). The zoom buttons OTOH only pass the zoom level. Hence possibly also the `ScaleBox.onComboChange` should only pass the zoom level?
- *Don't show a tooltip if the corresponding property is empty*: If the zoom button tooltip text is empty, don't show the tooltip arrow

[1] https://github.com/sourcepole/qwc2/blob/master/js/plugins/Measure.jsx
[2] https://github.com/sourcepole/qwc2/blob/master/js/plugins/BottomBar.jsx
[3] https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md

### DIFF
--- a/web/client/components/mapcontrols/mouseposition/CRSSelector.jsx
+++ b/web/client/components/mapcontrols/mouseposition/CRSSelector.jsx
@@ -6,9 +6,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 var React = require('react');
-var ReactDOM = require('react-dom');
-var BootstrapReact = require('react-bootstrap');
-var {Input} = BootstrapReact;
 var CoordinatesUtils = require('../../../utils/CoordinatesUtils');
 
 let CRSSelector = React.createClass({
@@ -41,22 +38,19 @@ let CRSSelector = React.createClass({
             }
         }
         return (
-              (this.props.enabled) ? (<Input
+              (this.props.enabled) ? (<select
                     id={this.props.id}
                     value={this.props.crs}
-                    type="select"
                     onChange={this.launchNewCRSAction}
                     bsSize="small"
                     {...this.props.inputProps}>
                     {list}
-                </Input>) : null
+                </select>) : null
 
         );
     },
-    launchNewCRSAction() {
-        var element = ReactDOM.findDOMNode(this);
-        var selectNode = element.getElementsByTagName('select').item(0);
-        this.props.onCRSChange(selectNode.value);
+    launchNewCRSAction(ev) {
+        this.props.onCRSChange(ev.target.value);
     }
 });
 
--- a/web/client/components/mapcontrols/scale/ScaleBox.jsx
+++ b/web/client/components/mapcontrols/scale/ScaleBox.jsx
@@ -7,7 +7,6 @@
  */
 
 const React = require('react');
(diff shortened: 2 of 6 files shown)


simboss:
Ciao Sandro,
Mauro is OOO and will be back on Mon. 

I would wait for his feedback before moving on.

manisandro:
Sure no problem, thanks.

web/client/components/mapcontrols/mouseposition/CRSSelector.jsx
```diff
@@ -41,22 +38,19 @@ let CRSSelector = React.createClass({
             }
         }
         return (
-              (this.props.enabled) ? (<Input
+              (this.props.enabled) ? (<select
```
mbarto: the Input control works with the react-bootstrap version we include in MS2. Can you give more details on the use case where it does not work for you?

web/client/components/mapcontrols/scale/ScaleBox.jsx
```diff
@@ -25,7 +24,7 @@ var ScaleBox = React.createClass({
     getDefaultProps() {
         return {
             id: 'mapstore-scalebox',
-            scales: mapUtils.getGoogleMercatorScales(0, 21),
+            scales: mapUtils.getGoogleMercatorScales(0, 28),
```
mbarto: Let's make this an overridable prop with a default value of 21

web/client/components/mapcontrols/scale/ScaleBox.jsx
```diff
@@ -49,9 +48,9 @@ var ScaleBox = React.createClass({
     render() {
         let control = this.props.readOnly ?
             <label>{this.props.template(this.props.scales[this.props.currentZoomLvl], this.props.currentZoomLvl)}</label>
-        : <Input type="select" label={this.props.label} onChange={this.onComboChange} bsSize="small" value={this.props.currentZoomLvl}>
```
mbarto: Also here: Input works for us, before replacing it we need to understand when it doesn't work

manisandro:
I've split up this PR into individual PRs, I'll follow up in these PRs.

PR #1049, opened 2016-09-29
Update fractional zoom

saidaipparla:


mbarto:
Please solve the conflicts so that I can merge

web/client/components/map/leaflet/__tests__/Map-test.jsx
```diff
@@ -169,7 +169,7 @@ describe('LeafletMap', () => {
 
         const leafletMap = map.map;
         expect(leafletMap.getZoom()).toBe(12);
-        map.setProps({zoom: 12, center: {y: 44, x: 10}});
+        map.setProps({zoom: 12.4, center: {y: 44, x: 10}});
         expect(leafletMap.getZoom()).toBe(12);
```
Gnafu: This test is to "check if the map changes when receive new props"
The zoom is 12 before the test, so you should expect another value after the setProps().

web/client/components/map/openlayers/__tests__/Map-test.jsx
```diff
@@ -168,7 +168,7 @@ describe('OpenlayersMap', () => {
 
         const olMap = map.map;
         expect(olMap.getView().getZoom()).toBe(12);
-        map.setProps({zoom: 12, center: {y: 44, x: 10}});
+        map.setProps({zoom: 12.4, center: {y: 44, x: 10}});
         expect(olMap.getView().getZoom()).toBe(12);
```
Gnafu: Same as before: here the test should expect a different value than before the setProps()

web/client/components/map/leaflet/__tests__/Map-test.jsx
saidaipparla: ok i will change

PR #1051, opened 2016-09-30
Pagination

saidaipparla:


web/client/actions/maps.js
```diff
@@ -178,7 +178,7 @@ function permissionsLoaded(permissions, mapId) {
     };
 }
 
-function loadMaps(geoStoreUrl, searchText="*", params={start: 0, limit: 20}) {
+function loadMaps(geoStoreUrl, searchText="*", params={start: 0, limit: 5}) {
```
mbarto: Why are we changing the limit to 5?

web/client/plugins/Maps.jsx
```diff
@@ -60,9 +60,9 @@ const PaginationToolbar = connect((state) => {
     }
     let {start, limit, results, loading, totalCount, searchText} = state.maps;
     let page = 0;
-    let total = totalCount || 0;
+    //    let total = totalCount || 0;
     if (results && totalCount) { // must be !==0 and exist to do the division
-        page = Math.ceil(start / total);
+        page = Math.ceil(start / limit);
```
mbarto: We should probably compare limit and totalCount and take the lowest?

web/client/actions/maps.js
saidaipparla: i changed for  easy  debugging  and i changed back in second commit.

web/client/plugins/Maps.jsx
```diff
@@ -60,11 +60,16 @@ const PaginationToolbar = connect((state) => {
     }
     let {start, limit, results, loading, totalCount, searchText} = state.maps;
     let page = 0;
-    //    let total = totalCount || 0;
-    if (results && totalCount) { // must be !==0 and exist to do the division
-        page = Math.ceil(start / limit);
+    let total = totalCount || 0;
```
mbarto: const total = Math.min(totalCount || 0, limit || 0);
const page = (results && total && Math.ceil(start / total)) || 0;

Isn't this simpler?

Let's use const as much as possible. Let is usually not necessary.

PR #1055, opened 2016-10-03
Only pass zoomLevel to onChange function, the changeZoomLevel action doesn't expect a scale number as second parameter

manisandro:
`changeZoomLevel` from `actions/map` is connected to `ScaleBox.props.onChange` in various examples. However, the parameters passed in `ScaleBox.onComboChange` (zoom level and scale) don't match what `changeZoomLevel` expects (zoom level and mapStateSource). The zoom buttons OTOH only pass the zoom level. Hence possibly also the `ScaleBox.onComboChange` should only pass the zoom level?

### DIFF
--- a/web/client/components/mapcontrols/scale/ScaleBox.jsx
+++ b/web/client/components/mapcontrols/scale/ScaleBox.jsx
@@ -37,7 +37,7 @@ var ScaleBox = React.createClass({
     },
     onComboChange(event) {
         var selectedZoomLvl = parseInt(event.nativeEvent.target.value, 10);
-        this.props.onChange(selectedZoomLvl, this.props.scales[selectedZoomLvl]);
+        this.props.onChange(selectedZoomLvl);
     },
     getOptions() {
         return this.props.scales.map((item, index) => {


web/client/components/mapcontrols/scale/ScaleBox.jsx
```diff
@@ -37,7 +37,7 @@ var ScaleBox = React.createClass({
     },
     onComboChange(event) {
         var selectedZoomLvl = parseInt(event.nativeEvent.target.value, 10);
-        this.props.onChange(selectedZoomLvl, this.props.scales[selectedZoomLvl]);
```
mbarto: This is meant to be used by "alternative" action bound to onChange. Does the additional parameter create any issue for you?

manisandro: It triggers a warning 
    warning.js:45 Warning: Failed propType: Invalid prop `mapStateSource` of type `number` supplied to `OpenlayersMap`, expected `string`. Check the render method of `Connect(OpenlayersMap)`.
since that value is assigned to the action field `mapStateSource` of `map.js:changeZoomLevel`. Alternatively one could write another action with matching signature.

PR #1057, opened 2016-10-03
Sync default maxLevel with default ZoomButton.maxZoom

manisandro:
With the default settings, the zoom button can zoom further in than the available scales in the scalebox

### DIFF
--- a/web/client/components/mapcontrols/scale/ScaleBox.jsx
+++ b/web/client/components/mapcontrols/scale/ScaleBox.jsx
@@ -25,7 +25,7 @@ var ScaleBox = React.createClass({
     getDefaultProps() {
         return {
             id: 'mapstore-scalebox',
-            scales: mapUtils.getGoogleMercatorScales(0, 21),
+            scales: mapUtils.getGoogleMercatorScales(0, 28),
             currentZoomLvl: 0,
             onChange() {},
             readOnly: false,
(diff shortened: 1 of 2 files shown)


web/client/components/mapcontrols/scale/ScaleBox.jsx
```diff
@@ -25,7 +25,7 @@ var ScaleBox = React.createClass({
     getDefaultProps() {
         return {
             id: 'mapstore-scalebox',
-            scales: mapUtils.getGoogleMercatorScales(0, 21),
+            scales: mapUtils.getGoogleMercatorScales(0, 28),
```
mbarto: Can we make this a property, with a default of 21?

manisandro: As far as I'm concerned, sure, but it is desired that with the default behaviour the zoom buttons can zoom in further than the available scales in the scalebox? An effect of this is that when one zooms in to max with the buttons, the corresponding scale does not exist as an option in the scalebox combo and consequently the first item of the combo will be set as current, even if it mismatches the actual map scale.

mbarto: Ok, fine let's default to 28, just make it a property

manisandro: The property would need to influence

    scales: mapUtils.getGoogleMercatorScales(0, 28),

in `getDefaultProps` which I don't think is possible? I think that users wanting to customize this should rather simply pass a custom value for `scales` directly, i.e. `mapUtils.getGoogleMercatorScales(0, 21)`.

PR #1070, opened 2016-10-04
Fix #1065 password buttons. Fix #832 password fb

### LINKED ISSUE #832: Change password modal has a wrong button label and doesn't give feedback
The button to change the password in the change password modal says 'Reset Password' it should probably say something like "Change Password". After changing the password no feedback is given and the modal just stays open:

anim

Note that the password is actually updated.

We should also write down the rules for password validation somewhere (probably as a tooltip in the password textfield). Now it's not clear how to enter a valid password.

### LINKED ISSUE #1065: When logged in "Change Password" window renders incorrectly
See attached image.

offtherailz:
* Fixes #1065 fixing I18N references
 * Fixes #832 adding feedback while editing password.

This is the result: 

image

### DIFF
--- a/web/client/components/security/forms/PasswordReset.jsx
+++ b/web/client/components/security/forms/PasswordReset.jsx
@@ -14,7 +14,7 @@
   */
 
 const React = require('react');
-const {Input} = require('react-bootstrap');
+const {Input, Alert} = require('react-bootstrap');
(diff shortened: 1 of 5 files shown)


offtherailz:
Tests seems to pass. Probably a connection error. Close and reopen pull request to check

PR #1108, opened 2016-10-10
fix #1095 changed modal layout  for both on startup and empty results

### LINKED ISSUE #1095: Avoid resize of the catalog when empty content
When search has empty results, or on startup the catalog appears too small. We should improve this, maybe using modal dialog that centers it automatically. 
Modal dialog means to use `<Dialog modal={true ...` . 
We should set up a fixed width (relative to the window) and responsive layout to view the items at different sizes depending on the size of the screen.

saidaipparla:
catalog

### DIFF
--- a/web/client/plugins/MetadataExplorer.jsx
+++ b/web/client/plugins/MetadataExplorer.jsx
@@ -80,7 +80,7 @@ const MetadataExplorerComponent = React.createClass({
                         {panel}
                     </Panel>);
                 }
-                return (<Dialog id="mapstore-catalog-panel" style={this.props.style}>
+                return (<Dialog modal id="mapstore-catalog-panel" style={assign({}, this.props.style, {display: "block" })}>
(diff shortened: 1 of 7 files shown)


web/client/product/assets/css/viewer.css
```diff
@@ -391,10 +391,7 @@ html, body, #container, .fill {
 
 #mapstore-catalog-panel {
     width: auto;
-    position: absolute;
-    right: 200px;
-    margin-top: 100px;
-    display: inline-block;
+    margin: 100PX;
```
offtherailz: **not** capital px. The style should be responsive (if the width of the window is smaller, it should be 0, for instance).

web/client/plugins/MetadataExplorer.jsx
```diff
@@ -55,6 +55,7 @@ const MetadataExplorerComponent = React.createClass({
             id: "mapstore-metadata-explorer",
             active: false,
             wrap: false,
+            modal: true,
```
offtherailz: This have no effect, the modal parameter should be passed to the `Dialog` component.

saidaipparla:
screenshots of final result
bigscreen
withresult
mapstore 2 homepage
screencapture-localhost-8081-1476175141360
screencapture-localhost-8081-1476175163910

offtherailz:
@simboss What do you think? In my opinion we should put on the same line CSW/WMS sekectir and the URL (at least for larger screens)
Maybe we can make it smaller in wider screens.

PR #1119, opened 2016-10-10
Catalog status

saidaipparla:
added action  "RESET_STATUS"  when the Catalog Widget is closed

### DIFF
--- a/web/client/actions/catalog.js
+++ b/web/client/actions/catalog.js
@@ -15,6 +15,7 @@ const RECORD_LIST_LOADED = 'RECORD_LIST_LOADED';
 const RECORD_LIST_LOAD_ERROR = 'RECORD_LIST_LOAD_ERROR';
 const CHANGE_CATALOG_FORMAT = 'CHANGE_CATALOG_FORMAT';
 const ADD_LAYER_ERROR = 'ADD_LAYER_ERROR';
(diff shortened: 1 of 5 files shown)


web/client/product/assets/css/viewer.css
```diff
@@ -391,10 +391,7 @@ html, body, #container, .fill {
 
 #mapstore-catalog-panel {
     width: auto;
-    position: absolute;
-    right: 200px;
-    margin-top: 100px;
-    display: inline-block;
+    margin: 100PX;
```
offtherailz: This is still wrong case.

web/client/actions/catalog.js
```diff
@@ -15,6 +15,7 @@ const RECORD_LIST_LOADED = 'RECORD_LIST_LOADED';
 const RECORD_LIST_LOAD_ERROR = 'RECORD_LIST_LOAD_ERROR';
 const CHANGE_CATALOG_FORMAT = 'CHANGE_CATALOG_FORMAT';
 const ADD_LAYER_ERROR = 'ADD_LAYER_ERROR';
+const RESET_STATUS = 'RESET_STATUS';
```
offtherailz: This is too generic. You can call the variable RESET_STATUS but the string should be unique, lets say `CATALOG_RESET`

PR #1127, opened 2016-10-11
Fixed #1126 : click on import data reloads the imports list

### LINKED ISSUE #1126: Click on Import Data should reload the imports list
see https://github.com/geosolutions-it/npa-cgg/issues/117

MV88:
Now when you click on import data the imports list is reloaded coming back to the initial state.

### DIFF
--- a/web/client/actions/manager.js
+++ b/web/client/actions/manager.js
@@ -1,0 +1,20 @@
+/**
+ * Copyright 2016, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const ON_STARTUP = 'ON_STARTUP';
+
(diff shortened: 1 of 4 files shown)


web/client/actions/manager.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright 2016, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const ON_STARTUP = 'ON_STARTUP';
```
offtherailz: Too generic string, call it MANAGER_ITEM_SELECTED

web/client/actions/manager.js
```diff
@@ -0,0 +1,20 @@
...
+
+function onStartUp(toolId) {
```
offtherailz: The function should be called something like itemSelected

PR #1132, opened 2016-10-11
fixed #1128 : owner defined when a map is created by SaveAsPlugin

### LINKED ISSUE #1128: Permission editor not displayed for new maps
The property "owner" is not defined for the new map.
This, does not allow the user to access the permission editor form for the new map.

MV88:
Added the definition of the attribute owner


offtherailz:
Let's try if save plugin needs also these changes before merge.

MV88:
Actually the save Plugin updates only the data property of the Resource MAP.
So if the owner wasn't defined it remains undefined. 
Since this PR will be merged we can use the following instructions to display the permission editor for map's owner: 
- the owner should open the map (that is not displaying the permission editor)
- use the saveAs option (from burger menu) to duplicate the map (with a different name)
- go back to the home deleting the old map.

offtherailz:
Ok, let's keep this in mind.

PR #1136, opened 2016-10-12
fix #1109  Moved the visibility icon to correct place

### LINKED ISSUE #1109: Move the visibility icon (the eye) to the side of the tools for the layer in the layer tree
Right now the eye icon that allows users to change the visibility of the layer is in a position which is strange, see below:
clipboard021

I would rather have something like this:
clipboard031

which can be taken from the SIRA product we are doing.

saidaipparla:


offtherailz:
Going to close this pull request, please redo it from your fork

PR #1139, opened 2016-10-12
fixed #1138 : style names encoded

### LINKED ISSUE #1138: Importer: Style names need to be encoded
After some investigation we discovered that the name of style needs to be encoded before sending it to the geoserver's importer.
see also https://github.com/geosolutions-it/npa-cgg/issues/109

MV88:
I also fixed some minor bugs other than solving the issue #1138

### DIFF
--- a/web/client/components/file/FileUploader.jsx
+++ b/web/client/components/file/FileUploader.jsx
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 const React = require('react');
-
+const {round} = require('lodash');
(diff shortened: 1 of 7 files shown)


web/client/components/file/FileUploader.jsx
```diff
@@ -75,7 +75,7 @@ const FileUploader = React.createClass({
                     </thead>
                     <tbody>
                         {this.state.fileList && this.state.fileList.map((file) =>
-                        (<tr>
+                        (<tr key="row">
```
offtherailz: Wrong key, should be parametric

PR #1142, opened 2016-10-13
fixed #1141 : saveAs plugin is displayed with user allowedRoles

### LINKED ISSUE #1141: SaveAs Plugin should be configurable
We need to customize the SaveAs Plugin using the roles of the user.
Provide an array of allowed roles that is checked in order to display, or not, the plugin.

MV88:
This fix is needed to allow the administrator to enable or disable saveAsPlugin for normal and admin users

### DIFF
--- a/web/client/plugins/SaveAs.jsx
+++ b/web/client/plugins/SaveAs.jsx
@@ -20,6 +20,7 @@ const {editMap, updateCurrentMap, errorCurrentMap, resetCurrentMap} = require('.
 const {mapSelector} = require('../selectors/map');
 const stateSelector = state => state;
 const {layersSelector} = require('../selectors/layers');
+const {indexOf} = require('lodash');
 
 
 const selector = createSelector(mapSelector, stateSelector, layersSelector, (map, state, layers) => ({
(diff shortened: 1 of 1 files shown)


web/client/product/appConfig.js
```diff
@@ -46,6 +46,9 @@ module.exports = {
                 drawer: {
                     enabled: false,
                     menu: "1"
+                },
+                saveAs: {
+                    allowedRoles: ["ADMIN", "USER"]
```
offtherailz: avoid to add this to the default state of mapstore 2

web/client/plugins/SaveAs.jsx
```diff
@@ -172,7 +174,8 @@ module.exports = {
             text: <Message msgId="saveAs"/>,
             icon: <Glyphicon glyph="floppy-open"/>,
             action: editMap.bind(null, {}),
-            selector: (state) => (state && state.security && state.security.user ? {} : { style: {display: "none"} })
+            selector: (state) => (
+                indexOf(state && state.controls && state.controls.saveAs && state.controls.saveAs.allowedRoles, state && state.security && state.security.user && state.security.user.role) !== -1 ? {} : { style: {display: "none"} })
```
offtherailz:
```
if (state && state.controls && state.controls.saveAs && state.controls.saveAs.allowedRoles) {
 // filter by role
} else {
  /   check if the user is present
}

PR #1148, opened 2016-10-14
Fix #1134 user manager group editing issues

### LINKED ISSUE #1134: Assign groups do not work in user manager
When I try to assign a group to a user, this operation is not applied. 
Also enabled checkbox can not be updated.

saidaipparla:
Fix issues #1134 with usergroup updates. 
 - Fix issue introduced by changes on the catalog that influenced the delete user modal  
- Keep the search filter when update or delete user


simboss:
next time squash the commits.

Issue #1145, opened 2016-10-14
Catalog config is not mantained on reset


offtherailz:
When some customization is applied to the catalog, they will disappear on catalog close.

PR #1152, opened 2016-10-14
Fix #1151 added proper keys and property types.

### LINKED ISSUE #1151: Remove warnings in debug
Due to missing keys or wrong property checks

image

offtherailz:
- Fix #1151 add defaults for key when missing, fixed issues with property types.
- Fixed also persistance of catalog configuration #1145


offtherailz:
We should not merge this for the moment, some noticed some issues with chrome

PR #1157, opened 2016-10-17
Add removeLayer action

manisandro:
Adds an easy to use removeLayer action complementary to addLayer.


web/client/reducers/layers.js
```diff
@@ -264,6 +264,23 @@ function layers(state = [], action) {
                     groups: newGroups
             };
         }
+        case REMOVE_LAYER: {
+            let layer = (state.flat || []).find(obj => obj.id === action.layerId);
```
mbarto: I don't think find is supported by all browsers. You can use filter and lodash head function to get the same result

PR #1159, opened 2016-10-17
Allow appending to existing search results

manisandro:
This allows having multiple search result providers feed results into the state.


manisandro:
Argh typo,  I'll push-force an update.

manisandro:
-> done

PR #1160, opened 2016-10-17
Use prop instead of state for searchText

manisandro:
Perhaps I missed the reason for storing searchText in the component state,
but it would be convenient for us to have the searchText field in the global
redux state reflect the actual search text to correctly restore the search
text from the url via a `dispatch(searchTextChanged(text))` in `loadMapConfig`.

### DIFF
--- a/web/client/components/mapcontrols/search/SearchBar.jsx
+++ b/web/client/components/mapcontrols/search/SearchBar.jsx
@@ -57,14 +57,8 @@ let SearchBar = React.createClass({
             searchText: ""
         };
     },
-    getInitialState() {
-        return {
-            searchText: this.props.searchText || ""
-            };
-    },
     onChange() {
         var text = this.refs.input.getValue();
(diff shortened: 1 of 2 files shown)


mbarto:
I agree. Please look at the failing tests

manisandro:
I suspect what's going on is that now that searchText is controlled by the global state and not by the component state, its value fails to update since action + reducer are not invoked in the test. Is this a possibility?

mbarto:
Yes, it's possible, we probably need to think of a different set of tests

manisandro:
Well the existing actions `TEXT_SEARCH_TEXT_CHANGE` and `TEXT_SEARCH_RESET` handle the `searchText` state just fine. Do you mean these should be used in the test?

mbarto:
Oh, ok, so actions and reducers for text change already existed also if state was used for internal handling. Sorry about asking. I'll do a quick local test and merge.

PR #1161, opened 2016-10-17
Honour queryable field in layer info when bulding WMS getFeatureInfo request

manisandro:
In QWC2, the entire project is published as one WMS with individual layers in
the QGIS project handled as sublayers of this WMS layer. We keep track of the
individual sublayers by extending the WMS layer params object with the fields:

    sublayers: Array
    opacities: Array
    queryable: Array

This commit makes MapInfoUtils::buildIdentifyWMSRequest honour the queryable
array to exclude sublayers from the query which are flagged as non-identifyable.

### DIFF
--- a/web/client/utils/MapInfoUtils.js
+++ b/web/client/utils/MapInfoUtils.js
@@ -152,12 +152,16 @@ const MapInfoUtils = {
         const center = {x: lngCorrected, y: props.point.latlng.lat};
         let centerProjected = CoordinatesUtils.reproject(center, 'EPSG:4326', props.map.projection);
         let bounds = MapInfoUtils.getProjectedBBox(centerProjected, resolution, rotation, size, null);
+        let queryLayers = layer.name;
+        if (layer.queryLayers) {
+            queryLayers = layer.queryLayers.join(",");
+        }
 
         return {
             request: {
                 id: layer.id,
                 layers: layer.name,
-                query_layers: layer.name,
+                query_layers: queryLayers,
                 styles: layer.style,
                 x: ((widthBBox % 2) === 1) ? Math.ceil(widthBBox / 2) : widthBBox / 2,
                 y: ((widthBBox % 2) === 1) ? Math.ceil(widthBBox / 2) : widthBBox / 2,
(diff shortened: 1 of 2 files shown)


web/client/utils/MapInfoUtils.js
```diff
@@ -152,12 +152,16 @@ const MapInfoUtils = {
         const center = {x: lngCorrected, y: props.point.latlng.lat};
         let centerProjected = CoordinatesUtils.reproject(center, 'EPSG:4326', props.map.projection);
         let bounds = MapInfoUtils.getProjectedBBox(centerProjected, resolution, rotation, size, null);
+        let queryLayers = layer.name;
+        if (layer.queryable) {
```
mbarto: Is the queryable attribute of layers something custom to your application? I get it's an array of layers to be queried instead of the original layer. I think queryable is a misleading name. Usually queryable is a flag to distinguish layers that can be queries.

manisandro: It is custom in the sense that the upstream there is not support for WMS sublayer handling AFAICS. Say you have a WMS layer published as

    somelayer
       |--> sublayer1
       |--> sublayer2
       etc

If you pass `somelayer` as layer parameter in the WMS request, it means the main layer including all possible sublayers. You can however select only certain sublayers by passing instead a comma-separated list of sublayer names, i.e. `sublayer1,sublayer2`. QGIS server (and possibly other WMS servers) can publish in the capabilities whether individual sublayers are "queryable" aka identifyable.

As for the name "queryable", I have no problems renaming it to "identifyable" or something else.

mbarto: Yes, please rename to something like queryLayers and I think it's good. If it's not too complex, please also add a unit test

manisandro: Actually I just noticed when grepping for queryable: there already are traces of a `queryable` field in the layer object [1], but I don't see other traces of it. Should I rename it also?

[1] https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/MapInfoUtils.js#L223

mbarto: That's the flag I was referring too. It's coming from WMS capabilities to tell you if a layer is queryable or not.

manisandro: Ahh ok.

PR #1162, opened 2016-10-17
Don't assume that vector layer geometries are EPSG:4326

manisandro:
Allow to specify which CRS the feature geometries are specified in, instead of assuming EPSG:4326.


manisandro:
Leaflet support for rendering vector features seems missing altogether.

Not sure if the added test helps much, but i think the only other thing one could do is somehow access the native OpenLayers layer and check whether the coordinate conversions are actually correct...

PR #1170, opened 2016-10-18
fix #1107 added button for deleting layers in simple way

### LINKED ISSUE #1107: Simpler way to remove existing layers from the layer tree
The standard way for removing layers from the layer tree is too complex, see image below:

clipboard02

I would like to have something similar to what we have done for SIRA, se below:
clipboard03

clipboard06

saidaipparla:
deletelayer

### DIFF
--- a/web/client/components/TOC/DefaultLayer.jsx
+++ b/web/client/components/TOC/DefaultLayer.jsx
@@ -12,8 +12,10 @@ var VisibilityCheck = require('./fragments/VisibilityCheck');
 var Title = require('./fragments/Title');
 var InlineSpinner = require('../misc/spinners/InlineSpinner/InlineSpinner');
 var WMSLegend = require('./fragments/WMSLegend');
+const ConfirmButton = require('../buttons/ConfirmButton');
 const LayersTool = require('./fragments/LayersTool');
 const SettingsModal = require('./fragments/SettingsModal');
+const {Glyphicon} = require('react-bootstrap');
 
 var DefaultLayer = React.createClass({
     propTypes: {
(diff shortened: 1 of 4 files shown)


web/client/components/TOC/DefaultLayer.jsx
```diff
@@ -46,6 +49,7 @@ var DefaultLayer = React.createClass({
             propertiesChangeHandler: () => {},
             onToggle: () => {},
             onSettings: () => {},
+            activateRemoveLayer: true,
```
mbarto: We should configure this as false, like the other tools

web/client/components/TOC/DefaultLayer.jsx
```diff
@@ -61,6 +65,18 @@ var DefaultLayer = React.createClass({
     },
     renderTools() {
         const tools = [];
+        if (this.props.activateRemoveLayer) {
+            tools.push(
+                <ConfirmButton key="removelayer" className="clayer_removal_button"
+                    text={(<Glyphicon glyph="1-close" />)}
+                    style={{"float": "right", cursor: "pointer", backgroundColor: "transparent", marginRight: 3, padding: 0, outline: "none"}}
+                    confirming={{text: "Sei sicuro",
```
mbarto: This has to be localized

web/client/components/TOC/__tests__/DefaultLayer-test.jsx
```diff
@@ -133,7 +133,7 @@ describe('test DefaultLayer module component', () => {
         const tool = ReactDOM.findDOMNode(TestUtils.scryRenderedDOMComponentsWithClass(comp, "glyphicon")[1]);
```
mbarto: You need to use a different index here to get the right tool, in this way you are simply skipping all the tests that were done before

web/client/components/TOC/__tests__/DefaultLayer-test.jsx
```diff
@@ -158,11 +158,11 @@ describe('test DefaultLayer module component', () => {
...
```
mbarto: Same here

web/client/components/TOC/__tests__/DefaultLayer-test.jsx
```diff
@@ -239,10 +239,10 @@ describe('test DefaultLayer module component', () => {
         expect(tool).toExist();
```
mbarto: Same here

Issue #804, opened 2016-07-04
Print Preview window does not adapt correctly to resizing


simboss:
See images below:

on the desktop:

![clipboard02](https://cloud.githubusercontent.com/assets/693880/16567859/ab047600-4223-11e6-9024-545067bef9c7.jpg)

again on the desktop....

![clipboard03](https://cloud.githubusercontent.com/assets/693880/16569443/cdd003ee-4236-11e6-8c9f-4a8934d81cdc.jpg)

and on the mobile:

![screenshot_2016-07-04-22-33-19](https://cloud.githubusercontent.com/assets/693880/16569497/78bd0bbc-4237-11e6-913d-cbf44d04ab8e.jpg)

PR #1173, opened 2016-10-18
Responsiveness updates to the print modal window - issue #804

robertkirsz:
Fixed radio inputs’ margins and made modal window scrollable and it’s width contained inside the main window. For issue #804

### DIFF
--- a/web/client/product/assets/css/viewer.css
+++ b/web/client/product/assets/css/viewer.css
@@ -444,7 +444,7 @@ html, body, #container, .fill {
 
 .modal-body {
     background-color: white;
-    height: calc(100vh - 190px);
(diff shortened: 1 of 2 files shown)


web/client/product/assets/css/viewer.css
```diff
@@ -435,6 +444,8 @@ html, body, #container, .fill {
 
 .modal-body {
     background-color: white;
+    height: calc(100vh - 190px);
```
mbarto: This makes the dialog height bigger than the original dialog.

image

robertkirsz: It should be fixed in the latest commit, can you check it again?

PR #1192, opened 2016-10-24
Add possibility to pass custom options to search action

manisandro:
To implement a coordinate search which can search in the currently specified coordinate system, the search handler needs to know the current CRS.

### DIFF
--- a/web/client/components/mapcontrols/search/SearchBar.jsx
+++ b/web/client/components/mapcontrols/search/SearchBar.jsx
@@ -38,7 +38,8 @@ let SearchBar = React.createClass({
         blurResetDelay: React.PropTypes.number,
         typeAhead: React.PropTypes.bool,
         searchText: React.PropTypes.string,
-        style: React.PropTypes.object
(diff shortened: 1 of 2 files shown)


web/client/components/mapcontrols/search/SearchBar.jsx
```diff
@@ -38,7 +38,8 @@ let SearchBar = React.createClass({
         blurResetDelay: React.PropTypes.number,
         typeAhead: React.PropTypes.bool,
         searchText: React.PropTypes.string,
-        style: React.PropTypes.object
+        style: React.PropTypes.object,
+        displaycrs: React.PropTypes.string
```
mbarto: Are you using this in a custom search action? Probably we could just introduce a generic searchOptions object (where you can put crs and other stuff if you need them) to be passed through to the action. What do you think?

manisandro: Yep makes sense, done.

PR #1198, opened 2016-10-24
Add option to control layer insertion position in addLayer action

manisandro:
Not sure if this is the proper approach, but I didn't find any way to control the position a layer is added. By default it seems new layers are always added behind all non-background layers, but we need a way to add the new layer on top.
If this approach is correct I'll add tests.


manisandro:
So this means that the approach is ok?

mbarto:
The standard behaviour is to have layers ordered from bottom to top in the order they are listed in state.layers.flat. This order is created in LayersUtils.sortLayers by browsing the TOC so that layers that are at the bottom of the TOC are also at the bottom of the map stacking order. So, in theory, your approach should work.

PR #1206, opened 2016-10-25
Remove redundant vector layer feature initialization

manisandro:
- Transforming and setting the features in `map/openlayers/plugins/VectorLayer.js` is redundant and leads to features added directly to the openlayers layer which are unmanaged by react: the features are already managed in the wrapper class in `map/openlayers/Feature.jsx`
- The `this.layer` member in `map/openlayers/Layer.jsx` needs to be a state, otherwise the component isn't updated when the actual layer is constructed in `this.addLayer`, leading to `const children = layer ? [...] : null` in `render()` yielding `children = null` the very first time the layer component is rendered, because the layer wasn't yet constructed. Unless `layer` is part of the state, the component won't get refreshed when `layer` is initialized, meaning that the layer component will only be correctly rendered on the next action which triggers a repaint.

### DIFF
--- a/web/client/components/map/openlayers/plugins/VectorLayer.js
+++ b/web/client/components/map/openlayers/plugins/VectorLayer.js
@@ -106,23 +106,7 @@ var styleFunction = function(feature) {
 
 Layers.registerType('vector', {
     create: (options) => {
-        let features;
-        let featuresCrs = options.featuresCrs || 'EPSG:4326';
-        let layerCrs = options.crs || 'EPSG:3857';
-        if (options.features) {
-            let featureCollection = options.features;
-            if (Array.isArray(options.features)) {
-                featureCollection = { "type": "FeatureCollection", features: featureCollection};
-            }
-            features = (new ol.format.GeoJSON()).readFeatures(featureCollection);
-            if (featuresCrs !== layerCrs) {
-                features.forEach((f) => f.getGeometry().transform(featuresCrs, layerCrs));
-            }
-        }
-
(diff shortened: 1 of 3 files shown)


web/client/components/map/openlayers/plugins/VectorLayer.js
```diff
@@ -106,23 +106,7 @@ var styleFunction = function(feature) {
 
 Layers.registerType('vector', {
     create: (options) => {
-        let features;
```
mbarto: This is needed to allow configuring the layer via JSON instead of using Feature.jsx components

manisandro: Uhm, there needs to be a way to suppress this though, since when adding a vector layer to the layers state, Feature.jsx objects are created automatically. This leads to those Feature.jsx objects being correctly managed whereas those added directly are unmanaged, meaning they are not removed anymore when changing the layer properties (i.e. changing the `features` field) without removing and re-adding the layer. I don't know the JSON use-case, do you see a solution?

mbarto: I need to do some testing, I will let you know...

manisandro: Unfortunately I still end up with unmanaged features using latest master.

mbarto:
Fixed the openlayers plugin with https://github.com/geosolutions-it/MapStore2/pull/1793, pending review of leaflet: https://github.com/geosolutions-it/MapStore2/issues/1794

PR #1211, opened 2016-10-26
Also store pixel position in state.mousePosition

manisandro:
Useful i.e. to display a hover tooltip in an "on-the-fly" identify mode.

### DIFF
--- a/web/client/components/map/leaflet/Map.jsx
+++ b/web/client/components/map/leaflet/Map.jsx
@@ -268,7 +268,8 @@ let LeafletMap = React.createClass({
         this.props.onMouseMove({
             x: pos.lng,
             y: pos.lat,
-            crs: "EPSG:4326"
+            crs: "EPSG:4326",
+            pixel: event.containerPoint
         });
     },
     registerHooks() {
--- a/web/client/components/map/leaflet/Map.jsx
+++ b/web/client/components/map/leaflet/Map.jsx
@@ -99,8 +99,14 @@ let LeafletMap = React.createClass({
         this.map.on('singleclick', (event) => {
             if (this.props.onClick) {
                 this.props.onClick({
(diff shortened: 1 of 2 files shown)


web/client/components/map/leaflet/Map.jsx
```diff
@@ -268,7 +268,8 @@ let LeafletMap = React.createClass({
         this.props.onMouseMove({
             x: pos.lng,
             y: pos.lat,
-            crs: "EPSG:4326"
+            crs: "EPSG:4326",
+            pixel: event.containerPoint
```
mbarto: I would copy coordinates also in this case, instead of storing Leaflet objects

manisandro: That's the smaller part, the bigger part is figuring out how to write a useful test about this ;)

manisandro:
I haven't found a way to add a sensible test for this. I suppose a proper test would be simulating a mouse move event and checking that the i.e. mouseMove event handler gets called with object one expects. However I can't find a way to simulate a mousemove event, below does not work:

    it('check that mouse position information is stored on mouse move', () => {
        const testHandlers = {
            onMouseMove: (mousePosition) => { return mousePosition; }
        };
        let spy = expect.spyOn(testHandlers, 'onMouseMove');

        const map = ReactDOM.render(
            <LeafletMap id="mymap" center={{y: 40.0, x: 10.0}} zoom={10} onMouseMove={testHandlers.onMouseMove} />,
            document.getElementById("container")
        );
        expect(map).toExist();

        ReactTestUtils.Simulate.mouseMove(document.getElementById("container"), {clientX: 0, clientY: 0, screenX: 0, screenY: 0});
        expect(spy.calls.length).toBe(1);
    });

PR #1226, opened 2016-10-27
fix #1151 fixed some warnings

### LINKED ISSUE #1151: Remove warnings in debug

saidaipparla:


web/client/components/maps/MapGrid.jsx
```diff
@@ -90,7 +90,7 @@ var MapGrid = React.createClass({
     renderMetadataModal() {
         if (this.props.metadataModal) {
             let MetadataModal = this.props.metadataModal;
-            return (<MetadataModal ref="metadataModal" show={this.props.currentMap && this.props.currentMap.displayMetadataEdit} onHide={this.props.resetCurrentMap}
+            return (<MetadataModal key="metadataModal" show={this.props.currentMap && this.props.currentMap.displayMetadataEdit} onHide={this.props.resetCurrentMap}
```
mbarto: You cannot remove the ref, it's used

web/client/plugins/CreateNewMap.jsx
```diff
@@ -17,15 +17,15 @@ const CreateNewMap = React.createClass({
         mapType: React.PropTypes.string,
         onGoToMap: React.PropTypes.func,
         colProps: React.PropTypes.object,
-        isLoggedIn: React.PropTypes.bool
+        isLoggedIn: React.PropTypes.object
```
mbarto: bool is correct. You need to change the connect to set a bool value

web/client/plugins/CreateNewMap.jsx
```diff
@@ -17,15 +17,15 @@ const CreateNewMap = React.createClass({
...
     },
     contextTypes: {
         router: React.PropTypes.object
     },
     getDefaultProps() {
         return {
             mapType: "leaflet",
-            isLoggedIn: false,
+            isLoggedIn: "",
```
mbarto: false is correct

web/client/components/maps/MapGrid.jsx
saidaipparla: ok i will keep ref and add key

saidaipparla:
yes as far i have seen there are other two warnings  
- when we will click on  print link from menu
- when we will click on settings icon in layer properties

web/client/plugins/CreateNewMap.jsx
saidaipparla: ok i will change this

mbarto:
Ok, then fix also the two remaining warnings

web/client/plugins/CreateNewMap.jsx
```diff
@@ -51,6 +51,6 @@ const CreateNewMap = React.createClass({
 module.exports = {
     CreateNewMapPlugin: connect((state) => ({
         mapType: (state.maps && state.maps.mapType) || (state.home && state.home.mapType),
-        isLoggedIn: state && state.security && state.security.user
+        isLoggedIn: state && state.security && state.security.user && state.security.user.enabled
```
mbarto: isLoggedIn: state && state.security && state.security.user && true || false

web/client/localConfig.json
```diff
@@ -385,7 +385,7 @@
 					"closeGlyph": "1-close",
 					"submitConfig": {
 						"buttonConfig": {
-							"bsSize": "medium",
+							"bsSize": "small",
```
mbarto: Have you checked that the button doesn't change?

mbarto:
Is the "when we will click on settings icon in layer properties" still there?

web/client/localConfig.json
saidaipparla: yes i checked its changed

web/client/plugins/CreateNewMap.jsx
saidaipparla: ok i will change

PR #1228, opened 2016-10-27
Fix #1227 Styler - Set default position for components

### LINKED ISSUE #1227: Color Picker closes on manual editing
Color picker do not allow to edit manually the values in the styler. If you try, it closes.

offtherailz:
This prevent the close event to be called.


offtherailz:
It still has some problems after a combination of clicks. Let me check better.

offtherailz:
Ok, now everything works in local, with the latest version of react-color

PR #1237, opened 2016-11-02
fixed #814 : cleaned up the viewer css

### LINKED ISSUE #814: Some css for viewer is applied only on product
Some of the css that defines the tools is applied on the product version only. 
i.e. Button sizes, Back to home button style... 
These changes should be applied in any instance of the viewer. We have to investigate which of them can be considered part of the theme and which are part of mapstore's viewer, separate and include them in the proper components or in the mapstore theme.

MV88:
fix #814 
- Moved into related css files the part of the viewer css 
- Moved some css code in the theme


web/client/components/misc/spinners/GlobalSpinner/css/GlobalSpinner.css
```diff
@@ -0,0 +1,6 @@
+#mapstore-globalspinner  {
+    margin: 0 !important;
```
offtherailz: Are !important needed?

web/client/product/assets/css/viewer.css
```diff
@@ -14,62 +14,19 @@ html, body, #container, .fill {
...
 }
-#page-manager #home-button {
-    float: right;
-}
-.hidden {
-    display: none !important;
-}
 
-.react-draggable .panel-heading {
+.react-draggable .draggable-header {
```
offtherailz: This should be included in the theme

web/client/product/pages/Maps.jsx
```diff
@@ -7,7 +7,7 @@
  */
 const React = require('react');
 
-require('../assets/css/viewer.css');
+// require('../assets/css/viewer.css');
```
offtherailz: Why did you removed it, there are still some classes specific for the product viewer, isn't it?

MV88: is not removed, is just the diff messing up.

web/client/product/assets/css/viewer.css
MV88: will do

web/client/components/misc/spinners/GlobalSpinner/css/GlobalSpinner.css
MV88: We will remove them as soon as we will have a new theme to switch

web/client/product/pages/Maps.jsx
offtherailz: My fault, this is not needed in Maps.jsx, you can remove it

PR #1239, opened 2016-11-03
fix #1167 for Infinite loading on new layer creation

### LINKED ISSUE #1167: Infinite loading on new map creation
This issue appears only if the first login is done from inside a map.

Step to reproduce:
- Open MapStore2 homepage (do no login) -> a list of maps is displayed
- Open a map -> the map opens
- Login -> the map is still open, the User button is green
- Click on Home button -> Home page is visible, "New map" button is visible
- Click on "New Map" button -> the map page opens but the background never loads

I noticed that the url contains "/viewer/openlayers/new" even if Leaflet was selected in the home page

saidaipparla:


mbarto:
Replaced by https://github.com/geosolutions-it/MapStore2/pull/1240

PR #1262, opened 2016-11-11
Add support to Cross-Layer Filtering

offtherailz:
Added to FilterUtils the possibility to parse cross layer filtering

### DIFF
--- a/web/client/utils/FilterUtils.jsx
+++ b/web/client/utils/FilterUtils.jsx
@@ -488,39 +488,52 @@ const FilterUtils = {
         ogc += this.ogcSpatialOperator[this.objFilter.spatialField.operation].endTag;
         return ogc;
     },
+    /**
+    *  processOGCCrossLayerFilter(object)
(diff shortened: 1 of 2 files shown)


web/client/utils/FilterUtils.jsx
```diff
@@ -475,6 +488,41 @@ const FilterUtils = {
...
     },
+    processOGCCrossLayerFilter: function(crossLayerFilter) {
+        let ogc = this.ogcSpatialOperator[crossLayerFilter.operation].startTag;
+        ogc +=
+            this.propertyTagReference[this.nsplaceholder].startTag +
+                crossLayerFilter.attribute +
+            this.propertyTagReference[this.nsplaceholder].endTag;
+
+        switch (crossLayerFilter.operation) {
+            case "INTERSECTS":
```
mbarto: Can we simply do this for ANY operation?

offtherailz: sure

PR #1264, opened 2016-11-11
Ability to create permalink

mricca:


web/client/utils/ConfigUtils.js
```diff
@@ -325,6 +325,11 @@ var ConfigUtils = {
     },
     getConfigProp: function(prop) {
         return defaultConfig[prop];
+    },
+    setPermalinkLayersVisibility: function(originalLayers, newLayers) {
+        return originalLayers.map((originalLayer) => {
+            return assign({}, originalLayer, newLayers.find((layer) => originalLayer.id === layer ? originalLayer.visibility = true : originalLayer.visibility = false));
```
mbarto: No side effects please:
return assign(
        {}, 
        originalLayer, 
        {
              visibility: newLayers.filter((layer) => originalLayer.id === layer).length > 0 ? true : false
        }
);

PR #1273, opened 2016-11-15
Add action to change map rotation

manisandro:
Adds an action to programmatically change the map rotation. Only OpenLayers supports map rotation.

### DIFF
--- a/web/client/components/map/openlayers/Map.jsx
+++ b/web/client/components/map/openlayers/Map.jsx
@@ -37,7 +37,8 @@ var OpenlayersMap = React.createClass({
         changeMeasurementState: React.PropTypes.func,
         registerHooks: React.PropTypes.bool,
         interactive: React.PropTypes.bool,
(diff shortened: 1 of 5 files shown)


web/client/components/map/openlayers/Map.jsx
```diff
@@ -37,7 +37,8 @@ var OpenlayersMap = React.createClass({
         changeMeasurementState: React.PropTypes.func,
         registerHooks: React.PropTypes.bool,
         interactive: React.PropTypes.bool,
-        onInvalidLayer: React.PropTypes.func
+        onInvalidLayer: React.PropTypes.func,
+        bbox: React.PropTypes.object
```
mbarto: Since we don't use the bbox, apart from the rotation, I would only have a rotation property

mbarto: Ok, nice

PR #1280, opened 2016-11-16
fix #1257 added react-select.css in missing places

### LINKED ISSUE #1257: react-select.css is not included correctly
Should be included by all the plugins using it

saidaipparla:


saidaipparla:
yes but i have not added there its already included before and i think that file is also using react-select

PR #1284, opened 2016-11-17
Added optional customExport function to FeatureGrid

kappu72:


web/client/components/data/featuregrid/FeatureGrid.jsx
```diff
@@ -129,7 +130,12 @@ const FeatureGrid = React.createClass({
         }
 
         if (this.props.toolbar.exporter) {
-            tools.push(<button key="exporter" onClick={() => {this.api.exportDataAsCsv(); }}>
+            tools.push(<button key="exporter" onClick={() => {
+                if (this.props.exportAction) {
+                    this.props.exportAction();
+                } else {
+                    this.api.exportDataAsCsv();
```
mbarto: Can't we set this as a default property and simplify this code?

kappu72: Done!

PR #1289, opened 2016-11-22
Fix #1193. Implementation of Group Manager

### LINKED ISSUE #1193: Usergroup management
Allow the admin to crud the usergoups
This will need an update of GeoStore to implement group update

offtherailz:
The group manager is accessible from the "Manage Groups" button, as in this mockup:
http://lf.mapstore2.geo-solutions.it/samples/map/manager.html

It is implemented using the users look and feel.
Access from here: 
image
Then you can create, search, edit and delete groups.
image

### Create or Edit
image 
image

**Note**: Editing allows to change only the user members, not group name or description (GeoStore do not provide this functionality).

### delete
image


offtherailz:
Only one thing doesn't seems immediate for a user. 
The pages of the users and the groups are very similar. This means that the user can be confused. 

image
image

Should we indicate somewhere on the top that we are watching users (or groups) page? Like this:
image

offtherailz:
Just set the "new" buttons to green, both for user and group manager as it is in the mockup.
image

web/client/components/manager/users/GroupCard.jsx
```diff
@@ -0,0 +1,72 @@
...
+const GridCard = require('../../misc/GridCard');
+const {Button, Glyphicon} = require('react-bootstrap');
+const Message = require('../../../components/I18N/Message');
+
+
+// const ConfirmModal = require('./modals/ConfirmModal');
+
+require('./style/usercard.css');
+
+const UserCard = React.createClass({
```
mbarto: GroupCard

web/client/components/manager/users/GroupDialog.jsx
```diff
@@ -0,0 +1,228 @@
...
+  renderMembers() {
+      let members = this.getCurrentGroupMembers();
+      if (!members || members.length === 0) {
+          return (<div style={{
+                  width: "100%",
+                  textAlign: "center"
+              }}><Message msgId="usergroups.noUsers"/></div>);
+      }
+      // NOTE: faking group Id
+      return (<UsersTable users={members.sort((u1, u2) => u1.name > u2.name)} onRemove={(user) => {
```
mbarto: sort is mutating members, can we avoid it?

web/client/components/manager/users/UsersTable.jsx
```diff
@@ -0,0 +1,45 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const React = require('react');
+const {Button, Glyphicon, Table, OverlayTrigger, Tooltip} = require('react-bootstrap');
+const Message = require('../../I18N/Message');
+var UsersGrid = React.createClass({
```
mbarto: UsersTable

MV88:
Btw something that tell the user where he is should be added imho.

offtherailz:
About the title, it is ok?

groups-users-ms2

PR #1306, opened 2016-11-30
doclets test config file

saidaipparla:


.doclets.yml
```diff
@@ -0,0 +1,3 @@
+dir: web/client/components/
+articles:
+  - Zoom Button: buttons/ZoomButton/README.md
```
mbarto: We do not have such a file, link an existing md (like the README)

PR #1311, opened 2016-11-30
Fix #1310. Improved vector gfi with buffer

### LINKED ISSUE #1310: Feature info do not work with local shapefile imports for points and lines
This because it uses `inside` turf function, that works only with polygons.

offtherailz:
- Create a buffer in px around the clicked point and try to intersect the geometry (bufferized) to find out if it is clicked.
 - uses intersect instead of inside
 - Fixes max height for feature info window.


web/client/reducers/mapInfo.js
```diff
@@ -27,7 +27,7 @@ const {RESET_CONTROLS} = require('../actions/controls');
 
 const assign = require('object-assign');
 const {head} = require('lodash');
-const {inside} = require('turf');
+const {intersect, buffer} = require('turf');
```
mbarto: We should use:

turf/node_modules/turf-intersect

and 

turf/node_modules/turf-buffer

to avoid including the complete turf bundle

offtherailz: no `node-modules` in my `node_modules/turf` dir.

Probably because I'm using most recent npm: 

There are only package.json and `index.js`. The index.js has this content: 
`npm: '3.5.2',`
```
/*eslint global-require: 0*/

/**
 * Turf is a modular geospatial analysis engine written in JavaScript. It performs geospatial
 * processing tasks with GeoJSON data and can be run on a server or in a browser.
 *
 * @module turf
 * @summary Geospatial analysis for JavaScript
 */
module.exports = {
    isolines: require('turf-isolines'),
    [...]
    buffer: require('turf-buffer'),
    [...]
    intersect: require('turf-intersect'),
    [...]
    };

var helpers = require('turf-helpers');
[...]
module.exports.point = helpers.point;
module.exports.polygon = helpers.polygon;
module.exports.lineString = helpers.lineString;
[...]
```

They are only in your node modules, with an oldest node_modules. 

So I think the good approch is the mine one, that is compatible with both versions, and optimized for the newer, that will be used in the future. 

What do you think?

offtherailz:
Ok, now it is optimized and should work with every npm version.

PR #1317, opened 2016-12-02
Fix #1314 Support to print vector layer

### LINKED ISSUE #1314: PDF support for vector layers (leaflet)
Support vector data (e.g. shape files) to be exported in pdf too.

### LINKED ISSUE #1315: PDF support for vector layers (OpenLayers)
Support vector data (e.g. shape files) to be exported in pdf too.

offtherailz:
- Support vector style for leaflet layers
 - Conversion to OL2 style format
 - add reprojectGeoJson utility method to CoordinateUtils
 - Print preview support
 - Markers
 - ForceUpdate when layer's created (needed to display features on vector layer mount). 
 - Optimized leaflet layer rendering
This fixes #1315 too, because I created a style abstraction that can be used in every map.

### DIFF
--- a/web/client/components/map/leaflet/Layer.jsx
+++ b/web/client/components/map/leaflet/Layer.jsx
@@ -8,6 +8,7 @@
 var React = require('react');
 var Layers = require('../../../utils/leaflet/Layers');
 var assign = require('object-assign');
+var {isEqual} = require('lodash');
 
 const LeafletLayer = React.createClass({
     propTypes: {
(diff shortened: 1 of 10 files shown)


web/client/components/map/leaflet/Layer.jsx
```diff
@@ -45,6 +46,26 @@ const LeafletLayer = React.createClass({
         }
         this.updateLayer(newProps, this.props);
     },
+    shouldComponentUpdate(newProps) {
+        // the reduce returns true when a prop is changed
+        // optimizing when options are equal ignorning loading key
+        return !(["map", "type", "srs", "position", "zoomOffset", "onInvalid", "onClick", "options"].reduce( (prev, p) => {
```
mbarto: Do  you have a use case where the layer was slow without this?

offtherailz: I noticed an increase of performances during debug, with many features in the map.

offtherailz: Honestly, I did not tested if it is perceptible in the normal application, but when you have both map and print preview (doubling in fact the number of features) it can become slow.

offtherailz: Can I go on with merge?

PR #1327, opened 2016-12-06
Fix #1325. Get only main map svg

### LINKED ISSUE #1325: Snapshot tool includes an orange square if leaflet map is used
See below some snapshots: the orange square should not be in the snapshot.
OpenLayers map is not affected by this bug.

snapshot
snapshot_2

offtherailz:
This retrieves only the correct svg, The minimap is contained in the controller div instead.


offtherailz:
close and reopen, locally test are passing both on chrome and firefox.

offtherailz:
Got it. Firefox 31 do not support query selectors v2 (the querySelector with `:scope` ).
see https://www.w3.org/TR/selectors-api2/

PR #1337, opened 2016-12-07
Intl not defined

manisandro:
On Firefox mobile our application fails to load with `Intl is not defined`. The suggested upstream fix [1] solved the issue.

[1] https://github.com/yahoo/react-intl/issues/17#issuecomment-56984521

### DIFF
--- a/web/client/components/I18N/Message.jsx
+++ b/web/client/components/I18N/Message.jsx
@@ -15,7 +15,7 @@ const FormattedMessage = ReactIntl.FormattedMessage;
 
 const LocaleUtils = require('../../utils/LocaleUtils');
 
-var Message = React.createClass({
+const Message = React.createClass({
     propTypes: {
         locale: React.PropTypes.string,
         messages: React.PropTypes.object,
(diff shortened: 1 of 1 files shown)


Gnafu:
Latest Firefox mobile (50.0.2) runs MapStore2 just fine, where does the issue appear?

manisandro:
True, upstream MapStore2 runs fine, but our downstream product fails... I suppose it is some unlucky load-ordering issue, and likely the fact that we don't load certain upstream plugins also plays a role.

mbarto:
This is done here for the MapStore2 product: https://github.com/geosolutions-it/MapStore2/blob/master/web/client/product/app.jsx

If you use a similar approach in your app, it should work. In particular this is the useful part:

if (!global.Intl ) {
    // Ensure Intl is loaded, then call the given callback
    LocaleUtils.ensureIntl(startApp);
}else {
    startApp();
}

manisandro:
Ok, I'll handle it in the app.

PR #1339, opened 2016-12-08
Add point measurement support

manisandro:
Adds a measurement mode for measuring point coordinates.

### DIFF
--- a/web/client/components/map/leaflet/MeasurementSupport.jsx
+++ b/web/client/components/map/leaflet/MeasurementSupport.jsx
@@ -7,6 +7,7 @@
  */
 
 const React = require('react');
+const assign = require('object-assign');
 var L = require('leaflet');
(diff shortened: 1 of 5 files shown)


web/client/components/map/leaflet/MeasurementSupport.jsx
```diff
@@ -105,6 +114,7 @@ const MeasurementSupport = React.createClass({
                 areaMeasureEnabled: this.props.measurement.areaMeasureEnabled,
                 bearingMeasureEnabled: this.props.measurement.bearingMeasureEnabled,
                 geomType: this.props.measurement.geomType,
+                point: null, // handled in onDraw.created
```
mbarto: Probably we also need a pointMeasureEnabled: ... here?

manisandro: Indeed, commit amended. BTW I'll have a followup PR which will clean up the code a bit.

PR #1343, opened 2016-12-12
fix #1331 added proxy call instead of original

### LINKED ISSUE #1331: PDF preview doesn't work without CORS headers on server
It looks like the final pdf file is loaded without a proper proxy, so it doesn't work if CORS headers are not enabled on geoserver.

saidaipparla:


web/client/plugins/print/index.js
```diff
@@ -7,7 +7,7 @@
 */
 const {connect} = require('react-redux');
 const {compose} = require('redux');
-
+var ConfigUtils = require('../../utils/ConfigUtils');
```
mbarto: const ConfigUtils

web/client/plugins/print/index.js
```diff
@@ -128,7 +128,7 @@ const PrintSubmit = connect((state) => ({
 }))(require('../../components/print/PrintSubmit'));
 
 const PrintPreview = connect((state) => ({
-    url: state.print && state.print.pdfUrl,
+    url: state.print && ConfigUtils.getProxyUrl(state.print.pdfUrl),
```
mbarto: You should use the proxied url only if it's not from the sameOrigin. Look here: https://github.com/geosolutions-it/MapStore2/blob/master/web/client/libs/ajax.js#L155

In addition: ConfigUtils.getProxyUrl(...) can return an object (look at the code linked above), you should take care of this. The best option would be to extract the code in ajax.js that handles proxyUrl, sameOrigin, useCORS and put it in a utility method.

mbarto:
Replaced by #1347

PR #1350, opened 2016-12-13
Also update measurements on mouse move

manisandro:
Also update the measurement values while moving the mouse. Includes some code cleanups along the way.


manisandro:
Uhm, it's a bit hard to inject properties into the MeasurementSupport components from the global config. Or shall I abuse `state.measurement` for this (i.e. state.measurement.updateOnMousemove)?

mbarto:
You should be able to use the Map plugin toolsOptions property to pass down properties to the tool

manisandro:
Ah missed that one, I'll have a look, thanks.

PR #1354, opened 2016-12-16
Fix issues in access manager

offtherailz:
This issue happens when trying to digit in the access manager
nimbus-record-video 1 convert-video-online com

 - Due to the update, we have to apply this fix to the RuleAttributeSelect to re-enable inline search

### DIFF
--- a/web/client/components/manager/rulesmanager/RuleAttributeSelect.jsx
+++ b/web/client/components/manager/rulesmanager/RuleAttributeSelect.jsx
@@ -41,7 +41,7 @@ const RuleAttributeSelect = React.createClass({
     },
     getDefaultProps() {
         return {
-            loadOptions: () => [],
+            loadOptions: () => {},
             onInputChange: () => {},
             onValueUpdated: () => {},
             options: [],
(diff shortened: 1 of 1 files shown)


web/client/components/manager/rulesmanager/RuleAttributeSelect.jsx
```diff
@@ -43,7 +43,7 @@ const RuleAttributeSelect = React.createClass({
         return {
             loadOptions: () => [],
```
mbarto: I would also fix this one

web/client/components/manager/rulesmanager/RuleAttributeSelect.jsx
```diff
@@ -114,9 +114,12 @@ const RuleAttributeSelect = React.createClass({
                     }
                     this.props.loadOptions();
                 }}
+                searchable={true}
+                matchPos="any"
+                matchProp="any"
                 clearable={this.props.clearable}
                 onInputChange={(input) => {
-                    if (!this.props.staticValues) {
+                    if (!this.props.staticValues && (!options && !this.props.paginated)) {
```
mbarto: I don't see when options could be falsy

offtherailz: I see, My fault, simply we should not display loading, because this when the operation is not async, (most of the times) this make the filtering do not work. Going to fix it too

PR #1364, opened 2016-12-21
fix #1300 mouseposition,map plugin doclets-doc

### LINKED ISSUE #1300: Document existing plugins using docma
We should document the existing plugins, for each of them:
 * available configuration properties, with explanation of usage
 * available containers (Toolbar, DrawerMenu, etc.) and related configuration options
 * styles / css
 * examples of usage

This should be done using proper JSDOC syntax (see http://usejsdoc.org/howto-commonjs-modules.html)
 * [ ] BackgroundSwitcher (https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/BackgroundSwitcher.jsx)
 * [ ] Identify (https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/Identify.jsx)
 * [ ] Map (https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/Map.jsx)

saidaipparla:


package.json
```diff
@@ -83,6 +83,7 @@
     "html2canvas": "0.5.0-beta4",
     "intl": "1.2.2",
     "ismobilejs": "0.4.0",
+    "jsdoc": "^3.4.3",
```
offtherailz: devDependencies for this

package.json
```diff
@@ -134,10 +135,10 @@
     "redux-undo": "0.5.0",
     "reselect": "2.5.1",
     "shpjs": "3.3.2",
+    "turf-bbox": "3.0.10",
```
offtherailz: Don't move up and down dependencies

web/client/plugins/index.js
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2016, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+/**
+* @overview list of plugins used for mapstore2
+* @name plugins
+* @module plugins
```
offtherailz: This do not include anything. It's confusing. We should include the available plugins.

The reason why the plugins included in the product is to optimize the bundle with only the plugins included. 
Here we can link the available plugins with links to the documentation of each of them.

package.json
saidaipparla: those are automatically done when i install jsdoc plugin :)

offtherailz:
Also link your version of docklets.io and details the choices we did for this proposal.

saidaipparla:
here is the latest version of doclet-docs
https://doclets.io/saidaipparla/MapStore2/doclets_test

PR #1377, opened 2017-01-10
Fix infinite loop when query panel is closed.

offtherailz:
This error happens when closing the query panel (in debug mode) .

image
This is due to the onClick event bind directly to the action, passing the wrong parameters (the mouseEvent instead of the search URL). 
In debug this causes an immutable error, in general the state becomes not consistent anymore.

### DIFF
--- a/web/client/plugins/Map.jsx
+++ b/web/client/plugins/Map.jsx
@@ -33,7 +33,7 @@ const MapPlugin = React.createClass({
     },
     getDefaultProps() {
         return {
-            mapType: 'cesium',
+            mapType: 'leaflet',
             actions: {},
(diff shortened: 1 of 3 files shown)


web/client/plugins/Map.jsx
```diff
@@ -33,7 +33,7 @@ const MapPlugin = React.createClass({
     },
     getDefaultProps() {
         return {
-            mapType: 'leaflet',
+            mapType: 'cesium',
```
mbarto: why?

offtherailz: didn't wanted to commit that. sorry, I'm going to revert this change

PR #1388, opened 2017-01-19
Added infinite nesting to TOC

kappu72:


web/client/utils/LayersUtils.js
```diff
@@ -33,58 +53,53 @@ var LayersUtils = {
...
     reorder: (groups, allLayers) => {
         return allLayers.filter((layer) => layer.group === 'background')
-            .concat(initialReorderLayers(groups, allLayers));
+            .concat(reorderLayers(groups, allLayers));
     },
     denormalizeGroups: (allLayers, groups) => {
-
         let getGroupVisibility = (nodes) => {
-            let visibility = true;
+            let visibility = false;
```
mbarto: I don't think this refactor is correct. The group visibility is true only if every node is visible

web/client/utils/LayersUtils.js
```diff
@@ -9,21 +9,41 @@
 const assign = require('object-assign');
 const {isObject, isArray} = require('lodash');
 
+const getGroup = (groupId, groups) => {
+    return groups.filter((subGroup) => isObject(subGroup) && subGroup.id === groupId).pop();
```
mbarto: use lodash head to get first element of array, not pop

web/client/reducers/layers.js
```diff
@@ -204,7 +149,29 @@ function layers(state = [], action) {
             });
             let originalNode = head(flatLayers.filter((layer) => { return (layer[selector] === action.node || layer[selector].indexOf(action.node + '.') === 0); }));
             if (!sameGroup && originalNode ) {
-                let newGroups = moveNode(state.groups, action.node, (action.options.group || 'Default'));
+                // Remove layers from old group
```
mbarto: Why did you remove moveNode?

web/client/utils/LayersUtils.js
kappu72: Fixed

web/client/utils/LayersUtils.js
kappu72: fixed

web/client/reducers/layers.js
kappu72: Fixed

web/client/utils/LayersUtils.js
kappu72: The id is autogenerated and is the same then before, are name and title that change.

web/client/components/TOC/DefaultLayerOrGroup.jsx
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright 2016, GeoSolutions Sas.
```
kappu72: fixed

web/client/components/TOC/fragments/GroupTitle.jsx
```diff
@@ -31,7 +31,7 @@ const GroupTitle = React.createClass({
         let expanded = (this.props.node.expanded !== undefined) ? this.props.node.expanded : true;
         let groupTitle = this.props.node && this.props.node.title || 'Default';
         return (
-            <div className="toc-group-title" onClick={() => this.props.onClick(this.props.node.name, expanded)} style={this.props.style}>
+            <div className="toc-group-title" onClick={() => this.props.onClick(this.props.node.id, expanded)} style={this.props.style}>
```
kappu72: Yes,  group id is mandatory, but is autogenerated in LayersUtils.getLayersByGroup.

web/client/reducers/layers.js
```diff
@@ -32,79 +34,23 @@ const deepChange = (nodes, findValue, propName, propValue) => {
...
-};
-
 const getNode = (nodes, name) => {
     if (nodes && isArray(nodes)) {
         return nodes.reduce((previous, node) => {
             if (previous) {
                 return previous;
             }
-            if (node && node.name === name) {
+            if (node && (node.name === name || node.id === name)) {
```
kappu72: fixed

web/client/reducers/layers.js
```diff
@@ -116,6 +62,32 @@ const getNode = (nodes, name) => {
...
+    // Remove node from old group
+    let newGroups = deepRemove(groups, node);
+    // Check if group to move to exists
+    let group = getNode(newGroups, groupId);
+    if (!group) {
+        // Create missing group
+        group = head(LayersUtils.getLayersByGroup([getNode(newLayers, node)]));
+        // check for parent group if exist
+        const parentGroup = groupId.split('.').reduce((tree, gName, idx) => {
+            const gId = groupId.split(".", idx + 1).join('.');
```
kappu72: parentGroup is the object parentGroup.id is the parentId

offtherailz:
Should we use `/` instead of `.` as a separator for groups? Does it make sense?
As a plus we should allow to escape the group separator somehow

PR #1396, opened 2017-01-24
fix #1326 : Added limits for group name and desc and a message

### LINKED ISSUE #1326: Groups cannot be created if the entered description is too long
In group manager, if the users inputs a long description the save can fail, so the error message appears:  

> "There was an error saving this groupInternal Server Error"

Is there a way to display the user the reason of the failure (description too long) ?

MV88:
fixed #1326
- provided a maxLength to the input field
- provided a localized message when the limit is reached for both the name and the description
- Header now display a long message correctly if there are some spaces

### DIFF
--- a/web/client/components/manager/users/GroupDialog.jsx
+++ b/web/client/components/manager/users/GroupDialog.jsx
@@ -44,6 +44,7 @@ const GroupDialog = React.createClass({
       closeGlyph: React.PropTypes.string,
       style: React.PropTypes.object,
       buttonSize: React.PropTypes.string,
+      descLimit: React.PropTypes.number,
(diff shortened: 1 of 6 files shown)


web/client/components/manager/users/GroupDialog.jsx
```diff
@@ -56,6 +57,7 @@ const GroupDialog = React.createClass({
           options: {},
           useModal: true,
           closeGlyph: "",
+          descLimit: 255,
```
mbarto: This is not used. You repeat 255 everywhere

MV88: I tried to use it but it was undefined and wasn't working

PR #1412, opened 2017-01-27
Fix 1144 Feature info allowed options should be a whitelist

allyoucanmap:
Replaced exclude options with include options on GetFeatureInfo WMS request

### DIFF
--- a/web/client/localConfig.json
+++ b/web/client/localConfig.json
@@ -281,13 +281,7 @@
             "continuous": false
           },
           "collapsible": false
-        },
-        "includeOptions": [
-            "buffer",
-            "cql_filter",
-            "filter",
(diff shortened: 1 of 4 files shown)


web/client/localConfig.json
```diff
@@ -282,23 +282,11 @@
...
-            "legend",
-            "statistics",
-            "columnDefs",
-            "advancedFilter",
-            "boundingBox",
-            "capabilities",
-            "describeCoverage",
-            "describeLayer",
-            "loading"
+        "includeOptions": [
```
offtherailz: This options should be the defaults, so they should not be present in the localConfig.json, but overridable in this way if needed.

PR #1426, opened 2017-02-08
Fix #861 Introduce wizard-like help system

### LINKED ISSUE #861: Introduce wizard-like help system
We should introduce a wizard like help system taking extra care that it will work nicely on the mobile where it will be the default help system.

allyoucanmap:
* added react-joyride library to controll the steps of tutorial
* added tutorial plugin | autostart or start from burger menu tutorial button
* added new ids to some components to target with joyride
* added tests for the new actions and reducer

main options in local configuration ("name": "Tutorial"):
- showCheckbox: true/false, shows or not the checkbox on intro to disable the tutorial autostart
- rawSteps: array of step objects

Step with intro selector represents the introduction and if it exists must be the first of rawSteps.
If there isn't introduction, auto start is disabled and the tutorial can be activated from the burger menu.

standard step object:

```
{
  "translation": "intro", // add a reference in the traslation file
  "selector": "intro", // could be a presetted selector (plugin/tutorial/selectors.js) or a custom id (#my-id)
  "position": "top" // position of the popup || "bottom" default
}
```

in translation file

```
"tutorial": {
   ... ,
  "intro": {
    "title": "Welcome on MapStore2",
    "text": "My description"
  }
}
```


web/client/localConfig.json
```diff
@@ -100,6 +100,29 @@
           }
         }
     }, {
+        "name": "Tutorial",
+        "cfg": {
+          "rawSteps": [
```
offtherailz: This should be not here by default, only if you want to change the steps

web/client/localConfig.json
```diff
@@ -100,6 +100,29 @@
...
+            {
+              "translation": "intro",
+              "selector": "intro"
```
offtherailz: You should place here the selector, and not a reference to it

web/client/localConfig.json
```diff
@@ -100,6 +100,29 @@
...
+            },
+            {
+              "translation": "lorem",
```
offtherailz: In the default configuration we should already have some minimal introduction, we can customize and extend later, not a lorem ipsum

web/client/localConfig.json
```diff
@@ -326,6 +349,33 @@
           }
         }
     }, {
+        "name": "Tutorial",
+        "cfg": {
+          "rawSteps": [
```
offtherailz: same here, remove cfg, use defaults for localConfig

web/client/localConfig.json
```diff
@@ -787,7 +837,38 @@
       {"name": "MapType"},
       {"name": "CreateNewMap"},
       {"name": "Maps"},
-      {"name": "Examples"}
+      {"name": "Examples"},
+      {
```
offtherailz: If you want 2 different default configurations for home and map, use something a "preset": <home>|<map> you can easily override with `localConfig.json`.
You can save that preset as separate files in the plugin folders. rawSteps configuration will override the steps in the preset.

web/client/plugins/Tutorial.jsx
```diff
@@ -0,0 +1,221 @@
...
+const defaultStep = {
+    title: '',
+    text: '',
+    selector: '',
+    position: 'bottom',
+    type: 'click',
+    allowClicksThruHole: true
+};
+
+const TutorialComponent = React.createClass({
```
offtherailz: This should be a separate component. To test.

web/client/plugins/tutorial/selectors.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright 2017, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const selectors = {
```
offtherailz: Remove this file, the selectors goes directly in the rawSteps array. Instead, create presets for home and map to use. 
So in your localConfig.json you will have something like` {name: "Tutorial", preset: "home"}` or ` {name: "Tutorial", preset: "map"}`

allyoucanmap: Could it be like this? Or we always have to declare the preset

preset home
```
{
 "name": "Tutorial",
 "cfg": {
  "preset": "home"
 }
}
```

preset map 

```
{
 "name": "Tutorial"
}
```

offtherailz: Yes, good

allyoucanmap:
there are plugin that disappear or change their structure based on events.
an example is the search bar

error0

error1

this doesn't permit to the library react-joyride to mount the popup to the target,
so I added a div with the id "error-tutorial" as target for every step with error.
When the user close the tutorial, every error detected is deleted.

error2

error3

web/client/actions/tutorial.js
```diff
@@ -0,0 +1,175 @@
...
+function setupTutorial(introSteps) {
+    return (dispatch, getState) => {
+        var state = getState();
+        var steps = assign([], introSteps);
+        var hasIntro = false;
+        var disabled = localStorage.getItem('mapstore.plugin.tutorial.disabled');
+        if (state.controls && state.controls.tutorial) {
+            dispatch(setControlProperty('tutorial', 'enabled', false));
+        }
+        steps.map((step) => {
```
mbarto: do not use map with side effects

mbarto: either use forEach or (better) reduce

web/client/actions/tutorial.js
```diff
@@ -0,0 +1,175 @@
...
+function changeStatusTutorial(status) {
+    return {
+        type: CHANGE_STATUS_TUTORIAL,
+        status: status
+    };
+}
+
+function setupTutorial(introSteps) {
+    return (dispatch, getState) => {
+        var state = getState();
```
mbarto: Please, never use var, either use const (preferred) or let

web/client/actions/tutorial.js
```diff
@@ -0,0 +1,175 @@
...
+
+function changeStatusTutorial(status) {
+    return {
+        type: CHANGE_STATUS_TUTORIAL,
+        status: status
+    };
+}
+
+function setupTutorial(introSteps) {
+    return (dispatch, getState) => {
```
mbarto: Using getState in actions is usually evil: reducers are the place where we manipulate state

web/client/actions/tutorial.js
```diff
@@ -0,0 +1,175 @@
...
+        return step.selector !== '#intro-tutorial';
+    });
+    dispatch(setStepsTutorial(steps));
+    if (tour.action !== 'next') {
+        dispatch(closeTutorial());
+    }
+}
+
+function updateTutorial(tour, updateSteps, error) {
+    return (dispatch, getState) => {
```
mbarto: This is a reducer, not an action

web/client/actions/tutorial.js
```diff
@@ -0,0 +1,175 @@
...
+            dispatch(changeStatusTutorial(tour.type));
+        }
+    };
+}
+
+function disableTutorial() {
+    return (dispatch, getState) => {
+        dispatch(toggleTutorial());
+        let state = getState();
+        let disabled = state && state.tutorial && state.tutorial.disabled;
```
mbarto: This is a reducer, not an action

web/client/actions/tutorial.js
```diff
@@ -0,0 +1,175 @@
...
+        localStorage.setItem('mapstore.plugin.tutorial.disabled', disabled);
+    };
+}
+
+function resetTutorial() {
+    return (dispatch, getState) => {
```
mbarto: This is a reducer, not an action

web/client/components/tutorial/Tutorial.jsx
```diff
@@ -0,0 +1,190 @@
...
+                onSetup: () => {},
+                onStart: () => {},
+                onUpdate: () => {},
+                onToggle: () => {},
+                onReset: () => {}
+            }
+        };
+    },
+    componentWillMount() {
+        let rawSteps = this.props.rawSteps.length > 0 ? this.props.rawSteps : this.props.presetList[this.props.preset] || [];
```
mbarto: We try to not use lifecycle methods, if not necessary. This is state initialization code, I presume so should be developed as actions / reducers code

web/client/components/tutorial/style/style.js
```diff
@@ -0,0 +1,149 @@
+/**
+ * Copyright 2017, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const introStyle = {
```
mbarto: Can this be just css?

allyoucanmap: We could use css to override the default style but we still need the style to change the tooltips of intro and error. Could we remove the style.js file and include the const introStyle and errorStyle in the Tutorial.jsx component?

offtherailz:
Looks very good. We should disable the intro until we have a stable version to show up. Please document the preset format.

PR #1439, opened 2017-02-10
#1438 Have binary package filename in accordance to release number

fernandinand:


Gnafu:
I see this will be built with:

> mvn clean install -Drelease.number=${branch} -P release

Looks good to me

PR #1446, opened 2017-02-13
Fix #1445. Configure travis to latest firefox version

### LINKED ISSUE #1445: Update Firefox version on travis
We can set the firefox version to use in travis by [changing travis configuration](https://muffinresearch.co.uk/using-the-latest-version-of-firefox-in-travis-ci/):

By default version 31 of firefox is used (2014). We should use the latest or at least a more recent version.

offtherailz:
This change configures trafis to use the latest firefox version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: node_js
 node_js:
   - 4
 script: npm run travis
+addons:  
+  firefox: latest
 before_install:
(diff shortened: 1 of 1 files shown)


offtherailz:
The previous fail is seems to be due to a recent change, not related with this change. Let's verify latest commits.

PR #1473, opened 2017-02-16
Fix #1460 measure panel and tooltip translations (#1467)

### LINKED ISSUE #1460: Measure panel in Drawer is not translated (immediately) when switching language
It happened to other components in the past. It's updated on the next action.

MV88:
* fixed measure panel and tooltip translations
* added a test for measurementSupport for leaflet


MV88:
no longer needed

Issue #1475, opened 2017-02-16
Investigate on react-docgen to create reference docs


mbarto:
[react-docgen](https://github.com/reactjs/react-docgen) is a library to create docs from react components in JSON format.

Other tools exist to create markdown from this JSON:
 * https://github.com/travelperk/react-docgen-markdown
 * https://github.com/IonicaBizau/json2md
 * https://github.com/thebuilder/react-docs-markdown
 * https://github.com/morlay/react-docgen-ui

We have to create a workflow to:
 * generate reference docs for our plugins (and components)
 * save them in MD format, in the docs/developer-guide/reference folder so that they become part of the github pages website
 * create an index of files for the docs/developer-guide/reference folder

Let's start documenting ONE plugin properly (e.g. MousePosition) and creating docs for it

offtherailz:
We may check docma instead

PR #1493, opened 2017-02-21
#1475 react-docgen node script

fernandinand:
The script takes as input a given plugins/components folder and a target output folder for generating a MD documentation file for each doc-gen ready jsx file recursively.
e.g. ```node genDocs web/client/components docs/developer-guide/reference```


genDocs
```diff
@@ -0,0 +1,81 @@
...
+  // FIXME: static output folder?
+  fs.writeFile('docs/developer-guide/reference/index.md', tocContent, (err) => {
+      if (err) throw err;
+    });
+}
+
+function checkExt(file, ext) {
+    if (path.extname(file) == ext) {
+        //filelist.concat(file);
+        const docgen = spawn('react-docgen', [ file ]);
```
mbarto: Why do we spawn an external process, we already have the react parser in memory with require, we should use that (by the way, this doesn't work on Windows)?

genDocs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+function buildDocs(doc, file) {
+  // Extract filename as component name
+  var name = getComponentName(file);
+  // Use fs to write the markdown to dist
+  var md = docsToMarkdown(doc, name);
+  // FIXME: static output folder?
+  fs.writeFile('docs/developer-guide/reference/' + name +'.md', md, (err) => {
```
mbarto: If reference folder does not exist (actual case) it breaks, we should create not existing folders

package.json
```diff
@@ -114,6 +117,8 @@
     "react-copy-to-clipboard": "4.1.0",
     "react-dnd": "2.1.3",
     "react-dnd-html5-backend": "2.1.2",
+    "react-docgen": "2.13.0",
```
mbarto: these should go in the devDependencies section

offtherailz:
We decided to use docma instead, going to close.

PR #1495, opened 2017-02-22
Fixes #1006: migrate react-intl to 2.x

### LINKED ISSUE #1006: Migrate to React-Intl 2.x.x
The new version implements something similar to our Localized component. The migration should refactor Localized, Message and the other I18N components (and LocaleUtils also)  to wrap the new react-intl functionality. Some work has already been done here: https://github.com/mbarto/MapStore2/tree/react_015

mbarto:


web/client/components/I18N/HTML.jsx
```diff
@@ -11,23 +11,14 @@ var FormattedHTMLMessage = ReactIntl.FormattedHTMLMessage;
...
     },
     render() {
-        var locale = this.props.locale || this.context.locale;
-        var messages = this.props.messages || this.context.messages;
-        let message = messages;
-        this.props.msgId.split('.').forEach(part => {
-            message = message[part];
-        });
-        return <FormattedHTMLMessage locales={locale} message={message} {...this.props.msgParams}/>;
+        return this.context.intl ? <FormattedHTMLMessage id={this.props.msgId} values={this.props.msgParams}/> : <span>{this.props.msgId || ""}</span>;
```
offtherailz: id should not be used here, we can have duplicated htmlMessages with the same msgId

web/client/components/I18N/__tests__/18N.HTML-test.jsx
```diff
@@ -57,7 +58,7 @@ describe('This test for I18N.HTML', () => {
         var currentData = data["it-IT"];
         var testMsg = currentData.messages[msgId];
 
-        const cmp = ReactDOM.render(<I18N.HTML msgId={msgId} messages={ita.messages} locale="it-IT"/>, document.getElementById("container"));
+        const cmp = ReactDOM.render(<Localized messages={ita.messages} locale="it-ITE"><I18N.HTML msgId={msgId}/></Localized>, document.getElementById("container"));
```
offtherailz: ITE or IT?

web/client/components/I18N/HTML.jsx
mbarto: id is a property of react-intl FormattedHTMLMessage component, not a DOM id

PR #1509, opened 2017-02-24
Fix #1497 Query panel shows a scrollbar when select spatial filter

### LINKED ISSUE #1497: Query panel shows a scrollbar when select spatial filter
The div should grow until it reaches the end of the page and then add a scrollbar. 
The dropdown menu contents should be visible without have to scroll. 

image_from_skype

allyoucanmap:
Changed behavior of query panel scrollbar

### DIFF
--- a/web/client/components/data/query/QueryBuilder.jsx
+++ b/web/client/components/data/query/QueryBuilder.jsx
@@ -47,7 +47,8 @@ const QueryBuilder = React.createClass({
         resultTitle: React.PropTypes.string,
         pagination: React.PropTypes.object,
         sortOptions: React.PropTypes.object,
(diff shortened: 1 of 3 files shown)


web/client/components/data/query/QueryBuilder.jsx
```diff
@@ -137,7 +137,7 @@ const QueryBuilder = React.createClass({
                     sortOptions={this.props.sortOptions}
                     hits={this.props.hits}
                     />
-                <div className="querypanel">
+                <div className="querypanel" style={{maxHeight: window.innerHeight - 170}}>
```
offtherailz: Check if it is possible to use the redux state to get the height.

PR #1536, opened 2017-03-10
Fixed #1472 handled multiple markers/points when bearing is activated

### LINKED ISSUE #1472: Bearing Measure allows more than 2 points
To reproduce:
- open measure tool and select measure bearing
- click quickly drawing multiple points
image

The problem is on both leaflet and ol3 (on ol3 a result is returned tho)
We should limit the number of point for the bearing to 2.

MV88:
Fixed #1472
The listener on the click on the map apply some sort of debouncing. This is causing sometimes to skip the case when 2 markers are added and passes from 1 to 3 (for example).

When this happens it stops drawing any other marker and it calculate the bearing on the first 2 inserted.

### DIFF
--- a/web/client/components/map/leaflet/MeasurementSupport.jsx
+++ b/web/client/components/map/leaflet/MeasurementSupport.jsx
@@ -9,6 +9,7 @@
 const React = require('react');
 const assign = require('object-assign');
 var L = require('leaflet');
+const {slice} = require('lodash');
 var CoordinatesUtils = require('../../../utils/CoordinatesUtils');
 
 require('leaflet-draw');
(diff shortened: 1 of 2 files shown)


web/client/components/map/leaflet/MeasurementSupport.jsx
```diff
@@ -121,9 +118,30 @@ const MeasurementSupport = React.createClass({
             this.props.map.removeLayer(this.lastLayer);
             this.drawControl.enable();
             this.drawing = true;
-
         } else {
-            this.updateMeasurementResults();
+            let bearingMarkers = this.drawControl._markers || [];
```
offtherailz: You can simplify : 
```
if (bearingMarkers.length >= 2 && this.props.measurement.geomType === 'Bearing') {
     this.drawControl._markers = slice(this.drawControl._markers, 0, 2);
     this.drawControl._poly._latlngs = slice(this.drawControl._poly._latlngs, 0, 2);
     this.drawControl._poly._originalPoints = slice(this.drawControl._poly._originalPoints, 0, 2);
     this.updateMeasurementResults();
     this.drawControl._finishShape();
     this.drawControl.disable();
} else {
     updateMeasurementResults()
}
```

PR #1537, opened 2017-03-10
Fix console warnings

Gnafu:


package.json
```diff
@@ -129,8 +129,8 @@
     "react-share": "1.8.0",
     "react-sidebar": "2.1.2",
     "react-sortable-items": "https://github.com/geosolutions-it/react-sortable-items/tarball/react15",
-    "react-spinkit": "1.1.6",
-    "react-swipeable-views": "0.6.3",
+    "react-spinkit": "2.1.1",
```
mbarto: Did you check that spinkit styles are still valid.?I tried to upgrade and had some spinners with wrong colors

Gnafu: I know, it now uses .sk-spinner instead of .spinner
We are using two Spinner components and we should clean up the styles

Gnafu: Done on the latest PR, please pull again.

PR #1542, opened 2017-03-13
Fix #1496 Support to include theme 2.0

### LINKED ISSUE #1496: Support to include theme 2.0
Include and support theme 2.0.

After this step, we can imagine to allow inclusion and dynamic switching with proper loaders. see this: https://github.com/geosolutions-it/MapStore2-theme/pull/99

allyoucanmap:
Removed CSS styles from components
Added Switcher
Added new themes


offtherailz:
The spinner is not correctly sized (probably due to an update of react-spinkit)
image

PR #1564, opened 2017-03-15
More warnings fix

Gnafu:
Note that I removed "defaultActiveKey={0}" from the generic "Container" parameters because it is not supported by the default implementation we are using.
Any other project using a different implementation should define it in its configuration file.


offtherailz:
duplicated pull request?

Gnafu:
This is for the master, or do you want to merge the branch after the deploy?

PR #1614, opened 2017-03-20
Fix  #1610. Now the epics are overridable

### LINKED ISSUE #1610: Epics in plugins should be overridable
Epics  in plugins should be overridable to : 
 * avoid duplications
 * allow replacement

offtherailz:


web/client/utils/PluginsUtils.js
```diff
@@ -120,15 +120,15 @@ const getPluginItems = (state, plugins, pluginsConfig, name, id, isDefault, load
 const getReducers = (plugins) => Object.keys(plugins).map((name) => plugins[name].reducers)
                             .reduce((previous, current) => assign({}, previous, current), {});
 const getEpics = (plugins) => Object.keys(plugins).map((name) => plugins[name].epics)
-                            .reduce((previous = [], current = []) => [ ...previous, ...current], []);
+                            .reduce((previous = [], current = []) => assign({}, previous, current), {});
```
mbarto: You don't need default values for reduce params. Moreover both should be empty objects, not array

offtherailz: good catch

PR #1615, opened 2017-03-20
Fix #1612. Extent calculation for GeoJson supported

### LINKED ISSUE #1612: Improve and Fix GeoJSon Extent calculation
GeoJson extent calculation is wrong and limited to the geometry (not supported feature and featurecollection).
We should support it and fix issues with validity of bounding box.

offtherailz:
Now GeoJson extent calculation support also `Feature` and `FeatureCollection` (before this only geometry was supported).

### DIFF
--- a/web/client/utils/CoordinatesUtils.js
+++ b/web/client/utils/CoordinatesUtils.js
@@ -240,12 +240,13 @@ const CoordinatesUtils = {
         if (geoJSON.coordinates) {
             if (geoJSON.type !== "Point" && geoJSON.type !== "GeometryCollection") {
                 const flatCoordinates = chunk(flattenDeep(geoJSON.coordinates), 2);
-                flatCoordinates.reduce((extent, point) => {
-                    extent[0] = (point[0] < newExtent[0]) ? point[0] : newExtent[0];
-                    extent[1] = (point[1] < newExtent[1]) ? point[1] : newExtent[1];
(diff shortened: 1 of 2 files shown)


web/client/utils/CoordinatesUtils.js
```diff
@@ -254,15 +255,34 @@ const CoordinatesUtils = {
...
-                geoJSON.geometies.reduce((extent, geometry) => {
+                return geoJSON.geometies.reduce((extent, geometry) => {
                     let ext = this.getGeoJSONExtent(geometry);
                     if (this.isValidExtent(ext)) {
-                        extent[0] = (ext[0] < newExtent[0]) ? ext[0] : newExtent[0];
-                        extent[1] = (ext[1] < newExtent[1]) ? ext[1] : newExtent[1];
-                        extent[2] = (ext[2] > newExtent[2]) ? ext[2] : newExtent[2];
-                        extent[3] = (ext[3] > newExtent[3]) ? ext[3] : newExtent[3];
+                        return [
+                            (ext[0] < extent[0]) ? ext[0] : extent[0],
```
mbarto: can we create a function with this stuff, that is repeated many times?

offtherailz: Sure, as you see the original one had many side effects. I didn't noticed they become exactly the same function after my refactor.

PR #1620, opened 2017-03-22
Fixes #1619, Fixes #1549, Fixes #1103: notification system and docume…

### LINKED ISSUE #1103: Refactor Embed API
The current embed API uses iFrames in HTML. This is considered a security threat in many environments hence we should go beyond this approach and introduce a more modern approach.

### LINKED ISSUE #1549: Document javascript api
Create a reference for it in docma

### LINKED ISSUE #1619: Add notification of actions and state changes to the js api
We should allow the embedding applications to be notified when:
 * an action is launched
 * the app state changes

mbarto:
…ntation of the js api


docma-config.json
```diff
@@ -131,12 +134,7 @@
         "./docs/**/*md",
         {
             "readme": "./README.md",
-            "site": "./docs/index.md",
-            "developer-guide": "./docs/developer-guide/index.md",
```
offtherailz: developer's guide should be produced, is referred by the top menu.

PR #1644, opened 2017-03-27
Added appEpics

kappu72:
This pull lets pass an object with app specific Epics
If Epic's name, in appEpics, is the same the one in plugins this will override plugin's one

### DIFF
--- a/web/client/stores/StandardStore.js
+++ b/web/client/stores/StandardStore.js
@@ -26,7 +26,7 @@ const {createEpicMiddleware} = require('redux-observable');
 const SecurityUtils = require('../utils/SecurityUtils');
 const ListenerEnhancer = require('@carnesen/redux-add-action-listener-enhancer').default;
 
-module.exports = (initialState = {defaultState: {}, mobile: {}}, appReducers = {}, plugins, storeOpts) => {
+module.exports = (initialState = {defaultState: {}, mobile: {}}, appReducers = {}, appEpics = {}, plugins, storeOpts) => {
     const allReducers = combineReducers(plugins, {
         ...appReducers,
         localConfig: require('../reducers/localConfig'),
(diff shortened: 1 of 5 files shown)


web/client/stores/StandardStore.js
```diff
@@ -26,7 +26,7 @@ const {createEpicMiddleware} = require('redux-observable');
 const SecurityUtils = require('../utils/SecurityUtils');
 const ListenerEnhancer = require('@carnesen/redux-add-action-listener-enhancer').default;
 
-module.exports = (initialState = {defaultState: {}, mobile: {}}, appReducers = {}, plugins, storeOpts) => {
+module.exports = (initialState = {defaultState: {}, mobile: {}}, appReducers = {}, appEpics = {}, plugins, storeOpts) => {
```
mbarto: New arguments at the end, for compatibility reasons, please

web/client/utils/__tests__/PluginUtils-test.js
```diff
@@ -72,5 +73,11 @@ describe('PluginsUtils', () => {
         expect(desc1.items[0].cfg).toExist();
 
     });
+    it('combineEpics', () => {
+        const plugins = {MapSearchPlugin: MapSearchPlugin};
+        const appEpics = {appEpics: (actions$) => actions$};
```
offtherailz: This sounds like an infinite loop. Is not executed but it should. Please remove it.

web/client/stores/StandardStore.js
offtherailz: This signature is better. This doesn't break the binding with StandardApp or other apps.

kappu72: Binding appEpics wouldn't be possible

PR #1650, opened 2017-03-28
Fixes #1645: custom connect to be used in plugins definition; propert…

### LINKED ISSUE #1645: Override properties used in connect methods from configuration
It is known that the properties used in connect method are not overridable from the config.
This can be achieved with some workaround ( set them in componentWillMount) but this is bad practice.

mbarto:
…ies in pluginCfg override state connected properties

### DIFF
--- a/web/client/utils/PluginsUtils.js
+++ b/web/client/utils/PluginsUtils.js
@@ -9,6 +9,8 @@
 const assign = require('object-assign');
 const {omit, isObject, head, isArray, isString} = require('lodash');
 const {combineReducers} = require('redux');
+const {connect} = require('react-redux');
+
 const {combineEpics} = require('redux-observable');
 
 const {memoize, get} = require('lodash');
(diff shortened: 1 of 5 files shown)


web/client/utils/PluginsUtils.js
```diff
@@ -122,6 +124,11 @@ const getReducers = (plugins) => Object.keys(plugins).map((name) => plugins[name
 const getEpics = (plugins) => Object.keys(plugins).map((name) => plugins[name].epics)
                             .reduce((previous, current) => assign({}, previous, current), {});
 
+const pluginsMergeProps = (stateProps, dispatchProps, ownProps) => {
+    const {pluginCfg, ...otherProps} = ownProps;
+    return assign({}, otherProps, stateProps, dispatchProps, pluginCfg || {});
```
offtherailz: Didn't understand how to configure it.
pluginsCfg always wind over stateProps?
Could it be that pluginsCfg can be overridden by stateProps? Or do you mean that all the other Props are overridable while pluginCfg is the way to set fixed config. 

can you document it?

mbarto: pluginCfg is the way to set fixed config, but plugins cfg can be dynamic, using expressions.

PR #1661, opened 2017-03-30
Fix #1660. Now results can fit to map size

### LINKED ISSUE #1660: Allow to fit search results to map size
Now search results do not fit to the window size, and even the z-index is wrong:
image

We should have something like this instead:
image

offtherailz:
- If results are too many, the result list become scrollable
 - The z-index is fixed to be above the toolbar

### DIFF
--- a/web/client/plugins/Search.jsx
+++ b/web/client/plugins/Search.jsx
@@ -73,7 +73,7 @@ const ToggleButton = require('./searchbar/ToggleButton');
  * @class Search
  * @memberof plugins
  * @prop {object} cfg.searchOptions initial search options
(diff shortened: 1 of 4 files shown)


web/client/plugins/Search.jsx
```diff
@@ -124,6 +125,7 @@ const SearchPlugin = connect((state) => ({
     selectedItems: state && state.search && state.search.selectedItems
 }))(React.createClass({
     propTypes: {
+        fitToResultsMapSize: React.PropTypes.bool,
```
mbarto: fitResultsToMapSize

Issue #1632, opened 2017-03-23
Review share plugin


offtherailz:
more options for the embedded IFrame should be configurable:
Iframe: 
- [x] Remove the layer tree by default, checkbox to display
- [x] Add a button to open in full mapstore site
- [x] remove setting from burger menu, help optional and responsive to the embedded view (e.g. show from the bottom)
- [x] Full screen
- [x] Zoom To Max extent
- [x] Locate Me

PR #1696, opened 2017-04-06
Add a button to open in full mapstore (see #1632)

offtherailz:
GoFull plugins autoshows when originalUrl is present in config (JsAPI use case) or when the url match with a regex (embedded iframe use case).
In both cases it opens MapStore full in a new tab (using originalUrl or parsing the current href to create the url to open.

### DIFF
--- a/web/client/plugins/GoFull.jsx
+++ b/web/client/plugins/GoFull.jsx
@@ -5,82 +5,24 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
-
-const React = require('react');
 const {connect} = require('react-redux');
 
-const {Button, Glyphicon, Tooltip} = require('react-bootstrap');
-const OverlayTrigger = require('../components/misc/OverlayTrigger');
-const Message = require('../components/I18N/Message');
 const {mapSelector} = require('../selectors/map');
 const {createSelector} = require('reselect');
-const ConfigUtils = require('../utils/ConfigUtils');
+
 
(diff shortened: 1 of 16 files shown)


web/client/plugins/GoFull.jsx
```diff
@@ -0,0 +1,100 @@
...
+        return {
+            glyph: "share",
+            tooltip: "fullscreen.viewLargerMap",
+            urlRegex: "^(.*?)embedded.html.*?#\\/(\\d?)",
+            urlReplaceString: "\\$1#/viewer/leaflet/\\$2"
+        };
+    },
+
+    render() {
+        if (!this.display()) return <noscript></noscript>;
```
mbarto: return null;

web/client/plugins/GoFull.jsx
```diff
@@ -0,0 +1,100 @@
...
+ * GoFull plugins. Shows a button that opens full MapStore2 in a new tab. Try to find the `originalUrl` in configuration or tries to guess the mapId and creates the proper URL.
+ * This plugins hides automatically if the originalUrl is not present in configuration and if the urlRegex do not match.
+ * @prop {string} cfg.glyph the glyph icon for the button
+ * @prop {string} cfg.tooltip messageId of the tooltip to use
+ * @prop {string} cfg.urlRegex. A regex to parse the current location.href. This regex must match if the originalUrl is not defined.
+ * @prop {string} cfg.urlReplaceString. The template to create the url link. Uses the `urlRegex` groups to create the final URL
+ * @memberof plugins
+ * @class
+ */
+const GoFull = React.createClass({
```
mbarto: Externalize the component, please

MV88: do some test also

PR #1698, opened 2017-04-06
Set optional drawer menu for share ( see #1632)

offtherailz:
- Documented drawer menu plugin
 - Added a forceDrawer and a hide option
 - Improved share to manage forceDrawer option
image

### DIFF
--- a/web/client/plugins/DrawerMenu.jsx
+++ b/web/client/plugins/DrawerMenu.jsx
@@ -20,8 +20,8 @@ const {Button, Glyphicon, Panel} = require('react-bootstrap');
 const Section = require('./drawer/Section');
 
 const {partialRight} = require('lodash');
-const url = require('url');
-const query = url.parse(window.location.href, true).query;
+
+
 const Menu = connect((state) => ({
     show: state.controls.drawer && state.controls.drawer.enabled,
     activeKey: state.controls.drawer && state.controls.drawer.menu || "1",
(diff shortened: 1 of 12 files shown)


web/client/plugins/DrawerMenu.jsx
```diff
@@ -20,7 +20,8 @@ const {Button, Glyphicon, Panel} = require('react-bootstrap');
 const Section = require('./drawer/Section');
 
 const {partialRight} = require('lodash');
-
+const url = require('url');
+const query = url.parse(window.location.href, true).query;
```
mbarto: I don't like plugins directly depending from the request. We should use dynamic plugin configuration of plugins to:
 * add dynamicity to hide (currently it doesn't support expressions)
 * add request params to properties available in expressions (e.g. {request.forceDrawer})

web/client/plugins/DrawerMenu.jsx
```diff
@@ -33,10 +34,26 @@ const Menu = connect((state) => ({
...
+ *   "name": "DrawerMenu",
+ *   "cfg": {
+ *   "hideButton": true
+ * }
+ */
 const DrawerMenu = React.createClass({
     propTypes: {
         items: React.PropTypes.array,
         active: React.PropTypes.string,
+        hideButton: React.PropTypes.bool,
```
mbarto: remove it, please

PR #1704, opened 2017-04-07
Fixed #1545 Various fixes to look and feel

### LINKED ISSUE #1545: Minor fixes to look and feel
- [x] Add a tooltip for the drawer menu (both open button and top buttons)
- [x] Change the icon of the drawer menu (arrow left for opening and right for closing)
- [x] In query form, attribute filter name combobox:  "Select" (text placeholder) should be something like "Select Attrbiute..." or "Attribute"
- [x] review position of combos on the home page for selecting map component and style

MV88:
- Added tooltips to drawer menu button and tools
- Changed placeholder from "Select" to "Attribute" in the query form
- Changed position of Theme Swithcer and MapType Switcher

### DIFF
--- a/web/client/components/data/query/ComboField.jsx
+++ b/web/client/components/data/query/ComboField.jsx
@@ -14,7 +14,7 @@ const OverlayTrigger = require('../../misc/OverlayTrigger');
 
 const {DropdownList, Multiselect} = require('react-widgets');
 
-const LocaleUtils = require('../../../utils/LocaleUtils');
+const Message = require('../../../components/I18N/Message');
 
 const ComboField = React.createClass({
     propTypes: {
@@ -22,6 +22,7 @@ const ComboField = React.createClass({
         style: React.PropTypes.object,
         valueField: React.PropTypes.string,
         textField: React.PropTypes.string,
+        placeholder: React.PropTypes.string,
         fieldOptions: React.PropTypes.array,
         fieldName: React.PropTypes.string,
         fieldRowId: React.PropTypes.number,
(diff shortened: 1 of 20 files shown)


web/client/localConfig.json
```diff
@@ -108,7 +108,7 @@
                     }
                 }
             }, "Login",
-            "OmniBar", "BurgerMenu", "Expander"
+            "OmniBar", "GridContainer", "BurgerMenu", "Expander"
```
mbarto: Is the GridContainer used in desktop and mobile, or only in the homepage?

web/client/localConfig.json
```diff
@@ -202,7 +202,7 @@
                     }
                 }
             },
-            "OmniBar", "Login", "Save", "SaveAs", "BurgerMenu", "Expander", "Undo", "Redo", "FullScreen"
+            "OmniBar", "GridContainer", "Login", "Save", "SaveAs", "BurgerMenu", "Expander", "Undo", "Redo", "FullScreen"
```
mbarto: Is the GridContainer used in desktop and mobile, or only in the homepage?

web/client/components/data/query/ComboField.jsx
```diff
@@ -64,6 +65,7 @@ const ComboField = React.createClass({
             style: {
                 width: "100%"
             },
+            placeholder: <Message msgId="queryform.attributefilter.combo_placeholder"/>,
             multivalue: false,
```
MV88: Probably we can use a different default as placeholder like
"Select..."

PR #1744, opened 2017-04-19
Fixed snapshot issue on IE browser

allyoucanmap:


web/client/components/map/leaflet/snapshot/GrabMap.jsx
```diff
@@ -142,10 +142,15 @@ let GrabLMap = React.createClass({
         let svgH;
         let svgW;
         let svgString;
-        if (svg && svg.outerHTML) {
-            svgOffsetX = svgs[0] ? svgs[0].getBoundingClientRect().left : 0;
-            svgOffsetY = svgs[0] ? svgs[0].getBoundingClientRect().top : 0;
-            svgString = svgs[0].outerHTML;
+        if (svg) {
+            // get svg translate position to avoid overlay issue on IE
+            let svgTranslate = svg.style.transform.replace(/\(|\)|translate3d|translate/g, '').split('px, ');
```
offtherailz: style & style.transform can be null? anyway you should check.

PR #1754, opened 2017-04-20
First Implementation of the 3d switcher.

offtherailz:
The GlobeSwitcher plugin allow to switch to Globe view. In this pull request I also syncronize the maptype with it's reducer to keep in sync the maptype forever in the state.
This is a first step to manage maptype's dependent tools (e.g. navigation tools) using react-redux system.
We still have to disable the zoom buttons and add proper navigation controls for the 3d view, but this is a first implementation that can be tested.


web/client/epics/globeswitcher.js
```diff
@@ -0,0 +1,48 @@
...
+module.exports = {
+    updateRouteOn3dSwitch
+};
+
+/**
+ * Epics for 3d switcher functionality
+ * @name epics.globeswitcher
+ * @type {Object}
+ */
+module.exports = {
```
mbarto: duplicated exports

PR #1763, opened 2017-04-24
Fix #1692 WFS spatial query works only if geometry field name is the_geom

### LINKED ISSUE #1692: WFS spatial query works only if geometry field name is the_geom
If a queriable layer does not have a "the_geom" field, the query tool fails to filter with spatial filters.

allyoucanmap:
Added geometry attribute selection on spatial filter
Added new attribute to layer if WMS (geometryField)

### DIFF
--- a/web/client/epics/layers.js
+++ b/web/client/epics/layers.js
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const {ADD_LAYER, changeLayerProperties} = require('../actions/layers');
+const {MAP_CONFIG_LOADED} = require('../actions/config');
+const Rx = require('rxjs');
+const WFS = require('../api/WFS');
+const _ = require('lodash');
+
+function getGeometryField(layer) {
+    return Rx.Observable.of(WFS.describeFeatureType(layer.url, layer.name).then((describeFeatureType) => {
+        const types = _.get(describeFeatureType, "complexType[0].complexContent.extension.sequence.element");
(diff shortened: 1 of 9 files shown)


web/client/epics/layers.js
```diff
@@ -0,0 +1,39 @@
...
+        const fields = _.head(types && types.filter( elem => (elem.name === "the_geom" || elem.type.prefix.indexOf("gml") === 0)));
+        const geometryField = fields && fields.name;
+        return geometryField ? changeLayerProperties(layer.id, {geometryField}) : changeLayerProperties(layer.id, {});
+    }));
+}
+
+const addGeometryFieldLayer = action$ =>
+    action$.ofType(ADD_LAYER).filter((action) => { return action.layer.type === 'wms'; }).switchMap(action => {
+        return getGeometryField(action.layer);
+    }).mergeAll();
```
offtherailz: mergeAll is not needed

web/client/epics/layers.js
```diff
@@ -0,0 +1,39 @@
...
+
+const addGeometryFieldConfig = action$ =>
+    action$.ofType(MAP_CONFIG_LOADED).switchMap(action => Rx.Observable.from(action.config.map.layers)
+    .filter((layer) => { return layer.type === 'wms'; }))
+    .switchMap(layer => {
+        return getGeometryField(layer);
+    }).mergeAll();
```
offtherailz: same as above

web/client/plugins/Map.jsx
```diff
@@ -286,5 +286,6 @@ const selector = createSelector(
 );
 module.exports = {
     MapPlugin: connect(selector)(MapPlugin),
-    reducers: { draw: require('../reducers/draw') }
+    reducers: { draw: require('../reducers/draw') },
+    epics: { addGeometryFieldLayer, addGeometryFieldConfig}
```
offtherailz: probably this side effect have to be added if the featureGrid is present @mbarto what do you think?

Issue #1780, opened 2017-05-03
Improve 3D map support


offtherailz:
We already introduce a 3D switcher. We should enhance the navigation tool to fit the 3D use case and disable tools that are not supported yet. 
Remove tools when 3D mode is enabled:
- [x] ScaleBox
- [x] Zoom In
- [x] Zoom Out
- [x] Print
- [x] Locate
- [x] Shapefile
- [x] Measure tool

Fixes/Improvements: 
 - [x] Attibution is not rendered as html

Tools to Add: 
 - [x] Navigation for cesium
 - [x] Hints for navigation

MV88:
Remove tools when 3D mode is enabled:
- [ ] ScaleBox
- [ ] Print
- [ ] Locate
- [ ] Shapefile

PR #1785, opened 2017-05-04
Added credits alternative for cesium baseLayers. See #1780

offtherailz:
Removes wrong html from attribution in cesium.

This pull request belongs to #1780 
From this: 
image

to this: 
image

### DIFF
--- a/web/client/components/map/cesium/plugins/TileProviderLayer.js
+++ b/web/client/components/map/cesium/plugins/TileProviderLayer.js
@@ -76,7 +76,7 @@ Layers.registerType('tileprovider', (options) => {
         proxy = !isCORS && proxyUrl;
     }
     const cr = opt.credits;
-    let credit = cr ? new Cesium.Credit(cr.text, cr.imageUrl, cr.link) : opt.attribution;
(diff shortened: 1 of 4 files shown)


web/client/components/map/cesium/plugins/TileProviderLayer.js
```diff
@@ -75,13 +75,15 @@ Layers.registerType('tileprovider', (options) => {
         const isCORS = useCORS.reduce((found, current) => found || url.indexOf(current) === 0, false);
         proxy = !isCORS && proxyUrl;
     }
+    const cr = opt.credits;
+    let credit = cr ? new Cesium.Credit(cr.text, cr.imageUrl, cr.link) : opt.attribution;
     return new Cesium.UrlTemplateImageryProvider({
```
MV88: use const if credit does not change anymore

web/client/utils/__tests__/TileConfigProvider-test.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright 2017, GeoSolutions Sas.
```
MV88: remove extra *

PR #1786, opened 2017-05-05
Prevent axios to request /null urls

Gnafu:
Every time the user closes the query panel, a FEATURE_TYPE_SELECTED action is dispatched.
When closing the panel, such action will not have an url set.
If the action is not filtered, axios will try to request a **/null** path, for example:
http://dev.mapstore2.geo-solutions.it/mapstore/null?service=WFS&version=1.1.0&request=DescribeFeatureType&typeName=null&outputFormat=application/json

To reproduce:
- add a queryable layer from the catalog
- open the query panel
- close the query panel

### DIFF
--- a/web/client/epics/wfsquery.js
+++ b/web/client/epics/wfsquery.js
@@ -52,7 +52,9 @@ const extractInfo = (data) => {
 };
 
 const featureTypeSelectedEpic = action$ =>
-    action$.ofType(FEATURE_TYPE_SELECTED).switchMap(action => {
+    action$.ofType(FEATURE_TYPE_SELECTED)
+    .filter(action => action.url !== null)  // Axios will try to request a /null url without this filter
+    .switchMap(action => {
         return Rx.Observable.defer( () =>
             axios.get(action.url + '?service=WFS&version=1.1.0&request=DescribeFeatureType&typeName=' + action.typeName + '&outputFormat=application/json'))
         .map((response) => {


web/client/epics/wfsquery.js
```diff
@@ -52,7 +52,9 @@ const extractInfo = (data) => {
 };
 
 const featureTypeSelectedEpic = action$ =>
-    action$.ofType(FEATURE_TYPE_SELECTED).switchMap(action => {
+    action$.ofType(FEATURE_TYPE_SELECTED)
+    .filter(action => action.url !== null)  // Axios will try to request a /null url without this filter
```
mbarto: I think we should NOT launch the action with a null url, if it doesn't make sense

Gnafu:
The action is null when the user closes the panel, so it's in a "reset" state.
I agree with avoiding unnecessary dispatch but I will still check for null in the epic

MV88:
Maybe you can return an empty observable  like Rx.Observable.empty()

2017-05-08 9:54 GMT+02:00 Lorenzo Pini <notifications@github.com>:

MV88:
Maybe you can return an empty observable like Rx.Observable.empty()
https://github.com/geosolutions-it/MapStore2/blob/master/web/client/epics/search.js#L141

Gnafu:
@MV88 , were do you think the empty Observable can be returned?

offtherailz:
When the user closes the panel an action like `FEATURE_TYPE_SELECTION_RESET`or `SEARCH_PANEL_CLOSED` should be performed, not an empty `FEATURE_TYPE_SELECTED` . So it will not be catched by the epic.

PR #1790, opened 2017-05-08
Fix #1776. Move measure into burger menu

### LINKED ISSUE #1776: Move Measure Tools to the burger menu
Currently the measure tools are located in a panel within the TOC.

That placement is at least strange, we should move them somewhere else, ideally as a section with subsections in the burger menu.

This requires some thinking and some mockups prior to implementation.

offtherailz:
The dialog for measure is now in the burger menu. This fixes #1776.

The old plugin will be kept in MeasurePanel + MeasureResult plugin and deprecated.
All the duplicated functionalities are now merged in a unique component.


web/client/components/mapcontrols/measure/MeasureComponent.jsx
```diff
@@ -10,10 +10,7 @@ const React = require('react');
 const {Panel, ButtonGroup, Tooltip, Glyphicon, Button} = require('react-bootstrap');
 const ToggleButton = require('../../buttons/ToggleButton');
 var NumberFormat = require('../../I18N/Number');
-
-const lineRuleIcon = require('./img/line-ruler.png');
-const areaRuleIcon = require('./img/area-ruler.png');
-const bearingRuleIcon = require('./img/bearing-ruler.png');
+var Message = require('../../I18N/Message');
```
mbarto: const

PR #1791, opened 2017-05-08
Hide tools not supported in 3d (see #1780)

offtherailz:
This pull request is a part of #1780 and hides the following tools when the maptype is cesium:
 - Zoom In/Out
 - ScaleBox
 - Print
 - Locate
 - ShapeFile
 - Measure

It adds the display function as a allowed member of tools in a toolcontainer. If this fuction is present, it will be used to filter the tools to display. the entry. It is useful to hide entries in the ToolContainer other then hiding the window itself.

Also some documentation improved (e.g locate plugin, selectors, external types).

### DIFF
--- a/docma-config.json
+++ b/docma-config.json
@@ -120,6 +120,10 @@
             "web/client/actions/maptype.js",
             "web/client/actions/search.js",
 
+            "web/client/selectors/index.jsdoc",
+            "web/client/selectors/search.js",
+            "web/client/selectors/maptype.js",
+
             "web/client/reducers/index.jsdoc",
             "web/client/reducers/controls.js",
             "web/client/reducers/globeswitcher.js",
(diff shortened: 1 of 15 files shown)


mbarto:
I prefer this alternative solution:

https://github.com/mbarto/MapStore2/tree/hide_tools_3d

PROS:
 * no changes in code (except for a fix)
 * fully done with configuration
 * doesn't bind plugins to mapping libraries

offtherailz:
Yes it is good, but I'm scared to add a thousand of exceptions and flags like this here and there in the configuration, making a MapStore2 configuration more complex that coding.
It makes sense if it is a temporary solution,like for tools that are going to be supported soon.

mbarto:
I think that since we built a configurable framework we should not be
scared to configure it, when this is the simplest solution. I have seen too
many cases of plugins that can only work in a specific scenario because of
hard coded behaviour.  Code should, in my opinion, open new possibilities,
not limit them. I am open to any solution that does not limit how I can use
a plugin. KISS

Il 08/mag/2017 19:00, "Lorenzo Natali" <notifications@github.com> ha
scritto:

offtherailz:
Ok, so let's close this pull request (maybe I'll port the other improvements like doc in a separte one). Please do a pull request with your solution.

PR #1792, opened 2017-05-08
Fix #1737 Query panel search fails if an id is not defined

### LINKED ISSUE #1737: Query panel search fails if an id is not defined
[GeoServer's only issue ]. Paginated WFS requests fail if there is not an ID configured on the server side, because GeoServer can't sort without this information. We could increase reliability of MapStore 2 with soft failing, forcing the sort attribute to the first argument of the request if we recognize this kind of error.

allyoucanmap:
- Added epic to handle WFS id error on query and forced sort on the first attribute.
- Removed thunk query from WFS action
- Added wfs query documentation


web/client/epics/wfsquery.js
```diff
@@ -72,6 +134,55 @@ const featureTypeSelectedEpic = action$ =>
         .catch(e => Rx.Observable.of(featureTypeError(action.typeName, e.message)));
     });
 
+/**
+ * Sends a WFS query, returns a response or handles request error
+ * in particular the NoApplicableCode WFS error with a forced sort option on the first attribute
+ * @param {Observable} action$ manages `QUERY`
```
offtherailz: external:Observable

PR #1799, opened 2017-05-09
fixed #1640 added singleTile option to Display Tab

### LINKED ISSUE #1640: singleTile option not available in layer properties
The 'singleTile' option should be introduced into the layer properties panel

MV88:
The new option is added for leaflet and openlayers.

### DIFF
--- a/web/client/components/map/leaflet/plugins/WMSLayer.js
+++ b/web/client/components/map/leaflet/plugins/WMSLayer.js
@@ -1,5 +1,5 @@
-/**
- * Copyright 2015, GeoSolutions Sas.
+/*
+ * Copyright 2017, GeoSolutions Sas.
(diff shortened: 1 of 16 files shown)


web/client/components/map/leaflet/plugins/WMSLayer.js
```diff
@@ -13,6 +13,7 @@ const L = require('leaflet');
 const objectAssign = require('object-assign');
 const {isArray, isEqual} = require('lodash');
 const SecurityUtils = require('../../../../utils/SecurityUtils');
+require('leaflet.nontiledlayer');
```
offtherailz: Check if this inclusion breaks the app if leaflet is missing or not loaded yet.

PR #1822, opened 2017-05-12
Fix #1392 added viewport into spatial filter list

### LINKED ISSUE #1392: Add viewport to the spatial filters list
Viewport, when selected, will get the current viewport bbox as a filter to perform search.

image

MV88:
The Viewport optios has been added to the spatial filter list.
image
Every time the view changes the geometry of the filterObj updates with an epic.

### DIFF
--- a/web/client/actions/queryform.js
+++ b/web/client/actions/queryform.js
@@ -16,6 +16,8 @@ const CHANGE_CASCADING_VALUE = 'CHANGE_CASCADING_VALUE';
 const EXPAND_ATTRIBUTE_PANEL = 'EXPAND_ATTRIBUTE_PANEL';
 const EXPAND_SPATIAL_PANEL = 'EXPAND_SPATIAL_PANEL';
 const SELECT_SPATIAL_METHOD = 'SELECT_SPATIAL_METHOD';
+const SELECT_VIEWPORT_SPATIAL_METHOD = 'SELECT_VIEWPORT_SPATIAL_METHOD';
+const UPDATE_GEOMETRY = 'UPDATE_GEOMETRY';
 const SELECT_SPATIAL_OPERATION = 'SELECT_SPATIAL_OPERATION';
 const CHANGE_SPATIAL_ATTRIBUTE = 'CHANGE_SPATIAL_ATTRIBUTE';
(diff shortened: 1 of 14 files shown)


web/client/actions/queryform.js
```diff
@@ -126,6 +128,18 @@ function selectSpatialMethod(method, fieldName) {
     };
 }
 
+function selectViewportSM() {
```
mbarto: don't be shy, selectViewportSpatialMethod

web/client/components/data/query/QueryBuilder.jsx
```diff
@@ -96,6 +96,7 @@ const QueryBuilder = React.createClass({
                 onChangeDrawingStatus: () => {},
                 onRemoveSpatialSelection: () => {},
                 onShowSpatialSelectionDetails: () => {},
+                onSelectViewportSM: () => {},
```
mbarto: don't be shy

web/client/actions/wfsquery.js
```diff
@@ -103,12 +103,11 @@ function createQuery(searchUrl, filterObj) {
     };
 }
 
-function query(searchUrl, filterObj, retry) {
+function query(searchUrl, filterObj) {
```
mbarto: is retry not used?

web/client/epics/wfsquery.js
```diff
@@ -227,6 +228,33 @@ const closeFeatureEpic = action$ =>
...
+        extent[0] = -180.0;
+        extent[2] = 180.0;
+    }
+    return extent;
+}
+const viewportSelectedEpic = (action$, store) =>
+    action$.ofType(SELECT_VIEWPORT_SPATIAL_METHOD, CHANGE_MAP_VIEW)
+        .switchMap((action) => {
+            // calculate new geometry from map properties
+            const map = action.type === CHANGE_MAP_VIEW ? action : store.getState().map.present;
```
mbarto: are you updating the geometry also if the spatial filter is not Viewport?

web/client/actions/wfsquery.js
MV88: nope, it's been removed in getFeatures method of DockerFeatureGrid component

PR #1823, opened 2017-05-12
Fix #1672 BackgroundSwitcher review

### LINKED ISSUE #1672: BackgroundSwitcher review
Background switcher floating on the map, with the current background preview showing. When clicked the tool should display the alternative backgrounds. Should be perfect if it is somehow merged with the overview map.

allyoucanmap:


web/client/plugins/BackgroundSelector.jsx
```diff
@@ -0,0 +1,61 @@
...
+  *       "side": 78,
+  *       "sidePreview": 104,
+  *       "frame": 3,
+  *       "margin": 5
+  *     }
+  *   }
+  * }
+  */
+
+const BackgroundSelectorPlugin = connect((state) => ({
```
offtherailz: use reselect

web/client/components/background/BackgroundSelector.jsx
```diff
@@ -0,0 +1,201 @@
...
+            size: {width: 0, height: 0},
+            desktop: {},
+            mobile: {},
+            onToggle: () => {},
+            propertiesChangeHandler: () => {},
+            setControlProperty: () => {}
+        };
+    },
+    componentWillUnmount() {
+        this.props.setControlProperty('backgroundSelector', 'currentLayer', {});
```
offtherailz: use a separate reducer or bind this method to background selector from the plugin. I don't want to make components aware of the setControlProperty method if it is possible.

web/client/components/background/PaginationButton.jsx
```diff
@@ -0,0 +1,41 @@
...
+    getDefaultProps() {
+        return {
+            side: 0,
+            direction: true,
+            vertical: false,
+            onClick: () => {}
+        };
+    },
+    render() {
+        let glyph = this.props.direction ? 'chevron-right' : 'chevron-left';
```
mbarto: const glyph = glyphs[this.props.vertical][this.props.direction];

web/client/components/background/PreviewIcon.jsx
```diff
@@ -0,0 +1,58 @@
...
+            layer: {},
+            currentLayer: {},
+            onToggle: () => {},
+            onClose: () => {},
+            setLayer: () => {}
+        };
+    },
+    render() {
+        const containerClass = this.props.vertical ? 'background-preview-icon-container-vertical' : 'background-preview-icon-container-horizontal';
+        let type = this.props.layer.visibility ? containerClass + ' bg-primary' : containerClass + ' bg-body';
```
mbarto: Use const, let is not needed. You can do everything without calculating type twice

web/client/plugins/DrawerMenu.jsx
```diff
@@ -78,9 +79,13 @@ const DrawerMenu = React.createClass({
             menuOptions: {},
             singleSection: true,
             buttonClassName: "square-button",
-            disabled: false
+            disabled: false,
+            resetWidth: () => {}
         };
     },
+    componentWillUnmount() {
+        this.props.resetWidth();
```
mbarto: I don't think this is needed. If the control is not enabled, width is 0, we don't need an additional property

PR #1847, opened 2017-05-19
fix #1845 save as - Master

### LINKED ISSUE #1845: Save As doesn't work

MV88:


offtherailz:
This will be taken from master branch

PR #1874, opened 2017-05-23
Fix #1873 Wrong Font Family and Size on Print plugin

### LINKED ISSUE #1873: Wrong Font Family and Size on Print plugin
The print output displays the wrong font family and size

allyoucanmap:


web/client/plugins/print/index.js
```diff
@@ -119,6 +119,7 @@ const Font = connect((state) => ({
     italic: state.print && state.print.spec && state.print.spec.italic
 }), {
     onChangeFamily: setPrintParameter.bind(null, 'fontFamily'),
+    onChangeName: setPrintParameter.bind(null, 'fontName'),
```
mbarto: fontName is not used, I think. You still use only fontFamily in PrintUtils. Maybe we don't need a duplicated property at all

PR #1876, opened 2017-05-23
Moved background selector

allyoucanmap:
bg

### DIFF
--- a/web/client/localConfig.json
+++ b/web/client/localConfig.json
@@ -37,6 +37,7 @@
             {
                 "name": "BackgroundSelector",
                 "cfg": {
+                    "bottom": 40,
                     "dimensions": {
                     "side": 65,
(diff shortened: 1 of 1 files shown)


web/client/localConfig.json
```diff
@@ -37,6 +37,7 @@
             {
                 "name": "BackgroundSelector",
                 "cfg": {
+                    "bottom": 40,
```
mbarto: Can't you use css?

offtherailz: Yes, we could, but most probably this change in each installation by plugins presence and so on. 
So configure it only in the product is better choice, IMHO.

mbarto: I don't like using config for css related stuff, but it's not a blocker for me

PR #1877, opened 2017-05-23
Fixes #1866: queryform example updated with epics and other related f…

### LINKED ISSUE #1866: Spatial Filter by Viewport does not enable the Search button in the QueryBuilder example
It works in the MapStore2 product

mbarto:
…ixes


web/client/components/data/query/QueryToolbar.jsx
```diff
@@ -70,7 +70,7 @@ const QueryToolbar = React.createClass({
             // fieldsWithoutValues ||
             fieldsExceptions ||
             !this.props.toolbarEnabled ||
-            (!fieldsWithValues && !this.props.spatialField.geometry);
+            (!fieldsWithValues && !this.props.spatialField.geometry && this.props.spatialField.method !== 'Viewport');
         const tooltip = <Tooltip id="query-warning-tooltip"><I18N.Message msgId="queryform.emptyfilter"/></Tooltip>;
```
MV88: Probably this is not needed with the fix made to the epic.

PR #1887, opened 2017-05-26
Some improvements to resizableGrid and DrawSupport of openlayers

MV88:
Also added some css for the even row selected and increased size of checkboxes
Changed toggle button with a different onclick behaviour
added some documentation


docs/developer-guide/maps-configuration.md
```diff
@@ -1,8 +1,36 @@
...
+        -20037508.34, -20037508.34,
+        20037508.34, 20037508.34
+    ],
+    "layers": [{...},{...}]
+}``
 # Layers option
 
+i.e.
+> ``{
+    "url": "http..."
```
offtherailz: significative sample if you have url, type or source osm are not relevant.

web/client/components/buttons/ToggleButtonV2.jsx
```diff
@@ -0,0 +1,101 @@
+/*
```
offtherailz: Duplicates v1. remove

PR #1908, opened 2017-06-01
Fix #1907. Support single feature styling in OL3

### LINKED ISSUE #1907: Support for feature styling missing in openlayers
You can style a vector layer but not a single feature with current Feature implementation.

We should: 
 - Unify the vector styling convention
 - Support styling of single features for OpenLayers

offtherailz:
This adds support to style the single feature with the style entry. Moved the style functionalities out of the VectorLayer use same utilities for Feature objects.

### DIFF
--- a/web/client/components/map/openlayers/Feature.jsx
+++ b/web/client/components/map/openlayers/Feature.jsx
@@ -9,10 +9,13 @@
 var React = require('react');
 var ol = require('openlayers');
 const {isEqual} = require('lodash');
+const {getStyle} = require('./VectorStyle');
 
 let Feature = React.createClass({
     propTypes: {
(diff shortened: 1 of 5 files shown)


web/client/components/map/openlayers/Feature.jsx
```diff
@@ -36,25 +39,35 @@ let Feature = React.createClass({
                 geometry: this.props.geometry,
                 id: this.props.msId});
             this._feature.forEach((f) => f.getGeometry().transform(this.props.featuresCrs, this.props.crs || 'EPSG:3857'));
+            if (this.props.style && (this.props.style !== this.props.layerStyle)) {
```
mbarto: Maybe we should use isEqual

PR #1911, opened 2017-06-08
Implemented filter builder

offtherailz:
Made a more generic filter generator. This allows to generate OGC Filters, using syntax like
```
property("a").like("*a*")
```
We can use the same syntax to create builders for ECQL or even some server side stuff.


web/client/utils/ogc/GML/index.js
```diff
@@ -1,5 +1,5 @@
 const {isArray} = require('lodash');
-const isGML1 = (version) => version === "1.0.0";
+const isGML2 = (version) => version.indexOf("2") === 0;
```
mbarto: indexOf('2.')

PR #1958, opened 2017-06-23
Fix #1957. Role filter for create new map

### LINKED ISSUE #1957: Create new map should appear only for certain roles
We should be able to configure a list of allowed roles that can see "Create a new map". Otherwise logged users are allowed to create new maps.

offtherailz:
Fix #1957. Now you can confiure the plugin to display only for the allowedRoles in configuration


web/client/plugins/CreateNewMap.jsx
```diff
@@ -17,14 +17,17 @@ const CreateNewMap = React.createClass({
...
+        allowedRoles: React.PropTypes.array,
+        user: React.PropTypes.object
     },
     contextTypes: {
         router: React.PropTypes.object
     },
     getDefaultProps() {
         return {
             mapType: "leaflet",
+            allowedRoles: ["ADMIN", "USER"],
```
mbarto: but... do we have other roles? maybe we should use groups...

offtherailz: By default all logged in, but we can deny for normal users via configuration.Using groups is a good idea for the future.

web/client/plugins/CreateNewMap.jsx
```diff
@@ -18,7 +18,9 @@ class CreateNewMap extends React.Component {
         mapType: PropTypes.string,
         onGoToMap: PropTypes.func,
         colProps: PropTypes.object,
-        isLoggedIn: PropTypes.bool
+        isLoggedIn: PropTypes.bool,
+        allowedRoles: React.PropTypes.array,
```
mbarto: Remove React.PropTypes and use PropTypes

PR #1965, opened 2017-06-26
Documentation typos and dependencies deduplication

Gnafu:


web/client/components/manager/rulesmanager/RuleAttributes.jsx
```diff
@@ -11,7 +11,6 @@ const React = require('react');
 const {Panel} = require('react-bootstrap');
 const _ = require('lodash');
 const Select = require('./RuleAttributeSelect');
-const {head} = require('lodash');
```
MV88: wasn't it better like it was before?

Gnafu: I'm trying to reduce clutter. The lodash wrapper is used anyway so another require is redundant.

PR #1981, opened 2017-06-27
Fix #1943 debounce autocomplete only on editing

### LINKED ISSUE #1943: Apply debouncing only when editing the combobox value
See the autocompleteField component

MV88:
see #1943

Moved css to ms2-theme
Used a delay for the epic passed via the action
Fixed a bug for is null operator (it was creating a undefined filter)
Migrated to react 15.5


web/client/themes/default/ms2-theme.less
```diff
@@ -1228,4 +1228,21 @@ select.form-control option {
     cursor: pointer;
     font-size: 40px;
     width: 44px;
+
```
offtherailz: didn't closed the bracket

PR #1985, opened 2017-06-28
Add resources to Developer Guide

Gnafu:
Adds instruction on own to customize the Theme and the search box services
Also make the standalone demo page accessible again

### DIFF
--- a/docs/developer-guide/API-usage.md
+++ b/docs/developer-guide/API-usage.md
@@ -1,0 +1,14 @@
+# MapStore2 API usage
+You can include MapStore2 in your application and interact with it via its JavaScript API
+
+## How to use
+
+1. Create a map using the standard installation
+1. Go to Share -> Embed
+1. Copy the API html code and paste it in your application page
+
+The map will now load inside your application
+
+    NOTE: If the map is using a Google Maps background you will have to provide your own API key.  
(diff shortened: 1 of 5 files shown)


mbarto:
This does not seem a user guide to me

offtherailz:
Yes it looks more like a work in progress. Should we place some "TODO" here and there to make it become a user guide?

mbarto:
Maybe we should simply remove what is not interesting to users, or redefine what a user is: a developer, a configurer, a designer (themes) or someone who uses the application?
This is not clear to me. From this PR it can be one of the first 3 roles, but not the last one.

Gnafu:
Whatever solution is fine for me, I just don't want to loose this informations as they are not documented anywhere (or they are buried in the mapstore-users mailing list)

mbarto:
I was wondering if the stuff you wrote is more developer oriented, than user oriented, and we already have a developer guide (where you can add new pages, if you like)

Gnafu:
Moved them to the developers guide

offtherailz:
Sorry if I take this too late, but the theme customization is for the dev guide, but the search services configuration is accessible from the web, so it's part user's guide, in my opinion. Should we merge this anyway?

PR #2039, opened 2017-07-18
Fix #2031 Trigger notification if a map is an old one asking for updating it

### LINKED ISSUE #2031: Trigger notification if a map is an old one asking for updating it
**requirements:**
- The user must have write access for the resource map.
- Must be a mapstore map (old mapstore)
- remove the refresh tools from the TOC

**behaviour:**
if the requirements are met then show a notification in top center position with warning style (yellow) containing:
- an update button that will update the map in background (with all the options selected). 
image

- a body text "This is an old map, thus not all the functionalities can be enabled. Click on update button in order to update the map or simply dismiss this notification if you don't want to update it."

allyoucanmap:
See #2031 
Added AutoMapUpdate plugin


web/client/components/misc/progressbars/OverlayProgressBar/OverlayProgressBar.jsx
```diff
@@ -0,0 +1,51 @@
+const PropTypes = require('prop-types');
```
mbarto: For new files, let's move this in the right place, please.

web/client/epics/automapupdate.js
```diff
@@ -0,0 +1,72 @@
...
+ * @memberof epics.automapupdate
+ * @return {external:Observable}
+ */
+
+const manageAutoMapUpdate = action$ =>
+    action$.ofType(MAP_CONFIG_LOADED)
+        .switchMap((mapConfigLoaded) =>
+            action$.ofType(MAP_INFO_LOADED)
+                .switchMap((mapInfoLoaded) => {
+                    const version = mapConfigLoaded.config && mapConfigLoaded.config.version || 1;
```
offtherailz: use selectors

web/client/actions/security.js
```diff
@@ -67,10 +68,15 @@ function logoutWithReload() {
 }
 
 function login(username, password) {
-    return (dispatch) => {
-        AuthenticationAPI.login(username, password).then((response) => {
+    return (dispatch, state) => {
+        return AuthenticationAPI.login(username, password).then((response) => {
```
offtherailz: This refreshes can go easily in an epic. Doing it here can cause unwanted side effects.

You should monitor the login/logout actions to reload MapInfo, until you change page.

web/client/actions/notifications.js
```diff
@@ -91,6 +91,15 @@ function clear() {
 }
 
 /**
+ * Dispatch a custom action on callback
+ * @memberof actions.notifications
+ * @returns {object}     action
+ */
+function dispatchAction(action) {
```
allyoucanmap: Yes, it's needed because we need to send custom action when the notification button has been toggled

web/client/epics/automapupdate.js
```diff
@@ -0,0 +1,95 @@
...
+                    const canEdit = mapInfoLoaded.info && mapInfoLoaded.info.canEdit || false;
+                    let layers = mapConfigLoaded.config && mapConfigLoaded.config.map && mapConfigLoaded.config.map.layers && mapConfigLoaded.config.map.layers.filter((l) => l.type === 'wms' && l.group !== 'background') || [];
+                    const options = {bbox: true, search: true, dimensions: true, title: true};
+                    return version < 2 && canEdit ?
```
MV88: change title to false

PR #2078, opened 2017-08-03
Changed cesium background color to white

MV88:
see https://github.com/geosolutions-it/webmapper-halliburton/issues/319


web/client/components/map/cesium/Map.jsx
```diff
@@ -144,6 +145,9 @@ let CesiumMap = React.createClass({
     },
     render() {
         const map = this.map;
+        if (this.map) {
+            this.map.scene.globe.baseColor = Cesium.Color.WHITE;
```
offtherailz: use a prop with defaults

PR #2107, opened 2017-08-10
fix #2096 restored spatial filter geometry if present when back to queryform

### LINKED ISSUE #2096: Show spatial filter on the map when re-open advanced search
When you generate a spatial filter the geometry is shown on the map until you search, then it disappears correctly. 
If you re-open the advanced search form, the spatial filter is selected in the form but it is not visible on the map. 
We should make it always visible when the query form is open and a spatial filter is selected.

MV88:
fix #2096

### DIFF
--- a/web/client/epics/featuregrid.js
+++ b/web/client/epics/featuregrid.js
@@ -33,7 +33,7 @@ const {SORT_BY, CHANGE_PAGE, SAVE_CHANGES, SAVE_SUCCESS, DELETE_SELECTED_FEATURE
 const {TOGGLE_CONTROL, resetControls} = require('../actions/controls');
 const {setHighlightFeaturesPath} = require('../actions/highlight');
 const {refreshLayerVersion} = require('../actions/layers');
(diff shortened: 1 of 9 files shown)


web/client/epics/featuregrid.js
```diff
@@ -424,9 +425,21 @@ module.exports = {
             return Rx.Observable.of(setControlProperty("drawer", "enabled", false), toggleTool("featureCloseConfirm", false));
         }),
     onOpenAdvancedSearch: (action$, store) =>
-        action$.ofType(OPEN_ADVANCED_SEARCH).switchMap(() =>
-            Rx.Observable.of(
+        action$.ofType(OPEN_ADVANCED_SEARCH).switchMap(() => {
```
offtherailz: move into query panel. Check if you can intercept better when the query panel is shown.

MV88:
Actually not TOGGLE_CONTROL is verified in order to redraw the spatialfilter because it is not used anywhere. If you prefer i can add also that case preventing future troubles

offtherailz:
Please add also TOGGLE_CONTROL check, thanks.
We should provide a utility function to use with let. 
Something like: 
```
action$.let(check_control("querypanel"))
```
What do you think.

PR #2113, opened 2017-08-21
Fixes #2083: removed escaping of single quotes when saving

### LINKED ISSUE #2083: MapStore 2 escapes single quotes when saving
Try to change the title of a layer of the map as: 
```
CITTA' 
```
The map title will be converted into: 
```
"title": "CITTA\' "
```
That is not a valid JSON (because of the \').
Probably because of this: 
https://github.com/mathiasbynens/jsesc#quotes

mbarto:


web/client/api/GeoStoreDAO.js
```diff
@@ -29,6 +29,10 @@ let parseUserGroups = (groupsObj) => {
     return groupsObj.User.groups.group.filter(obj => !!obj.id).map((obj) => _.pick(obj, ["id", "groupName", "description"]));
 };
 
+const encodeContent = function(content) {
+    return jsesc(content, {json: true, wrap: false, quotes: 'backtick'});
```
offtherailz: So the backtick will be escaped? Is there a way to disable quotes escaping totally ?

mbarto: Not with jsesc (unless we patch it). The options are:
 * forking and making a PR to the project
 * reimplementing the escaping code only for unicode chars, without using this lib

offtherailz: So if a user insert a backquote in a resource for instance, to break everything. We are planning to use markdown somewere for rich text, where backquote is part of the syntax.
I still didn't get why we have to use that lib. Did we do all the checks, as explained here? 
https://www.w3.org/International/questions/qa-changing-encoding

mbarto: Yes, it seems to be working only if we encode unicode chars with hex sequences. jsesc is doing this, but also does more and you cannot disable everything (like quote encoding). As I said, we can:
 * implement a flag in jsesc to only do unicode chars encoding
 * implement in MS2 a piece of code doing only that

offtherailz: This pull request moves the problem to a less used char. I think we should solve the problem at all.

PR #2121, opened 2017-08-25
Fixed #2103 disabled edit and search tool in 3d and closed querypanel

### LINKED ISSUE #2103: Disable featuregrid editing and spatial filter with Cesium
Featuregrid editing doesn't work with cesium and also the spatial filter.

We can go in two directions.
- remove the feature grid when switching to cesium (this will also remove the possibility to reach the queryform)
- just disable/remove some functionalities like spatial filter and edit mode

MV88:
see #2103

Also closed query panel if it was opened and the user switches to 3D mode.

BTW, if a button is disabled it cannot have a tooltip, do you really want it? @offtherailz

### DIFF
--- a/web/client/components/data/featuregrid/toolbars/Toolbar.jsx
+++ b/web/client/components/data/featuregrid/toolbars/Toolbar.jsx
@@ -16,19 +16,19 @@ const getSaveMessageId = ({saving, saved}) => {
     return "featuregrid.toolbar.saveChanges";
 };
 
-module.exports = ({events = {}, mode = "VIEW", selectedCount, hasChanges, hasGeometry, hasNewFeatures, isSimpleGeom, isDrawing = false, isEditingAllowed, saving = false, saved = false, isDownloadOpen, isColumnsOpen, disableToolbar, isCesium} = {}) =>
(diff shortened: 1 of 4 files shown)


web/client/components/data/featuregrid/toolbars/Toolbar.jsx
```diff
@@ -16,19 +16,19 @@ const getSaveMessageId = ({saving, saved}) => {
...
 };
 
-module.exports = ({events = {}, mode = "VIEW", selectedCount, hasChanges, hasGeometry, hasNewFeatures, isSimpleGeom, isDrawing = false, isEditingAllowed, saving = false, saved = false, isDownloadOpen, isColumnsOpen, disableToolbar} = {}) =>
+module.exports = ({events = {}, mode = "VIEW", selectedCount, hasChanges, hasGeometry, hasNewFeatures, isSimpleGeom, isDrawing = false, isEditingAllowed, saving = false, saved = false, isDownloadOpen, isColumnsOpen, disableToolbar, isCesium} = {}) =>
     (<ButtonGroup id="featuregrid-toolbar" className="featuregrid-toolbar featuregrid-toolbar-margin">
         <TButton
             id="edit-mode"
             tooltip={<Message msgId="featuregrid.toolbar.editMode"/>}
-            disabled={disableToolbar}
+            disabled={disableToolbar || isCesium}
```
mbarto: Components do not have to be aware of cesium. disableToolbar is enough, set it accordingly in the plugin

Issue #2136, opened 2017-09-04
Feedback on new footer


simboss:
Some feedback inline below; let's discuss and the schedule eventual tasks that come out of it.

- [x] Not sure about the gray aspect. At a second look is not probably the gray but the fact it does not really look in synch with the L&F of the other components: it misses profondity (it looks too flat) and it is easy to confuse it with the browser own toolbar (see below).
- [x] The mouse icon: too small, no tooltip, difficult to understand what it does!
- [x] Labels can be improved; mouseCoordinates and mousePositionCRS, wait shouldn't this be mouseCoordinates and mouseCoordinatesCRS or mousePosition and mousePositionCRS? Maybe we want shorter names? Let's think about this.
- [x] The GetFeatureInfo is now hidden, should we make it on by default? This would be important to have right away, I believe; we are confusing users right now IMHO.

@tdipisa Check my feedback and let's decide together what we do with this.

![ms2_1](https://user-images.githubusercontent.com/693880/30036614-4037207e-91b5-11e7-89ba-3dd87bcd29c2.jpg)

tdipisa:
In latest versions we have a reviewed footer. Closed for inactivity. We will open a new issue if needed io the future

PR #2141, opened 2017-09-06
Fixes #2023: improvements to the footer toolbar tools

### LINKED ISSUE #2023: Add a new toolbar to clean up the lower right corner area
A complete review of the lower right corner area is needed adding a new toolbar that contains only useful controls:

- the preview map control should be removed as default 
- the info tool should be moved into the right toolbar

The new toolbar contains:

- scale and scalebar
- coordinates
- coordinate reference system

tools-footer

mbarto:
* showLabels: false css fixes
 * google attribution not visible
From #2136 
 * tooltip on mouse coords enable/disable
 * Identify tool always visible in toolbar

### DIFF
--- a/web/client/localConfig.json
+++ b/web/client/localConfig.json
@@ -207,7 +207,8 @@
                 },
                 "override": {
                   "Toolbar": {
-                    "position": 11
+                    "position": 11,
+                    "alwaysVisible": true
                   }
                 }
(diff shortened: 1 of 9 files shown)


Gnafu:
Does this pull request fixes the mobile layout as well?

PR #2163, opened 2017-09-11
Closes #1639 Extend max leaflet zoom

### LINKED ISSUE #1639: Leaflet Map cannot be zoomed after scale 1:2257
The bug can be replicated using the scalebar.
This works fine using openlayers.

kappu72:
Closes #1639.
Understanding Scales and Zooming in MapStore2 (Leaflet)
Leflet part:
In the Leaflet library scales are calculated from the map minZoom and maxZoom properties.
These properties can be passes as options at map creation.
If ommitted are computed by taking the maximum maxZoom of the all layers and the minimum minZoom of the all layers.
Defautl value for Leaflet.TileLayer class are maxZoom = 18 , minZoom = 0.
If  the map current zoom is bigger then layer's maxZoom the layer isn't rendered.
But the layer has also another option,  the maxNativeZoom property.
When the map current zoom is bigger then maxNativeZoom but smaller then maxZoom, the layer tiles are stretched and not requested to the server.

MapStore2 part:
In MapStore2 if the resolutions property is ommitted this is computed by default from 0 to 23. 
The scales in the scale component are computed from Map.resolutions property.
So we have as default 24 zooms level from 0 to 23.
By default the leflet map maxZoom option is setted to 23 but can be overridden in the mapOptions object.
This setting  syncs Leflet maxZoom whit mapStore2 maxZoom.
All layers also has the maxZoom property setted to 23 and the maxNativeZoom properly setted.
This allow users to zoom until 23 and layers are always shown.
If the layer dosen't have that resolution the tiles are stretched.
Google layer is an exception, it doesn't works after zoom = 20. (same as in openlayers)
If resolutions are changed It's important to reconfigure the map maxZoom and each layer maxZoom property.

### DIFF
--- a/web/client/components/map/leaflet/plugins/GoogleLayer.js
+++ b/web/client/components/map/leaflet/plugins/GoogleLayer.js
@@ -8,7 +8,10 @@
 
 var Layers = require('../../../../utils/leaflet/Layers');
 var Google = require('leaflet-plugins/layer/tile/Google');
-
+Google.prototype._checkZoomLevels = function() {
+    // Avoid map zoom setting when current zoom is greatr then  the google's max zoom
+};
 Layers.registerType('google', (options) => {
-    return new Google(options.name, {zoomOffset: options.zoomOffset || 0});
+    return new Google(options.name, {zoomOffset: options.zoomOffset || 0, maxNativeZoom: options.maxNativeZoom || 18,
+        maxZoom: options.maxZoom || 20});
 });
--- a/web/client/components/map/leaflet/Map.jsx
+++ b/web/client/components/map/leaflet/Map.jsx
@@ -90,7 +90,7 @@ class LeafletMap extends React.Component {
             doubleClickZoom: false,
             boxZoom: false,
             tap: false
-        }, this.props.mapOptions, this.crs ? {crs: this.crs} : {});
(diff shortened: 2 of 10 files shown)


web/client/components/map/leaflet/plugins/GoogleLayer.js
```diff
@@ -8,7 +8,10 @@
...
 var Layers = require('../../../../utils/leaflet/Layers');
 var Google = require('leaflet-plugins/layer/tile/Google');
-
+Google.prototype._checkZoomLevels = function() {
+    // Avoid map zoom setting when current zoom is greatr then  the google's max zoom
+};
 Layers.registerType('google', (options) => {
-    return new Google(options.name, {zoomOffset: options.zoomOffset || 0});
+    return new Google(options.name, {zoomOffset: options.zoomOffset || 0, maxNativeZoom: options.maxNativeZoom || 18,
+        maxZoom: options.maxZoom || 20});
```
mbarto: Shouldn't this be 23?

kappu72: Google layer is the only one that will not zoom more then 20. So here we have to decide if is better to show wrong tile, (set maxZoom = 23) or nothing (maxZoom= 20).
Let me know what you prefer.

mbarto: Ok, I think 20 is fine, then

PR #2164, opened 2017-09-11
Fixes #2148 removed unwanted safari's zoom 

### LINKED ISSUE #2148: 3D On Mobile with IPhone is hard to use
User's feedback again.

Using the 3D version on the mobile (i.e. IPhone here) is very difficult. The zoom in and out with double pinch interact badly with the double pinch of the browse and instead of zooming the map we get the browser zoom (see below).

**Before**
file 07-09-17 15 11 54

**After**
file 07-09-17 15 11 40

I have no idea if this works any better on Android, but on IPhone I would rather not have it enabled by default.

MV88:
#2148 it's just removed from mobile config for the moment

### DIFF
--- a/web/client/components/map/cesium/Map.jsx
+++ b/web/client/components/map/cesium/Map.jsx
@@ -56,6 +56,12 @@ class CesiumMap extends React.Component {
     state = { };
 
     componentWillMount() {
+        /*
+         this prevent the Safari browser to zoom and mess up with the view.
+         added only for Safari's broswers (mobile and not) bescause from safari 10 it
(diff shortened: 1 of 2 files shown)


web/client/components/map/cesium/Map.jsx
```diff
@@ -109,6 +120,8 @@ class CesiumMap extends React.Component {
         this.clickStream$.complete();
         this.pauserStream$.complete();
         this.hand.destroy();
+        // see comment in componentWillMount
+        document.removeEventListener('gesturestart', function() {});
```
offtherailz: You have to refer the same function to correctly unregister event listener.

PR #2166, opened 2017-09-11
fix #2151 removed full screen button if safari is detected

### LINKED ISSUE #2151: Full Screen button does not work on IPhone
Not sure this happens on Android but on IPhone it does nothing, see below.

**Nothing happens :( **
file 07-09-17 15 17 20

MV88:
#2151

### DIFF
--- a/web/client/components/app/StandardApp.jsx
+++ b/web/client/components/app/StandardApp.jsx
@@ -13,6 +13,7 @@ const proj4 = require('proj4');
 const {changeBrowserProperties} = require('../../actions/browser');
 const {loadLocale} = require('../../actions/locale');
 const {localConfigLoaded} = require('../../actions/localConfig');
+const {setControlProperty} = require('../../actions/controls');
 const {loadPrintCapabilities} = require('../../actions/print');
 
 const ConfigUtils = require('../../utils/ConfigUtils');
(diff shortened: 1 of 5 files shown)


web/client/components/app/StandardApp.jsx
```diff
@@ -105,6 +106,11 @@ class StandardApp extends React.Component {
     }
     init = (config) => {
         this.store.dispatch(changeBrowserProperties(ConfigUtils.getBrowserProperties()));
+        if (navigator.userAgent.indexOf('Safari') !== -1 && navigator.userAgent.indexOf('Chrome') === -1) {
```
mbarto: can't we amend the browser state slice with safari info, instead of doing this?

web/client/components/buttons/FullScreenButton.jsx
```diff
@@ -44,6 +45,7 @@ class FullScreenButton extends React.Component {
         help: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
         glyphicon: PropTypes.string,
         active: PropTypes.bool,
+        hide: PropTypes.bool,
```
mbarto: Can we use disablePluginIf?

MV88: sure, it is better

Issue #2062, opened 2017-07-24
Unavailable layers in TOC


tdipisa:
When a layer is unavailable in TOC there's a ban-circle icon to represents the related status.
Anyway the layer should not be requested when the user perform a pan or zoom (for example if the layer is missing in the WMS server) and should be disabled during printing (in order to avoid white printed maps).

Issue #2061, opened 2017-07-24
Intermittent icon status in TOC when loading


tdipisa:
The animated gif below to clarify the issue:

![issue](https://user-images.githubusercontent.com/1280027/28520595-727ef8f6-7070-11e7-8a78-018e1c502405.gif)

This issue could be fixed during the TOC review ( issue #2025 )

Issue #1502, opened 2017-02-24
TOC should manage long layers name


MV88:
on the left the actual situation - on the right a possible solutions
![image](https://cloud.githubusercontent.com/assets/11991428/23299428/d6e401ac-fa81-11e6-9b42-ef04422c1f15.png)

If you agree I can prepare a PR to fix this.

I also noticed that the name of the layer is not perfectly aligned with the icons
![image](https://cloud.githubusercontent.com/assets/11991428/23299612/7c0c7d58-fa82-11e6-9ea6-ec3a91e5b7e3.png)

simboss:
Is this a bug?

I mean the name on the left seems properly truncated to me. I am not so sure that showing the entire name in the TOC will help when that name is too long since it will make the TOC huge difficult to use.

MV88:
Yes, but can exist some cases when two or more layers start with the same text. Just wanted to highlight this situation.

Gnafu:
Maybe we can expand the name when the "additional tools" icon is clicked

simboss:
Good point @Gnafu.

The default view should be compact, that is my point.

MV88:
Ok, let's move with the Gnafu's solution. 
Also limit the name to three rows when the "additional tools" is closed. 
When it is expanded we can show the entire name of the layer.

Let's wait the theme integration for the alignment of the group and the icons.

PR #2170, opened 2017-09-12
Fixes #2025 TOC review, #1502 TOC should manage long layers name, #2062 Unavailable layers in TOC, #2061 Intermittent icon status in TOC when loading

### LINKED ISSUE #2025: TOC Review
The TOC component must be reviewed:

- [ ] Add a main toolbar on top, here the available actions should be
- [ ] Remove the action buttons available per layers: the layer must be selected first and then the top toolbar can be used to perform the available actions for the layer 
- [ ] The internal layout of the TOC must be accordingly reviewed to make also more clear the view ([here](http://demo.geo-solutions.it/mockups/mapstore2/toc/) a mockup)

allyoucanmap:
#2025 TOC review

- Added main toolbar to manage layer, group, multiple layer and multiple group
- Added filter form to filter layers by title
- Updated layout and styles of TOC
- Removed tools from single layer component
- Added Title map name on top of the TOC 
- Fixed nested groups settings management
- Now the expanded status of collapsible panel (Legend Tool) is saved

screenshot from 2017-09-12 17-06-18

#1502 TOC should manage long layers name
screenshot from 2017-09-12 16-50-56

#2062 Unavailable layers in TOC
After a loading error, the percentage of errors (due to filed tile request) set a flag
'Error' if the percentage errors is greater than 50% (or equal)
in this status the layer is invisible and can not be printed
screenshot from 2017-09-12 16-53-15
'Warning' if the percentage errors is less than 50%
in this status the layer is visible bat can not be printed
screenshot from 2017-09-12 16-53-57

#2061 Intermittent icon status in TOC when loading
aaa


web/client/components/TOC/fragments/Title.jsx
```diff
@@ -16,21 +16,34 @@ class Title extends React.Component {
...
+        currentLocale: 'en-US',
+        filterText: ''
     };
 
+    getFilteredTitle = (title) => {
+        const splitTitle = title.toLowerCase().split(this.props.filterText.toLowerCase());
+
+        return this.props.filterText ? splitTitle.reduce((a, b, idx) => {
+            if (idx === splitTitle.length - 1) {
+                a.push(b);
```
mbarto: side effects here

offtherailz: yes use = a = [...a, b]

web/client/utils/LayersUtils.js
```diff
@@ -71,10 +71,64 @@ const isSupportedLayer = (layer, maptype) => {
...
+
+const getGroupNodes = (node) => {
+    if (node && node.nodes) {
+        return node.nodes.reduce((a, b) => {
+            let nodes = [].concat(a);
+            if (b.nodes) {
+                nodes = a.concat(getGroupNodes(b));
+            }
+            if (isString(b)) {
+                nodes.push(b);
```
offtherailz: possible side effect. Avoid side effect as much as you can. This function returns the new nodes, so it doesn't have to mutate the existing value.

PR #2176, opened 2017-09-13
Fixes #2172 viewport spatial filter creation

### LINKED ISSUE #2172: Viewport query filter uses wrong srsName
Viewport option uses srsName="EPSG:900913" but then writes the polygon in EPSG:4326

MV88:
#2172


web/client/epics/wfsquery.js
```diff
@@ -253,26 +256,43 @@ const wfsQueryEpic = (action$, store) =>
             );
         });
 
+
+function validateExtent(extent, proj) {
+    const minxEpsg4326 = -180.0;
+    const maxxEpsg4326 = 180.0;
+    if (extent[0] <= CoordinatesUtils.reproject([minxEpsg4326, 0], "EPSG:4326", proj ).x) {
```
offtherailz: try to use the bbox inverted.

PR #2192, opened 2017-09-19
Fix #2172 Viewport query filter uses wrong srsName

### LINKED ISSUE #2172: Viewport query filter uses wrong srsName

allyoucanmap:
## Description
Added a control on leaflet map to change srs 'EPSG:4326'.
Openlayers return a srs 'EPSG:900913' with correct values.
Due to values positioned outside the worlds limits, here has been introduced a function to split in two multipolygon the bound and extent on the Viewport spatial filter.
screenshot from 2017-09-19 12-20-04

## Requirements

**What kind of change does this PR introduce?** (check one with "x")
 - [x] Bugfix

issue #2172

**What is the new behavior?**
- Added a control on leaflet map to change srs 'EPSG:4326' (Viewport)
- split bound in two polygon when view contains 180 -180 values (Viewport)

**Does this PR introduce a breaking change?** (check one with "x")
- [x] No

**Other information**:

### DIFF
--- a/web/client/epics/wfsquery.js
+++ b/web/client/epics/wfsquery.js
@@ -254,6 +254,13 @@ const wfsQueryEpic = (action$, store) =>
             );
         });
 
+/**
+ * Generate a spatial filter geometry from the view bounds
+ * @memberof epics.wfsquery
+ * @param {external:Observable} action$ manages `SELECT_VIEWPORT_SPATIAL_METHOD` and `CHANGE_MAP_VIEW`
+ * @return {external:Observable}
+ */
+
 const viewportSelectedEpic = (action$, store) =>
     action$.ofType(SELECT_VIEWPORT_SPATIAL_METHOD, CHANGE_MAP_VIEW)
         .switchMap((action) => {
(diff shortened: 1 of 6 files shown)


web/client/epics/wfsquery.js
```diff
@@ -261,21 +263,11 @@ const viewportSelectedEpic = (action$, store) =>
             if (action.type === SELECT_VIEWPORT_SPATIAL_METHOD ||
                 action.type === CHANGE_MAP_VIEW && spatialFieldMethodSelector(store.getState()) === "Viewport") {
 
+                const projection = isLeaflet(store.getState()) ? 'EPSG:4326' : map.projection;
```
mbarto: bounds should also contain its srs, if it doesn't we should add it and not have knowledge of the mapping lib in use

web/client/utils/CoordinatesUtils.js
```diff
@@ -56,6 +57,100 @@ let crsLabels = {
     "EPSG:3857": "EPSG:3857"
 };
 
+const normalizePoint = (point) => {
```
mbarto: You moved this, did you change anything?

web/client/utils/CoordinatesUtils.js
```diff
@@ -56,6 +57,100 @@ let crsLabels = {
...
+    return {
+        x: point.x || 0.0,
+        y: point.y || 0.0,
+        srs: point.srs || 'EPSG:4326'
+    };
+};
+
+const reproject = (point, source, dest, normalize = true) => {
```
mbarto: You moved this, did you change anything?

web/client/utils/CoordinatesUtils.js
allyoucanmap: no, i didn't change the function

web/client/utils/CoordinatesUtils.js
allyoucanmap: no, i didn't change the function

PR #2197, opened 2017-09-20
Closes #1918. Filter WMS layer in sync with FeatureGrid

### LINKED ISSUE #1918: Add optional 'show on map' support to feature grid
The 'show on map' option is required into the FeatureGrid toolbar as action to show on map the results retrieved in the current page.

kappu72:
Added sync on map button to toolbar of featuregrid plugin.
When active the selected layer showed on map is reloaded with the properly configured cql filter.
Is it possible to esclude from sync the quick filter query, adding disableQuickFilterSync = true in featuregrid state.

Docs:
state param: 
1) query.syncWmsFilter default false. if active the current filterObj is pasted to 
 the featuregrid.selectedLayer options.
2) featuregrid.disableQuickFilterSync default false. When true syncing of quick filters is avoided

Leaflet, Ol and Cesium WMS/T plugins generate properly configurate cql_filter if filterObj is present in layer.options.
Same is done in mapinfo/ wms and wmst utils during the creation of the request params.

### DIFF
--- a/web/client/epics/featuregrid.js
+++ b/web/client/epics/featuregrid.js
@@ -163,13 +163,6 @@ const createInitialQueryFlow = (action$, store, {url, name} = {}) => {
     );
 };
 
-const featureGridIsClosed = ( action$ ) =>
- action$.ofType(ASK_CLOSE_FEATURE_GRID_CONFIRM)
-      .switchMap(() =>
-         action$.ofype(CLOSE_FEATURE_GRID)
-             .takeUntil(action$.ofype(FEATURE_GRID_CLOSE_CONFIRMED))
-             .take(1));
-
 // Create action to add filter to wms layer
 const addFilterToWMSLayer = (layer, filter) => {
     return changeLayerProperties(layer, {filterObj: filter});
@@ -539,17 +532,6 @@ module.exports = {
         action$.ofType(OPEN_FEATURE_GRID)
         .switchMap(() => {
             return Rx.Observable.of(purgeMapInfoResults());
-        }),
-    /**
-     * Close sync filter wms control on faturegrid close,
-     * not on advanced search open
-     */
-    disableSyncWmsOnGridClose: (action$, store) =>
-        action$.ofType(OPEN_FEATURE_GRID).
-        switchMap(() => {
-            return action$.let(featureGridIsClosed)
-            .filter(() => isSyncWmsActive(store.getState()))
-            .map(() => removeFilterFromWMSLayer(store.getState()));
         }),
     /**
      * start sync filter with wms layer
--- a/web/client/epics/featuregrid.js
+++ b/web/client/epics/featuregrid.js
@@ -18,7 +18,7 @@ const {changeDrawingStatus, GEOMETRY_CHANGED} = require('../actions/draw');
 const requestBuilder = require('../utils/ogc/WFST/RequestBuilder');
 const {findGeometryProperty} = require('../utils/ogc/WFS/base');
 const {setControlProperty} = require('../actions/controls');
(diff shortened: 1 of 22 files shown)


web/client/plugins/featuregrid/panels/index.jsx
```diff
@@ -35,8 +35,9 @@ const Toolbar = connect(
         disableToolbar: state => state && state.featuregrid && state.featuregrid.disableToolbar,
         disableDownload: state => (resultsSelector(state) || []).length === 0,
         isDownloadOpen: state => state && state.controls && state.controls.wfsdownload && state.controls.wfsdownload.enabled,
+        isSyncActive: state => state && state.controls && state.controls.syncwmsfilter && state.controls.syncwmsfilter.enabled,
         isColumnsOpen: state => state && state.featuregrid && state.featuregrid.tools && state.featuregrid.tools.settings,
-        isSearchAllowed: (state) => !isCesium(state),
+        isSearchAllowed: (state) => true || !isCesium(state),
```
mbarto: this is always true

web/client/epics/featuregrid.js
```diff
@@ -470,6 +483,17 @@ module.exports = {
...
         }),
+    /**
+     * Close sync filter wms control on faturegrid close,
+     * not on advanced search open
+     */
+     toggleSyncControl: (action$, store) =>
+         action$.ofType(OPEN_ADVANCED_SEARCH, CLOSE_FEATURE_GRID)
+            .filter(() => isSyncWmsActive(store.getState()))
+            .scan((acc, cur) => {
+                return cur.type === CLOSE_FEATURE_GRID && acc.type !== OPEN_ADVANCED_SEARCH ? toggleControl("syncwmsfilter") : cur;
```
offtherailz: This intercepts all the : 
 - CLOSE_FEATURE_GRID --> CLOSE_FEATURE_GRID
Instead you want to stop sync on feature grid close. You should see the action ASK_CLOSE_FEATURE_GRID_CONFIRM is triggered on close. There is an epic that intercepts this and follows a close grid, if no confirm is required. If the user see the confirm and deny it emits FEATUREGRID:FEATURE_GRID_CLOSE_CONFIRMED (strange name, maybe let's check and change the name of this action later. 

You can intercept this sequence of actions: 
ASK_CLOSE_FEATURE_GRID_CONFIRM --> FEATUREGRID:CLOSE_GRID -->  take until FEATUREGRID:FEATURE_GRID_CLOSE_CONFIRMED
something like this should work: 
```js
      .filter( () => isSyncWmsActive(store.getState()))
      action$.ofType(ASK_CLOSE_FEATURE_GRID_CONFIRM)
      .switchMap(() => 
         action$.ofype(FEATUREGRID:CLOSE_GRID)
             .takeUntil(action$.ofype(FEATUREGRID:FEATURE_GRID_CLOSE_CONFIRMED))
             .take(1)
             .map(() => stopSync() )
```

You could make a function to use in less. Should be something like this:
```js
const featureGridIsClosed = ( action$ ) => 
 action$.ofType(ASK_CLOSE_FEATURE_GRID_CONFIRM)
      .switchMap(() => 
         action$.ofype(FEATUREGRID:CLOSE_GRID)
             .takeUntil(action$.ofype(FEATUREGRID:FEATURE_GRID_CLOSE_CONFIRMED))
             .take(1)
```

And use it with let
```js
   // .. intercept feature grid first opening
            .switchMap ( 
                action$.let(fetaureGridIsClosed).map(() => stopSync() )
```

web/client/epics/featuregrid.js
```diff
@@ -470,6 +483,17 @@ module.exports = {
...
+            }, {type: ''})
+            .filter((a) => a.type === TOGGLE_CONTROL),
```
offtherailz: Then trigger a proper action

web/client/plugins/featuregrid/panels/index.jsx
kappu72: Sorry Mauro, I used this trick to test the sync with Cesium and I forgot to remove!

PR #2230, opened 2017-09-27
Annotations main view

### LINKED ISSUE #2157: Annotations main view
Implement the annotations list as Card layout. The annotations list resides in a side panel that can be accessed by the Options menu.

mbarto:
## Description
This pull request completes the annotations functionalities, adding the main panel with annotations list / detail / editing.

## Issues
 - Fix #2157

 - [x] Feature

The annotations panel is not open when clicking on the related burger menu option

**What is the new behavior?**
The panel is open and working as expected.

 - [x] No

**Other information**:


web/client/epics/annotations.js
```diff
@@ -58,34 +59,38 @@ const mergeGeometry = (features) => {
...
+        geometry: action.geometry,
+        style: action.style
+    };
 };
 
 module.exports = (viewer) => ({
     addAnnotationsLayerEpic: (action$, store) =>
     action$.ofType(MAP_CONFIG_LOADED)
         .switchMap(() => {
             const annotationsLayer = head(store.getState().layers.flat.filter(l => l.id === 'annotations'));
```
offtherailz: use layers selector and then filter

web/client/plugins/Annotations.jsx
```diff
@@ -10,53 +10,41 @@ const React = require('react');
...
+    onSetStyle: setStyle,
+    onDeleteGeometry: removeAnnotationGeometry
+})(require('../components/mapcontrols/annotations/AnnotationsEditor'));
 
 const AnnotationsInfoViewer = connect((state) => ({
-    fields: state.annotations && state.annotations.fields,
+    config: state.annotations && state.annotations.config,
     editing: state.annotations && state.annotations.editing,
     drawing: state.annotations && !!state.annotations.drawing,
     styling: state.annotations && !!state.annotations.styling,
```
offtherailz: All state slices are replicated in many places. You could simply use createSelector

web/client/plugins/Annotations.jsx
```diff
@@ -74,7 +62,112 @@ const AnnotationsInfoViewer = connect((state) => ({
...
+
+const Annotations = connect((state) => ({
+    removing: state.annotations && state.annotations.removing,
+    mode: state.annotations && state.annotations.editing && 'editing' || (state.annotations && state.annotations.current && 'detail' || 'list'),
+    editor: AnnotationsEditor,
+    config: state.annotations && state.annotations.config,
+    annotations: state.layers && state.layers.flat && head(state.layers.flat.filter(l => l.id === 'annotations')) && head(state.layers.flat.filter(l => l.id === 'annotations')).features || [],
+    current: state.annotations && state.annotations.current || null,
+    editing: state.annotations && state.annotations.editing,
+    filter: state.annotations && state.annotations.filter || ''
```
offtherailz: same as above,  some of thees should go in a selector

web/client/plugins/Annotations.jsx
```diff
@@ -10,53 +10,41 @@ const React = require('react');
...
-const AnnotationsPlugin = connect((state) => ({
-    removing: state.annotations && state.annotations.removing,
-    editing: state.annotations && !!state.annotations.editing
-}), {
-    onCancelRemove: cancelRemoveAnnotation,
-    onConfirmRemove: confirmRemoveAnnotation
-})(require('../components/mapcontrols/annotations/Annotations'));
+
+const AnnotationsEditor = connect((state) => ({
+    config: state.annotations && state.annotations.config,
```
offtherailz: use selector

offtherailz:
Some notes about usage:
 - While editing a new, if I close the annotation panel, it should ask confirm before closing. 
 - If I start drawing points and I click on thrash button, It seems like I'm able to create points (the tool is active and the map shows the same tooltip) but I can't. 
 - It is not clear how can I stop adding point. I propose to allow adding only one point each time and disable button when added. If a user want to add another point, he have to click again on the button.
 - I can not edit title or description when I clicked first on "add point". Maybe on click on the annotation panel you could disable editing
 - Is difficoult to exit from the edit mode, maybe we should allow to exit by clicking on the panel and/or hitting "ESC" (or right click). 
- With disabled featureinfo, hovering the mouse on annotation, I see I can click but I can't.
- Buttons are too near to the bottom of the page, so they are not too visible
- Errors should be more visible (e.g. we should use red color)

mbarto:
Done the following:
 - Added confirm before closing in editing mode (both on panel and identify)
 - restored correct drawing mode after geometry delete
 - set single point by default (multiGeometry enabled with configuration)
 - disabled edting fields during geometry drawing
- removed cursor pointer css from annotations markers when not in identify mode
- moved buttons on the top of the page
- changed css of errors to show them in red
- removed annotations layer when there are no more annotations on the map

About the following, this is my opinion:
 - Annotations layer should be always on top? NO
 - Does it make sense to show it in TOC? YES, because you may want to show/hide annotations

simboss:
I am happy to have annotations as a separate layer in the TOC.

In the future we might have the concept of per layer annotations.

offtherailz:
Good, it makes sense.
I'm going to review latest changes

offtherailz:
I found a regression with editing of multigeom. Try to edit 1 geome of an existing multigeom and save. The modified disappears. I checked and it is working on current stable.

PR #2246, opened 2017-09-29
Fix #2244 reset controls now skip correctly some controls

### LINKED ISSUE #2244: resetcontrols functionality always sets all the controls to false without considering the skip parameter
We need to fix this the controls.js reducer.

MV88:
## Description
Now reset controls rest correctly the controls. previously it was ignorign the "skip" param.

## Issues
 - Fix #2244

 - [x] Bugfix

controls are always reset to a false value and if we cannot avoid to reset one or more controls.

**What is the new behavior?**
now one or more controls can be skipped from resetting.

 - [x] No

### DIFF
--- a/web/client/reducers/__tests__/controls-test.js
+++ b/web/client/reducers/__tests__/controls-test.js
@@ -109,7 +109,7 @@ describe('Test the controls reducer', () => {
         const state = controls(
             {
                 c1: { enabled: true},
-                c2: { enabled: false},
+                c2: { enabled: true},
                 c3: { idonthaveenabledfield: "whatever"}
             }, {
                 type: RESET_CONTROLS,
--- a/web/client/reducers/__tests__/controls-test.js
+++ b/web/client/reducers/__tests__/controls-test.js
@@ -10,7 +10,17 @@ const expect = require('expect');
 const controls = require('../controls');
 const {TOGGLE_CONTROL, SET_CONTROL_PROPERTY, RESET_CONTROLS} = require('../../actions/controls');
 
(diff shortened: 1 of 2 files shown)


web/client/reducers/__tests__/controls-test.js
```diff
@@ -58,21 +68,60 @@ describe('Test the constrols reducer', () => {
...
         expect(state.c3.enabled).toNotExist();
+        expect(state.c3.idonthaveenabledfield).toExist();
+        expect(state.c3.idonthaveenabledfield).toBe("whatever");
+    });
+
+    it('reset the controls with skip c1,c3', () => {
+        const state = controls(
+            {
+                c1: { enabled: true},
+                c2: { enabled: false},
```
mbarto: The test would be better with true here

web/client/reducers/controls.js
```diff
@@ -67,7 +67,11 @@ function controls(state = {}, action) {
             })
         });
     case RESET_CONTROLS: {
-        const resetted = Object.keys(state).reduce((previous, controlName) => {
+        let newControls = Object.keys(state);
```
mbarto: There is always a way to use const:
const newControls = Object.keys(state).filter(c => (action.skip || []).indexOf(c) === -1)

MV88: ok...

web/client/reducers/__tests__/controls-test.js
MV88: right

PR #2264, opened 2017-10-05
Fixes #2227: shows error message on adding WMTS layer if the current …

### LINKED ISSUE #2227: Unable to load public WMTS
I was trying to load layers from [this public WMTS](https://tiles.maps.eox.at/wmts/1.0.0/WMTSCapabilities.xml) without any luck.

Specifically the Sentinel 2 cloudless, we should investigate where the problem is.

mbarto:
…SRS is not supported by the layer

## Description
Fixing WMTS SRS support in Catalog.

## Issues
 - Fix #2227

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

When a WMTS service supports a list of SRS, but a specific layer has a shorter list, the full SRS list is used to decide if the layers can be added to map.

**What is the new behavior?**
The supported SRS list is filtered for each layer, so that an error message is shown if the current map SRS is not supported for that layer.

 - [ ] Yes
 - [x] No

**Other information**:

### DIFF
--- a/web/client/utils/WMTSUtils.js
+++ b/web/client/utils/WMTSUtils.js
@@ -20,7 +20,7 @@ const WMTSUtils = {
         return matrixIds;
     },
     getMatrixIds: (matrix, srs) => {
-        return (isObject(matrix) && matrix[srs] || matrix).map((el) => el.identifier);
+        return ((isObject(matrix) && matrix[srs]) || isArray(matrix) || []).map((el) => el.identifier);
     },
     limitMatrix: (matrix, len) => {
         if (matrix.length > len) {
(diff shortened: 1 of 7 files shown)


web/client/utils/WMTSUtils.js
```diff
@@ -20,7 +20,7 @@ const WMTSUtils = {
         return matrixIds;
     },
     getMatrixIds: (matrix, srs) => {
-        return (isObject(matrix) && matrix[srs] || matrix).map((el) => el.identifier);
+        return ((isObject(matrix) && matrix[srs]) || isArray(matrix) || []).map((el) => el.identifier);
```
offtherailz: if `(isObject(matrix) && matrix[srs])` is true the `.map` iterates over an object.

web/client/utils/__tests__/CatalogUtils-test.js
```diff
@@ -77,6 +77,82 @@ describe('Test the CatalogUtils', () => {
 
     });
 
+    it('wmts with tilematrix filtered', () => {
```
offtherailz: test the Object use case, if it possible.

web/client/utils/WMTSUtils.js
mbarto: matrix is an object, matrix[srs] is an array

offtherailz: You are right. ok

PR #2288, opened 2017-10-16
Fixes #2290 Changes to search services API

### LINKED ISSUE #2290: Search Services should be able to return more info
Add a flag which allows to return the full response and not only an array of features.

MV88:
## Description
Search services were returning only the results without some other useful info like total results useful for pagination.

## Issues 
Fix 
- #2290

the search services return onl an array of the fethced data without returning for example the total features matched.

**What is the new behavior?**
now a flag enable the possibility to return the full response in order to fetch

 - [x] No

**Other information**:

### DIFF
--- a/web/client/api/searchText.js
+++ b/web/client/api/searchText.js
@@ -27,7 +27,11 @@ const defaultFromTextToFilter = ({searchText, staticFilter, blacklist, item, que
     filter = filter ? filter.concat(staticFilterParsed) : staticFilterParsed || null;
     return filter;
 };
-
+/*
+ * The API returns a promise for each search service.
+ * These search services have a particular option that specify how the response is returned.
+ * 'returnFullResponse' is a boolean option that if true a the full reponse is returned, otherwise an array o fearures.
+*/
 let Services = {
     nominatim: (searchText, options = {
         returnFullResponse: false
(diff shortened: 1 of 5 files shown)


web/client/api/searchText.js
```diff
@@ -27,12 +27,15 @@ const defaultFromTextToFilter = ({searchText, staticFilter, blacklist, item, que
     filter = filter ? filter.concat(staticFilterParsed) : staticFilterParsed || null;
     return filter;
 };
+
 let Services = {
-    nominatim: (searchText, options = {}) =>
+    nominatim: (searchText, options = {
```
offtherailz: if nominatim ignores it, you don't have to set a default. You can specify otherwise, documenting this file, how the result of the promise is shaped.

web/client/plugins/Search.jsx
```diff
@@ -99,14 +100,17 @@ const ToggleButton = require('./searchbar/ToggleButton');
  *        "sortBy": "ID",
  *        "srsName": "EPSG:4326",
  *        "maxFeatures": 4,
- *        "blackist": [... an array of strings to exclude from the final search filter ]
+ *        "blackist": [... an array of strings to exclude from the final search filter ],
+ *        "returnFullResponse": false
```
offtherailz: `returnFullResponse` is not a property that should be configured in search plugin. 
This example is wrong and seems to suggest to have more info, instead it breaks the functionality at all. 
Document the property in the right place.

PR #2302, opened 2017-10-18
Add read only mode on annotations

allyoucanmap:
## Description
Added read only mode on annotations

 - [x] Bugfix
 - [x] Feature

annotations cannot be disabled

**What is the new behavior?**
annotations can be disabled with readOnly flag

 - [x] Yes

**Other information**:

### DIFF
--- a/web/client/components/data/identify/viewers/JSONViewer.jsx
+++ b/web/client/components/data/identify/viewers/JSONViewer.jsx
@@ -25,7 +25,7 @@ class JSONViewer extends React.Component {
         const RowViewer = (this.props.layer && this.props.layer.rowViewer) || this.props.rowViewer || PropertiesViewer;
         return (<div className="mapstore-json-viewer">
                 {(this.props.response.features || []).map((feature, i) => {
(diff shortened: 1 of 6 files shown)


web/client/components/data/identify/viewers/JSONViewer.jsx
```diff
@@ -25,7 +25,7 @@ class JSONViewer extends React.Component {
         const RowViewer = (this.props.layer && this.props.layer.rowViewer) || this.props.rowViewer || PropertiesViewer;
         return (<div className="mapstore-json-viewer">
                 {(this.props.response.features || []).map((feature, i) => {
-                    return <RowViewer key={i} title={feature.id} exclude={["bbox"]} {...feature.properties}/>;
+                    return <RowViewer key={i} readOnly={feature.readOnly} title={feature.id} exclude={["bbox"]} {...feature.properties}/>;
```
mbarto: I think we should generalize this and pass down a more generic feature={feature}. For the future I would change the way properties from the features are passed down. I don't like the {...feature.properties} at all. For now, let's add feature as a property for the "full" object and let the annotations rowviewer use that to get the readonly value

Issue #2222, opened 2017-09-26
Circle spatial filter shows different selections on high denominator scales


allyoucanmap:
tested on leaflet.
to reproduce:
- select a WMS layer
- open feature grid
- select advanced search
- select circle spatial filter and draw on map
- click on search
- back to advanced search
now the circle has a different geometry (high denominator scales)

while drawing the filter
![circle1](https://user-images.githubusercontent.com/19175505/30850229-1fd7d728-a2a5-11e7-9853-abc74d83bf9a.png)

back to advanced search
![circle2](https://user-images.githubusercontent.com/19175505/30850231-20ff4b40-a2a5-11e7-862c-ad9fb72ab123.png)

difference between above images
![circle](https://user-images.githubusercontent.com/19175505/30850182-fc9defcc-a2a4-11e7-815e-597cd2d94827.png)

allyoucanmap:
Update: 
after back to advanced search and changed the selection details
circle geometry is a point on map (center)

![screenshot from 2017-09-28 14-51-50](https://user-images.githubusercontent.com/19175505/30967530-e10dc98a-a45c-11e7-8ffe-47adcca98f45.png)

offtherailz:
About the main issue, The cause is mainly this:
https://github.com/geosolutions-it/MapStore2/blob/8a70613ff895e04742a565c726ec10250324a07a/web/client/utils/CoordinatesUtils.js#L476
That doesn't properly fit the leaflet case. 
Take a look at [this](https://github.com/w8r/L.Circle.toPolygon) maybe it can fix the leaflet case. 
About the update in comment, i opened a separate issue for that #2321

MV88:
note: issue #2321 has been closed

MV88:
I found a solution for this issue, but it remains the problem while switching from ol to leaflet and viceversa beacuse of the different radius calculated. I'm opening an issue for that

PR #2326, opened 2017-10-24
Fix #2321. Keep circle on change

### LINKED ISSUE #2321: Editing of circle spatial filter details fails
Editing of circle spatial filter details makes the search and the UI not consistent. 

### Steps to reproduce
 - open http://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/leaflet/1740
 - Open the attribute table
 - Open advanced search
 - Select spatial filter type: circle
 - click on details
 - edit center or radius

#### Expected result 
the circle changes in radius and center

#### Effective result
 - When I edit the center or the radius I see a marker instead of the circle's moving on the map.
 - We I perform the search again I have no result. 
 - When I re-open the circle filter, I see an oval instead of a circle (maybe this is somehow correct, but it can be misleading)

offtherailz:
## Description
This pull request fixes 2 kind of issues in Advanced Search circle spatial filter: 
 - When you edit the circle's details the style doesn't change to marker
 - When you save the edited circle, the geometry is valid. 
## Issues
 - Fix #2321

 - [x] Bugfix

In advanced search, when you try to edit the circle's details, a marker appears on the map instead of circle. When save the changes, the geometry is not valid and so the search produces no results.

**What is the new behavior?**
Now you can change the details and the geometry remains a circle. 

 - [x] No

**Other information**:
The #2222 still exists. So the real geometry used for the filter is not the same of the one drown on the map.

### DIFF
--- a/web/client/components/data/query/GeometryDetails.jsx
+++ b/web/client/components/data/query/GeometryDetails.jsx
@@ -144,12 +144,20 @@ class GeometryDetails extends React.Component {
                 center.x - this.circle.radius, center.y - this.circle.radius,
                 center.x + this.circle.radius, center.y + this.circle.radius
             ];
-
+            let coordinates = CoordinatesUtils.calculateCircleCoordinates(center, this.circle.radius, 100);
+            if (!this.props.useMapProjection) {
+                // reproject into "EPSG:3857" to draw and circle coordinates
+                const projection = "EPSG:3857";
+                const tempCenter = CoordinatesUtils.reproject(center, "EPSG:4326", projection);
+                const tempCoordinates = CoordinatesUtils.calculateCircleCoordinates(tempCenter, this.circle.radius, 100);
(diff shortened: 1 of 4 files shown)


web/client/components/data/query/GeometryDetails.jsx
```diff
@@ -144,12 +144,20 @@ class GeometryDetails extends React.Component {
                 center.x - this.circle.radius, center.y - this.circle.radius,
                 center.x + this.circle.radius, center.y + this.circle.radius
             ];
-
+            let coordinates = CoordinatesUtils.calculateCircleCoordinates(center, this.circle.radius, 100);
+            if (this.props.geometry && this.props.geometry.projection === "EPSG:4326") {
```
mbarto: Can't we reproject from any projection that is not the current map projection to the current map projection?

offtherailz: I could try, but don't know if it works, because of #2222

PR #2359, opened 2017-11-03
Fixes #2322 Custom WMTS gridsets are not supported, Fixes #2342 Tiles errors management

### LINKED ISSUE #2322: Custom WMTS gridsets are not supported
Layers using a custom gridset are incorrectly displayed on the map, even if the gridset information is available in the GetCapabilities document.

To reproduce:
 - Open a map
 - Open the catalog
 - Select the GeoSolutions DEMO WMTS service
 - Search for tasmania_test layer and add it to the map

Expected result: the Tasmania polygons are placed south of Australia, with the correct size.

Actual result: the Tasmania polygons are displayed spanning from Canada to Antarctica.

The tasmania_test layer is set to have a custom gridset with a boundary smaller than the projection maximum bounds.

### LINKED ISSUE #2342: Tiles errors management
We should improve the management of errors generated from tile loading.
Now the layer is disabled if all the tile loaded on screen throw an error but in some cases it doesn't mean that the layer is broken.
e.g. When some zoom levels are disabled from server

allyoucanmap:
## Description
Now custom gridset are loaded an displayed at the right dimension.
About tasmania_test:
I tried this fix with another custom grid set and the world grid set and it works with both (right position and size).
I noticed also an issue with the same layer in WMS (tasmania_test).
Could it be that the gridset isn't configured correctly?
Could it be that on high scale denominator there's a loss of precision in the gridset pyramid?
- About last question I noticed that at small scale denominator the precision increases.

## Issues
 - Fix #2322
 - Fix #2342 

 - [x] Feature

Cutsom WMTS grids are displayed world wide.
#2322
#2342 

**What is the new behavior?**
Cutsom WMTS grids are displayed at the correct position.
Try to reload layers with errors on zoom.

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/map/leaflet/__tests__/Layer-test.jsx
+++ b/web/client/components/map/leaflet/__tests__/Layer-test.jsx
@@ -189,7 +189,13 @@ describe('Leaflet layer', () => {
             "name": "nurc:Arc_Sample",
             "group": "Meteo",
             "format": "image/png",
-            "tileMatrixSet": "EPSG:900913",
+            "tileMatrixSet": [
+                {
+                    "TileMatrix": [],
+                    "ows:Identifier": "EPSG:4326",
+                    "ows:SupportedCRS": "urn:ogc:def:crs:EPSG::900913"
+                }
+            ],
             "url": "http://sample.server/geoserver/gwc/service/wmts"
         };
         // create layers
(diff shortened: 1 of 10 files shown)


web/client/components/map/leaflet/__tests__/Layer-test.jsx
```diff
@@ -189,7 +189,13 @@ describe('Leaflet layer', () => {
             "name": "nurc:Arc_Sample",
             "group": "Meteo",
             "format": "image/png",
-            "tileMatrixSet": "EPSG:900913",
+            "tileMatrixSet": [
+                {
+                    "TileMatrix": [],
+                    "ows:Identifier": "EPSG:4326",
```
mbarto: Is this a typo?

PR #2372, opened 2017-11-08
Fix #2369 correct parsing of url fetched from GS removing authkey param

### LINKED ISSUE #2369: Url used for wfs request contains old params like ms2-authkey
### Description
Wfs url fetched from geoserver contains old authkey param that will invalid the request.
In the layer params there is the previous old authkey param that must not be saved.

- [x] Bug

*Browser Affected*
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*
Chrome 61.0.3163.100

*Steps to reproduce*
**Note:** this issue was thrown for project with custom authentication method like client C040
- enter c027 project homepage and login as admin
- open the map sociosanitarie
- open the TOC and the feature editor for the layer v_sociosanitarie

*Expected Result*
The feature grid correctly opened and populated with results

*Current Result*
Now an error occurs in console and no results are returned
image
this error is thrown because [here](https://github.com/geosolutions-it/MapStore2/blob/c1d2cb4cd00aff8bf3cf194e14b3df01c47cd20b/web/client/epics/wfsquery.js#L223) it does not load correctly the describeLayer properties and [here](https://github.com/geosolutions-it/MapStore2/blob/c1d2cb4cd00aff8bf3cf194e14b3df01c47cd20b/web/client/epics/wfsquery.js#L102) it throws the exception

--------------------------------------------------------

image
no results are displayed

GeoServer is returning the search.url containing the ms2-authkey param that should be removed from the url.
We should also check how to filter params saved in the layer properties, because actually the authkey param is saved in the map configuration.

MV88:
## Description
Added some utils to parse the url that allow to filter out some params.
When adding a new layer from the catalog now it parses the url filtering out the authkey specified in the configuration.
Also when the featuregrid is opened and a describelayer is performed

## Issues
- Fix #2369 

 - [x] Bugfix

Now if using csw service geoserver returns a url containing a double ? it contains an old authkey that make fail the wfs request when opening the featuregrid.

**What is the new behavior?**
now the urls is parsed correctly

 - [x] No

### DIFF
--- a/web/client/actions/catalog.js
+++ b/web/client/actions/catalog.js
@@ -15,7 +15,9 @@ var API = {
 const {addLayer, changeLayerProperties} = require('./layers');
 
 const LayersUtils = require('../utils/LayersUtils');
+const ConfigUtils = require('../utils/ConfigUtils');
 const {find} = require('lodash');
+const {authkeyParamNameSelector} = require('../selectors/catalog');
 
 const RECORD_LIST_LOADED = 'CATALOG:RECORD_LIST_LOADED';
 const RESET_CATALOG = 'CATALOG:RESET_CATALOG';
@@ -178,6 +180,7 @@ function describeError(layer, error) {
         error
     };
 }
+
 function addLayerAndDescribe(layer) {
     return (dispatch, getState) => {
         const state = getState();
@@ -190,9 +193,10 @@ function addLayerAndDescribe(layer) {
                 if (results) {
                     let description = find(results, (desc) => desc.name === layer.name );
                     if (description && description.owsType === 'WFS') {
+                        const filteredUrl = ConfigUtils.filterUrlParams(ConfigUtils.normalizeUrl(description.owsURL), authkeyParamNameSelector(state));
                         dispatch(changeLayerProperties(id, {
                             search: {
-                                url: description.owsURL,
+                                url: filteredUrl,
                                 type: 'wfs'
                             }
                         }));
--- a/web/client/actions/catalog.js
+++ b/web/client/actions/catalog.js
@@ -193,7 +193,7 @@ function addLayerAndDescribe(layer) {
                 if (results) {
                     let description = find(results, (desc) => desc.name === layer.name );
                     if (description && description.owsType === 'WFS') {
-                        const filteredUrl = ConfigUtils.filterUrlParams(ConfigUtils.normalizeUrl(description.owsURL), authkeyParamNameSelector(state));
+                        const filteredUrl = ConfigUtils.filterUrlParams(ConfigUtils.cleanDuplicatedQuestionMarks(description.owsURL), authkeyParamNameSelector(state));
                         dispatch(changeLayerProperties(id, {
                             search: {
                                 url: filteredUrl,
--- a/web/client/components/catalog/RecordItem.jsx
+++ b/web/client/components/catalog/RecordItem.jsx
@@ -14,6 +14,7 @@ const {head, memoize, isObject} = require('lodash');
 const assign = require('object-assign');
 
 const CoordinatesUtils = require('../../utils/CoordinatesUtils');
+const ConfigUtils = require('../../utils/ConfigUtils');
 
 const defaultThumb = require('./img/default.jpg');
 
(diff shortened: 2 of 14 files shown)


web/client/components/catalog/RecordItem.jsx
```diff
@@ -211,7 +213,7 @@ class RecordItem extends React.Component {
     };
 
     addLayer = (wms) => {
-        const {url, params} = removeParameters(wms.url, ["request", "layer", "service", "version"]);
+        const {url, params} = removeParameters(ConfigUtils.normalizeUrl(wms.url), ["request", "layer", "service", "version"].concat(this.props.authkeyParamNames));
```
offtherailz: This `removeParameters` seems to be a copy of the one in ConfigUtils, isn't it?
If it is, we should remove this local definition.

web/client/utils/ConfigUtils.js
```diff
@@ -50,6 +50,38 @@ const getConfigurationOptions = function(query, defaultName, extension, geoStore
...
+            if (skip.indexOf(param[0].toLowerCase()) === -1) {
+                let addAnd = i === (pars.length - 1) ? "" : "&";
+                paramsFiltered += param.join("=") + addAnd;
+            }
+        });
+    }
+    return !!paramsFiltered ? urlparts[0] + "?" + paramsFiltered : urlparts[0];
+};
+/**
+ * WORKAROUND: it removes the extra ? when the authkey param is present
```
offtherailz: Please specify that the problem of the extra ? is due to a configuratbility problem of the csw plugin of GeoServer.

I give some you some details to insert in the comment, even to open a JIRA on GeoServer in the future for this:

The problem happens when you have this in the `Record.properties` file of the csw folder in GeoServer, with csw plugin installed: 
```
references.scheme='OGC:WMS'
references.value=strConcat('${url.wms}?service=WMS&request=GetMap&layers=', prefixedName)
```
That is ok when you are not using an authkey, but If you have the authkey module installed and you get record with a proper authkey parameter the ${url.wms} URL will have also `?authkey=...` and so the final URL is something like: 
```
http://domain.org/geoserver/?autkey=abcdefghijklmnopqrstuvz1234567890?service=WMS&request=GetMap&layers=LAYER_NAME
```
 `${url.wms}` is replaced with the wms URL after the execution of `strConcat`, so the url can not be parsed in any way  to solve the problem via configuration, because you can not know if ${url.wms} contains `?` or not.

web/client/actions/catalog.js
```diff
@@ -190,9 +193,10 @@ function addLayerAndDescribe(layer) {
                 if (results) {
                     let description = find(results, (desc) => desc.name === layer.name );
                     if (description && description.owsType === 'WFS') {
+                        const filteredUrl = ConfigUtils.filterUrlParams(ConfigUtils.normalizeUrl(description.owsURL), authkeyParamNameSelector(state));
```
offtherailz: what's the different between this and the removeParameters? Maybe you can clarify the difference by documenting them in ConfigUtils.

web/client/utils/ConfigUtils.js
```diff
@@ -50,6 +50,38 @@ const getConfigurationOptions = function(query, defaultName, extension, geoStore
...
+ * the url was like         http......?authkey=....?service=....&otherparam
+ * that should become like  http......?authkey=....&service=....&otherparam
+*/
+const normalizeUrl = (urlToNormalize) => {
```
offtherailz: the name doesn't remind what you are doing. Call it something like `cleanDuplicatedQuestionMarks` and normalizeUrl can be the combination of removeParameters + cleanDuplicatedQuestionMarks (I've seen you used it often)

web/client/components/catalog/RecordItem.jsx
MV88: this one returns an object containing:
- a first part of the url till the querystring
- the params not removed

web/client/actions/catalog.js
MV88: this removeParameters returns only a shrinked url without the skip params.
i'll change this name in order to clarify that.

PR #2380, opened 2017-11-10
Fix #2378 removed translate interaction for polygons in ol

### LINKED ISSUE #2378: Translate interaction for polygons in openlayers creates conflicts when editing a feature
### Description
When both interactions of modify and translate are enabled in **Openlayers** it becomes hard to insert new vertices because translate interaction trigger when the blue circle is shown.
image

- [x] Bug

*Browser Affected*
- [ ] Internet Explorer
- [x] Chrome 
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*
Chrome 62.0.3202.89

*Steps to reproduce*
- open a map with a layer with polygon features
- enter edit mode for that layer
- try to insert a new vertex moving the perimeter of the polygon (posittion the pointer a little bit inside the polygon) 

*Expected Result*
a blue circle appear as posted above and while dragging that point a new vertex is added

*Current Result*
actually if the clicked point is inside the polygon the translate interaction 

Maybe we can intercep the event when the tranlate interaction starts and calculate with [this](http://turfjs.org/docs#pointToLineDistance) the distance to a linestring if it is over a certain threshold (longer than ray of the blue circle) and it is inside the polygon then stop the propagation of the event in order to perform only the edit of the polygon. doing so if the point is far in the center of the polygon only a translate interaction will be exectuted.

NOTE: For now the translate interaction will be disabled for polygons.

MV88:
## Description
Removed translate interaction in order to avoid conflict when a polygon is being edited.
I also tested that a point is anyway moved eithout this interaction.

## Issues
 - Fix #2378

 - [x] Bugfix

see issue #2378 for details

**What is the new behavior?**
Actually it does not anymore move the polygon, only add or remove vertices.

 - [x] No

### DIFF
--- a/web/client/components/map/openlayers/DrawSupport.jsx
+++ b/web/client/components/map/openlayers/DrawSupport.jsx
@@ -422,7 +422,7 @@ class DrawSupport extends React.Component {
         if (newProps.options.editEnabled) {
             this.addModifyInteraction();
         }
-        if (isSimpleGeomType(newProps.drawMethod) && getSimpleGeomType(newProps.drawMethod) === "Point") {
+        if (getSimpleGeomType(newProps.drawMethod) !== "Polygon") {
             this.addTranslateInteraction();
         }
         if (newProps.options.drawEnabled) {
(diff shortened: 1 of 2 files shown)


web/client/components/map/openlayers/DrawSupport.jsx
```diff
@@ -421,7 +421,6 @@ class DrawSupport extends React.Component {
         }
         if (newProps.options.editEnabled) {
             this.addModifyInteraction();
-            this.addTranslateInteraction();
```
mbarto: can annotations still be dragged?

PR #2383, opened 2017-11-13
Fix #2382 added a new type check for empty background

### LINKED ISSUE #2382: Allow backgrounds with type "OpenLayers.Layer" to be selectable
### Description
There is a background layer which acts as "empty background" which has type "OpenLayers.Layer" and is not selectable because is treated as not supported

- [x] New Feature

### In case of New Feature (otherwise remove this paragraph)
*Acceptance Criteria - AC*

- Empty background layers with type  "OpenLayers.Layer" or "ol" should be selectable

MV88:
## Description
Added the new type "OpenLayers.Layer" for checking if a layer is a empty background.
Added tests 

## Issues
 - Fix #2382

 - [x] Feature

Now if a background layer has type "OpenLayers.Layer" is not selectable from backgroundSelector tool.

**What is the new behavior?**
Now it does.

 - [x] No

### DIFF
--- a/web/client/reducers/config.js
+++ b/web/client/reducers/config.js
@@ -10,7 +10,6 @@ const {MAP_CONFIG_LOADED, MAP_INFO_LOAD_START, MAP_INFO_LOADED, MAP_INFO_LOAD_ER
 const {MAP_CREATED} = require('../actions/maps');
 
 const assign = require('object-assign');
-const {set} = require('lodash');
 const ConfigUtils = require('../utils/ConfigUtils');
 
 function mapConfig(state = null, action) {
@@ -22,14 +21,17 @@ function mapConfig(state = null, action) {
         let hasVersion = action.config.version && action.config.version >= 2;
             // we get from the configuration what will be used as the initial state
         let mapState = action.legacy && !hasVersion ? ConfigUtils.convertFromLegacy(action.config) : ConfigUtils.normalizeConfig(action.config.map);
-        const newMapState = assign({}, set(mapState, "layers", mapState.layers.map( l => {
-            if (l.group === "background" && (l.type === "ol" || l.type === "OpenLayers.Layer")) {
-                l.type = "empty";
-            }
-            return l;
-        })));
+        let newMapState = {
+            ...mapState,
+            layers: mapState.layers.map( l => {
+                if (l.group === "background" && (l.type === "ol" || l.type === "OpenLayers.Layer")) {
+                    l.type = "empty";
+                }
+                return l;
+            })
+        };
         newMapState.map = assign({}, newMapState.map, {mapId: action.mapId, size, version: hasVersion ? action.config.version : 1});
-            // we store the map initial state for future usage
+        // we store the map initial state for future usage
         return assign({}, newMapState, {mapInitialConfig: {...newMapState.map, mapId: action.mapId}});
     case MAP_CONFIG_LOAD_ERROR:
         return {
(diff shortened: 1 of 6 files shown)


web/client/utils/LayersUtils.js
```diff
@@ -64,8 +64,8 @@ const isSupportedLayer = (layer, maptype) => {
     if (layer.type === "mapquest" || layer.type === "bing") {
         return Layers.isSupported(layer.type) && layer.apiKey && layer.apiKey !== "__API_KEY_MAPQUEST__" && !layer.invalid;
     }
-    // type 'ol' represents 'No background' layer
-    if (layer.type === 'ol') {
+    // type 'ol' or 'OpenLayers.Layer' represents 'No background' layer
+    if (layer.type === 'ol' || layer.type === 'OpenLayers.Layer') {
```
mbarto: This is not very MapStore2...ish, can we make the ol / OpenLayers.Layer type something that is handled by the legacy converter and use a better name for the standard type (e.g. empty)

MV88: is it fine if I parse the type and convert it to **empty** after this [line] ?(https://github.com/geosolutions-it/MapStore2/blob/master/web/client/reducers/config.js#L23)

mbarto: I would do it here: https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/ConfigUtils.js#L151

web/client/reducers/config.js
```diff
@@ -21,7 +21,12 @@ function mapConfig(state = null, action) {
         let hasVersion = action.config.version && action.config.version >= 2;
             // we get from the configuration what will be used as the initial state
         let mapState = action.legacy && !hasVersion ? ConfigUtils.convertFromLegacy(action.config) : ConfigUtils.normalizeConfig(action.config.map);
-
+        mapState.layers = mapState.layers.map( l => {
+            if (l.group === "background" && (l.type === "ol" || l.type === "OpenLayers.Layer")) {
+                l.type = "empty";
```
mbarto: You are mutating an object.
Moreover I would not do the conversion in a reducer, but here:  https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/ConfigUtils.js#L151

MV88: the problem is that new maps have these old values as type for empty background. so checking in the legacy converter only is not sufficient.  i'll provide a change to avoid mutate the object using assign()

web/client/reducers/config.js
```diff
@@ -21,10 +22,15 @@ function mapConfig(state = null, action) {
         let hasVersion = action.config.version && action.config.version >= 2;
             // we get from the configuration what will be used as the initial state
         let mapState = action.legacy && !hasVersion ? ConfigUtils.convertFromLegacy(action.config) : ConfigUtils.normalizeConfig(action.config.map);
-
-        mapState.map = assign({}, mapState.map, {mapId: action.mapId, size, version: hasVersion ? action.config.version : 1});
+        const newMapState = assign({}, set(mapState, "layers", mapState.layers.map( l => {
+            if (l.group === "background" && (l.type === "ol" || l.type === "OpenLayers.Layer")) {
+                l.type = "empty";
```
offtherailz: Change new.json to align everything to the type "empty"

MV88: done

offtherailz:
```
const newMapState = {
    ...mapState,
   layers : ...
}
```

You are doing something like: 
```
var temp = mapState.layers = ...
{
  ...mapState: 
 layers: temp
}
```
so you are still mutating

PR #2388, opened 2017-11-13
Fixes #2084 Rethink the layout and behaviour of the sidebar on the right.

### LINKED ISSUE #2084: Rethink the layout and behaviour of the sidebar on the right.
We need to decide how to change some tools of mapstore when the feature grid is opened which hides some of the following tools.

in particular these tools
image

allyoucanmap:
## Description
Changed layout behaviours when dock panels are open

## Issues
 - Fix #2084

 - [x] Feature

dock panels hide other plugins

**What is the new behavior?**
dock panels don't hide other plugins

 - [x] No

**Other information**:


web/client/selectors/controls.js
```diff
@@ -2,5 +2,21 @@ const {get} = require('lodash');
...
+        const drawer = get(state, "controls.drawer.enabled") && ' mapstore-drawer-open' || '';
+
+        const featuregrid = get(state, "featuregrid.open") && ' mapstore-featuregrid-open' || '';
+        const queryPanel = get(state, "controls.queryPanel.enabled") && ' mapstore-queryPanel-open' || '';
+
+        const annotation = get(state, "controls.annotations.enabled") && ' mapstore-annotation-open' || '';
+        const metadataexplorer = get(state, "controls.metadataexplorer.enabled") && ' mapstore-metadataexplorer-open' || '';
+        const identify = get(state, "mapInfo.requests") && state.mapInfo.requests.length > 0 && ' mapstore-identify-open' || '';
+
+        const rightPanel = metadataexplorer || annotation || identify ? ' mapstore-right-panel-open' : '';
```
offtherailz: Get Feature Info panel is not well supported (because it's width is responsive). In addition any additional panel (even for custom plugins) should modify this selector. And any of these class is in css, and so resizable panel will not work.

IMHO we should use a higher level solution to this. Personally I don't like this solution because it is not pluggable at all, only makes the demo work. 
MS2 is framework to do custom applications, that typically will have some custom panels on the map. This solution do not allow the usage with custom panels.

web/client/plugins/Widgets.jsx
```diff
@@ -40,37 +40,31 @@ const WidgetsView = connect(
...
+        return this.props.enabled ? (<ContainerDimensions>{({width, height}) => <WidgetsView width={width} height={height}/>}</ContainerDimensions> ) : null;
 
     }
 }
 
-module.exports = {
-    WidgetsPlugin: assign(Widgets, {
+const WidgetsPlugin = connect(
+    createSelector(
+        state => checkConditionsSelector(state, [
```
offtherailz: move this in maplayout selectors and name it better. rightPanelOpenSelector, bottomPanelOpenSelector and put them together here.

PR #2397, opened 2017-11-15
Fixed #2210 cql creation for spatial filter in ol

### LINKED ISSUE #2210: Sync map with Filter sometimes disables the layer
Sometimes CQL_FILTER generated are not valid. So, when the sync map and table is active, sometimes the invalid request makes the layer disable. 
Example wrong filters: 
empty:
sometimes with filter = 0.
```
() 
```
This(don't know what is wrong in this filter, single conditions work, together with AND generate error): 
```
(strToLowerCase(NOME_LIVELLO) LIKE '%e%' AND ID='1')
```

MV88:
## Description
Cql Filter was wrongly created if a patial filter was present. the crs was not the native one of the layer

## Issues
 - Fix #2210

 - [x] Bugfix

if sync wasa active and it was drawn a spatial filter then the layer were hidden. (cql messed up)

**What is the new behavior?**
actually the layer is correctly filtered.

 - [x] Maybe

I'm not sure 100% that this PR introduces a breaking change in the projects that uses this. There could be some problem related to the fix made in drawsupport when a polygon is drawn because in ol it was missing the last vertix

**Other information**:
For testing this try to draw a spatial filter and then activate the sync. the features should be filtered based on the drawn spatial filter.


MV88:
creating another one due to some changes..

PR #2413, opened 2017-11-20
Fixed #2287 allow setup languages from configuration

### LINKED ISSUE #2287: Set available languages from configuration
The list of available languages should come from the configuration and not within the app.

The setting of supportedLocales in app.jsx should be removed from there

MV88:
## Description
Now it is possible to set up languages via configuration.

## Issues
 - Fix #2287 

 - [x] Feature

Previously languages could be set only hard coding them in the app.

**What is the new behavior?**
Actually the languages can be set via configuration

 - [x] No

**Other information**:
If a user add a locale which is not supported then the app does not crash and it does not show that locale in the lang list
if the flag is present but not the translations then an error message appear


mbarto:
main.jsx should be good

offtherailz:
@MV88 can you do this small change ?

MV88:
k, opening a new PR

PR #2427, opened 2017-11-23
Closes #2203. GetFeatureInfo active by default

### LINKED ISSUE #2203: Always-on GetFeatureInfo
We should evaluate the default design of having the Info tool always active.

User behaviour shows that info-on-click is expected.

Related to #1125

kappu72:
## Description
The GFI tool has become active by default. The user just needs to click on map to get the feature info. The GFI is stopped if Annotation or Grid plugins are editing, draw or measure support are active.
## Issues
 - Fix #2203 

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

The GFI is enabled from the btn in toolbar. It's disabled by some other tolls but not reenabled.

**What is the new behavior?**
The GFI is enabled by default, and disabled/enabled under some circumstance.

 - [ ] Yes
 - [x] No

**Other information**:
It is possible to control the GFI, by setting the state variables: enabled and disableAlwaysOn.

### DIFF
--- a/web/client/epics/identify.js
+++ b/web/client/epics/identify.js
@@ -7,13 +7,26 @@
 */
 const Rx = require('rxjs');
 
-const {LOAD_FEATURE_INFO, GET_VECTOR_INFO} = require('../actions/mapInfo');
+const {LOAD_FEATURE_INFO, GET_VECTOR_INFO, FEATURE_INFO_CLICK} = require('../actions/mapInfo');
 const {closeFeatureGrid} = require('../actions/featuregrid');
+const {CHANGE_MOUSE_POINTER, CLICK_ON_MAP} = require('../actions/map');
+const {MAP_CONFIG_LOADED} = require('../actions/config');
+const {stopGetFeatureInfoSelector} = require('../selectors/mapinfo');
 
 module.exports = {
     closeFeatureGridFromIdentifyEpic: (action$) =>
         action$.ofType(LOAD_FEATURE_INFO, GET_VECTOR_INFO)
         .switchMap(() => {
             return Rx.Observable.of(closeFeatureGrid());
+        }),
+    changeMapPointer: (action$, store) =>
+        action$.ofType(CHANGE_MOUSE_POINTER)
+        .filter(() => !(store.getState()).map)
+        .switchMap((a) => action$.ofType(MAP_CONFIG_LOADED).mapTo(a)),
+    onMapClick: (action$, store) =>
+        action$.ofType(CLICK_ON_MAP).filter(() => {
+            const {disableAlwaysOn = false} = (store.getState()).mapInfo;
+            return disableAlwaysOn || !stopGetFeatureInfoSelector(store.getState() || {});
         })
+        .map(({point, layer}) => ({type: FEATURE_INFO_CLICK, point, layer}))
 };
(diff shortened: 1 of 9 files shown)


web/client/epics/identify.js
```diff
@@ -7,13 +7,26 @@
...
 
 module.exports = {
     closeFeatureGridFromIdentifyEpic: (action$) =>
         action$.ofType(LOAD_FEATURE_INFO, GET_VECTOR_INFO)
         .switchMap(() => {
             return Rx.Observable.of(closeFeatureGrid());
+        }),
+    changeMapPointer: (action$, store) =>
+        action$.ofType(CHANGE_MOUSE_POINTER)
+        .filter(() => !(store.getState()).map)
```
MV88: Everytime you access the state, it is preferrable to use a selector

web/client/epics/identify.js
```diff
@@ -7,13 +7,26 @@
...
+        .switchMap((a) => action$.ofType(MAP_CONFIG_LOADED).mapTo(a)),
+    onMapClick: (action$, store) =>
+        action$.ofType(CLICK_ON_MAP).filter(() => {
+            const {disableAlwaysOn = false} = (store.getState()).mapInfo;
```
MV88: add or use existing selector

web/client/selectors/mapinfo.js
```diff
@@ -32,7 +34,27 @@ const mapInfoRequestsSelector = state => get(state, "mapInfo.requests") || [];
  */
 const generalInfoFormatSelector = (state) => get(state, "mapInfo.infoFormat", "text/plain");
 
+const measureActiveSelector = (state) => get(state, "measurement.lineMeasureEnabled") || get(state, "measurement.areaMeasureEnabled") || get(state, "measurement.bearingMeasureEnabled");
+const drawSupportActiveSelector = (state) => {
+    const drawStatus = get(state, "draw.drawStatus", false);
+    return drawStatus && drawStatus !== 'clean' && drawStatus !== 'stop';
+};
+const gridEditingSelector = (state) => get(state, "featuregrid.mode") === 'EDIT';
```
MV88: the feature grid mode selector already exists, please use that as part of this selector

web/client/epics/identify.js
```diff
@@ -7,13 +7,26 @@
...
+const {MAP_CONFIG_LOADED} = require('../actions/config');
+const {stopGetFeatureInfoSelector} = require('../selectors/mapinfo');
 
 module.exports = {
     closeFeatureGridFromIdentifyEpic: (action$) =>
         action$.ofType(LOAD_FEATURE_INFO, GET_VECTOR_INFO)
         .switchMap(() => {
             return Rx.Observable.of(closeFeatureGrid());
+        }),
+    changeMapPointer: (action$, store) =>
```
MV88: not sure about this epic, can you explain it?

kappu72: It could throw the change-map-pointer action before the map is configured so the mouse pointer action will fail

web/client/epics/identify.js
kappu72: Ok, but here we are not computing any derivate value, so whats the meaning of using a selector?

web/client/selectors/mapinfo.js
kappu72: Ok I'll use that.

web/client/epics/identify.js
MV88: maybe add a comment above this epic with the answer you give me. :)

Issue #2443, opened 2017-11-30
Load locally available KML/KMZ, GPX files into the map and show the data (geometry and attributes)


mricca:
### Description
The users shall be able to upload KML/KMZ, GPX data to the UI and the data shall be displayed in the map.
The result of the import will be a (vector) layer in the map, with all the usual functionalities (included the ability to do an identify and show attributes of the vector data).

### Please indicate if this issue is related to a bug or a new feature request
- [ ] Bug
- [X] New Feature

### In case of New Feature (otherwise remove this paragraph)

*Acceptance Criteria - AC*

 - upload KML/KMZ or GPX files
 - data shall be displayed in the map
 - getFeatureInfo functionality available

PR #2444, opened 2017-11-30
Issue #2443. Load locally available KML/KMZ, GPX files into the map a…

mricca:
…nd show the data (geometry and attributes)

## Description
The users shall be able to upload KML/KMZ, GPX data to the UI and the data shall be displayed in the map.
The result of the import will be a (vector) layer in the map, with all the usual functionalities (included the ability to do an identify and show attributes of the vector data).

## Issues
 - #2443 

 - [X] Feature

You can only upload Shapefiles in zip archive

**What is the new behavior?**
Now you can upload also KML/KMZ, GPX vector files.

 - [ ] Yes
 - [X] No

### DIFF
--- a/web/client/components/shapefile/SelectShape.jsx
+++ b/web/client/components/shapefile/SelectShape.jsx
@@ -57,9 +57,8 @@ class SelectShape extends React.Component {
     checkFileType = (file) => {
         return new Promise((resolve, reject) => {
             const ext = FileUtils.recognizeExt(file.name);
-            const type = file.type || FileUtils.MIME_LOOKUPS[ext];
-            if (type instanceof Array ||
-                type === 'application/x-zip-compressed' ||
+            const type = FileUtils.MIME_LOOKUPS[ext];
+            if (type === 'application/x-zip-compressed' ||
                 type === 'application/zip' ||
                 type === 'application/vnd.google-earth.kml+xml' ||
                 type === 'application/vnd.google-earth.kmz' ||
--- a/web/client/components/shapefile/SelectShape.jsx
+++ b/web/client/components/shapefile/SelectShape.jsx
@@ -56,12 +56,13 @@ class SelectShape extends React.Component {
 
     checkFileType = (file) => {
         return new Promise((resolve, reject) => {
-            const ext = file.name.split('.').slice(-1)[0];
+            const ext = FileUtils.recognizeExt(file.name);
             const type = file.type || FileUtils.MIME_LOOKUPS[ext];
-            if (file.type === 'application/zip' ||
-                file.type === 'application/x-zip-compressed' ||
-                file.type === 'application/vnd.google-earth.kml+xml' ||
-                file.type === 'application/vnd.google-earth.kmz' ||
+            if (type instanceof Array ||
+                type === 'application/x-zip-compressed' ||
+                type === 'application/zip' ||
+                type === 'application/vnd.google-earth.kml+xml' ||
+                type === 'application/vnd.google-earth.kmz' ||
                 type === 'application/gpx+xml') {
                 resolve();
             } else {
--- a/web/client/components/shapefile/SelectShape.jsx
+++ b/web/client/components/shapefile/SelectShape.jsx
@@ -57,7 +57,7 @@ class SelectShape extends React.Component {
     checkFileType = (file) => {
         return new Promise((resolve, reject) => {
             const ext = FileUtils.recognizeExt(file.name);
-            const type = FileUtils.MIME_LOOKUPS[ext];
+            const type = file.type || FileUtils.MIME_LOOKUPS[ext];
             if (type === 'application/x-zip-compressed' ||
                 type === 'application/zip' ||
                 type === 'application/vnd.google-earth.kml+xml' ||
--- a/web/client/components/shapefile/SelectShape.jsx
+++ b/web/client/components/shapefile/SelectShape.jsx
@@ -49,13 +49,20 @@ class SelectShape extends React.Component {
 
     tryUnzip = (file) => {
         return FileUtils.readZip(file).then((buffer) => {
-            return new JSZip(buffer);
+            var zip = new JSZip();
+            return zip.loadAsync(buffer);
         });
     };
 
-    isZip = (file) => {
+    checkFileType = (file) => {
         return new Promise((resolve, reject) => {
-            if (file.type === 'application/zip' || file.type === 'application/x-zip-compressed') {
+            const ext = file.name.split('.').slice(-1)[0];
+            const type = file.type || FileUtils.MIME_LOOKUPS[ext];
+            if (file.type === 'application/zip' ||
+                file.type === 'application/x-zip-compressed' ||
+                file.type === 'application/vnd.google-earth.kml+xml' ||
+                file.type === 'application/vnd.google-earth.kmz' ||
+                type === 'application/gpx+xml') {
                 resolve();
             } else {
                 this.tryUnzip(file).then(resolve).catch(reject);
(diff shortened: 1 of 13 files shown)


web/client/components/shapefile/ShapefileUploadAndStyle.jsx
```diff
@@ -81,6 +98,22 @@ class ShapeFileUploadAndStyle extends React.Component {
 
     getGeomType = (layer) => {
         if (layer && layer.features && layer.features[0].geometry) {
+            let unique;
+            const geometryType = layer.features.filter((element) => {
+                if (element.geometry.type === "GeometryCollection") {
+                    return element.geometry.geometries;
+                }
```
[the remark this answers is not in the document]
mbarto: filter function should return true / false

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
```diff
@@ -81,6 +98,22 @@ class ShapeFileUploadAndStyle extends React.Component {
...
+            });
+            if (geometryType.length > 0) {
+                let arr = [];
+                geometryType.forEach((item) => {
```
mbarto: Don't use forEach if you can get the same result using map / reduce

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
```diff
@@ -152,7 +185,7 @@ class ShapeFileUploadAndStyle extends React.Component {
         this.props.shapeLoading(true);
         let queue = this.props.readFiles(files);
         Promise.all(queue).then((geoJsons) => {
-            let ls = geoJsons.reduce((layers, geoJson) => {
+            let ls = geoJsons.filter((element) => {if (element[0].features.length !== 0) {return element; }}).reduce((layers, geoJson) => {
```
mbarto: What's the purpose of this filter? As usual, filter should returns true / false

web/client/utils/FileUtils.js
```diff
@@ -9,9 +9,18 @@
...
+const JSZip = require('jszip');
 const {Promise} = require('es6-promise');
+const parser = new DOMParser();
 
 const FileUtils = {
+    MIME_LOOKUPS: {
+        'gpx': 'application/gpx+xml',
+        'kmz': 'application/vnd.google-earth.kmz',
+        'kml': 'application/vnd.google-earth.kml+xml',
+        'zip': ['application/x-zip-compressed', 'application/zip']
```
mbarto: Where do you handle MIME_LOOKUPS['zip'] being an array? It seems to me that MIME_LOOKUPS['zip'] is expected to be a string

web/client/utils/FileUtils.js
```diff
@@ -23,14 +32,55 @@ const FileUtils = {
     shpToGeoJSON: function(zipBuffer) {
         return [].concat(shp.parseZip(zipBuffer));
     },
+    kmlToGeoJSON: function(xml) {
+        let geoJSON = [].concat(tj.kml(xml));
```
mbarto: const geoJSON = [].concat(tj.kml(xml).map(item => assign({}, item, {fileName: xml.getElementsByTagName('name')[0].innerHTML})));

everywhere please, don't use forEach if you can use map / reduce

web/client/utils/FileUtils.js
```diff
@@ -23,14 +32,55 @@ const FileUtils = {
...
+            reader.readAsText(file);
+        });
+    },
+    readKmz: function(file) {
+        const zip = new JSZip();
+        return new Promise((resolve, reject) => {
+            zip.loadAsync(file).then(files => {
+                files.filter(elem => {
+                    return elem.indexOf('kml') !== -1;
+                }).map(elem => {
```
mbarto: Here you have to use forEach, instead, because you have side effects, it doesn't make sense to use map if you ignore the resulting array

web/client/utils/FileUtils.js
mricca: I'm changing it

web/client/components/shapefile/SelectShape.jsx
```diff
@@ -49,13 +49,20 @@ class SelectShape extends React.Component {
...
+            return zip.loadAsync(buffer);
         });
     };
 
-    isZip = (file) => {
+    checkFileType = (file) => {
         return new Promise((resolve, reject) => {
-            if (file.type === 'application/zip' || file.type === 'application/x-zip-compressed') {
+            const ext = FileUtils.recognizeExt(file.name);
+            const type = FileUtils.MIME_LOOKUPS[ext];
```
mbarto: We don't use file.type anymore. I think we should do:
const type = file.type || FileUtils.MIME_LOOKUPS[ext];

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
```diff
@@ -58,10 +58,38 @@ class ShapeFileUploadAndStyle extends React.Component {
...
-                return FileUtils.shpToGeoJSON(buffer);
-            });
+        readFiles: (files) => files.map((file) => {
+            const ext = FileUtils.recognizeExt(file.name);
+            const type = FileUtils.MIME_LOOKUPS[ext];
+            if (type === 'application/vnd.google-earth.kml+xml') {
+                return FileUtils.readKml(file).then((xml) => {
+                    return FileUtils.kmlToGeoJSON(xml);
+                }).catch((e) => {
+                    return e.message;
```
mbarto: Is this used to show an error to the user at all?

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
```diff
@@ -81,6 +109,20 @@ class ShapeFileUploadAndStyle extends React.Component {
 
     getGeomType = (layer) => {
         if (layer && layer.features && layer.features[0].geometry) {
+            let unique;
```
mbarto: I was thinking to something similar to this (not tested):

```
const type = layer.features.reduce((previous, f) => {
    const currentType = getGeometryType(f.geometry);
    return currentType === previous ? previous : 'GeometryCollection';
}, null);

getGeometryType = (geometry) => {
    if (geometry && geometry.type === 'GeometryCollection') {
        return geometry.geometries.reduce((previous, g) => {
            if (g && g.type === previous) {
                return previous;
            }
            return g.type;
        }, null);
    }
    if (geometry) {
        return geometry.type;
    }
    return null;
}
```

web/client/components/shapefile/SelectShape.jsx
mricca: Mistake

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
mricca: it seems that it always returns 'GeometryCollection'. I'm going to check

web/client/components/shapefile/ShapefileUploadAndStyle.jsx
mricca: This is an additional catch, the error is catched and show here:
https://github.com/mricca/MapStore2/blob/a00b1a4063fcc81b03bd929a8d268135f7998041/web/client/components/shapefile/ShapefileUploadAndStyle.jsx#L207

what is the most logical thing to do?

mbarto: Uhm... if it's already catched elsewhere, it's not so useful to have another catch here

mricca: ok, I will remove it

PR #2505, opened 2018-01-10
Fixes #2291 If a map save fails, user is not warned

### LINKED ISSUE #2291: If a map save fails, user is not warned
Saving a map with the same name of another one will fail by design, but the user is not warned.

To reproduce: 
- Log in in the home page
- Click "Create map"
- From the menu select "Save as.."
- Input a name and Save
- Edit the map
- From the menu select "Save as.." again
- Input the same name and Save

Expected result: An error message informs the user that the same name cannot be used.

Actual result: The saving window closes as if the map was correctly saved.

allyoucanmap:
## Description
Added feedback messages on save and save as.

success save

Recognized errors:  403 - 404 - 409 - 500
error save

## Issues
 - Fix #2291

 - [x] Bugfix

issue #2291

**What is the new behavior?**
Notifications feedback will be shown after saving

 - [x] No

**Other information**:

### DIFF
--- a/web/client/utils/LocaleUtils.js
+++ b/web/client/utils/LocaleUtils.js
@@ -51,6 +51,27 @@ const DATE_FORMATS = {
     "it-IT": "dd/MM/yyyy",
     "nl-NL": "dd/MM/yyyy"
 };
+
+const errorParser = {
+    geostore: {
+        mapsError: e => {
+            if (e.status === 403 || e.status === 404 || e.status === 409 || e.status === 500) {
+                return {
+                    title: 'map.mapError.errorTitle',
+                    message: 'map.mapError.error' + e.status
+                };
+            }
+            return {
+                title: 'map.mapError.errorTitle',
+                message: 'map.mapError.errorDefault'
+            };
+        }
+    }
+};
+/**
+ * Utilities for locales.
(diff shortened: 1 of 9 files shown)


web/client/utils/LocaleUtils.js
```diff
@@ -51,6 +51,27 @@ const DATE_FORMATS = {
     "it-IT": "dd/MM/yyyy",
     "nl-NL": "dd/MM/yyyy"
 };
+
+const errorParser = {
+    geostore: {
```
mbarto: I would register this handler in the GeoStoreAPI code, and put here a function to register the parsers

web/client/utils/LocaleUtils.js
```diff
@@ -51,6 +51,18 @@ const DATE_FORMATS = {
     "it-IT": "dd/MM/yyyy",
     "nl-NL": "dd/MM/yyyy"
 };
+
+const servicesErrorParser = {
+    geostore: require('../api/GeoStoreDAO').errorParser
```
mbarto: Can we do the opposite?
 * registerParser = (type, parser) => {} in LocaleUtils
 * registerParser('geostore', parser) in GeoStoreDAO

PR #2506, opened 2018-01-10
Fixes #2248 Leaflet doesn't detect tiles errors

### LINKED ISSUE #2248: Leaflet doesn't detect tiles errors
Even if the layer is disabled the layer component doesn't change to error status

to reproduce the issue:
- hide visibility of a layer from TOC
- show visibility of a layer from TOC
- now `Block request domain` of the layer from dev tools to simulate a not accessible layer

Current result: 
- even if the layer is not accessible, the TOC doesn't modify its status.

Expected result:
- TOC should recognized the errors of tile and shows the exclamation point

peek 2017-09-29 16-23

allyoucanmap:
## Description
Updated observables management on leaflet layers.

After we had Annotation Plugin to MapStore2 the lifecycle of leaflet layers has changed.
Now they add only one time  the `load`, `loading` and `tileerror ` events and the load observables because of the new flag [ms2Added ](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/map/leaflet/Map.jsx#L150L165).
When a layer has been removed it triggers the `layerremove` event that it complete the stream of the observable of the layers so if we will add again the layer, the observable are not created again (e.g. toggle visibility of the layer cause this behaviours)

## Issues
 - Fix #2248

 - [x] Bugfix

Issue #2248

**What is the new behavior?**
Correct detection of tiles errors on leaflet

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/map/leaflet/Map.jsx
+++ b/web/client/components/map/leaflet/Map.jsx
@@ -148,6 +148,8 @@ class LeafletMap extends React.Component {
         this.map.on('layeradd', (event) => {
             // we want to run init code only the first time a layer is added to the map
             if (event.layer._ms2Added) {
+                const isStopped = event.layer.layerLoadingStream$ && event.layer.layerLoadingStream$.isStopped;
+                this.addLayerObservable(event, isStopped);
                 return;
             }
             event.layer._ms2Added = true;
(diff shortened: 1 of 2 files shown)


offtherailz:
Is not clear what the problem was. Can you clarify the problem and how you solved it quickly in the PR description?

allyoucanmap:
Description updated.

Issue #2512, opened 2018-01-15
Queryform redraws Circle filter differently when moving from ol to leaflet and viceversa


MV88:
### Description
There are variations based on the latitude because leaflet circle is an approximation (see [here](http://leafletjs.com/reference-1.3.0.html#circle))

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*
any

*Steps to reproduce*
- Open "WFS Test Map" with openlayers
- Draw a circle and save a screenshot of it (for future comparison)
- Go to home and open map with leaflet
- Open the queryform

*Expected Result*
Same circle drawn with ol

*Current Result*
Different circle (bigger)

### Other useful information (optional):
With the same radius they draw different circles.
North America/Canada
![image](https://user-images.githubusercontent.com/11991428/34938037-1c783e6c-f9e7-11e7-8ff1-774c41f8754b.png)
on the left openlayers (radius ~740806.4155330788) 
on the right leaflet (radius ~217131.55189546535)

MV88:
fixed by #2544

PR #2513, opened 2018-01-15
Fix #2222 leaflet circle redraw

### LINKED ISSUE #2222: Circle spatial filter shows different selections on high denominator scales

MV88:
## Description
A few sentences describing the overall goals of the pull request' s commits.

## Issues
 - Fix #2222 

 - [x] Bugfix

The queryform was not drawing the same circle

**What is the new behavior?**
Now the queryform redraws the same circle.

 - [x] No

**Other information**:
While fixing this issue I found out a new one present from long time (#2512)

### DIFF
--- a/web/client/epics/wfsquery.js
+++ b/web/client/epics/wfsquery.js
@@ -21,6 +21,7 @@ const ConfigUtils = require('../utils/ConfigUtils');
 const assign = require('object-assign');
 const {spatialFieldMethodSelector, spatialFieldGeomTypeSelector, spatialFieldGeomCoordSelector, spatialFieldGeomSelector, spatialFieldGeomProjSelector} = require('../selectors/queryform');
 const {changeDrawingStatus} = require('../actions/draw');
+const {isLeaflet} = require('../selectors/maptype');
 const {INIT_QUERY_PANEL} = require('../actions/wfsquery');
 
 const {isObject} = require('lodash');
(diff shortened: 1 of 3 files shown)


web/client/epics/wfsquery.js
```diff
@@ -212,15 +213,39 @@ const redrawSpatialFilterEpic = (action$, store) =>
...
+        let feature;
+        let proj;
+        if (method === "Circle" && geom ) {
+            proj = "EPSG:3857";
+            const point = CoordinatesUtils.reproject(geom.center, geom.projection, proj);
+            const center = [point.x, point.y];
+            feature = {
+                type: "Feature",
+                geometry: {
+                    type: isLeaflet(state) ? "Point" : "Polygon",
```
offtherailz: This should not be aware of maptype

MV88:
The radius calculated with the projected point has been used as the model. this breaks how leaflet redraws circles because it calculates the radius with Haversine Formula, see [this](https://github.com/Leaflet/Leaflet/blob/v0.7.7/src/geo/LatLng.js#L48)
There are still problem when editing properties of the circle like radius and lat,lon and the query result that are not so consitent with the drawn circle.

another note:
[here ](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/data/query/GeometryDetails.jsx#L147) we use this function that creates a smaller polygon not so close to the drawn one, for this reason i introduced a new [circleToPolygon ](https://github.com/geosolutions-it/MapStore2/pull/2513/files#diff-f322b2ea07061749fc77ec3ebed17825R16) module but it works with point represented with 4326

There is also a limitation on number of decimals in the geometry details to consider

at this stage on Ol works all

PR #2544, opened 2018-01-22
Fix #2222. Support for circle reprojection in leaflet

### LINKED ISSUE #2222: Circle spatial filter shows different selections on high denominator scales

offtherailz:
## Description
This changes make the DrawSupport mask the different radius management for circles in leaflet. 
From the external point of view, draw support now is identical to openlayer's one, and circle is virtually drawn on projected system (so the geometry will be a real circle on EPS:3857).
Internal conversion provides proper scaling of leaflet radius to draw circles in leaflet.
## Issues
 - Fix #2222

 - [x] Bugfix

 - Circles are not reloaded correctly in leaflet. 
 - Switching from OL to leaflet doesn't preserve the same circle as filter
 - Searching using the circle drawn provides wrong results (e.g. some features that seems to be inside the circle on the map are not shown in results). 

**What is the new behavior?**
 - The circle is correctly reloaded (I noticed on first load the `DrawSupport` shows the 100 points polygon, then when you change, for instance from the query panel, some parameter, you will see again a real circle. I wasn't able to remove this behavior this time. Anyway is doesn't seems to create any problem)
 - You can switch from OL to Leaflet preserving the circle filter
 - Search results are now consistent with the map

 - [ ] Yes
 - [x] No

**NOTE**: Don't think it is a breaking change at all. The filter in leaflet now will be generated as with openlayers, ad a geometry in EPSG:3857, before it was a circle in EPSG:4326 that caused many issues in terms of search result expectation, for instance. 

**Other information**:
`DrawSupport` is a little too complicated. We should define and document an API to interact with it and reduce duplication.

### DIFF
--- a/web/client/components/map/leaflet/DrawSupport.jsx
+++ b/web/client/components/map/leaflet/DrawSupport.jsx
@@ -39,6 +39,7 @@ const toProjectedCircle = (mRadius, center, projection) => {
         const checkPoint = CoordinatesUtils.reproject([center[0] + lonRadius, center[1]], "EPSG:4326", projection);
         return {
             center: projCenter,
+            srs: projection,
             radius: Math.sqrt(Math.pow(projCenter.x - checkPoint.x, 2) + Math.pow(projCenter.y - checkPoint.y, 2))
         };
     }
--- a/web/client/components/map/leaflet/DrawSupport.jsx
+++ b/web/client/components/map/leaflet/DrawSupport.jsx
@@ -17,7 +17,72 @@ const assign = require('object-assign');
 const CoordinatesUtils = require('../../../utils/CoordinatesUtils');
 
 const VectorUtils = require('../../../utils/leaflet/Vector');
+/**
+ * Converts the leaflet circle into the projected circle (usually in 3857)
+ * @param  {number} mRadius leaflet radius of circle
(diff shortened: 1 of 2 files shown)


web/client/components/map/leaflet/DrawSupport.jsx
```diff
@@ -17,7 +17,72 @@ const assign = require('object-assign');
...
+            radius: mRadius
+        };
+    }
+
+    // calculate
+    const lonRadius = (mRadius / 40075017) * 360 / Math.cos(L.LatLng.DEG_TO_RAD * (center[1]));
+    const projCenter = CoordinatesUtils.reproject(center, "EPSG:4326", projection);
+    if (lonRadius) {
+        const checkPoint = CoordinatesUtils.reproject([center[0] + lonRadius, center[1]], "EPSG:4326", projection);
+        return {
```
mbarto: srs is missing

PR #2581, opened 2018-02-02
DrawSupport:Replace does not add drawinteraction by default

ndufrane:
## Description
Hi,
 
This PR change a little bit the drawsupport replace function as proposed here https://groups.google.com/forum/#!topic/mapstore-developers/s3A77gAHleU

It add the possibility to choose to add interactions whit the drawEnabled option. By default, it's false.

- [ x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md

 - [X] Refactoring (no functional changes, no api changes)

DrawSupport : When the draw layer is not defined and replace function is called. The draw layer is initialized with draw interactions by default : call to add layer with True hardcoded

**What is the new behavior?**
DrawSupport : When the draw layer is not defined and replace function is called. The draw layer is NOT initialized with draw interactions by default : call to add layer use  options.drawEnabled parameter. By default the value is False/

 - [x ] Yes

This PR may introduce breaking change. As described before the current behavior is to add by default the interactions when using replace. And this PR set the default to false. If it is not possible or correct to use "False" by default , i can change this PR to have the same default behaviour

**Other information**:

If tests case are needed, please let me know

### DIFF
--- a/web/client/components/map/openlayers/DrawSupport.jsx
+++ b/web/client/components/map/openlayers/DrawSupport.jsx
@@ -152,7 +152,7 @@ class DrawSupport extends React.Component {
 
     replaceFeatures = (newProps) => {
         if (!this.drawLayer) {
-            this.addLayer(newProps, true);
+            this.addLayer(newProps, newProps.options && newProps.options.drawEnabled || false);
         } else {
             this.drawSource.clear();
             this.addFeatures(newProps);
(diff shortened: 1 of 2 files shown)


mbarto:
The PR looks good. Any chance you can add a unit test?

ndufrane:
Ok for the unit test.  I will update the pr when it's done 📦

PR #2720, opened 2018-03-09
Fix #2668 Increase precision of Query Panel Circle/BBox Details

### LINKED ISSUE #2668: Increase precision of Query Panel Circle/BBox Details
### Description
Query Panel Geometry details has a decimal precision fixed to 2 digits. It is not enough. User can not edit precise circle filter or bbox filters at high zoom levels. 

 - When coordinates are in 4326 (`useMapProjection === false` in `GeometryDetails`)
when you change the radius of a circle the center of the circle moves to the nearest point with coordinates multiple of 0.01. 
 - When you click on input arrows, the step of 1 degree is too big. It should be smaller (may depend from the current map resolution, but a simple solution can be to setting the step of 0.01). 
 - Because of live update of the geometry, if the user deletes a digit and insert another one, the geometry may jump here and there. We may introduce a debounce time (1 sec) for geometry update.

*Browser Affected*
any

*Steps to reproduce*

- Open the map
- Add a vector layer to the map
- Search for "Colosseo" and select the first result
- Open feature grid, then filter
- Select circle filter and draw a circle on the map centered in the coliseum center
image

- Click on pencil to open filter's geometry details
- change radius value

*Expected Result*
- The circle only changes radius

*Current Result*
 - The circle moves it's center to the nearest point with coordinates multiple of 0.01 
Something like this:
image

This is due to a wrong approximation + to fast live editing + wrong step. 
We should: 
 - [ ] increase precision in case of 4326.
 - [ ]  debounce user input to update with 1s of delay.
 - [ ]  Increase step for the input component at least to 0.01
 - [ ]  Specify the unit of measure of the circle radius (m)

kappu72:
## Description
If Query Panel Geometry details has proj 4326 the precision off numeric fields  is increased to 6th decimal, the step is related to the current scale, and the update is debounced by 1 sec.
## Issues
 - Fix #2668

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

All numeric fields has tow decimal precision, and the step is always one

**What is the new behaviour?**
If proj is 4326, coordinates fields has 6 decimal precision, the step is related to the scale and the update is debounced by one sec. The radius has 2 decimal precision.

 - [ ] Yes
 - [x] No

**Other information**:


web/client/components/data/query/GeometryDetails.jsx
```diff
@@ -168,28 +169,44 @@ class GeometryDetails extends React.Component {
...
-        } else if (this.props.type === "Circle") {
-            this.resetCircle();
-        }
-
+        this.resetGeom();
         this.props.onShowPanel(false);
     };
-
+    roundValue = (val, prec = 1000000) => Math.round(val * prec) / prec;
+    getStep = (zoom = 1) => {
```
mbarto: I think we can translate this into a formula, in any case we shouldn't use ifs

web/client/components/data/query/GeometryDetails.jsx
```diff
@@ -168,28 +169,44 @@ class GeometryDetails extends React.Component {
...
+        }else if( zoom >= 6) {
+            return 1
+        }
+            return 10;
+    };
+    getStepCircle = (zoom, name) => {
+        const step = this.getStep(zoom);
+        return name === 'radius' && step * 100000 || step;
+    };
+    isWGS84 = () => (this.props.geometry || {}).projection === 'EPSG:4326' || !this.props.useMapProjection;
```
mbarto: I think we can use proj4 to know the uom of the projection, and generalize EPSG:4326 to uom === 'degrees'

web/client/components/data/query/GeometryDetails.jsx
```diff
@@ -77,7 +79,7 @@ class GeometryDetails extends React.Component {
             projection: this.props.geometry.projection
         };
 
-        this.props.onChangeDrawingStatus("replace", undefined, "queryform", [geometry]);
+        this.props.onChangeDrawingStatus({geometry: [geometry]});
```
mbarto: is CHANGE_DRAWING_STATUS without a "command" (e.g. replace) supported?

web/client/components/data/query/enhancers/geometrydetails.js
```diff
@@ -0,0 +1,24 @@
...
+*
+* This source code is licensed under the BSD-style license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+const {compose, createEventHandler, defaultProps} = require('recompose');
+
+const propsStreamFactory = require('../../../misc/enhancers/propsStreamFactory');
+
+
+const dataStreamFactory = $props => {
```
mbarto: Why do we need this?

kappu72: Done to debounce the change.

mbarto: it also changes the interface to call onChangeDrawingStatus, which is not immediately clear, look at the other question a bove

web/client/components/data/query/GeometryDetails.jsx
kappu72: maybe was better to change the name, but isn't directly calling the onChangeDrawingStatus action. That action is debounced and then called by enhancer

web/client/components/data/query/enhancers/geometrydetails.js
kappu72: se answer above

web/client/components/data/query/GeometryDetails.jsx
mbarto: I mean, if you put a "transparent" debounce in the middle, it should be transparent, if I remove it the same code will work, without debouncing. We want to enhance, not to change.

web/client/components/data/query/enhancers/geometrydetails.js
mbarto: I mean, if you put a "transparent" debounce in the middle, it should be transparent, if I remove it the same code will work, without debouncing. We want to enhance, not to change.

web/client/components/data/query/GeometryDetails.jsx
mbarto: Actions have a standard interface, you are changing it here

web/client/components/data/query/GeometryDetails.jsx
```diff
@@ -168,28 +169,42 @@ class GeometryDetails extends React.Component {
...
-        } else if (this.props.type === "Circle") {
-            this.resetCircle();
-        }
-
+        this.resetGeom();
         this.props.onShowPanel(false);
     };
-
+    getStep = (zoom = 1) => {
+        if ( zoom >= 21 ) {
```
mbarto: Can we translate this into a formula?

web/client/components/data/query/enhancers/geometrydetails.js
```diff
@@ -0,0 +1,16 @@
...
+* Copyright 2018, GeoSolutions Sas.
+* All rights reserved.
+*
+* This source code is licensed under the BSD-style license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+const {compose, withHandlers} = require('recompose');
+const {debounce} = require("lodash");
+
+module.exports = compose(withHandlers((initProp) => {
```
mbarto: Can't we generalize this into a "debouncing" enhancer? I don't get the purpose of having single use enhancers in separate files

web/client/components/misc/enhancers/debounce.js
```diff
@@ -0,0 +1,27 @@
...
+* All rights reserved.
+*
+* This source code is licensed under the BSD-style license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+const {withHandlers} = require('recompose');
+const {debounce} = require("lodash");
+const emptyFunc = () => {};
+/**
+ * This enhancer debounce an action or of a passed debounceTime
```
offtherailz: This enhancer de-bounce a method passed as prop of the given time.

PR #2721, opened 2018-03-09
Fix #2661 First implementation of map widget

### LINKED ISSUE #2661: GeoDashboard: implement map widget
### Description
Implement a map widget for dashboard. ([mockup](https://demo.geo-solutions.it/mockups/mapstore2/#/dashboard))
### Acceptance Criteria AC
 - The map should be created from an existing map on mapstore or crated from an empty map.
 - In a second step you can configure layers by adding them (you will see a TOC in the wizard with the possibility to add layers from an embedded catalog)
 - Take into account future connections can be added to the map state to filter by viewport #2662

offtherailz:
## Description
A first implementation of map widget, that simply allows to select a map and add to the geodashboard 
## Issues
 - Fix #2661

 - [x] Feature

No map widget

**What is the new behavior?**
you can add to the dashboard (not to the map) map widget, browsing existing maps.

 - [x] No

**Other information**:
For devs: 
 I created a `BaseMap` that is an abstract representation of map component without implementation. The implementation must be passed as `plugins` prop. 

Then 2 enhancers:
 - `mapType`: passing the `mapType` string prop, enhances a `BaseMap` with the implementation.
 - `autoMapType`. connect to app state `mapType` to set the `mapType` string prop have to be used with mapType and BaseMap
So you can use them this way: 
```
const NoTypeMap = mapType(BaseMap);
const TypedMap = compose(
    autoMapType, 
   mapType
)(BaseMap);
 // render
<NoTypeMap mapType="openlayers" layers={...} />
<TypedMap layers={...}  />
```
For the map there is also a autoResize that enhances the map to autotrigger the resize by itself using a polyfill for resize observer (implemented by the `withResizeSpy` enhancer, for general divs, in misc).


web/client/components/map/__tests__/BaseMap-test-chrome.jsx
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright 2018, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const React = require('react');
```
mbarto: Why it doesn't work for Firefox?

web/client/components/map/BaseMap.jsx
```diff
@@ -0,0 +1,146 @@
+/*
```
mbarto: Can this component replace this one: https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/Map.jsx#L119 ?

web/client/components/map/__tests__/BaseMap-test-chrome.jsx
offtherailz: maptype is cesium

mbarto: Ok, got it, because of Cesium

web/client/components/map/BaseMap.jsx
offtherailz: Yes I developed it also for this purpose. It's a simplified and more clear version of it. Some property mapping, transformation should be enough to replace it in the future. 
Didn't want to replace a so important element at this stage.

mbarto: I would go for replace it now, unless we want to replace it "mai"

offtherailz: Yes, I know. But now base map has a minimal support and is not enough tested. if I replace it I've got to fully support cesium and other minor things that are out of scope.

offtherailz: we could open a separate issue for that

PR #2737, opened 2018-03-13
Temp Fix for #2736. Disable Tutorial for IE

### LINKED ISSUE #2736: Tutorial can not be closed in internet explorer
### Description
Using internet explorer the tutorial doesn't react on click on "skip" or "next".

*Browser Affected*
- [x] Internet Explorer
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

- Tested on v11

*Steps to reproduce*

- Open MapStore with Internet explorer 11 (empty browser cache before this to prevent cookies about skip tutorial invalidate the test)
-  Open a map and then the tutorial

*Expected Result*

- I can click on skip or next

*Current Result*

- Clicking on "skip" or "next" doesn't have any effect

Notes: in some installation (e.g. c040) we have a welcome page that grays out the page and invites the user to make the tutorial. Also in this case the 2 buttons do not work, and this situation is more critical because it seems like you can't do anything else. The only thing to do is to select the checkbox of (don't ask again) and refresh the page. 

I suggest to disable the plugin for IE using `disablePluginIf`

offtherailz:
## Description
This disabled tutorial for IE users
## Issues
 - Fix #2736

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: generic configuration of plugins

see #2736

**What is the new behavior?**
The tutorial is not present for IE users

 - [ ] Yes
 - [x] No

### DIFF
--- a/web/client/plugins/Tutorial.jsx
+++ b/web/client/plugins/Tutorial.jsx
@@ -122,6 +122,7 @@ const Tutorial = connect(tutorialPluginSelector, (dispatch) => {
 
 module.exports = {
     TutorialPlugin: assign(Tutorial, {
+        disablePluginIf: "{state('browser') && (state('browser').ie || state('browser').ie11)}",
         BurgerMenu: {
             name: 'tutorial',
             position: 1000,


mbarto:
Did we understand why it doesn't work for IE?

offtherailz:
No, I didn't. I'm afraid it's the lib used for tutorial that need to be updated: 
see this: https://github.com/gilbarbara/react-joyride/issues/238

simboss:
Why we are disabling the tutorial? Can't we fix this?

offtherailz:
This PR is only a quick fix applied only to users of Internet Explorer that avoids to lock the screen of the user. 
I can investigate more, but I didn't thing that internet explorer users (10%  of world users, in better estimation, decreasing) absolutely need the tutorial, even broken ( maybe a tutorial to change browser :) ).
Jokes apart, let me know if we need to investigate more. I ported this workaround from a custom project, so it took me 0 time

tdipisa:
@offtherailz in my opinion we can consider this just a temporary solution to move forward with other most important pending activities. Anyway the real fix should be obviously implemented within the current milestone.

mbarto:
In my opinion if the fix needs to be made before the release, we don't need a temporary fix

simboss:
I agree with @mbarto, if we are going to implement a fix I don't see why we should disable it meanwhile as we are on a development branch. This is just confusing, next time let's move forward in a more linear way.

PR #2818, opened 2018-04-11
Fix #2812 parseInitialState function 

### LINKED ISSUE #2812: Cannot customize initial state with array of strings
### Description
I tried to customize initial state with the default state of featuregrid reducer and ended up with an error that was raised on opening of the feature grid.
array of string are not handled correctly [here ](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/app/StandardApp.jsx#L124)
it ended to transform the array ["ADMIN"] into {0: "ADMIN"}

there was also a problem with toChangesMap function in FeatureGridUtils because changes was an empty object instead of an array.

*Browser Affected*
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

*Steps to reproduce* (on a local environment)

- Copy default state of feature grid reducer in localconfig
- try to open a map (nothing shows)

once you have fixed the FeatureGridUtils problem
- try to open the feature grid 

*Expected Result*

- the feature grid opened correctly with no errors

*Current Result*
- you have an error in the console

here is the fix for FeatureGridUtils
```
const {isArray} = require('lodash');
const toChangesMap = (changesArray) => isArray(changesArray) ? changesArray.reduce((changes, c) => ({
    ...changes,
    [c.id]: {
        ...changes[c.id],
        ...c.updated
    }
}), {}) : {};
```

MV88:
## Description
Now it is possible to add ["string value"] inside initalState configuration
it is parsed correctly.

## Issues
 - Fix #2812

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see issue #2812

**What is the new behavior?**
Now it is possible for example to copy the default state of reducer from featureGrid to the initialState inside configuration with no error raising in the console.
it loads correctly the map if we do that.

 - [ ] Yes
 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/app/StandardApp.jsx
+++ b/web/client/components/app/StandardApp.jsx
@@ -121,11 +121,10 @@ class StandardApp extends React.Component {
         });
     };
     /**
-     * A function used to parse the initialState in the localConfig.json
-     * If an array of simple values is parsed then just return their values, otherwise return the recursive func
-     * @param {object} state the piece of state to be parsed
-     * @param {object} context contains the mode to be used
-     * @return {object} the object parsed.
+     * It returns an object of the same structure of the initialState but replacing strings like "{someExpression}" with the result of the expression between brackets.
+     * @param {object} state the object to parse
+     * @param {object} context context for expression
+     * @return {object} the modified object
     */
     parseInitialState = (state, context) => {
         return Object.keys(state || {}).reduce((previous, key) => {
--- a/web/client/utils/FeatureGridUtils.js
+++ b/web/client/utils/FeatureGridUtils.js
@@ -35,7 +35,7 @@ const getRow = (i, rows) => rows[i];
 
 /* eslint-disable */
 
-const toChangesMap = (changesArray) => isArray(changesArray) ? changesArray.reduce((changes, c) => ({
+const toChangesMap = (changesArray = []) => isArray(changesArray) ? changesArray.reduce((changes, c) => ({
     ...changes,
     [c.id]: {
         ...changes[c.id],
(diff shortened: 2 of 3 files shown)


web/client/components/app/StandardApp.jsx
```diff
@@ -120,9 +120,19 @@ class StandardApp extends React.Component {
             this.store.dispatch(action());
         });
     };
+    /**
+     * A function used to parse the initialState in the localConfig.json
+     * If an array of simple values is parsed then just return their values, otherwise return the recursive func
```
offtherailz: it doesn't return the recursive func:

It returns an object of the same structure of the `initialState` but replacing strings like `"{someExpression}"` with the result of the expression between brackets. 
state the object to parse
context context for expression
return the modified object.

web/client/utils/FeatureGridUtils.js
```diff
@@ -35,13 +35,13 @@ const getRow = (i, rows) => rows[i];
 
 /* eslint-disable */
 
-const toChangesMap = (changesArray) => changesArray.reduce((changes, c) => ({
+const toChangesMap = (changesArray) => isArray(changesArray) ? changesArray.reduce((changes, c) => ({
```
offtherailz: what's the reason of this change?

MV88: because moving the reducer default state of featuregrid caused that changesArray to passed as empty object initially, this makes the map loading forever

offtherailz: You may use a default = [] that seems to be more clear. 
Did you changed the initial state? It should be anyway consistent.

PR #2819, opened 2018-04-11
Fix #2790 improved wizard messages

### LINKED ISSUE #2790: GeoDashboard DEV Review
### Description
the tasks below should be part of the first geodashboard release:

- [x] If I cannot use a layer for a specific widget the tooltip containing the warning must appears immediately and the icon must be properly changed using the standard one (triangle + exclamation mark)
- [x] Remove the gauge widget from configuration
- [x] #2673 - The wizard must allows to go back to the previous step (not only the next one) 
- [x] Vertical and Horizontal align of the number inside the counter widget
- [x] #2768 - Allow creating a map starting from an empty one
- [x] The table widget must show a proper message when empty
- [x] Sometime when panning a map some charts in the widget list are redrawn
- [x] Improve the titles in wizard

Bugs revealed during testing:
 - [x] edit existing map doesn't update the map position.  #2794
 - [x] Maps with google background sometime fail #2786 
 - [x] selecting chart widget and a layer in the Widget Catalog you can proceed without selecting a Chart type, the result configuring chart options is that you have a line chart by default and this description fields:
immagine

offtherailz:
## Description
Improve widget builder messages. Also localized missing strings and improved usability. 
## Issues
 - Fix #2790

 - [x] Feature

 - [x] No

**Other information**:
introduced `localizeStringMap` that localized props like layer or group titles as properties.
```
const obj1 = {
"default": "String",
"en-US": "String localized in en-US"
...
}
const str = "TEST"
const CMP = localizeStringMap('testProp')(MyCmp);
<CMP testProp={obj1 || str} /> // the nested component will receive testProp="TEST" if the passed object is  str. Passing obj1 the nested component will receive  testProp="String localized in en-US" is current locale is testProp="en-US" or default other wise.
```

### DIFF
--- a/web/client/components/misc/enhancers/localizeStringMap.js
+++ b/web/client/components/misc/enhancers/localizeStringMap.js
@@ -30,7 +30,7 @@ const accumulate = (props, locale) => (acc = {}, propName) => ({
  * @param  {string|[string]} propNames Name of the prop(s) to replace. can be an array or a single prop
  * @return {HOC}         An HOC that replaces the prop string with localized string.
  * @example
- * const Input = localizeProps('title')(TitleBar);
+ * const Input = localizeStringMap('title')(TitleBar);
  * // render
  * //...
  * <Input placeholder={{title}} />
--- a/web/client/components/misc/enhancers/localizeStringMap.js
+++ b/web/client/components/misc/enhancers/localizeStringMap.js
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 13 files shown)


web/client/components/misc/enhancers/localizeStringMap.js
```diff
@@ -0,0 +1,46 @@
...
+ *  "en-US": "TEST en-US"
+ * }
+ * ```
+ * @name localizeStringMap
+ * @memberof components.misc.enhancers
+ * @param  {string|[string]} propNames Name of the prop(s) to replace. can be an array or a single prop
+ * @return {HOC}         An HOC that replaces the prop string with localized string.
+ * @example
+ * const Input = localizeProps('title')(TitleBar);
+ * // render
```
allyoucanmap: shouldn't it be localizeStringMap?

PR #2830, opened 2018-04-18
Fix #2610 clipped long descriptions of catalog cards

### LINKED ISSUE #2610: Long descriptions on cards of catalog enable y scrollbar
### Description
Long descriptions on cards of catalog enable y scrollbar amd this behaviour should be avoided.

cat

*Browser Affected*
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

- Chrome v63

*Steps to reproduce*

- open a map
- open catalog  

*Expected Result*

- cards of catalog with long description should be clipped with a text-overflow ellipsis

*Current Result*

- cards of catalog with long description enable y scrollbar

Mloweedgar:
## Description
Clipped long descriptions of catalog cards with text-overflow ellipsis.

## Issues
 - Fix #2610

 - [x] Bugfix

 - [x] No

### DIFF
--- a/web/client/components/catalog/RecordItem.css
+++ b/web/client/components/catalog/RecordItem.css
@@ -27,6 +27,11 @@
     overflow: hidden;
 }
 
+.record-item-description {
+    width: 350px;
+    white-space: nowrap;
+}
+
 .record-item .preview{
     float: left;
     width: 150px;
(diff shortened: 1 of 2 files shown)


web/client/components/catalog/RecordItem.css
```diff
@@ -27,6 +27,11 @@
     overflow: hidden;
 }
 
+.record-item-description {
+    width: 350px;
```
allyoucanmap: If possible we prefer to avoid fixed width, in particular for nested elements.
This because children elements should adapt their size based on parent element, in this case the catolog panel.
Could you verify if it still works removing width in particular with description without white spaces?

Mloweedgar:
It doesn't work when I remove width. I am trying a different approach to fix this. I want to use lodash

web/client/components/catalog/RecordItem.css
Mloweedgar: It didn't work when I removed width, therefore have done another implementation which uses lodash to achieve the same results .

PR #2841, opened 2018-04-24
Fixes #2824: enable elevation layer only for cesium or when mouseposition is enabled

### LINKED ISSUE #2824: Terrain layer should download only when needed
### Description
The current solution for elevation is add an hidden layer that downloads image/bil tiles with the special flag useForElevation. 

This causes the download of tiles even when not necessary. We should set the layer initially not visible and enable it only when the elevation tool is active. This should help openlayers and leaflet to avoid to download unnecessary tiles, and should be not a problem for cesium that uses the layer in a different way. 
See https://github.com/geosolutions-it/MapStore2/pull/2816

The best solution is to trigger  1 tile download only on demand, when mouse stops, for instance. An intermediate solution is to make the layer visible (and so download tiles) only when mouseposition is enabled.

mbarto:
## Issues
 - See title

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

elevation layer is always active and loaded at every zoom

**What is the new behavior?**
elevation layer is active only when mouseposition plugin is enabled or the map type is cesium

 - [ ] Yes
 - [x] No

**Other information**:

### DIFF
--- a/web/client/plugins/Map.jsx
+++ b/web/client/plugins/Map.jsx
@@ -152,7 +152,8 @@ class MapPlugin extends React.Component {
         toolsOptions: PropTypes.object,
         actions: PropTypes.object,
         features: PropTypes.array,
-        securityToken: PropTypes.string
+        securityToken: PropTypes.string,
+        elevationEnabled: PropTypes.bool
     };
 
     static defaultProps = {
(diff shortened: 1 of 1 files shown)


offtherailz:
@mbarto long story short, I'll merge it. but we should open an issue to move this functionality out from the map, as for other additional layers. Could you?

PR #2938, opened 2018-05-22
Fix #2306: preserving changes on map after login

### LINKED ISSUE #2306: Map changes are lost on login
On the previous release an user could:
- Access MapStore2 as anonymous user
- Open a public map
- Add/remove layers
- Login
- Save as a new map

Currently this is not possible because when a user logs in from an open map, that map reloads, lossing any change made in the meanwhile.

Is this by design?

Mloweedgar:
## Description
Changes made by an anonymous user on map are preserved when that user logs in 
## Issues
 - Fix #2306 

 - [x] Bugfix

 - [x] No


web/client/selectors/maps.js
```diff
@@ -24,6 +24,8 @@ const mapDescriptionSelector = (state, id) => mapFromIdSelector(state, id) && ma
 const mapDetailsUriFromIdSelector = (state, id) => mapFromIdSelector(state, id) && mapFromIdSelector(state, id).details || "";
 const mapPermissionsFromIdSelector = (state, id) => mapFromIdSelector(state, id) && mapFromIdSelector(state, id).permissions || "";
 const mapThumbnailsUriFromIdSelector = (state, id) => mapFromIdSelector(state, id) && mapFromIdSelector(state, id).thumbnail || "";
+const hasMapAccessLoadingError = (state) => get(state, state.maps && "maps.loadingError");
```
offtherailz: get already checks if some of the elements in the path are null, so you should simply use get(state, "maps.loadingError")

web/client/epics/login.js
```diff
@@ -42,6 +43,7 @@ const refreshTokenEpic = (action$, store) =>
 const reloadMapConfig = (action$, store) =>
     action$.ofType(LOGIN_SUCCESS, LOGOUT)
     .filter(() => pathnameSelector(store.getState()).indexOf("viewer") !== -1)
+    .filter(() => hasMapAccessLoadingError(store.getState))
```
offtherailz: What happens on logout?  
Try to reload the resource metadata (canEdit, canView) for anonymous user. If the permission are view only it's ok (Check that you are not able to save anyway).
If you can not see the map anymore, you could redirect the user to the home page or show the same error.

web/client/epics/login.js
```diff
@@ -42,6 +43,9 @@ const refreshTokenEpic = (action$, store) =>
 const reloadMapConfig = (action$, store) =>
     action$.ofType(LOGIN_SUCCESS, LOGOUT)
     .filter(() => pathnameSelector(store.getState()).indexOf("viewer") !== -1)
+    .filter((data) => {
+        return data.type !== "LOGOUT" ? hasMapAccessLoadingError(store.getState) : true;
```
offtherailz: store.getState is a function, I'm afraid `hasMapAccessLoadingError(store.getState) will always return false

web/client/selectors/maps.js
offtherailz: I think loadingError is in `mapinitialConfig` state. Isn't it? did you tried this?

web/client/selectors/mapInitialConfig.js
```diff
@@ -0,0 +1,19 @@
...
+* All rights reserved.
+*
+* This source code is licensed under the BSD-style license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+const {get} = require('lodash');
+/**
+ * Checks for an error when a map is accessed
+ * @function
+ * @memberof selectors.map
```
offtherailz:
```
@memberof selectors.mapInitialConfig
```

PR #2959, opened 2018-05-29
Fix #2465: fixed problem in crosslayer filtering when url of layers look different

### LINKED ISSUE #2465: Problem in crosslayers filtering (MS2) when urls of layers are different
### Description
Problem in crosslayer filtering (MS2) when url of layers are different:
the application checks the url of the layers to verify what layers are available for crossLayer filter (they have to be the same). If urls are different the crosslayer filtering found only the current layer in Target layer despite. 
For example adding this 2 layers:
- Spearfish roads
- USA Population

urls are:
https://demo.geo-solutions.it:443/geoserver/wfs"
https://demo.geo-solutions.it/geoserver/wfs
This is probably because the US States map was saved a long time ago and in the meanwhile something changed search URL save procedure, so now you have also the port.

It is not a problem of the crossLayer itself, anyway we can make it smarter and understand that if https is present, the 443 port is the same of no port, and the same for http and port 80.

- [x] Bug
- [ ] New Feature

### In case of Bug:

*Browser Affected*
all

*Steps to reproduce*
- open this test map: https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/leaflet/3382
- check the problòem with the crosslayers filtering: in Target layer only one layer is available despite 2.

*Expected Result*
- both layer present

*Current Result*
- only the current layer present

url_croslayersfilt

Mloweedgar:
## Issues
 - Fix #2465 

 - [x] Bugfix

 - [x] No

### DIFF
--- a/web/client/components/data/query/CrossLayerFilter.jsx
+++ b/web/client/components/data/query/CrossLayerFilter.jsx
@@ -18,8 +18,19 @@ const Select = require('react-select');
 const GeometricOperationSelector = require('./GeometricOperationSelector');
 const GroupField = require('./GroupField');
 
+const isSameUrl = (originalUrl, otherUrl) => {
 
-const isSameOGCServiceRoot = (origSearchUrl, {search, url} = {}) => origSearchUrl === url || origSearchUrl === (search && search.url);
+    if (originalUrl !== otherUrl) {
+        let originalUrlArray = originalUrl.match(/([^:]*:)\/\/([^:]*:?[^@]*@)?([^:\/\?]*):?([^\/\?]*)/);
+        let otherUrlArray = otherUrl.match(/([^:]*:)\/\/([^:]*:?[^@]*@)?([^:\/\?]*):?([^\/\?]*)/);
+        let isSameProtocol = originalUrlArray[1] === otherUrlArray[1];
+        let isSameDomain = originalUrlArray[3] === otherUrlArray[3];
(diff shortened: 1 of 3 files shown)


web/client/components/data/query/CrossLayerFilter.jsx
```diff
@@ -18,8 +18,19 @@ const Select = require('react-select');
...
-const isSameOGCServiceRoot = (origSearchUrl, {search, url} = {}) => origSearchUrl === url || origSearchUrl === (search && search.url);
+    if (originalUrl !== otherUrl) {
+        let originalUrlArray = originalUrl.match(/([^:]*:)\/\/([^:]*:?[^@]*@)?([^:\/\?]*):?([^\/\?]*)/);
+        let otherUrlArray = otherUrl.match(/([^:]*:)\/\/([^:]*:?[^@]*@)?([^:\/\?]*):?([^\/\?]*)/);
+        let isSameProtocol = originalUrlArray[1] === otherUrlArray[1];
+        let isSameDomain = originalUrlArray[3] === otherUrlArray[3];
+        let isSamePath = originalUrl.slice(originalUrlArray[0].length) === otherUrl.slice(otherUrlArray[0].length);
+        return isSameDomain && isSameProtocol && isSamePath;
+    }
+    return originalUrl === otherUrl;
```
mbarto: this is always true, so return true is enough

PR #2968, opened 2018-06-01
Support for time dimension (C125)

offtherailz:
## Description
Support for time dimension (porting of #2967)
support for issue [3](https://github.com/geosolutions-it/austrocontrol-C125/issues/3)

### DIFF
--- a/web/client/epics/layers.js
+++ b/web/client/epics/layers.js
@@ -8,8 +8,11 @@
 
 const Rx = require('rxjs');
 const Api = require('../api/WMS');
-const {REFRESH_LAYERS, layersRefreshed, updateNode, layersRefreshError} = require('../actions/layers');
+const { REFRESH_LAYERS, UPDATE_LAYERS_DIMENSION, layersRefreshed, updateNode, layersRefreshError, changeLayerParams} = require('../actions/layers');
+const {getLayersWithDimension} = require('../selectors/layers');
+
 const LayersUtils = require('../utils/LayersUtils');
+
 
 const assign = require('object-assign');
 const {isArray, head} = require('lodash');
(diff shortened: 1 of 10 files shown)


web/client/reducers/layers.js
```diff
@@ -102,15 +102,24 @@ function layers(state = [], action) {
             });
             return assign({}, state, {refreshing: newLayers});
         }
+        case CHANGE_LAYER_PARAMS:
```
mbarto: So, is CHANGE_LAYER_PARAMS just an alias of CHANGE_LAYER_PROPERTIES ?

web/client/epics/layers.js
```diff
@@ -86,7 +89,26 @@ const refresh = action$ =>
...
         });
 
-
+/**
+ * Update dimension to all layers that have that dimension set, or for the layers indicated in the action.
+ * @memberof epics.layers
+ * @param {external:Observable} action$ manages `UPDATE_LAYERS_DIMENSION`
+ * @return {external:Observable}
+ */
+const updateDimension = (action$, {getState = () => {}} = {}) =>
```
mbarto: What is the side effect here?

web/client/reducers/layers.js
offtherailz: Yes, the action creators are different in terms of entries they add to the action object, and this is the reason of different . 
I don't know if it is easier to use the same action name with 2 different action creators or use 2 different names. Do you think I should change it?

web/client/epics/layers.js
offtherailz: What do you mean? Triggers a changeLayerParams for the layers passed as property or all the layer that have the dimension, if layers array is empty.
Notice that I have extended CHANGE_LAYER_PARAMS (PROPERTIES) to accept 1 string or an array of string for layers to change, to make multiple updates easier.

web/client/plugins/AutoMapUpdate.jsx
```diff
@@ -78,5 +77,5 @@ const AutoMapUpdatePlugin = connect(autoMapUpdateSelector, {
 module.exports = {
     AutoMapUpdatePlugin,
     reducers: {},
-    epics: {manageAutoMapUpdate, refresh}
```
MV88: no longer needed this?

offtherailz: No, refresh is now included in defaults

PR #3013, opened 2018-06-19
Import export porting

offtherailz:
## Description
This PR ports Import and Export plugins to main branch

Note: we should remove the ShapeFile plugin
TODO: 
 - [ ] Review text strings with @tdipisa


web/client/actions/mapimport.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * Copyright 2016, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const SET_LAYERS = 'IMPORT::SET_LAYERS';
```
mbarto: Why double colon?

mbarto:
Aren't we replicating a lot of code / functionality that is already in the Shapefile plugin without removing / aliasing it?

offtherailz:
@mbarto, yes I did the PR so we can discuss if remove or alias shapefile.

web/client/components/import/dragZone/enhancers/processFiles.jsx
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const Rx = require('rxjs');
+const { compose, mapPropsStream, createEventHandler} = require('recompose');
+const FileUtils = require('../../../../utils/FileUtils');
+const LayersUtils = require('../../../../utils/LayersUtils');
+
+const JSZip = require('jszip');
```
mbarto: All this code is already in the Shapefile plugin, isn't it?

web/client/components/import/style/StylePanel.jsx
```diff
@@ -0,0 +1,198 @@
+const PropTypes = require('prop-types');
```
mbarto: Is this completely different than the styling panel used by the Shapefile plugin?

web/client/actions/mapimport.js
offtherailz: Good catch, I will fix it together with shapefile stuff

web/client/components/import/style/StylePanel.jsx
offtherailz: No, it is the same, but more isolated from import stuff, so we may be able to improve and reuse it in a easier way.

offtherailz:
@tdipisa and me discussed about the opportunity to maintain the shapefile as a wrapper. 
We have decided that it is better to delete the shapefile plugin and add to the existing configuration Import and Export map plugin. 
We have to schedule time to 
 - remove ShapeFile plugin 
 - replace existing configurations with the new one
 - Document migration in migration guide

MV88:
As a side note, the supported GeoJSON format does not support the type **Feature** but instead requires an object containing an array of Feature under the features props, i.e a "FeatureCollection" something like:
`{features: [{GeoJSON feature}, ...]}`
and by the way it does not check the mandatory presence of any **type, properties, and geometry** properties and it does check for example the Right hand rule
IMO: we should add/use a GeoJSON lint functionality when importing a GeoJSON Object

PR #3033, opened 2018-06-26
Improved radius with unit of measure

offtherailz:
## Description
This is for c125 adding radius for UOM

 -  Required by https://github.com/geosolutions-it/austrocontrol-C125/issues/84

 - [x] Feature

You had the annotation's radius only in meters

**What is the new behavior?**
You can select different unit of measure for radius. You can configure unit of measure for radius in initialState

 - [ ] Yes
 - [x] No

### DIFF
--- a/web/client/components/mapcontrols/annotations/MeasureEditor.jsx
+++ b/web/client/components/mapcontrols/annotations/MeasureEditor.jsx
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2018, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const React = require('react');
+const { FormControl, FormGroup } = require('react-bootstrap');
+const { isNumber } = require('lodash');
+const { convertUom } = require('../../../utils/MeasureUtils');
+
+// convert to valueUom if it is a valid number
+const toValue = (value, uom, valueUom) => (isNumber(parseFloat(value)) && !isNaN(parseFloat(value)))
+    ? convertUom(parseFloat(value), uom, valueUom)
+    : value;
+// convert to local uom if it is a valid number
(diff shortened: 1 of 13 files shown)


web/client/components/mapcontrols/annotations/MeasureEditor.jsx
```diff
@@ -0,0 +1,93 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+const React = require('react');
+const { FormControl, FormGroup } = require('react-bootstrap');
+const { isNumber } = require('lodash');
+const { convertUom } = require('../../../utils/MeasureUtils');
+
+// convert to valueUom if it is a valid number
+const toValue = (value, uom, valueUom) => (isNumber(parseFloat(value)) && !isNaN(parseFloat(value)))
```
MV88: Can these functions be moved into some utils (measureUtils)?

web/client/components/mapcontrols/annotations/MeasureEditor.jsx
```diff
@@ -0,0 +1,93 @@
...
+// convert to local uom if it is a valid number
+const toLocalValue = (value, uom, valueUom) =>
+    isNumber(parseFloat(value)) && !isNaN(parseFloat(value))
+        ? parseFloat(convertUom(value, valueUom, uom).toFixed(4))
+        : value;
+
+const { compose, withHandlers, withPropsOnChange, withState, withStateHandlers, defaultProps} = require('recompose');
+
+
+module.exports = compose(
```
MV88: why not a separate enhancer?

offtherailz: If needed in other contexts (for instance spatial filter) this can be moved in a common folder and we may have to separate logic from the component (only if we need to append a different logic, or for testability reasons). 
Now is not needed, so I prefer to keep the logic near the component.

PR #3037, opened 2018-06-28
Update .travis.yml

offtherailz:
## Description
Try to update travis.yml to avoid recent timeout errors, extending stale time to 20 minutes.  
This is only a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_install:
     - node -v
     - npm -v
 branches:
+install: travis_wait npm install
   only:
     - master
     - /^[cC]\d{3}[\w-]*/


offtherailz:
This seems to work. The problem is caused by a process in install that takes more then 10 minutes without any output.@mbarto, is it normal for you? Is there some pre or post install process that takes so long?

offtherailz:
in this test  I had a timeout with 20 minutes: `https://travis-ci.org/offtherailz/MapStore2/builds/397825895?utm_source=email&utm_medium=notification`

offtherailz:
Seems not needed anymore. going to close

PR #3080, opened 2018-07-13
Fix #3066. Replaced LRU cache with a IE11 compliant lib (Porting)

### LINKED ISSUE #3066: Cannot open maps with Internet Explorer 11
### Description
`lru-cache` is not compatible with IE11. So opening a map causes a failure.

*Browser Affected*
- [x] Internet Explorer

*Browser Version Affected*

- 11

*Steps to reproduce*

- Open a map with Internet Explorer 

*Expected Result*

- You can see the map

*Current Result*

- You have a blank screen

offtherailz:
## Description
See (#3067)

## Issues
 - Fix #3066

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "leaflet.locatecontrol": "0.62.0",
     "leaflet.nontiledlayer": "1.0.7",
     "lodash": "4.16.6",
-    "lru-cache": "4.1.2",
+    "lrucache": "1.0.3",
     "moment": "2.21.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -121,7 +121,7 @@
     "leaflet.locatecontrol": "0.62.0",
     "leaflet.nontiledlayer": "1.0.7",
     "lodash": "4.16.6",
-    "lru-cache": "4.1.2",
```
MV88: verify that also standard package.json is fixed

PR #3124, opened 2018-07-27
Moved geostrore api in persistence api

kappu72:
## Description
It becomes possible to implement and plug different api to store resources in backend.
The default implementation is the geostore one.
The new api needs to implement 4 methods getResource, createResource, updateResource and deleteResource.
It's possible to select the added implementation configuring persistenceApi property in localConfing or by manually setting the current api by the setApi method of the persistence object.
## Issues
 - ...

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

Mapstore2 uses gestore backend and api

**What is the new behavior?**
It allows to implement and select different backend/api

 - [ ] Yes
 - [x] No

**Other information**:


web/client/observables/persistence/index.js
```diff
@@ -0,0 +1,94 @@
+/**
```
mbarto: Not clear why this should be an observable. It looks more like an API (provider) to me.

offtherailz: Good observation. 
What if we move persistence API (this index.js) object in API folder, leaving here the implementation of `geostore` ? @mbarto @kappu72

mbarto: Not sure, does the developer calling this (the GeoStore impl of this) need to know that this is implemented using observables? I mean, does the caller need to use RxJs or not? Looking at the interface, I cannot understand it.

kappu72: CreateResource and getResource methods return observables, so I guess that RxJs is needed

mbarto: Then I think we should document how the caller should use those methods

offtherailz: I agree.
So, summarizing: 
 - [ ] document the interface where missing. updateResource and deleteResource should describe the observable returned (create and get have the `@returns`). 
 - [ ] Move the index.js into api folder
 - [ ] Fix the white space issue ( I introduced it trying to enhance documentation, you can remove it applying these changes)
Did I missed anything else @mbarto  ?

mbarto: No, I think it's everything, thanks

PR #3181, opened 2018-09-06
Fixes #3104: issues with printing

### LINKED ISSUE #3104: Print a map is unstable with respect to zoom levels or Sheet size
### Description
Changing the zoom level in the print preview then printing or changing the Sheet size gives two errors types:
1) Stretched layout 
2) Blank layout

*Browser Affected*
- All

*Steps to reproduce*

1) Stretched layout
- Open a map with a vector layer, specifically open this one https://dev.mapstore.geo-solutions.it/mapstore/#/viewer/openlayers/3264
- zoom in/out the map  
- From the main menu, click on print. Leave the default options then print.
  
image

- You will get a stretched layout

image

2) Blank layout
- Zoom out in order to see all the north american continent in the preview

image

- Leave the default options then print, you will get a blank layout
image

Try it also changing from A4 to A3 and Portrait to Landscape. 
Moreover try to zoom out in order to see all the world map in the preview. You will get this error

image

mbarto:
## Description
#3104

## Issues
 - Fix #3104 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

geodetic print on by default

**What is the new behavior?**
geodetic print off by default

 - [x] Yes
 - [ ] No

Applications in need for geodetic true should override the param explicitly.

**Other information**:

### DIFF
--- a/web/client/utils/PrintUtils.js
+++ b/web/client/utils/PrintUtils.js
@@ -151,7 +151,7 @@ const PrintUtils = {
             "layout": PrintUtils.getLayoutName(spec),
             "dpi": parseInt(spec.resolution, 10),
             "outputFilename": "mapstore-print",
-            "geodetic": true,
+            "geodetic": false,
             "mapTitle": spec.name || '',
             "comment": spec.description || '',
             "layers": PrintUtils.getMapfishLayersSpecification(spec.layers, spec, 'map'),
--- a/web/client/utils/__tests__/PrintUtils-test.js
+++ b/web/client/utils/__tests__/PrintUtils-test.js
@@ -202,7 +202,7 @@ describe('PrintUtils', () => {
         expect(printSpec).toExist();
         expect(printSpec.dpi).toBe(96);
         expect(printSpec.layers.length).toBe(1);
-        expect(printSpec.geodetic).toBe(true);
+        expect(printSpec.geodetic).toBe(false);
     });
     it('from rgba to rgb', () => {
         const rgb = PrintUtils.rgbaTorgb("rgba(255, 255, 255, 0.1)");


mbarto:
About the first issue: what does it mean that a printed map should be measurable? Please, explain.

The second issue is about making the configuration overridable, so making the geodetic parameter true by default was not requested / needed.

offtherailz:
About the 1st issue, it means geodetic (measures by hand should be the same of the measure with the online measure tool and the same of real world measures, within the obvious limits of this phrase). 

About the second issue, of course, geodetic parameters should work. We made it true by default because around 90% of the users expect to to be able to measure distances on the printed map. 

Removing a functionality by default does not mean solving an issue. If we want to remove that functionality, please ask confirm to @tdipisa

mbarto:
Measuring a printed map like we do on the web app is something that, in general, cannot be expected.
I know that users expect that, but it doesn't mean they are doing the right thing,

A decade ago I had to put an explicit notice in printed maps that they cannot be measured, because people insisted to do it, without using a proper projection. There are projections that preserve distance (or areas) and you can expect to use a ruler with those projections, but not with the most common used ones.

When we do a measure in MapStore2 we apply calculations to get a real metric measure that a human is not going to do when using a ruler on a piece of paper.

EPSG:3857 maps (the ones we use most) have a distortion that makes this (by hand) operation almost impossible.

The geodetic option in mapfish-print is used to get a better scalebar (that probably is good only on one direction, the horizontal one, and only in the center of a map at certain scales) and to transform the center - scale parameters to a bbox for wms layers.
The geodetic option eventually has bugs, in doing that.

Having in mind that we will not be able to make a map measurable, we could eventually:

 * try to identify mapfish-print bugs in geodetic: true mode, to avoid distortions and blank prints
 * try to send a bbox to mapfish-print, instead of center + scale, moving the logic into MapStore2

offtherailz:
Yes I know how it works. I only asked that if we want to remove this functionality, you have to ask to @tdipisa first.

offtherailz:
We could even explain to each customer how it works and propose to print in a different projection. 
I'm just stressing that it's a test that is required and diabilitarla would cause some discomfort. Improve it is a valid option

PR #3246, opened 2018-10-17
Fix #3243 circle radius problem for openlayers

### LINKED ISSUE #3243: Circle radius calculated after confirm in geometry details is wrong (OL only)
### Description
This problem happens if you have configurated to draw a geodesic Circle 
The radius is re-calculated from the approximated polygon and it is different from the starting one use to create that polygon.

inside spatialMethodOptions of cfg of QueryPanel add geodesic :true
`{"id": "Circle", "name": "queryform.spatialfilter.methods.circle", "geodesic": true},`

*Browser Affected*
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*
Chrome 69.0.3497.100

*Steps to reproduce*
- Open any map with a queryable layer "US Population Statistics" for example.
- Open featureGrid and query form
- Draw a Circle
- Enter geometry details and change its value to 10000 (zoom in/out to see it comfortably)
- Click on confirm ( the V icon)
- Re-enter the geometry details

*Expected Result*
- To see the same radius previously inserted

*Current Result*
- a different value is present. 

that is because of [this](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/map/openlayers/DrawSupport.jsx#L565-L568)

MV88:
## Description
see title

## Issues
 - Fix #3243 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

now the radius keeps the value inserted  in the geometry details form.

**What is the new behavior?**
see #3243 

 - [ ] Yes
 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/map/openlayers/DrawSupport.jsx
+++ b/web/client/components/map/openlayers/DrawSupport.jsx
@@ -176,8 +176,8 @@ class DrawSupport extends React.Component {
         if (olFeature) {
             const feature = this.fromOLFeature(olFeature);
             if (newProps.drawMethod === "Circle" && newProps && newProps.features && newProps.features.length && newProps.features[0] && newProps.features[0].radius >= 0) {
-                // this is needed because the calculation of the radius is wrong
-                // because it is done starting from the approximated polygon
+                // this prevents the radius coming from `fromOLFeature` to override the radius set from an external tool
+                // this is because `endDrawing` need to impose the radius value, without any re-calculation or approximation
                 feature.radius = newProps.features[0].radius;
             }
             this.props.onEndDrawing(feature, newProps.drawOwner);
--- a/web/client/components/map/openlayers/DrawSupport.jsx
+++ b/web/client/components/map/openlayers/DrawSupport.jsx
@@ -176,8 +176,8 @@ class DrawSupport extends React.Component {
         if (olFeature) {
             const feature = this.fromOLFeature(olFeature);
             if (newProps.drawMethod === "Circle" && newProps && newProps.features && newProps.features.length && newProps.features[0] && newProps.features[0].radius >= 0) {
-                // this is needed because th calculation of the radius is wrong
-                // and is not needed when the geometry details is edited
+                // this is needed because the calculation of the radius is wrong
+                // because it is done starting from the approximated polygon
                 feature.radius = newProps.features[0].radius;
             }
             this.props.onEndDrawing(feature, newProps.drawOwner);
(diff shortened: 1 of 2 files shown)


web/client/components/map/openlayers/__tests__/DrawSupport-test.jsx
```diff
@@ -22,6 +23,14 @@ describe('Test DrawSupport', () => {
         setTimeout(done);
     });
 
+    /**
+     * used to reduce a bit the boilerplate and ReactDom stuff
+     * TODO extend it to the other tests
+    */
+    const renderDrawSupport = (props = {}) => {
+        ReactDOM.render(<DrawSupport {...props}/>, document.getElementById("container"));
```
mbarto: You don't reduce the boilerplate if you use it only in one place / test.
It's useful to refactor code in external functions if they are used at least in a bunch of different places, expecially in tests, where being able to read the full test at once is more important than duplication.
If you have too much code in one test it's better to split it in two different tests.

web/client/components/map/openlayers/DrawSupport.jsx
```diff
@@ -175,6 +175,11 @@ class DrawSupport extends React.Component {
         const olFeature = this.replaceFeatures(newProps);
         if (olFeature) {
             const feature = this.fromOLFeature(olFeature);
+            if (newProps.drawMethod === "Circle" && newProps && newProps.features && newProps.features.length && newProps.features[0] && newProps.features[0].radius >= 0) {
+                // this is needed because the calculation of the radius is wrong
```
offtherailz:
```suggestion
                // this prevents the radius coming from `fromOLFeature` to override the radius set from an external tool
```

web/client/components/map/openlayers/DrawSupport.jsx
```diff
@@ -175,6 +175,11 @@ class DrawSupport extends React.Component {
...
+                // because it is done starting from the approximated polygon
```
offtherailz:
```suggestion
                // this is because `endDrawing` need to impose the radius value, without any re-calculation or approximation
                // but fromOLFeature re-calculates the radius using some distance formula
                // TODO: externalize radius calculation out from `fromOLFeature`
```

web/client/components/map/openlayers/__tests__/DrawSupport-test.jsx
MV88: I created this function Because i wanted to write less code  in particular the parts
ReactDOM.render(<COMPONENT {props}/>, document.getElementById("container"));
I could generalize it so we can use it in general and also in other places, but I see that you prefer the older version.
Going to refactor this.

PR #3256, opened 2018-10-18
Fixes #3155 and Fixes #3266 and loads fontAwesome when the application starts

### LINKED ISSUE #3155: First point annotation has unknown icon [ ]
### Description
font-awsome icon is not correctly render on first Annotation point

*Browser Affected*
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

- Chrome v68

*Steps to reproduce*

- open a map
- refresh the page
- open annotations panel from burger menu
- add a new point Annotation
- add a point geometry on map
- click on map

*Expected Result*

-  The icon should be the one showed in the select form

*Current Result*

-  2018-08-09 14_45_59-mapstore homepage

### LINKED ISSUE #3266: Introduce a loading font mechanism
### Description
The issue #3155 has shown the need to preload esternal fonts before rendering.

We need to introduce a load font mechanism based on the fonts being used by openlayers when rendering vector layers. This is not stricly correlated to the annotations plugin because you can load a map with annotations on the embedded mode.

check out these resources
- https://hacks.mozilla.org/2016/06/webfont-preloading-for-html5-games/
- https://stackoverflow.com/questions/41852594/dynamically-load-fonts
- https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API ( only a draft atm)
- https://fontawesome.com/how-to-use/on-the-web/using-with/react

MV88:
## Description
This PR will temporarily fix the wrong render of the marker when : 
- loading a map with annotation
- after drawing a marker from the annotations plugin

and it will introduce a loading font mechanism configurable using fonts cfg in Map plugin
and it will check by default if the maptype is openlayers, but it can be overridden by the shouldLoadFont flag from configuration for the map plugin in general.

**NOTE** :  i also added the possibility to use a template string and passing the values to be used to substitute in the string.

**NOTE** :  Now the map Plugin is connected with the connect taken from PluginsUtils, be aware of that

## Issues
 - Fix #3155 
 - Fix #3266

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

a [] is displayed instead of the glyph inside the marker. because the font is not available on the very first render.

**What is the new behavior?**
The first marker you insert it shows correctly the gliph
if you load a map with annotation after cleaning the cache it loads the marker correctly

 - [ ] Yes
 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/notifications/NotificationContainer.jsx
+++ b/web/client/components/notifications/NotificationContainer.jsx
@@ -8,6 +8,7 @@
 const React = require('react');
 const NotificationSystem = require('react-notification-system');
 const PropTypes = require('prop-types');
+const {injectIntl, intlShape, defineMessages} = require('react-intl');
 
 var LocaleUtils = require('../../utils/LocaleUtils');
 
@@ -24,7 +25,10 @@ var LocaleUtils = require('../../utils/LocaleUtils');
  *    uid: "1234" // a unique identifier (if not present, current time is used),
  *    action: {
  *      label:  "label.translation.path" // or the message directly
- *    }
+ *    },
+ *    autoDismiss: 5 // Delay in seconds for the notification go away. Set this to 0 to not auto-dismiss the notification
+ *    position: tr // Position of the notification. Available: tr (top right), tl (top left), tc (top center), br (bottom right), bl (bottom left), bc (bottom center),
+ *    values: {...} // an object used for templating the string. if undefined the formatMessage will return the original string
  *  }
  * ```
  * @see https://github.com/igorprado/react-notification-system#creating-a-notification for othe options
@@ -34,6 +38,7 @@ var LocaleUtils = require('../../utils/LocaleUtils');
 class NotificationContainer extends React.Component {
     static propTypes = {
         notifications: PropTypes.array,
+        intl: intlShape.isRequired,
         onRemove: PropTypes.func,
         onDispatch: PropTypes.func
     };
@@ -78,10 +83,25 @@ class NotificationContainer extends React.Component {
         });
         notifications.forEach(notification => {
             if (systemNotifications.indexOf(notification.uid) < 0) {
+                let id = notification.message;
+                let str = LocaleUtils.getMessageById(this.context.messages, id);
+                let message;
+                if (notification.values) {
+                    // this add support the template string using the same library react-intl used for i18n
+                    const {formatMessage} = this.props.intl;
+                    message = defineMessages({
+                        id,
+                        defaultMessage: str
+                    });
+                    message = formatMessage(message, notification.values);
+                } else {
+                    message = str || id;
+                }
+
                 this.system().addNotification({
                   ...notification,
                   title: LocaleUtils.getMessageById(this.context.messages, notification.title) || notification.title,
-                  message: LocaleUtils.getMessageById(this.context.messages, notification.message) || notification.message,
+                  message,
                   action: notification.action && {
                       label: LocaleUtils.getMessageById(this.context.messages, notification.action.label) || notification.action.label,
                       callback: notification.action.dispatch ? () => { this.props.onDispatch(notification.action.dispatch); } : notification.action.callback
@@ -96,4 +116,4 @@ class NotificationContainer extends React.Component {
     };
 }
 
-module.exports = NotificationContainer;
+module.exports = injectIntl(NotificationContainer);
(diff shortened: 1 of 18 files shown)


MV88:
@offtherailz @mbarto 

You are right, we can move it in a better place.

since there is this problem only in openlayers we can : 
1. move it to the render method op openlayers/Map.jsx (tested and working)
1. move it to openlayers/plugins/VectorLayers.js (tested and **not** working)

we can add a check in the Map component to see if the annotations layer will be rendered, but i see some caveats:
- you have to pass layers info to the map component
- this does not prevent this problem when you create an annotation for the first time and the font is not loaded yet.

In my opinion we can leave it in the map render, but we have to add a the same div also to the AnnotationsEditor for example.  

fontAwesome size is 75 kB

offtherailz:
It is a little ugly to see some code specific for annotations in VectorLayer.
Anyway if we want the annotations to work also with existing maps, we have to do it, if otherwise this is not a problem, we can somehow add to vector layer the ability to pre-load some fonts for instance and in annotation's object provide the a fontUsed entry that is used in the vector's layer render method (adding something that effectively pre-load the font before rendering the layer).

MV88:
Ok, for the moment we decided to move it to the map component. 

in the future we will implement a cleaner solution. See #3266

offtherailz:
@mbarto good notes. This was a quick fix, but if using FontObservable is enough fast, we can do it. Please sync with @MV88 about implementation details.
Yes placing it in the plugin instead of the component is a good idea, if you don't have to use react to render anything.

tdipisa:
@mbarto, can you please review? Thank you.

web/client/components/notifications/NotificationContainer.jsx
```diff
@@ -78,10 +83,25 @@ class NotificationContainer extends React.Component {
         });
         notifications.forEach(notification => {
             if (systemNotifications.indexOf(notification.uid) < 0) {
+                let id = notification.message;
+                let str = LocaleUtils.getMessageById(this.context.messages, id);
```
mbarto: I would add formatting support with values directly in LocaleUtils.getMessageById(context, id, values)

web/client/plugins/Map.jsx
```diff
@@ -146,10 +168,12 @@ class MapPlugin extends React.Component {
         loadingSpinner: PropTypes.bool,
         loadingError: PropTypes.string,
         tools: PropTypes.array,
+        fonts: PropTypes.array,
         options: PropTypes.object,
         mapOptions: PropTypes.object,
         projectionDefs: PropTypes.array,
         toolsOptions: PropTypes.object,
+        onErrorFont: PropTypes.func,
```
mbarto: onFontError

web/client/plugins/Map.jsx
```diff
@@ -165,6 +189,10 @@ class MapPlugin extends React.Component {
         tools: ["measurement", "locate", "scalebar", "draw", "highlight"],
         options: {},
         mapOptions: {},
+        fonts: {
+            'ThisFontDoesNotExist': {},
```
mbarto: remove

web/client/plugins/Map.jsx
```diff
@@ -186,10 +214,35 @@ class MapPlugin extends React.Component {
...
+    };
+    state = {
+        canRender: true
     };
 
     componentWillMount() {
+        const {mapType, fonts} = this.props;
+
+        // load each font before rendering (see issue #3155)
+        if (mapType === "openlayers" && fonts) {
```
mbarto: The Map plugin is not aware of openlayers

web/client/plugins/Map.jsx
```diff
@@ -112,29 +114,49 @@ const {handleCreationLayerError, handleCreationBackgroundError, resetMapOnInit}
...
+ *         "useForElevation": true,
+ *         "nodata": -9999,
+ *         "hidden": true
+ *      }]
+ *   }
+ * }
  *
  */
+
+const updateCanRenderState = (canRender) => () => ({
```
mbarto: ???

MV88: This function updates the internal state, [setState accepts also functions](https://medium.freecodecamp.org/functional-setstate-is-the-future-of-react-374f30401b6b).
since i use it twice, i've created a function, i'll add a description above.

web/client/plugins/Map.jsx
MV88: Sure, it was there for testing

web/client/components/notifications/NotificationContainer.jsx
MV88: you cannot do that with react-int api, because they are injected from an HOC and can't be used in the LocaleUtils.
for what you are suggesting we should look at [Intl MessageFormat](https://github.com/yahoo/intl-messageformat)

mbarto: ok, for now

Issue #3154, opened 2018-08-09
Annotation icon select form should not be clearable


allyoucanmap:
### Description
Clear button on select form of Annotation icon doesn't clear the value and throws an error.
![clear value](https://user-images.githubusercontent.com/19175505/43898816-49b05b58-9be0-11e8-8c73-7e65c6c6a2c0.gif)

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

- Chrome v68

*Steps to reproduce*

- open a map
- open annotations panel from burger menu
- add a new point annotation
- edit the style of annotation
- try to clear icon select form with **X** button

*Expected Result*

- If the select form is not clearable that button should be removed 

*Current Result*
- Nothing happens on the select form
error in console:
`Uncaught TypeError: Cannot read property 'value' of null`

Issue #3121, opened 2018-07-26
Catalog search text is cleared but still effective


Gnafu:
### Description
Catalog should not filter the results without showing any filter text.

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
- [Not Tested] Internet Explorer
- [x] Chrome
- [Not Tested] Firefox
- [Not Tested] Safari

*Browser Version Affected*

- 67.0.3396.99

*Steps to reproduce*

- Open a map
- Open select "Catalog" from the menu
- Select "Demo CSW Service" service
- Type "kkk" in the search bar
- Click "Search" (No Results are shown)
- Click the "+" icon
- Type "https://demo.geo-solutions.it/geoserver/ows?service=wms&version=1.3.0&request=GetCapabilities" in the URL field
- Type "test" in the name field
- Save
- Click "Search"

*Expected Result*

- 68 pages of results should show up

*Current Result*

- No Results are shown

### Other useful information (optional):

The "kkk" filter is still active even if nothing is showing in the search bar.  
Type anything in the search bar and delete it to have proper results.

PR #3262, opened 2018-10-19
Porting of the fix for #3121 on 2018.02.xx

MV88:
## Description
Porting of the fix for #3121 (PR #3154) on 2018.02.xx


MV88:
this was wrongly created

PR #3269, opened 2018-10-22
Introduce withRef connect option in plugins connected components

mricca:
## Description
Use Refs to store references to an object on a component instance.

View this thread for more information:
https://groups.google.com/forum/?hl=it#!topic/mapstore-developers/DcAjwl6bJXc

 - [X] Feature

 - [ ] Yes
 - [X] No

**Other information**:

### DIFF
--- a/web/client/components/plugins/__tests__/PluginsContainer-test.jsx
+++ b/web/client/components/plugins/__tests__/PluginsContainer-test.jsx
@@ -10,15 +10,26 @@ const expect = require('expect');
 const React = require('react');
 const ReactDOM = require('react-dom');
 const PluginsContainer = require('../PluginsContainer');
+const {Provider} = require('react-redux');
+const {connect} = require('react-redux');
 
 class My extends React.Component {
     render() {
         return <div/>;
     }
+    myFunc() {}
 }
 const plugins = {
     MyPlugin: My,
-    OtherPlugin: My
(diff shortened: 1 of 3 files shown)


web/client/components/plugins/__tests__/PluginsContainer-test.jsx
```diff
@@ -10,15 +10,26 @@ const expect = require('expect');
...
 const plugins = {
     MyPlugin: My,
-    OtherPlugin: My
+    OtherPlugin: My,
+    WithGlobalRefPlugin: connect(
+        null,
+        null,
+        null,
+        {
+            withRef: true
```
mbarto: Is this needed at all? I don't think so, and it has the wrong name, so I would remove it to avoid confusion

mricca: withRef: true is part of the options argument of connect.
https://github.com/reduxjs/react-redux/blob/master/docs/api.md#arguments-1

You have to set this option to true to store references. When you write your plugin you have to insert this options.

withGlobalRef is a cfg properties to tell mapstore2 to call setRefToWrappedComponent only if it is true.

mbarto: Got it, you are right.

PR #3351, opened 2018-11-20
Fix #3267. Sync with map layout. other minor fixes

### LINKED ISSUE #3267: Solve conflicts between Timeline and widgets or other components
### Description
 - The timeline and widgets are overlapping. It's up to the user to place the widgets correctly (for the future we can plan to resize widgets area accordingly).
 - Hide only on editing. Timeline should be always visibile and resized accordingly when panels from every side. 

image
 How to manage these cases? 

depends on #3642

offtherailz:
## Description
This PR sync the timeline with map layout. 

NOTE: Actually the widgets and the timeline are overlapping. We have to find out a way to avoid this in the future. @giohappy can we skip widgets/timeline overlapping issues in favor of other tasks for the moment?

other minor fixes: 
 - Following toggle is sync with animation
 - Temporary disabled resize button hiding during loading to avoid issues found with some layers.


giohappy:
ok @offtherailz

PR #3353, opened 2018-11-20
geonode integration align with master

offtherailz:
Align GeoNode branch with current master.

**DO NOT SQUASH**


offtherailz:
@allyoucanmap, @kappu72 is there any reason to why you din't update it?
It was around 1 month ago since last merge

PR #3381, opened 2018-11-30
Use Alpine in the Dockerfile

Gnafu:
This produces smaller docker images (50% less) because of Alpine, not Ubuntu.
Updates Tomcat to 7.0.92
Specifies jre7
EVERY war file in the docker folder will be copied to the tomcat instance.

**Breaking changes**
- I removed the APP_NAME build argument: if you need to rename your mapstore.war file, rename it before the docker build.
- I removed the INCLUDE_MS_WAR build argument: you can always override the webapps folder with a custom volume mount

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,43 +1,26 @@
-FROM tomcat:7.0.75
+FROM tomcat:7.0.92-jre7-alpine
 MAINTAINER geosolutions<info@geo-solutions.it>
 
-RUN  export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


randomorder:
https://build.geo-solutions.it/jenkins/job/MapStore2_Dev_Build/994/console
looks like there is an extra ";" [here](https://github.com/geosolutions-it/MapStore2/pull/3381/files#diff-3254677a7917c6c01f55212f86c57fbfR16)

PR #3394, opened 2018-12-06
Writing epics dev guide

offtherailz:
Writing epics. Dev guide: first part

### DIFF
--- a/docs/developer-guide/writing-epics.md
+++ b/docs/developer-guide/writing-epics.md
@@ -83,7 +83,7 @@ This function **must return** a new stream that emits the action we want to send
 
 > **NOTE**: **redux-observable** middleware is already added to the MapStore2's StandardStore and StandardApp, so a developer should only take care of creating his own epics and add them to MapStore.
 
-Typically the stream returned by an epic is always listening for new actions and sends other actions:
+Typically the stream returned by an epic is always listening for new actions and dispatches other actions:
 
 > ** actions in, actions out **.
 
--- a/docs/developer-guide/writing-epics.md
+++ b/docs/developer-guide/writing-epics.md
@@ -79,7 +79,7 @@ The epic function has 2 arguments:
 - `action$`: the stream of of redux action. Every time an action is triggered redux, it is emitted as an event on the `action$` stream.
 - `store`. A *small version* of redux store, that contains essentially only one important method called `getState()`. This method returns the current redux state object.
 
-This function **must return** a new stream that emits the action we want to send back to redux. The **redux-observable** middleware subscribes the epics returned streams so the action will be automatically triggered on redux when they are emitted by the stream.
+This function **must return** a new stream that emits the actions we want to dispatch to redux. The **redux-observable** middleware subscribes the epics returned streams so the action will be automatically triggered on redux when they are emitted by the stream.
 
 > **NOTE**: **redux-observable** middleware is already added to the MapStore2's StandardStore and StandardApp, so a developer should only take care of creating his own epics and add them to MapStore.
 
--- a/docs/developer-guide/writing-epics.md
+++ b/docs/developer-guide/writing-epics.md
@@ -23,7 +23,7 @@ X is an error
 **Observable** is the core entity of RxJS and, more generically, of the whole reactive programming paradigm. Basically is an entity that emit events and can be subscribed, so then the subscribers can intercept the events emitted. It this the entity that implements the concept of **stream** (so *stream* and *Observable* are almost used as synonym).
 
 Subscribing observables can be hard, so RxJs provides a lot of **operators** to help manipulating and combining observables (so, *streams*).
-here an example of how operators allow to manipulate event streams to count clicks:
+here an example of how operators allow manipulating an event streams to count clicks:
 
 ```text
 clickStream:    ---c----c--c----c------c--> <-- Stream of clicks
--- a/docs/developer-guide/writing-epics.md
+++ b/docs/developer-guide/writing-epics.md
@@ -20,7 +20,7 @@ X is an error
 
 ```
 
-**Observable** is the core entity of RxJS and, more generically, of the whole reactive programming paradigm. Basically is an entity that emit events and can be subscribed, so then the subscribers can intercept the events emitted. It this the entity that implements the concept of **stream** (so *stream* and *Observable* are almost used as synonym).
+**Observable** is the core entity of RxJS and, more generically, of the whole reactive programming paradigm. Basically is an entity that emit events and can be subscribed to, so that subscribers can intercept the events emitted. This is the entity that implements the concept of **stream** (so *stream* and *Observable* are almost used as synonym).
 
 Subscribing observables can be hard, so RxJs provides a lot of **operators** to help manipulating and combining observables (so, *streams*).
 here an example of how operators allow manipulating an event streams to count clicks:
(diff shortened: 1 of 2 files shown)


docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+a, b, c, d are emitted values
+X is an error
+| is the 'completed' signal
+---> is the timeline
+
+```
+
+**Observable** is the core entity of RxJS and, more generically, of the whole reactive programming paradigm. Basically is an entity that emit events and can be subscribed, so then the subscribers can intercept the events emitted. It this the entity that implements the concept of **stream** (so *stream* and *Observable* are almost used as synonym).
+
+Subscribing observables can be hard, so the RxJs provides also a lot of **operators** that help to manipulate and combine observables (so, *streams*).
```
mbarto:
```suggestion
Subscribing observables can be hard, so RxJs provides a lot of **operators** to help manipulating and combining observables (so, *streams*).
```

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+here an example of how operators allow to manipulate event streams to count clicks:
```
mbarto:
```suggestion
here an example of how operators allow manipulating an event streams to count clicks:
```

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+                ---1----1--1----1------1--> <-- new stream returned by the operator
+                vvvvvvvvv scan(+) vvvvvvvvv <-- opeartor that does the sum
+counterStream:  ---1----2--3----4------5--> <-- click count stream returned by the operator
+```
+
+The final stream can be finally subscribed to update, for instance, a counter on the UI.
+
+### Versions
+
+At the time of writing this documentation MapStore2 are using RxJS 5.1.1 and redux-observable 0.13.0.
```
mbarto: I would not confuse the developer with too many details, just a note on the versions we use (and the link to the official documentation for those is enough in my opinion).

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+
+So make you sure to check the documentation about the current versions of these libraries.
+
+## What is an epic
+
+An **epic** is basically nothing more that:
+
+ *a function that returns a stream of redux actions to emit.*
+
+A simple epic in mapstore can be this:
```
mbarto:
```suggestion
A simple epic in mapstore can be this one:
```

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+    });
+);
+```
+
+The epic function has 2 arguments:
+
+- `action$`: the stream of of redux action. Every time an action is triggered redux, it is emitted as an event on the `action$` stream.
+- `store`. A *small version* of redux store, that contains essentially only one important method called `getState()`. This method returns the current redux state object.
+
+This function **must return** a new stream that emits the action we want to send back to redux. The **redux-observable** middleware subscribes the epics returned streams so the action will be automatically triggered on redux when they are emitted by the stream.
```
mbarto:
```suggestion
This function **must return** a new stream that emits the actions we want to dispatch to redux. The **redux-observable** middleware subscribes the epics returned streams so the action will be automatically triggered on redux when they are emitted by the stream.
```

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+
+> **NOTE**: **redux-observable** middleware is already added to the MapStore2's StandardStore and StandardApp, so a developer should only take care of creating his own epics and add them to MapStore.
+
+Typically the stream returned by an epic is always listening for new actions and sends other actions:
```
mbarto:
```suggestion
Typically the stream returned by an epic is always listening for new actions and dispatches other actions:
```

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Doing async
+
+Typical async operation involve operators like:
+
+- [switchMap](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-switchMap)
+- [mergeMap](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-mergeMap)
+- many others
+
+The base concept of all these solutions is to project to the main stream the result of some other observable.
```
mbarto: I would explain the meaning of "project"

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+...............A....A....A....A...B...A...B....B....B....B........    id
+--------------{5}--{4}--{3}--{2}-{5}-{1}-{4}--{3}--{2}--{1}------->   value
+
+```
+
+## Doing AJAX
+
+Ajax calls in MapStore should all pass by `libs/ajax.js`. This is an `axios` instance that add the support for using proxies or CORS.
+
+Axios is a library that uses es6 Promises to do ajax calls. Luckily RxJs allow to use Promises instead of streams in most of the cases. In the other cases, there are some a specific creational operators called `fromPromise` or `defer` you can use to wrap your Promise into a stream.
```
mbarto:
```suggestion
Axios is a library that uses es6 Promises to do ajax calls. Luckily RxJs allow to use Promises instead of streams in most of the cases. In the other cases, there are specific operators called `fromPromise` or `defer` that you can use to wrap your Promise into a stream.
```

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+
+> NOTE: is perfectly normal to consider the concept of Promise as a special case of a stream, that emit one value, then closes.
```
mbarto:
```suggestion
> NOTE: it is perfectly normal to consider the concept of Promise as a special case of a stream, that emit one value, then closes.
```

docs/developer-guide/writing-epics.md
```diff
@@ -0,0 +1,221 @@
...
+
+
+So,everytime you have to do an ajax call, you will need to use axios:
+
+Example with `defer`:
```
mbarto: Why we have two choices and what is the difference?
Shouldn't we stick to one if they are equivalent?

offtherailz: They are not the same. Basically defer is more generic and can be used in more use cases (e.g. error handling operators like `retry` and `retryWhen` work well with `defer`.
Most of the times it is converted implicitly (e.g. when retrurned in switchMap, mergeMap, defer). defer is only the simplest way to do it. 
After a small investigation  i noticed that `fromPromise` has been removed in RxJs 6, so we can only cite defer.

docs/developer-guide/writing-epics.md
offtherailz: Yes, but I'm afraid the original documentation will disappear soon, so I'd like to provide at least some basic information about how to read the actual code to posterity :).

docs/developer-guide/writing-epics.md
offtherailz: I changed a little bit the purpose of this paragraph and so this phrase, check if does it seem better

docs/developer-guide/writing-epics.md
mbarto: I hope the documentation of older versions will remain, maybe it will be more difficult to find them. If they are going to disappear, blame to the RxJS project. In any case, I am looking for a compromise: can we move this libraries-related docs to different page and link it from here?

docs/developer-guide/writing-epics.md
mbarto: good

docs/developer-guide/writing-epics.md
offtherailz: Hehe, I use to skip details I'm not interested, but if you think it's confusing, I'll remove it.

PR #3403, opened 2018-12-10
Fix range operator "><" for ogcDateField

serkejal:
There is a simple typo during assignement of const endIso.

## Description
Fix the use of daterange picker while filtering on datetime column

 - [X] Bugfix

- [X] No

### DIFF
--- a/web/client/utils/FilterUtils.jsx
+++ b/web/client/utils/FilterUtils.jsx
@@ -78,7 +78,7 @@ const ogcDateField = (attribute, operator, value, nsplaceholder) => {
     if (operator === "><") {
         if (value.startDate && value.endDate) {
             const startIso = value.startDate.toISOString ? value.startDate.toISOString() : value.startDate; // for Compatibility reasons. We should use ISO string to store data.
-            const endIso = value.startDate.toISOString ? value.startDate.toISOString() : value.startDate; // for Compatibility reasons. We should use ISO string to store data.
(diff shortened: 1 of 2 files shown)


offtherailz:
Hi @serkejal , Thank you for your contribution. 
I added a unit test tp check the bug you notified. This new test is failing on master, and should pass because of your fix. If it passes, we can merge your pull request. 
( I prefer to add unit tests for bug fixes, to prevent future regressions )

offtherailz:
**Note for testers**: to test this fix you need :
 - to have a layer with date-time attributes
 - Add it to the map
 - Try to filter with start-end date (filterpanel in featuregrid)

Expected Result: 
 - The data is filtered correctly

Current Result: 
 - The end date of the filter is not filtered correctly (the filter sent to the server uses start Date also as end date of the filter)

serkejal:
Hi, thanks for merging.
I've found this bug many time inside the source code. Now i'm extending some functionalities and so i fix this in my override files but i have a simple method to looking for the issue:

just use function of an IDE like "search in path" and type this

`const endIso = value.startDate.toISOString` 

:D

offtherailz:
hi @serkejal , thank you very much. I opened an issue for this. #3763

One question Looking in the code I found **only 1 time**, this
https://github.com/geosolutions-it/MapStore2/blob/f847346d60bfde009ddb71ab287379e2838b724f/web/client/utils/FilterUtils.jsx#L729

Do you have found other points where there is a problem like this? 

If so, please notify it commenting the issue #3763 . Thank you very much

PR #3419, opened 2018-12-19
Geonode Integration + Styler + Timeline 

### LINKED ISSUE #1372: Implement TimeSlider functionality for WMS
## Description

As a user I would like to browse time information on a layer that has a TIME dimension.

The major target of the whole issue is to avoid to call getCapabilities requests that, by specific, allows list of values, continuous interval or time/resolution. 
This doesn't fit well if we have a lot of irregular point in time, because a list may become very big. We want to use WMTS multidimensional extension to retrieve time dimension data. 

**note**: accordingly to the standard continuous interval means that we have date value formatted as `min/max/0` (mapstore should also understand `min/max` without resolution even if it is not required by specification ). This accordingly to [ISO/DIS 19128](http://www.demis.nl/wp-content/uploads/2017/03/wms_1_3_0.pdf) (page 46) 

> A resolution value of zero (as in min/max/0) means that the data are effectively at infinitely-fine resolution for
the purposes of making requests on the server. For instance, an instrument which continuously monitors
randomly-occurring data may have no explicitly defined temporal resolution.

### Acceptance criteria
- [ ] Timeslider should allow the user to have an overview of where data is present in time dimension
- [ ] User can zoom time slider and select a time and see data present in that moment
- [ ] User should be able to animate data (playback)
- [ ] User is able to select a range of time instead of a single instant
- [ ] Where data granules are too many, the timeslider, if possibile, should be enough smart to show aggregate data (strikes) and, when the user zoom, display single instant
- [ ] User should be able to see an histogram of how data is disseminated across time

After a first phase of investigation and creation of a prototype, we can split this user story into smaller features to reach the target of this issue incrementally.

## Tech details
We have several points where to work on: 
 
### General consideration
To avoid to send getCapabilities request we need to know as soon as possible if the layers we add to the map has some time dimension. This information will be stored in the layer to retrieve data only for the layer with this dimension, when the user open a map that contains them. 

### Initialization
One of the major problems is the initialization of the layers. At the layer configuration creation, we need to understand if the layer has the time dimension. 

#### WMS Catalogs
If the layer comes from a WMS catalog (so we already did a getCapabilities), we already have all the information about time data, so we will need at least to refresh these information if we want to have updated data (on map load) but this is it. 
In this case the map config actually contains the following: 
```
dimensions: [{
   "values": [ /* a list, interval or time-resolution*/],
    "name":"time",
    "units":"ISO8601"
}
}]
```
For the moment we can replace this list of values with a "source" entry for the "dimensions" record to refresh the dimensions on map load, looking for layer capability or, as a fallback, general capability.

```
dimensions: [{
    "source": {
         type: "wms",
         url: "..." // base url for getCapabilities
    },
    "name":"time",
    "units":"ISO8601"
}
}]
```

#### WMTS/CSW Catalogs
For other catalogs we could ask, on layer add event, we may ask for WMTS multidim extension to get dimension info. 
If this works, we can configure the layer this way: 

```
dimensions: [{
    "source": {
         type: "multidim-extension",
         url: "..." // base url where to send DescribeDomain/getHistrogram...
    },
    "name":"time",
    "units":"ISO8601"
}
}]
```

We should not go forward to getCapabilities fallback if multidim extension is not present, or we may have the same issue found MapStore1/geoExplorer. 

Anyway we can post-pone this as a future enhancement, but for the moment we should not support it.

#### Static configuration
In this case the map configuration should contain the `dimensions` array in layer . Following the same rules as before. We can support also list of values in layer configuration, like: 
```
dimensions: [{
   "values": [ /* a list, interval or time-resolution*/],
    "name":"time",
    "units":"ISO8601"
}
}]
```

### Time Data visualization/Retrival
The client need to know time dimension data for the following reasons: 
 a. Display information in timeline about where the data is present
 b. Snap to the selected time points (if user select a single time)
 c. Know time frames for animations

We have 2 cases now: 
1. Perfect knowledge of position of each value of time dimension (List of values/continuous interval/time resolution )
2. Service to retrieve time dimension (WM(T)S multidim extension)

In the case 1 the we can calculate every time point. This means that point b and c are solved. We can even display timeline, if we take care to show time/inetrval single points only when they are not too many.

The case 2 is the most complicated. We can follow this procedure: 
2.1. Ask to DescribeDomains the `time` dimension without any limit, except `ExpandLimit` set to a **proper value** (i.e. 100, or whatever the client is able to manage) that we will call **client limit**.
2.2. If the server provides the list of values, we are in case 1 (perfect knowledge)
2.3 If not the application should ask for histogram of the visible range and show the proper strikes where data is present
2.4 On every timeline zoom/pan the application the histogram request is updated, until the sum of all data in the histogram response for the current interval is not less then **client limit**. In this case a DescribeDomains can be performed to get single data. 

### Limit of this solution
 - Snapping is not allowed when data is aggregated using histogram.
 - If client limit > server ExpandLimit, then you will not be able to see points until the number of point in the view is less of the server side limit. 

### Animation
To get animation steps we need a paginated back-end service to provides single time domain elements (animation steps).
As an alterative, the application should use getHistogram + DescribeDomain recursively to get domain values, than continue using this method during the animation.

### LINKED ISSUE #3224: Style Editor Plugin (GeoCSS)
### Description
Create a new plugin to edit styles in MapStore.

features:
- [x] selection of an available style
- [x] editing of an available style in Textarea (Codemirror)
- [x] handle GeoCSS mode with hints and grammar highligth (Codemirror)
- [x] inline widgets (eg: color picker)
- [x] create new styles from a list of templates
- [x] hendle SLD style in a basic Textarea (Codemirror, no hint, standard xml mode)
- [x] live preview on editing/creation phases

workflows:
- create:
create

- selection:
selection

- editing:
editing

### LINKED ISSUE #3237: MapStore2 api and router
### Description
MapStore2 api shouldn't use the StandardRouter
 

*Browser Affected*
- [ ] Internet Explorer
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

*Steps to reproduce*
When the route location isn't just / the api doesn't render anything.
*Expected Result*

The api has to  ignore the route

*Current Result*

- Describe here the current behavior

### LINKED ISSUE #3289: Other - OpenLayers Glitch

offtherailz:
## Description
This pull request ports the functionalities of Geonode integration in the master branch: 
## Issues
 - Fix #3237 - Remove standard router from API
-  Fix #1372 - Implementation of timeline
 - Fix #3224 - Style editor plugin

Minor fixes: 
 - Fix #3289 - Openlayers rendering glitch

Note:
Due to avoid 1 big commit for separated functionalities, maybe we should merge this PR without squashing.


allyoucanmap:
Note:
Style editor needs geoserver rest API for styles endpoint in order to work.
If rest endpoint is missing and the user credentials are incorrect It will still works as list of available styles but a default basic auth form will popup from the browser (even for GET request).

offtherailz:
Yes, @allyoucanmap  we have to take into account these cases: 
 - Non GeoServer ogc services, that doesn't have the rest interface
 - Default setup of rest interface in geoserver requires Basic Authentication to work. So if you try to do any kind of request to the `/rest/*` path, by default you have a basic authentication prompt by the browser.

We should find a way to test if GeoServer is configured correctly without prompting basic auth. 
Otherwise we may disable the styler for demo or we may activate it with a button, after asking the user if he is sure every thing is configured correctly. Of course, this annoying explicit activation should be enabled/disabled by configuration, to avoid these steps in a well known installation where everything is already configured fine.

PR #3438, opened 2019-01-09
Fix #3405. Updated react-data-grid to latest version

### LINKED ISSUE #3405: Feature grid do not work with Chrome > 70
### Description
Things that do not work
 - Horizontal scroll
 - When scroll sometimes the whole grid becomes white
 - When select from autocomplete, the selection is not applied. 

*Browser Affected*
- [ ] Internet Explorer
- [x] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

-  70.0.3538.67

*Steps to reproduce*

- Open the feature grid for a layer that have enough auttributes to scroll horizontally

*Expected Result*

- The feature grid scroll horizontally
- Doing login and editing works 

*Current Result*
- Horizontal scroll do not work
- When resizing the feature grid, sometimes the grid becames all white.
- Editing autocomplete do not work
- Editing single row sometimes fails

Probably we have to update to this fixes:
https://github.com/adazzle/react-data-grid/pull/1346

Unfortunately the fix is applied to a new version that is not compatible with current mapstore. Trying to apply the fix to the current version (in node_modules folder) makes the scroll bar to appear. 

So a possible strategy is to: 
 - Test and verify that feature grid works with this fix,in all browsers.
 - Fork react-data-grid on github
 - Checkout to the current release commit 
( unfortunately there is not a tag for this version, use the commit 03f0eb7a0587393b49fcd6e72721dd75fdfb086d that should be the 2.0.59 version)
 - Apply the fix and publish on npm ( something like `@geosolutions-it/react-data-grid`)
 - Change package.json package to our fixed version. 
 - Propose the fix (not sure that version is maintained anymore)

An (better, but harder) alternative is to update to the latest version of the library (also `react-data-grid-addons` have to be updated).

offtherailz:
## Description
The issue for horizontal scrolling has been fixed in the latest version (see issue). 
I had also to port some extension to the latest version: 
 - The new version requires `immutable` dependency (it doesn't work without)
 - automatic scroll up on sort/filter: It was a workaround to avoid empty rows when virtual scroll is active. Now the `Grid` object has been changed and the old overrides that implemented this didn't work anymore. Implemented instead a `forceScrollTop` enhancer + `scrollToTopCounter` property to use to force scroll. 
 - Custom `RowRenderer` have to use `renderBaseRow` to preserve `ref` function and so support frozen (ex - blocked) columns (tipically selection column and "zoom to feature" button in the feature grid). Otherwise the frozen columns didn't worked anymire
 - Updated tests

## Issues
 - Fix #3405

 - [x] Bugfix

Grid can not be scrolled horizontally

**What is the new behavior?**

 - [x] No

**Other information**:
I suggest to re-do all the integration tests on the feature grid.


web/client/components/data/grid/forceScrollTop.js
```diff
@@ -0,0 +1,50 @@
...
+        mapPropsStream( props$ => {
+            const { handler: scrollToTop, stream: scrollToTop$} = createEventHandler();
+            return props$.combineLatest(
+                // add an incremental `scrollToTopCounter` variable each time scrollToTop handler is called
+                scrollToTop$
+                    .throttle((time = 0) => Rx.Observable.timer(time))
+                    .scan((acc) => acc + 1, 0)
+                    .map(scrollToTopCounter => ({scrollToTopCounter}))
+                    .startWith({}),
+                (props, newProps) => ({
```
mbarto: I only see one Observable, why props and newProps?

package.json
```diff
@@ -107,6 +107,7 @@
     "fs-extra": "3.0.1",
     "history": "4.6.1",
     "html2canvas": "0.5.0-beta4",
+    "immutable": "4.0.0-rc.12",
```
mbarto: needed by the new react-grid?

offtherailz: I think is because of this issue, closed but still very popular:
https://github.com/adazzle/react-data-grid/issues/80

web/client/components/data/grid/forceScrollTop.js
offtherailz: They are 2 observable merged into one. Maybe newProps is misleasing. Should i call it forceScrollProps?

package.json
mbarto: Ok, got it

web/client/components/data/grid/forceScrollTop.js
mbarto: Yes, maybe, should make it more clear, thanks

PR #3449, opened 2019-01-15
Fix #3443. Various improvements to timeline UI

### LINKED ISSUE #3443: Timeline - More feedbacks on user actions
### Description
1 Many times the tool is not so responsive, especially when the range or the tick element are moved.

timeline1

 The range element should respond to the mouse movement synchronously or the timeline should be blocked showing a loading gif to make the user aware that something happen.
 
2. Some time the tick element freeze without snapping:

timeline2

Generally speaking there should be instant feedback to user actions.

### LINKED ISSUE #3451: Minor issues with date selector
### Description
 - Cursor icon in time selector is now clickable only if the timeline is expanded (otherwise it doesn't make sense to move the view to the cursor, without view)
 - Editing the current time selector do not allow (and so do not trigger errors) when you try to set a date out of the max JavaScript date range ( from Apr 20 -271821 to 13 Sep 275760)

*Browser Affected*

- any

*Steps to reproduce*

- Open a map with time dimension to display the timeline
- 1. Try to click on the cursor's icon when the timeline is still collapsed
- 2. Try to write in the year a number bigger than `275761`

*Expected Result*

- Nothing should happen

*Current Result*

- Errors in console, for both case 1 and 2

offtherailz:
## Description
This PR improve timeline with some minor changes: 
- Now during snap time the timeline has a loading mask (loading mask appears also when animation is initializing, to notify the user that something is happening) As required by @tdipisa i changed the mask background color to white. 
- Fixed issues  that sometime didn't snap 
- Other minor fixes notified by @gioscarda (#3451) : 
   - Cursor icon in time selector is now clickable only if the timeline is expanded (otherwise it doesn't make sense to move the view to the cursor, without view)
   - Editing the current time selector do not allow (and so do not trigger errors) when you try to set a date out of the max JavaScript date range ( from Apr 20 -271821 to 13 Sep 275760)

**Important Note**: drag-drop range issue notified in #3443 can not be fixed without customizing visjs-timeline library or changing the logic we manage ranges. I suggest avoid this in this phase and postpone this improvement in a second phase (opened issue #3463 for this purpose).

## Issues
 - Fix #3443
 - Fix  #3451

 - [x] Bugfix
 
- See  #3443
- Inline time selector triggered errors when : 
    - Click on cursor icon with the timeline collapsed
    - Change the date out of max JavaScript date range ( **from** Apr 20 -271821 **to** 13 Sep 275760)

**What is the new behavior?**
- Now during snap time the timeline has a loading mask
- Snap always works
- Inline time selector errors above are fixed

 - [x] No

### DIFF
--- a/web/client/components/misc/enhancers/withMask.js
+++ b/web/client/components/misc/enhancers/withMask.js
@@ -12,10 +12,10 @@ const {branch, nest} = require('recompose');
  * @param {function} showMask gets props as argument and returns true if the enhancer should be applied
  * @param {*} maskContent the content of the mask
  */
-const maskEnhancer = (showMask, maskContent, { maskContainerStyle, maskStyle, className }) => (A) => nest(
+const maskEnhancer = (showMask, maskContent, { maskContainerStyle, maskStyle, className, white }) => (A) => nest(
     (props) => (<div className={`ms2-mask-container ${className || ''} ${!showMask(props) && 'ms2-mask-empty' || ''}`} style={maskContainerStyle} >
         {props.children}
-        {showMask(props) ? <div className="ms2-mask" style={maskStyle} >
+        {showMask(props) ? <div className={"ms2-mask" + (white ? " white-mask" : "")} style={maskStyle} >
             {maskContent(props)}
         </div> : null}
     </div>),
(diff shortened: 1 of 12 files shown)


offtherailz:
@gioscarda If you open an issue for the 2 minor problems you found, we may link them here

PR #3505, opened 2019-02-01
Fixes to align WMS gridsets

offtherailz:
## Description
This pull request fixes some issues that didn't allow GWC caching HIT on WMS using projections different from EPSG:4326 and EPSG:3857. (Issue found on EPSG:3003) 

## Issues
 - Fix #3410 

 - [x] Bugfix
 

Tiles in some srs (like EPSG:3003) do not hit cache. 

**What is the new behavior?**
Tiles in srs EPSG:3003 hit cache. 

 - [x] No

**Other information**:
There were 2 problems: 
1. Origin was set to minX,minY, instead to use minX,maxY. This problem was hidden using EPSG:3857 and EPSG:4326. These 2 projection have `minX= k * maxX, k Integer`, and so setting the this wrong origin make it aligned anyway to the grid
2. The second problem was that the resolutions was calculated with `maxResolution=max(extentWidth, extentHeight)`. Instead, to have the correct resolution aligned to the grid, you must use extent width. This again didn't caused any issue in EPSG:4326 and EPSG:3857 because in these cases `extentWidth >= extentHeight`

### DIFF
--- a/web/client/components/map/openlayers/Map.jsx
+++ b/web/client/components/map/openlayers/Map.jsx
@@ -63,10 +63,12 @@ class OpenlayersMap extends React.Component {
     };
 
     componentDidMount() {
-        var center = CoordinatesUtils.reproject([this.props.center.x, this.props.center.y], 'EPSG:4326', this.props.projection);
         this.props.projectionDefs.forEach(p => {
             projUtils.addProjections(ol, p.code, p.extent, p.worldExtent);
         });
+        // It may be a good idea to check if CoordinateUtils also registered the projectionDefs
+        // normally it happens ad application level.
+        let center = CoordinatesUtils.reproject([this.props.center.x, this.props.center.y], 'EPSG:4326', this.props.projection);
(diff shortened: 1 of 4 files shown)


web/client/components/map/openlayers/Map.jsx
```diff
@@ -287,10 +289,10 @@ class OpenlayersMap extends React.Component {
         const projection = this.map.getView().getProjection();
         const extent = projection.getExtent();
         const size = !extent ?
-            // use an extent that can fit the whole world if need be
+            // use the max width
             360 * ol.proj.METERS_PER_UNIT[ol.proj.Units.DEGREES] /
                 ol.proj.METERS_PER_UNIT[projection.getUnits()] :
-            Math.max(ol.extent.getWidth(extent), ol.extent.getHeight(extent));
+            ol.extent.getWidth(extent);
```
mbarto: We change the current behaviour. Is this needed?

PR #3566, opened 2019-02-22
Fix #3562. Fixed API map load action sequence

### LINKED ISSUE #3562: Problem with embedded map using APIs
### Description
The embedded code for map sharing through API doesn't work with Firefox browser.
Refer to [this test](https://app.hiptest.com/projects/48097/test-runs/237259/folder-snapshots/1982098/scenario-snapshots/7653718/test-snapshots/9887161) .
### In case of Bug

*Browser Affected*
- [x] Internet Explorer
- [x] Chrome
- [x] Firefox
- [ ] Safari

*Browser Version Affected*

- All

*Steps to reproduce*

- Open a map in MapStore
- Open the SHARE panel from the Burger Menù
- Got o the EMBED tab
- Copy the text inside the "Using APIs" box
- Paste it into a new file and save it as HTML
- Open the html file in a Firefox browser 

*Expected Result*
The map is visible in an embedded window

*Current Result*
The map is not visible, a blank page is displayed

- Describe here the current behavior
When you open the html file in a Firefox browser a blank page is displayed

offtherailz:
## Description
The issue was caused by an INIT_MAP action triggered before map load process. This caused the map to be cleared from all the layers and a blank mask to be shown on it.

To solve the problem I had to put the initMap in the initialActions of Map API.

To test it I had to clean up some debug and test parts to avoid errors during unit tests ( e.g. undefined __DEV_TOOLS__ variable), and allow to add epics from API (not official feature, so I didn't documented it) to check results.

I also added some tests for JsAPI methods. 

Another error happens using JSAPI as it is presented in the share tool, about not found version.txt. 

It doesn't seems to affect the final result. Anyway if we need it we may need to include the `versionURL` it in the API share template.

## Issues
 - Fix #3562

 - [x] Bugfix
 

JsAPI do not load map sometimes

**What is the new behavior?**
JsAPI loads the map

 - [x] No


web/client/jsapi/MapStore2.js
```diff
@@ -93,7 +94,7 @@ let stateChangeListeners = [];
 const getInitialActions = (options) => {
     if (!options.initialState || !options.initialState.defaultState.map) {
         if (options.configUrl) {
-            return [loadMapConfig.bind(null, options.configUrl || defaultConfig, options.mapId)];
+            return [initMap.bind(null), loadMapConfig.bind(null, options.configUrl || defaultConfig, options.mapId)];
```
allyoucanmap: could it be `initMap` instead of `initMap.bind(null)`

offtherailz:
```suggestion
            return [initMap, loadMapConfig.bind(null, options.configUrl || defaultConfig, options.mapId)];
```

Issue #3109, opened 2018-07-24
Improvements on floating legend plugin


allyoucanmap:
### Description
Improvements:
- Resizing of legeng should start only when it recieve all lagend images to fit correctly the content
- Layers should not disappear from Floating Legend but should display their status ( eg. layer with error should be displayed similar to TOC with exclamation mark icon)

in general we should avoid unexpected visual changes (eg.: layer disappearing due to change of status)

tdipisa:
@allyoucanmap Can we also use the same dimensions of the legend in TOC

tdipisa:
The floating lagend plugin is no longer maintained, I'm going to close this issue for now: we are reviewing the entire MS's layout (#4254).

PR #3577, opened 2019-02-26
Remove FloatingLegend for the release

offtherailz:
## Description
The plugin should not be included in the release


simboss:
why we are removing it?

tdipisa:
> why we are removing it?

Because we have agreed to disable it by default until the issue #3109 will be solved (this issue is currently planned for the next milestone of minor release 2019.01.01).

PR #3616, opened 2019-03-13
Externalized in epic getFeatureInfo retrieval logic

offtherailz:
## Description
This pull request moves the request build and trigger logic from the FeatureInfoContainer to the `identify` epic. This
 - simplify the logic following SRP
 - Allow to intercept and modify the request sequence in an efficient and simple way. 

This refactor is required to allow clicked feature highlight (#3617)

No changes

**What is the new behavior?**
Same as before. The only difference is about multiSelection (disabled by default: 
 - The `multiSelection` flag is has been moved temporary from the Component configuration onto the action (is not configurable, we need a little more work to make configurable from the app state, but I didn't want to complicate things for `c125_annotations` merge too much, for a feature that is not used neither documented).

 - [x] No


web/client/plugins/Identify.jsx
```diff
@@ -94,28 +87,18 @@ const identifyDefaultProps = defaultProps({
...
-        "propertyName"
-    ],
     panelClassName: "modal-dialog info-panel modal-content",
     headerClassName: "modal-header",
     bodyClassName: "modal-body info-wrap",
     dock: true,
     headerGlyph: "",
     closeGlyph: "1-close",
     className: "square-button",
-    allowMultiselection: false,
```
mbarto: I would not remove this without an explicit  reason.

offtherailz: I partially maintained it (that it means that is not configurable anymore, but the code base for this flag is still maintained and tested). Ideally, after this changes, it should go in the application state. 
If you want I can finalize the support for this flag in the mapInfo state.

offtherailz: See this https://github.com/geosolutions-it/MapStore2/pull/3616/files#diff-e7312b552c083e1cbff765fe521d4d0bR82

PR #3635, opened 2019-03-21
Fix #3617. Add feature zoomTo and highlight to Identify

### LINKED ISSUE #3617: Highlight GetFeatureInfo results on map
### Issues
 - depends on #3616

### Acceptance criteria
- A toggle button should be present in the top toolbar of the featureinfo button, called "highlight features"
- The button visibility should be configurable
- When the highlight features button is activated, the clicked geometry should be highlighted on map, shown together with the marker, on feature info click.
- The geometry should be highlighted only if the following conditions are satisfied (Otherwise only the marker will be on map) : 
    - The server supports GFI JSON format
    -  The layer is a vector layer and so the GFI can return also the geometry
- If more then one layer has feature info response, only the geometries shown also in the GFI panel will be visible (so only one layer at the same time). 

Preview ( [mockup](https://demo.geo-solutions.it/mockups/mapstore2/annotation_styles/#/viewer/openlayers/new) ): 
image

offtherailz:
## Description
This changes add the button highlight feature the identify toolbar (on top of identify panel). When highlight is available, you can zoom to the selected feature by clicking on the button. 

Limitations: 
 - This functionality is limited to openlayers, with remote vector data. 
 - Leaflet / Cesium map will not highlight show the button. Anyway zoom to feature may be available.
 - Local features (imported shape files) do not have this functionality yet. 
## Issues
 - Fix #3617

 - [x] Feature

this feature is not present

**What is the new behavior?**
You can highlight the features 

 - [x] No


web/client/actions/mapInfo.js
```diff
@@ -193,6 +195,20 @@ function featureInfoClick(point, layer) {
     };
 }
 
+function highlightFeature(enabled) {
```
mbarto: This enables / disables the highlightFeature functionality, it doesn't highlight a feature, I would change the name of the action and action creator.

web/client/actions/mapInfo.js
```diff
@@ -193,6 +195,20 @@ function featureInfoClick(point, layer) {
...
+    return {
+        type: HIGHLIGHT_FEATURE,
+        enabled
+    };
+}
+
+function changeIndex(index) {
```
mbarto: Changes the index of what?

web/client/actions/mapInfo.js
offtherailz: ok

web/client/actions/mapInfo.js
offtherailz: Index of the current response shown in the feature info panel in the list of valid responses (sub-group of responses). 
The property in the state is mapInfo.index. Should I document the action and the reducer's parts?

web/client/actions/mapInfo.js
offtherailz: any suggestion?

Issue #2971, opened 2018-06-01
Fix tests randomly failing on the build


offtherailz:
### Description
Some tests sometimes fail becouse of slowdown or some wrong race condition:
Mainly in 2 or 3 different failure sequences: 
-  [Sample 1](https://travis-ci.org/geosolutions-it/MapStore2/builds/380775634)
-  [Sample 2](https://travis-ci.org/geosolutions-it/MapStore2/builds/380747290)
-  [Sample 3](https://travis-ci.org/geosolutions-it/MapStore2/builds/380773425)
- Another error found is : `StandardApp creates a default app with onInit FAILED`
Sample 1 and 2 are similar, so they may have the same reason. 
### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [x] Firefox

*Browser Version Affected*

- Latest (on travis)

*Steps to reproduce*

- Run test on travis, sometimes. Need to investigate the reasons. 

*Expected Result*

- Build always has success

*Current Result*

- Random fails of the travis build

MV88:
Here a list of travis build that have failed:

**Measurement leaflet**
it was a problem with measurement leaflet before each hook and causes the build to stop
- https://travis-ci.org/geosolutions-it/MapStore2/builds/553656098
not sure we are using the setTimeout [here](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/map/leaflet/__tests__/MeasurementSupport-test.jsx#L41) since by removing them all the tests are successfull

**StandarApp**
these builds run all the tests but fail because of this test or something that happen "before" it
```
Firefox 67.0.0 (Linux 0.0.0) StandardApp creates a default app with onInit FAILED
	Error: spy was not called (build/tests-travis.webpack.js line 5688 > eval:29)
```
- https://travis-ci.org/geosolutions-it/MapStore2/builds/553357344#L2822
- https://travis-ci.org/geosolutions-it/MapStore2/builds/553318784#L2834
- https://travis-ci.org/geosolutions-it/MapStore2/builds/552651458#L2835
- https://travis-ci.org/geosolutions-it/MapStore2/builds/551310545#L2829

### Other useful information (optional):
It can be related to tests using the setTimeout not calling the (done)

PR #3649, opened 2019-03-26
Connected to #2971. Use mock lib to speed up wfstable test

offtherailz:
## Description
Use axios mock to speed-up the data load and avoid random test failures in wfsTable fetch data enhancer.

During tests I noticed that repeating these tests, sometimes they take up to 1 second finish using Firefox. After this fix, I didn't notice this slow down anymore. 
This, in the Travis context may be the cause of the failure due 2 secs exceed, because of the lower performance. This is my (not confirmed) hypothesis about the cause of it.
This is why I replaced old test with a mock one. The mock should reply faster (because of the pre-loaded JSON) so the test should not fail anymore. 

Anyway the random nature of the failures, that can not replicated locally, don't ensure the issue to be solved, so the error may not be solved.  

## Issues
 - Connected to #2971 - Should fix one of the problems in this issue

 - [x] CI related changes

This Unit test sometimes fails on travis

**What is the new behavior?**
This tests should not fail anymore

 - [x] No

### DIFF
--- a/web/client/components/widgets/enhancers/wfsTable/__tests__/wfsTable-test.js
+++ b/web/client/components/widgets/enhancers/wfsTable/__tests__/wfsTable-test.js
@@ -8,26 +8,39 @@
 
 const React = require('react');
 const ReactDOM = require('react-dom');
-const { createSink, setObservableConfig} = require('recompose');
+const { createSink, setObservableConfig } = require('recompose');
 const expect = require('expect');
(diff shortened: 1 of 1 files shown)


tdipisa:
@offtherailz the issue #2971 is assigned to the milestone 2019.03.00, if this PR fixes one of the reported problems there, split the issue and change the assignment using the new one.

PR #3651, opened 2019-03-27
Fix #3585 date-time and time support

### LINKED ISSUE #3585: QueryPanel DateTime filter issues
### Description
 - [x] DateTime attribute time is not supported, only date. (not sure if time is supported) 3PT. We need to add datetime support
 - [x] If you edit one date time Filter and then change the operator (e.g. > to <), without changing the date, the filter will not be applied. 1PT
 - [x] The hour of the filter in the final timestamp is set randomly (current hour when create the first attribute filter). It should be 00:00:00 (or use date-only format). 1PT

*Browser Affected*
any

*Steps to reproduce*

- Add a new Layer with dates and date-time attributes
- Open the query panel

*Expected Result*

- I can create filters for all attributes
- filters should work

*Current Result*

- Datetime (timestamp) attributes are not available
- When I change the operator after changing the date, the filter is not applied: 
   - Add an attribute filter on a date field
   - Select operator > 
   - Select a date using the date picker (e.g . today)
   - Change the opearator to <
   - Click on search ( thie filter is not applied, the button is not green, if you re-open the query panel you can not see the attribute fiter you did)
- Dates are relative to my date time instead of absolute. 
   - open dev tools
   - perform a search with a date attribute filter
   - check network tab in dev tool
   - select the request o and check the effective xml: I see some random hours (It should be 00:00:00).
screenshot

Notes for developers: 
Only for dates (not dateTime), we should check if the filter is correctly applied as expected with > or >= (we are using timestamp for simple dates: 
 -Example:  2018:01:01 (original attribute) > 2018-01-01:00:00:00 (timestamp filter) is misleading.

MV88:
## Description
This pr add the support for date-time and time attributes.
Now you can use them in the Attribute Filter field.

- [x] When changing the operator it will reset the value
- [x] It starts with time 00:00:00 when selecting a date
- [x] updated formats for date and hours
- [x] requests are done in date UTC (the same you see in the input field)

## Issues
 - Fix #3585 

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

date-time and time are not supported
We are not using utc date in the date picker and in the requests they are shifted to UTC 
Hours are set to actual date if date only is used

**What is the new behavior?**
see description

 - [ ] Yes
 - [x] No

**Other information**:
Please note that the date formats are a bit changed since it was not displaying correctly the Year,
and for the hours it was using month instead of minutes


web/client/components/misc/enhancers/utcDateWrapper.js
```diff
@@ -0,0 +1,83 @@
...
+} = require('../../../utils/TimeUtils');
+
+/**
+ * Converts local date as it was UTC.
+ * Useful wrapper for calendar tools that internally use date object localized,
+ * but you want to use the date as it was UTC.
+ * @param {object} options to pass to the enhancer
+ * @param {string} options.dateType type choosed from ("date", "time", "date-time")
+ * @param {string} options.dateField date to use
+ * @param {string} options.setDateField function to use to update the date
```
offtherailz: call them props, not fields. 
`dateProp`, `setDateProp`

web/client/components/misc/enhancers/utcDateWrapper.js
```diff
@@ -0,0 +1,83 @@
...
+ * useUTCOffset is used to translate the data as it is defined in UTC.
+ * for example when using the useUTCOffset=true 20 March 2019 will correspond to 2019-03-20T00:00:00.000Z
+ * instead without it will be used as 2019-03-19T23:00:00.000Z (in cql filter), but in the UI it will be used the
+ * actual timezone
+ */
+module.exports = ({dateType = "type", dateField = 'date', setDateField = 'onSetDate'} = {}) => compose(
```
offtherailz: "type" is not one of the allowed values. use `datetime`

web/client/utils/FeatureTypeUtils.js
```diff
@@ -25,7 +25,10 @@ const types = {
     // 'xsd:gMonthDay': 'date',
     // 'xsd:gYear': 'date',
     // 'xsd:gYearMonth': 'date',
-    // 'xsd:time': 'date',
+    // date-time
+    'xsd:date-time': 'date-time',
```
offtherailz: [xsd:dateTime](http://books.xmlschemata.org/relaxng/ch19-77049.html) is the correct format. Please replace 'date-time' everywhere. Make you sure you receive this also in the DescribeFeatureType (check also the camel case).

How did it work?

web/client/utils/__tests__/LocaleUtils-test.js
```diff
@@ -80,4 +80,17 @@ describe('LocaleUtils', () => {
...
+    it('getDateFormat', () => {
+        expect(LocaleUtils.getDateFormat()).toBe("YYYY/MM/DD");
+        expect(LocaleUtils.getDateFormat("en-US")).toBe("MM/DD/YYYY");
+        expect(LocaleUtils.getDateFormat("it-IT")).toBe("DD/MM/YYYY");
+        expect(LocaleUtils.getDateFormat("nl-NL")).toBe("DD/MM/YYYY");
+        expect(LocaleUtils.getDateFormat("zh-ZH")).toBe("YYYY/MM/DD");
+        expect(LocaleUtils.getDateFormat("hr-HR")).toBe("DD/MM/YYYY");
+        expect(LocaleUtils.getDateFormat("pt-PT")).toBe("DD/MM/YYYY");
+    });
+    it('DATE_FORMATS', () => {
```
offtherailz: This test is not useful at all.

web/client/utils/__tests__/LocaleUtils-test.js
```diff
@@ -80,4 +80,17 @@ describe('LocaleUtils', () => {
         expect(LocaleUtils.getLocalizedProp('it-IT', localizedObjectProp)).toBe('titolo');
         expect(LocaleUtils.getLocalizedProp('fr-FR', localizedObjectProp)).toBe('title');
     });
+    it('getDateFormat', () => {
+        expect(LocaleUtils.getDateFormat()).toBe("YYYY/MM/DD");
```
offtherailz: I suppose this is a check for unknown locale. Please describe in a comment

web/client/utils/TimeUtils.js
```diff
@@ -150,7 +151,72 @@ const getStartEnd = (startTime, endTime) => {
...
+    let seconds = dateToParse.getUTCSeconds();
+    seconds = seconds < 10 ? "0" + seconds : seconds;
+    return `${hours}:${minutes}:${seconds}`;
+};
+
+/**
+ * @param {object} date to get only date parts (without time parts)
+ * @return {string} used to compose a UTC date
+*/
+const getUTCDateParts = (date) => {
```
offtherailz: Is singular: `getUTCDatePart`

web/client/utils/TimeUtils.js
```diff
@@ -150,7 +151,72 @@ const getStartEnd = (startTime, endTime) => {
...
+*/
+const getTimezoneOffset = (date) => {
+    return (date).getTimezoneOffset() * 60000;
+};
+
+/**
+ * @param {object} date to get only time parts (without date parts)
+ * @return {string} used to compose a UTC date
+*/
+const getUTCTimeParts = (date) => {
```
offtherailz: singular `getUTCTimePart`

web/client/utils/TimeUtils.js
```diff
@@ -150,7 +151,72 @@ const getStartEnd = (startTime, endTime) => {
...
+ * get time zone offset for a given date (september and march have different tzoffset)
+ * @param {object} date
+ * @return {number} the offset in milliseconds
+*/
+const getTimezoneOffset = (date) => {
+    return (date).getTimezoneOffset() * 60000;
+};
+
+/**
+ * @param {object} date to get only time parts (without date parts)
```
offtherailz:
```suggestion
 * @param {Date|string} date to parse
```

web/client/utils/TimeUtils.js
```diff
@@ -150,7 +151,72 @@ const getStartEnd = (startTime, endTime) => {
...
+ * @return {string} used to compose a UTC date
```
offtherailz:
```suggestion
 * @return {string} time part of the timeStamp
```

web/client/utils/TimeUtils.js
```diff
@@ -150,7 +151,72 @@ const getStartEnd = (startTime, endTime) => {
     };
 };
 
+/**
+ * get time zone offset for a given date (september and march have different tzoffset)
+ * @param {object} date
+ * @return {number} the offset in milliseconds
+*/
+const getTimezoneOffset = (date) => {
```
offtherailz: `getTimezoneOffsetMillis`, or can be confused with original one

web/client/utils/TimeUtils.js
```diff
@@ -150,7 +151,72 @@ const getStartEnd = (startTime, endTime) => {
...
+    hours = hours < 10 ? "0" + hours : hours;
+    let minutes = dateToParse.getUTCMinutes();
+    minutes = minutes < 10 ? "0" + minutes : minutes;
+    let seconds = dateToParse.getUTCSeconds();
+    seconds = seconds < 10 ? "0" + seconds : seconds;
+    return `${hours}:${minutes}:${seconds}`;
+};
+
+/**
+ * @param {object} date to get only date parts (without time parts)
```
offtherailz:
```suggestion
 * @param {Date|string} date to parse
```

web/client/utils/TimeUtils.js
```diff
@@ -150,7 +151,72 @@ const getStartEnd = (startTime, endTime) => {
...
+ * @return {string} used to compose a UTC date
```
offtherailz:
```suggestion
 * @return {string} date part of the TimeStamp
```

web/client/components/data/featuregrid/filterRenderers/__tests__/DateTimeFilter-test.jsx
```diff
@@ -30,7 +30,7 @@ describe('Test for DateTimeFilter component', () => {
         expect(el).toExist();
     });
     it('render with value', () => {
-        ReactDOM.render(<DateTimeFilter type="date" value={"2017-01-05Z"}/>, document.getElementById("container"));
+        ReactDOM.render(<DateTimeFilter type="date" value="2017-01-05T04:04:04Z"/>, document.getElementById("container"));
```
offtherailz: `DateTimeFilter` should be enhanced, or I'm afraid it will not work with dates or times only

web/client/components/data/query/DateField.jsx
```diff
@@ -44,39 +60,57 @@ class DateField extends React.Component {
...
         onUpdateExceptionField: () => {}
     };
-
     render() {
+        // these values are already parsed by the enhancer
+        const startdate = this.props.fieldValue && this.props.fieldValue.startDate || null;
+        const enddate = this.props.fieldValue && this.props.fieldValue.endDate || null;
+
+        // needed to initialize the time parts to 00:00:00
+        const defaultCurrentDate = moment().startOf("day").toDate();
```
offtherailz: Don't know why this is here (and in date filter) if is needed to make UTCDateTimePicker work, it should be inside it (enhancer, setting, for instance moment().startOf("day").toDate() as default value.

If has another reason, please explain better.

web/client/components/misc/enhancers/utcDateWrapper.js
MV88: dateType is where to check in the props the type prop ('time', 'date', 'date-time'. 
I would call it **dateTypeProp**

offtherailz: Ah, ok. Yes, fine. That's a better name

web/client/utils/__tests__/LocaleUtils-test.js
MV88: what if you accidentally erase the it-IT date format customization. it will use the default one.
this checks the presence of the customization for the some locales. i'll update the it description

web/client/utils/__tests__/LocaleUtils-test.js
MV88: for getting the default one

web/client/components/data/featuregrid/filterRenderers/__tests__/DateTimeFilter-test.jsx
MV88: you mean it should handle partial date or time?

web/client/utils/FeatureTypeUtils.js
MV88: date-time was the type returned from capabilities as you have also checked

web/client/components/data/query/DateField.jsx
MV88: This is needed to initialize the tiem part

Imagine the "date" case:
DateTimePicker component is initialized with a new Date based on that moment and it includes the time part.

then you select a date from the calendar but the time part will remain untouched.
with defaultCurrentDate it changes the time part of the initial value 

notice that it does that also for date-time case (we may want to not do it for that case)
i think we can put it in the enhancer enabled by a flag or by the type

web/client/components/data/featuregrid/filterRenderers/__tests__/DateTimeFilter-test.jsx
MV88: btw it is using the utcDateWrapper enhancer

offtherailz:
Tested with [this map](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10103).
It works :+1: 
Only the last comment is still pending @MV88

web/client/components/data/query/DateField.jsx
MV88: i think that the way we are using "defaultCurrentDate" is not needed in general,
but is needed only if you want to start with a 00:00 hh:ss which is not always the case

MV88: @offtherailz I'll update the enhnacer with the defaultCurrentDate = moment().startOf("day").toDate();  at least for "date" format

PR #3652, opened 2019-03-27
Fix #3585 date-time and time support (2019.01.xx)

### LINKED ISSUE #3585: QueryPanel DateTime filter issues

MV88:
## Description
port on stable branch of the pr #3651

### DIFF
--- a/web/client/components/data/featuregrid/filterRenderers/BaseDateTimeFilter.jsx
+++ b/web/client/components/data/featuregrid/filterRenderers/BaseDateTimeFilter.jsx
@@ -8,21 +8,21 @@
 
 const React = require('react');
 const PropTypes = require('prop-types');
-const LocaleUtils = require('../../../../utils/LocaleUtils');
-
 require('react-widgets/lib/less/react-widgets.less');
-
 const {DateTimePicker} = require('react-widgets');
+const moment = require('moment');
 
+const LocaleUtils = require('../../../../utils/LocaleUtils');
+const {getDateTimeFormat} = require('../../../../utils/TimeUtils');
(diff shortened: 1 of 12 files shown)


MV88:
Closing because outdated.
Going to create a new one

PR #3662, opened 2019-04-01
Fix #3617 minor bug fixes and improvements

### LINKED ISSUE #3617: Highlight GetFeatureInfo results on map

offtherailz:
## Description
This PR fixes some minor issues on highlight feature info feature functionality: 
 - [ ] Moved navigation buttons from tool bar on the right side, to prevent appearing/disappearing buttons to shift.
 - [ ] Allow configurability (using localConfig) of highlight color
-  [ ] Fixed german translations

## Issues
 - Fix #3617 

 - [x] Bugfix

- Toolbar makes hard to navigate features, when more than one layer is present
- Tooltip for german translation is inccorrect  "Stoppen Sie die Hervorhebung von Funktionen" 

**What is the new behavior?**
- Navigation and tools are separated. Navigation buttons, when present, are only disabled on first/last page (before they was not visible).
- Highlight color can be configured using localConfig.json (documented how to in mapinfo reducer). 
- Tooltip above replaced with : "Stoppen Sie die Markierung von Features".

 - [x] No

### DIFF
--- a/web/client/components/data/identify/IdentifyContainer.jsx
+++ b/web/client/components/data/identify/IdentifyContainer.jsx
@@ -79,7 +79,7 @@ module.exports = props => {
     }
     const Viewer = viewer;
     // TODO: put all the header (Toolbar, navigation, coordinate editor) outside the container
-    const buttons = getToolButtons({...props, lngCorrected, validResponses, latlng});
+    const toolButtons = getToolButtons({...props, lngCorrected, validResponses, latlng});
     const missingResponses = requests.length - responses.length;
     const revGeocodeDisplayName = reverseGeocodeData.error ? <Message msgId="identifyRevGeocodeError"/> : reverseGeocodeData.display_name;
     return (
@@ -113,7 +113,7 @@ module.exports = props => {
                             <Col key="tools" xs={12}>
                                 <Toolbar
                                     btnDefaultProps={{ bsStyle: 'primary', className: 'square-button-md' }}
-                                    buttons={buttons}/>
+                                    buttons={toolButtons}/>
                             </Col>
                         <div key="navigation" style={{
                                 position: "absolute",
--- a/web/client/components/data/identify/IdentifyContainer.jsx
+++ b/web/client/components/data/identify/IdentifyContainer.jsx
@@ -23,7 +23,8 @@ const Portal = require('../../misc/Portal');
  * @prop {dock} dock switch between Dockable Panel and Resizable Modal, default true (DockPanel)
  * @prop {function} viewer component that will be used as viewer of Identify
  * @prop {object} viewerOptions options to use with the viewer, eg { header: MyHeader, container: MyContainer }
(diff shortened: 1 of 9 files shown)


web/client/components/data/identify/IdentifyContainer.jsx
```diff
@@ -106,14 +109,22 @@ module.exports = props => {
                         onChangeFormat={onChangeFormat}
                     /> || null,
                     <GeocodeViewer latlng={latlng} revGeocodeDisplayName={revGeocodeDisplayName} {...props}/>,
-                    buttons.length > 0 ? (
```
MV88: you sure about this? if i remember correctly it was messing up the layout (try it with IE11)

web/client/components/data/identify/IdentifyContainer.jsx
```diff
@@ -76,10 +78,10 @@ module.exports = props => {
         lngCorrected = lngCorrected - 360 * Math.floor(lngCorrected / 360 + 0.5);
     }
     const Viewer = viewer;
-    const buttons = getButtons({...props, lngCorrected, validResponses, latlng});
+    // TODO: put all the header (Toolbar, navigation, coordinate editor) outside the container
+    const buttons = getToolButtons({...props, lngCorrected, validResponses, latlng});
```
MV88: let's call it toolsButtons or buttonsTools

web/client/selectors/mapInfo.js
```diff
@@ -6,7 +6,7 @@
 * LICENSE file in the root directory of this source tree.
 */
 
-const { get, isArray } = require('lodash');
+const { get, omit, isArray } = require('lodash');
```
MV88: they say that pluck is to be preferred to omit for performance reason.
(if you know precisely which props to pass. i mean)

web/client/components/data/identify/IdentifyContainer.jsx
offtherailz: :ok:

web/client/selectors/mapInfo.js
offtherailz: We should prefer readability over premature optimization

web/client/components/data/identify/IdentifyContainer.jsx
offtherailz: Checked, buttons array is never empty, so this check dosn't make sense

PR #3668, opened 2019-04-02
Wip -  fix on Measure digits fixes (c125_annotations)

MV88:
## Description

this PR changes the default of direction from W to E

there is still a problem with Input field of type number where the locale is taken from the browser.


MV88:
duplicate

PR #3675, opened 2019-04-05
autoDetectCORS option for proxy config

mbarto:
## Description
Optional configuration to introduce autodetect of CORS on services used by MapStore2.
If enabled, when contacting any service, MS2 will try to contact it directly, on error will fallback to using the proxy (and continue to use the proxy for following requests).

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

CORS is enabled only if an host is listed in useCORS configuration option.

**What is the new behavior?**
CORS is autodetected and proxy is only used as a fallback.

 - [ ] Yes
 - [x] No

**Other information**:

### DIFF
--- a/web/client/libs/ajax.js
+++ b/web/client/libs/ajax.js
@@ -7,6 +7,7 @@
  */
 
 const axios = require('axios');
+const url = require('url');
 const ConfigUtils = require('../utils/ConfigUtils');
 
 const SecurityUtils = require('../utils/SecurityUtils');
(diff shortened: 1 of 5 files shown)


web/client/libs/ajax.js
```diff
@@ -31,6 +32,8 @@ function addHeaderToAxiosConfig(axiosConfig, headerName, headerValue) {
     axiosConfig.headers = assign({}, axiosConfig.headers, {[headerName]: headerValue});
 }
 
+const corsDisabled = [];
```
offtherailz: I'm afraid this is static. Maybe you meant to make it configurable like `proxyConfig.useCORS` array ?

mbarto: No, it's not configurable, it records urls that do not support CORS so the test is not repeated and proxy is used for those.

mbarto:
@offtherailz I probably forgot to commit package.json, I have to check again...

web/client/libs/ajax.js
offtherailz: ok, sorry, I didn't get it. So... fine. No problem :smile:

PR #3678, opened 2019-04-08
Revert "autoDetectCORS option for proxy config"

mbarto:
Reverts geosolutions-it/MapStore2#3675


mbarto:
Need to check conflicts with user info stored in localStorage

PR #3690, opened 2019-04-12
Update PULL_REQUEST_TEMPLATE.md

offtherailz:
## Description
To remember that breaking changes have to be documented in migration guide


MV88:
@offtherailz what if we also update the ISSUE Template

removing the Browser Affected section and leaving only the Browser Affected Version ?

offtherailz:
@MV88 as @tdipisa said, this can be done in a second time

Issue #3622, opened 2019-03-14
User Guide Update for new features


tdipisa:
### Description
Since some new features have been introduced in master, the user guide needs to be updated.

### Other useful information (optional):

The new tools/functionalities/updates to document are the following:

- [x] Update doc related to the layer filtering (see #3785)
- [x] Search Tool updates (consider to review the doc related to the  WFS custom services that has been moved from Settings) Parent of #3523
- [x] Catalog tool updates #3824 
- [x] Tooltip settings in layer properties Parent of  #3608 
- [x] Info tool with input coordinates Parent of #3548
- [x] New annotations tool #3715 
- [x] CRS selector tool #3716 (**only on master for this release 2019.02.00**)

PR #3604, opened 2019-03-08
Fix #3531 collapsible hidden widgets

offtherailz:
## Description
This pull request allow to collapse hidden widgets, when the user can see them. 

## Note for testers
There is no way to test this if you don't have the hide tool activated. dev server don't have it activated

### Issues
 - Fix #3531

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)

 - [x] Bugfix
 
**What is the current behavior?** (You can also link to an open issue here)
Admin can see hidden widgets but can not collapse them

**What is the new behavior?**
Hidden widgets are as the other widgets for the admin.

**Does this PR introduce a breaking change?** (check one with "x", remove the other)

 - [x] No

PR #3535, opened 2019-02-14
Fix #3531 Add collapse/pin/hide functionalities to the widgets

offtherailz:
## Description
With this changes:
 - widgets can be collapsed in the new "WidgetTray" plugin. 
 - widgets can be pinned (see the related issue for definition of `pin`)
 - widgets can be hidden to certain users (via localConfig.json)
 - hide and pin tools visibility and policies can be configured in localConfig.json. 
 - text widgets can contain images

### Pin
These tools are implemented as icons on the top left corner of a widget
image
Pinned widgets look like this: 
image
The pin tool is not visible to users that are not able to pin/unpin widgets. By default, only admin can pin (configurable). 

### Collapse
Collapsed widget appear in a tray
image
When collapse a widget, the tray expands. You can collapse all widgets at once using the "collapse all" button in the tray

### Hide
The hide functionality is disabled by default. If configured (in localConfig.json), a user that belongs to a certain role can be able to hide a widget to other users. 

### Add image to text widget
When you edit/create a widget, the WYSIWYG editor have the tool to add images. 

image

## Issues
 - Fix #3531
 - Connected to #2655 (only for collapse widget task)

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Feature

**What is the current behavior?** (You can also link to an open issue here)
Nor of features in  #3531 is present

**What is the new behavior?**
Widgets are collapsible

**Does this PR introduce a breaking change?** (check one with "x", remove the other)

 - [x] No

**Other information**:

PR #3691, opened 2019-04-12
Porting of widgets tools #3531 for 2019.01.xx

### LINKED ISSUE #3531: Widgets collapse/pin/hide functionalities
### Acceitance Criteria
- [x] The user must be able to collapse/expand widgets: one by one or all together, in a tray
- [x] The admin should be able to pin widget: a pinned widget can not be moved, resized or collapsed. A normal user can not pin/unpin widgets.
- [x] when MapStore is configured accordingly in `localConfig.json`, certain users should be able hide widgets. An hidden user is not visible to other users. 
 - [x] All the properties and the visibility of these tools should be configurable by user role. 
 - [x] User should also be able to add images to the text widget content

offtherailz:
Porting of pull requests #3535, #3604 (Feature #3531) for branch 2019.01.xx
## Issues
 - Fix #3531

NOTE: @tdipisa theoretically this feature is not documented yet (#3622).

### DIFF
--- a/docma-config.json
+++ b/docma-config.json
@@ -120,6 +120,7 @@
             "web/client/components/misc/cardgrids/SideGrid.jsx",
             "web/client/components/misc/enhancers/emptyState.jsx",
             "web/client/components/misc/enhancers/tooltip.jsx",
+            "web/client/components/misc/enhancers/security/accessRuleParser.jsx",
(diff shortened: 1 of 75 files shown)


offtherailz:
As discussed the documentation is going do be written and back-ported

PR #3719, opened 2019-05-06
Fix #3415 Layer properties are not accessible if its style is invalid

### LINKED ISSUE #3415: Layer properties are not accessible if its style is invalid
### Description
In some cases styles are created and updated correctly but when apply to the layer return an exception.
If GetMap has exception TOCItemSettings is not accessible and the style parameters can't be changed.

* Following style cause the issue
```
@styleTitle 'Solid fill';
@styleAbstract '';

[this_is_not_a_param = 'a'] {
	fill: #aaaaaa;
}
```

*Browser Affected*
- [ ] Internet Explorer
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

- all

*Steps to reproduce*

- open a map with vector layer
- select the layer and open the settings of layer
- select style tab
- create a new style and edit it
- write the style* that cause the error

*Expected Result*

- There should be a way to access the parameters to update the layer parameters or at least reset the layer parameters to ensure that the error depends on wrong paramaters

*Current Result*

- Layer parameters can't be chaged and settings aren't accessible

allyoucanmap:
## Description
This PR introduces the possibility to access the settings of a layer even if it has error so if a setting causes an error is possible to update it.

## Issues
 - Fix #3415

 - [x] Bugfix

#3415 

**What is the new behavior?**
It's possible to access layer properties even if there is an error

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/TOC/Toolbar.jsx
+++ b/web/client/components/TOC/Toolbar.jsx
@@ -124,7 +124,16 @@ class Toolbar extends React.Component {
     isLoading = () => {
         return head(this.props.selectedLayers.filter(l => l.loading));
     }
-
+    /**
+     * retrieve current status based on selected layers and groups
+     * 'DESELECT' no selection
+     * 'LAYER' single layer selection
+     * 'LAYERS' multiple layer selection
+     * 'GROUP' single group selection, it select also children layers
+     * 'GROUPS' multiple group selection, it select also children layers
+     * 'LAYER_LOAD_ERROR' single layer selection with error
+     * 'LAYERS_LOAD_ERROR' multiple layer selection with error, all selected layer have an error
+     */
     getStatus = () => {
         const {selectedLayers, selectedGroups} = this.props;
         const isSingleGroup = this.isNestedGroup();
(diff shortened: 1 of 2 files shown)


web/client/components/TOC/Toolbar.jsx
```diff
@@ -133,7 +133,7 @@ class Toolbar extends React.Component {
         status = isSingleGroup ? 'GROUP' : status;
         status = selectedLayers.length > 1 & selectedGroups.length === 0 ? 'LAYERS' : status;
         status = selectedGroups.length > 1 && !isSingleGroup ? 'GROUPS' : status;
-        status = this.props.selectedLayers.length > 0 && this.props.selectedLayers.filter(l => l.loadingError === 'Error').length === this.props.selectedLayers.length ? 'LAYERS_LOAD_ERROR' : status;
+        status = this.props.selectedLayers.length > 0 && this.props.selectedLayers.filter(l => l.loadingError === 'Error').length === this.props.selectedLayers.length ? `${status}_LOAD_ERROR` : status;
```
offtherailz: please document getStatus. Is not too much clear what are the possible states of status. Please include all the definitions

tdipisa:
@allyoucanmap please provide a backport to 2019.01.xx

Issue #3175, opened 2018-08-29
Allow to manage map backgrounds


offtherailz:
### Description
Many times the user needs to add or remove a background to the maps. Now the only way to do it is to edit the map configuration manually. 

We should allow to manage backgrounds of maps using MapStore UI. 

Proposed features (draft): 
 - Add backgrounds
   - from catalog
   - From a list of available tile providers (OSM/Google/Bing...). The list should show a thumbnail of each background available, maybe grouped by category
 - Edit current backgrounds 
   - Remove 
   - Customize (if it comes to from a catalog service)
      - Thumbnail
      - Label
      - Format
      - Style 
      - Custom params
 - Reset to defaults (replacing current backgrounds with ones from `new.json`. This may be useful to quickly change all backgrounds to many maps). 
 
### Other useful information (optional):
See also  geosolutions-it/MapStore2#3170

Mockup available [here](https://demo.geo-solutions.it/mockups/mapstore2/backgroundedit/#/viewer/openlayers/new) that includes also the part related to #3170

Some evaluation have been already performed by @allyoucanmap for the thumbnails management [here](https://docs.google.com/document/d/1DNotOKY4xlujA43p9ndHMfybUwURINzgTW2Afvd1SY8/edit)

This issue does not includes the implementation of the support for the catalogs basic auth, that work is part of another issues not connected now #3380.

PR #3727, opened 2019-05-06
WIP Background selector ( backgrounds management ) #3175

tdipisa:
See origin PR for this #3614


offtherailz:
(from the previous PR comment): 

## Bugs 

Testing I found these bugs: 
 - [ ] Add background don't work
  - Steps to reproduce:
    - Open map 4093.
    - Add background
    - search for regioni
    - Add the layer "Regioni Italiane"
  - Expected result: the layer is added and set as background 
  - Result: 
     - I see instead geosolutions:OrtofotoRegione2010 set as background layer
     - The background is enabled together with the other backgrounds (so I see 2 backgrounds enabled at the same time
wrong_double_background

- [ ] Map 464, open the backgrounds list. One thumbnail is not visible. on dev it works
image
- [ ] Backgrounds Catalog is accessible only in new map (not existing maps). It should be accessible by any other existing map.
Background Dialog
- [ ] Thumbnail is not picked if I click on the drop-zone instead of dragging the thumbnail in. 
- [ ] Style selection do not work
- [ ] There is a Type selection that is not useful (parameters are always strings, and the only type I can select is string)

### Design Issues

@tdipisa @allyoucanmap I didn't look at the background selector mock-up until now. I only said some opinions during the creation of it, but didn't see the final version until now. I think there are some design issues to fix and some additional tasks to do before to add this functionalities : 
- [ ] There should be a way to configure what Layers appear in the backgrounds catalog (e.g. google background should not be shown in most of the cases, at least until we configure the API key. They are more then 50% of the current  available background, and they do not work without API key). 
- [ ] Parameters should be configurable for any WMS layer, so it's natural place is in the layer properties. I can't figure out the utility of params in backgrounds
- [ ] Maybe Catalog and background selector should close when the background is add, or the user can do other things while the background selection is still open, creating confusion
- [ ] I can not remove layers with errors or disabled because they can not be selected, and you can delete only the selected layer.

In addition, some opinions about how it could be better:
- [ ] I think they should be grouped somehow by source (maybe different sources: google, bing, OpenStreetMap, Nasa, instead of a unique "backgrounds" source). Add TMS to the available Catalog types and then configure some of well known ones is a better choice. Then add Google, Bing... 
- [ ] There are a lot of tile provider available, but they still can't be selected. See this list [here](https://mapstore2.readthedocs.io/en/latest/developer-guide/maps-configuration/#tileprovider)
- [ ] Management for background thumbnail as separated resources requires a lot of effort, maintenance and possible issues. I should have instead add the possibility to use a tile URL (or at least save the base64 of the thumbnail directly in the map, if the URL is not available. 

I think you should take these considerations into account, because the mock-up seems good, but in the real world we have a lot of variants and possible issues not considered in the mock-up. 
We should discuss these things before start coding, defining the requirements and identifying what implementation should be better.

tdipisa:
>> Testing I found these bugs: 
we need to fix the bugs reported by @offtherailz

>> There should be a way to configure what Layers appear in the backgrounds catalog (e.g. google background should not be shown in most of the cases, at least until we configure the API key. They are more then 50% of the current  available background, and they do not work without API key).  **TO ESTIMATE --> design **

The **Map Backgrounds** section is the default one that appears to the user, it contains the default list of backgrounds. This list must be configurable from configuration file to be customized by the administrator. The default list of available backgrounds should be the one that we have now in new.json. We need also to change the **Map Backgrounds** label in the dropdown with somethign like **MapStore Backgrounds** or **Default Backgrounds**

>> Parameters should be configurable for any WMS layer, so it's natural place is in the layer properties. I can't figure out the utility of params in backgrounds **2pt**

We need this available in background settings now, so nothing to do except for the pending fixes on it reported above by @offtherailz

>> Maybe Catalog and background selector should close when the background is add, or the user can do other things while the background selection is still open, creating confusion **DECIDE 0pt**

I agree, we have to do that.

>> I can not remove layers with errors or disabled because they can not be selected, and you can delete only the selected layer. **TO ESTIMATE: --> design a different way to delete the backgrounds <--**

I agree with that, this need discussion with @allyoucanmap.

>> I think they should be grouped somehow by source (maybe different sources: google, bing, OpenStreetMap, Nasa, instead of a unique "backgrounds" source). Add TMS to the available Catalog types and then configure some of well known ones is a better choice. Then add Google, Bing... **opinion**

Nothing to do here.

>> There are a lot of tile provider available, but they still can't be selected. See this list [here](https://mapstore2.readthedocs.io/en/latest/developer-guide/maps-configuration/#tileprovider) **TO ESTIMATE --> please list backgrounds to use and design how to select them (by category, what options or alternatives)** 

Nothing to do here.

>> Management for background thumbnail as separated resources requires a lot of effort, maintenance and possible issues. I should have instead add the possibility to use a tile URL (or at least save the base64 of the thumbnail directly in the map, if the URL is not available. **3pt**

Ok, we can consider to change the current implementation but the thumbnail must be added to the user in the same way (drag&drop/selection area). We can highly compress the image selected by the user to have a short base64 to put directly inside the map configuration (the size of the preview is small so is doesn't need an high definition).

offtherailz:
Closing in favor of the new PR updated #4343

PR #3797, opened 2019-05-24
#2215: add group in TOC toolbar

### LINKED ISSUE #2215: Allow creation of new groups / subgroups from the TOC
We should have:
 - [ ] an "Add group" button next to "Add layer" to add first level groups
 - [ ] an "Add group" tool when a group is selected to create subgroups
 - [ ] an "Add layer" tool when a group is selected to create a layer and assign it directly to the selected group
 - [ ] a plugin configuration option to limit the level of nesting allowed
 - [ ] use icons instead of text in the buttons at all times

We should also change the group selection combo behaviour in layer settings, to only allow changing the group of layer and remove the ability to create new groups from there.

mbarto:
## Description
Improvements to the TOC:
 * NEW add group button (when status is DESELECTED or a single group is selected)
 * add layer button (when status is DESELECTED or a single group is selected) --> adds a layer directly to the selected group

## Issues
 - Fix #2215

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

Adding a new group or adding a layer to an existing group is complicated.

**What is the new behavior?**
Adding a new group or adding a layer to an existing group is easier.

 - [x] No

**Other information**:


web/client/actions/controls.js
```diff
@@ -52,6 +55,21 @@ function setControlProperty(control, property, value, toggle) {
...
 }
 
+/**
+ * Sets a list of properties at once
+ * @memberof actions.controls
+ * @param {string} control  control name
+ * @param {object} properties the properties to set, as key (property name), value pairs
+ * @return {object} of type `SET_CONTROL_PROPERTIES` with control and properties
+ */
+function setControlProperties(control, ...properties) {
```
offtherailz: Why not directly the object? I think is more explicit that the array:
```
setControlProperties('mycontrol', 'a', 1, 'b', 2); // less intuitive
setControlProperties('mycontrol', {a:1, b: 2}) // more intuitive
```
Notice. The documentation tells something different. I didn't try yet, but documenting spread operator may work using this syntax `@param {...object} myParam` (JSDoc 3.4.3). Please try (or refactor as suggested)

mbarto:
@offtherailz "When I open the catalog from the TOC the layer is added to the selected group, when I do it from the menu, the layer is added to the "Default" group. This may confuse the user."
What you suggest instead?

web/client/actions/controls.js
mbarto: Because using ...properties is more functional friendly (partial application) and in fact my first usage of it took advantage of partial application. Eventually we can also support both syntaxes

offtherailz:
@mbarto I'd like to see the opinion of @tdipisa about it. Anyway I can suggest to set the tooltip "Add layer to the selected group", that clarifies a little bit. I think that it could be enough.

web/client/actions/controls.js
offtherailz: Now I see. That's cool :+1: 
I should not do both to avoid confusion. 
Maybe here we are creating an ad-hoc API to solve a specific problem (binding connect with partial application). 

So you could do 2 different things: 
1. Change the documentation to specify better how to use the action
2. Change the `.bind` with something like ( properties ) => setControlProperties('myControl', { ...properties

PR #3832, opened 2019-06-13
Fix importer

offtherailz:
## Description
Update importer API to work with latest versions of GeoServer
Support for External page and manager Menu

 - [x] Feature

Importer worked only in user manager, run import caused errors.

**What is the new behavior?**
Run interaction has been fixed and Importer follows the same Strategy of RuleManager to be included in the application. 

 - [x] No


web/client/api/geoserver/Importer.js
```diff
@@ -55,9 +55,15 @@ var Api = {
         let url = geoserverBaseUrl + "imports/" + importId + "/tasks/" + taskId + "/target";
         return axios.get(url, options);
     },
-    runImport: function( geoserverBaseUrl, importId, options) {
+    runImport: function( geoserverBaseUrl, importId, options = {}) {
         let url = geoserverBaseUrl + "imports/" + importId + "?async=true";
-        return axios.post(url, null, options);
+        return axios.post(url, null, {
+            ...options,
```
mbarto: I think you are duplicating headers, this way. I would extract headers from options

web/client/plugins/manager/ManagerMenu.jsx
```diff
@@ -108,6 +121,7 @@ class ManagerMenu extends React.Component {
 module.exports = {
     ManagerMenuPlugin: assign(connect((state) => ({
         enableRulesManager: isRulesManagerConfigured(state),
+        enableImporter: () => isImporterConfigured(state),
```
mbarto: why not simply isImporterConfigured(state) ?

web/client/product/pages/Importer.jsx
```diff
@@ -0,0 +1,84 @@
...
+  *         // ...
+  *         {
+  *             "name": "Importer",
+  *            "cfg": {} // see plugin configuration
+  *         }
+  *     ]
+  * }
+*/
+
+class ImporterPage extends React.Component {
```
mbarto: Maybe it's time to have some basic page to reuse, I think most of the pages we have are now identical, except some configuration option. What do you think?

web/client/selectors/importer.js
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright 2019, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+export const isImporterConfigured = state => state.localConfig && state.localConfig.plugins && !!state.localConfig.plugins.importer;
```
mbarto: Maybe we can create some generic selector to look for configured plugins, instead of creating one for each plugin. What do you think?

web/client/product/pages/Importer.jsx
offtherailz: There are some differences that can not be put at common during this activity. I would prefer to do a task like this using different, dedicated time. . 

The only ones that look really the same are Importer and RulesManager (with anyway some exclusion that is not trivial). The other have some actions, lifecycle, actions that are not easy to put together.

web/client/selectors/importer.js
offtherailz: I'd like to do the same as above. Think about pages and do it once, in a separate task. Not so difficult but requires some plan and time.

web/client/plugins/manager/ManagerMenu.jsx
offtherailz: Because I did an error, good catch :+1:

web/client/api/geoserver/Importer.js
offtherailz: yes, good proposal, it could be more readable.

web/client/product/pages/Importer.jsx
mbarto: Please, open a separated issue for this

web/client/selectors/importer.js
mbarto: Maybe this could work:

export const isPluginConfigured = (plugin) => ((state) => state.localConfig && state.localConfig.plugins && !!state.localConfig.plugins[plugin]);

offtherailz: Where should I place it. `web/client/selectors/{,,,}.js` ?
plugins.js , localConfig.js ? What do you think is a better name for this selector?

web/client/product/pages/Importer.jsx
offtherailz: did it. here it is
https://github.com/geosolutions-it/MapStore2/issues/3841

web/client/plugins/manager/ManagerMenu.jsx
offtherailz: This should be solved

web/client/selectors/importer.js
mbarto: plugins.js

web/client/selectors/plugins.js
```diff
@@ -0,0 +1,15 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Check if an entry in localConfig is configured for a specific entry of `plugins`
+ * @param {string} entry the name of the entry
+ * @return {boolean} true if the entry exists, false otherwise
+ */
+export const isPluginsEntryConfigured = plugin => state => state.localConfig && state.localConfig.plugins && !!state.localConfig.plugins[plugin];
```
mbarto: I think the originally suggested isPluginConfigured name is better. It explains what the selector returns (is my plugin included in configuration?), not the technical details of entries in a JSON object.

offtherailz: Yes I remember, but I though that is a page or a part of plugins entry that contains many plugins inside it. So even isPluginConfigured is a little obscure. (I expect to see if a single plugin is configured, not an entire page). I see also that the first argument as a misleading name. 

I can suggest 
- isPageConfigured pageName => ...
- isPluginsEntryConfigured entry => 
- isPluginsSetConfigured setName =>

mbarto: You are right, this is about a page, not a single plugin, hasPagePlugins? isPageConfigured is good too

offtherailz: ok, done with isPageConfigured

PR #3839, opened 2019-06-14
Geofence Integrated Rules Manager

offtherailz:
## Description
Add support to manage GeoFence integrated in GeoServer.
 - Remove old actions and Components of the fist version of RuleManager
 - Support API to interact with users, rules and layers
 - Switch between the two APIs
 - Some fixes to the UI due to the updates of the Grid Component
 - Documentation of GeoFence Page

 - [x] Feature

 - [x] No


web/client/api/geoserver/geofence/RuleService.js
```diff
@@ -0,0 +1,169 @@
...
+};
+
+/**
+ *
+ * @param {object} object the rule
+ */
+const clearNullEntries = object => pickBy(object, v => v !== null);
+
+// TODO: missing documentation
+const cleanConstraints = (rule) => {
```
mbarto: Looks like there is a lot of duplicated code, here and in the other RuleService.js

offtherailz: In fact it is the differences game :smile: 

image

There are many small differences but you are right, 3 methods are equal. 

I didn't think they have to be put in common because the API on GeoServer is going to be re-implemented soon using yet another library for JavaToJson conversion. Therefore I suppose the differences will increase and so it makes sense to keep it as it is. 

What do you think?

PR #3904, opened 2019-07-03
Fix for standardApp and MeasurementSupport leaflet test failure

MV88:
## Description
This pr wants to remove some setTimeout that can be the cause of the travis build failures

## Issues
 -  partial fix of #2971 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

travis build failing too often

**What is the new behavior?**
hopefully the build stops failing randomly

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/app/__tests__/StandardApp-test.jsx
+++ b/web/client/components/app/__tests__/StandardApp-test.jsx
@@ -53,16 +53,14 @@ describe('StandardApp', () => {
         expect(app).toExist();
     });
 
-    it('creates a default app with onInit', (done) => {
+    it('creates a default app with onInit', () => {
         const init = {
-            onInit: () => { }
+            onInit: (cfg) => {
+                expect(cfg).toExist();
+            }
         };
-        const spy = expect.spyOn(init, 'onInit');
-        ReactDOM.render(<StandardApp onInit={init.onInit}/>, document.getElementById("container"));
-        setTimeout(() => {
-            expect(spy).toHaveBeenCalled();
-            done();
-        }, 100);
+        let app = ReactDOM.render(<StandardApp onInit={init.onInit}/>, document.getElementById("container"));
+        expect(app).toExist();
     });
 
     it('creates a default app with the given store creator', (done) => {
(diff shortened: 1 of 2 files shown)


web/client/components/app/__tests__/StandardApp-test.jsx
```diff
@@ -53,16 +53,14 @@ describe('StandardApp', () => {
...
+                expect(cfg).toExist();
+            }
         };
-        const spy = expect.spyOn(init, 'onInit');
-        ReactDOM.render(<StandardApp onInit={init.onInit}/>, document.getElementById("container"));
-        setTimeout(() => {
-            expect(spy).toHaveBeenCalled();
-            done();
-        }, 100);
+        let app = ReactDOM.render(<StandardApp onInit={init.onInit}/>, document.getElementById("container"));
```
mbarto: I think this changes the original test: you are removing the spy test of onInit. If onInit is not called this test is successful anyway.

MV88: you're right. but there is a way to avoid to use the setTimeout?

MV88: i was not able to find a final and good solution. Probably we can increase the timeout but that is not the solution

mbarto: did you try to remove the spy AND move done() inside the onInit function ?

MV88: yes, but as you said before: "If onInit is not called this test is successful anyway."  the done() will not be called either

mbarto: But then it will fail for the right reason: bacause onInit is not called. If onInit is called the test will pass, whatever the time needed (within the 2 seconds max timeout of every test)

MV88: Dammit I was so close!  

You are superb!

Issue #3924, opened 2019-07-09
Graphic glitch when filtering features by region


gioscarda:
### Description
When filtering the features of a vector layer by region, the region geometry should not be visible after the search has been triggered.

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [x] Internet Explorer
- [x] Chrome
- [x] Firefox
- [x] Safari

*Browser Version Affected*

- All

*Steps to reproduce*

- open a map with vector layers
- open the TOC and select a vector layer
- open the layer Attribute Table
- open the Advanced Search tool
- select 'Filter Type' and 'Geometric Operation' in the 'region of interest' section
- draw the region geometry on the map
- click on search

*Expected Result*

- the advanced search window closes, the feature grid opens and the region geometry is not visible on the map
- if the filter is configured to stay in the map it should not have the white fill

*Current Result*

- the advanced search window closes, the feature grid opens and the region geometry is still visible on the map (it has an opaque fill also which is wrong)

![Selection_041](https://user-images.githubusercontent.com/43670920/60881809-cb13e580-a246-11e9-9372-d9c0c649c6df.png)

MV88:
addendum: if the filter is configured to stay in the map it should not have the white fill

PR #3948, opened 2019-07-11
Fix Glich of query panels roi styles

MV88:
## Description
this pr fixes the behaviour of query panel rois.

if showFilteredObject is configured to be true it keeps the filter when pressing search
and the the style has a white fill with 0.2 opacity

## Issues
 - #3924 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

filter stays in map with a wrong style

**What is the new behavior?**
style is correct if filter remains in map.
by default filter wil lnot remain (reducer change)

 - [x] No

**Other information**:
TODO  we should handle a fill configured as RGBA(R, G, B, A)

### DIFF
--- a/web/client/plugins/FeatureEditor.jsx
+++ b/web/client/plugins/FeatureEditor.jsx
@@ -144,6 +144,7 @@ const FeatureDock = (props = {
             {getDialogs(props.tools)}
             <Grid
                 onMount={props.onMount}
+                showFilteredObject={props.showFilteredObject}
                 editingAllowedRoles={props.editingAllowedRoles}
                 initPlugin={props.initPlugin}
                 customEditorsOptions={props.customEditorsOptions}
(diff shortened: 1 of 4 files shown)


web/client/plugins/FeatureEditor.jsx
```diff
@@ -144,6 +144,7 @@ const FeatureDock = (props = {
             {getDialogs(props.tools)}
             <Grid
                 onMount={props.onMount}
+                showFilteredObject={props.showFilteredObject}
                 editingAllowedRoles={props.editingAllowedRoles}
```
allyoucanmap: we could remove this props because they are not used anymore for grid
```
onMount={props.onMount}
showFilteredObject={props.showFilteredObject}
```
they have been moved to the lifecycle of plugin

Issue #3985, opened 2019-07-18
Autosearch for Catalog after a delay


MV88:
### Description
- [ ]  Add autosearch in Catalog
- [ ] The delay can be configurable, default 1 sec

### Other useful information (optional):
It will be needed a refactor since catalog is an old plugin because we need to introduce observables in order to handle this kind of interaction (delay etc). In that way we will handle asyncronously the text flow and api anwsers in an optimal way

PR #3988, opened 2019-07-19
Fix 3985 add catalog autosearch

MV88:
## Description
This pr adds AutoSearch to the filter of catalog form
Also loading state has been moved from local to the global state
The delay can be configured in the initialState of localConfig, for example: 
```
"initialState": {
  "defaultState": {
    "catalog": {
      "delayAutoSearch" : 1000
    }
  }
}
```

## Issues
 - #3985

 - [x] Feature

When you type in the filter you have to click on "search button" in order to trigger the search

**What is the new behavior?**
after 1 second it triggers automatically the search when idle for 1s

 - [x] No 

**Other information**:
notes this could be harmful for projects that are customizing the catalog plugin (components and plugin changes needs to be ported too)

### DIFF
--- a/web/client/epics/__tests__/catalog-test.js
+++ b/web/client/epics/__tests__/catalog-test.js
@@ -12,10 +12,10 @@ const API = {
     csw
 };
 import catalog from '../catalog';
-const {getMetadataRecordById, autoSearchEpic} = catalog(API);
+const {getMetadataRecordById} = catalog(API);
 import {SHOW_NOTIFICATION} from '../../actions/notifications';
-import {testEpic, TEST_TIMEOUT, addTimeoutEpic} from './epicTestUtils';
-import {getMetadataRecordById as initAction, changeText, SET_LOADING} from '../../actions/catalog';
+import {testEpic} from './epicTestUtils';
+import {getMetadataRecordById as initAction} from '../../actions/catalog';
 
 describe('catalog Epics', () => {
     it('getMetadataRecordById', (done) => {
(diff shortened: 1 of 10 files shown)


web/client/epics/catalog.js
```diff
@@ -162,5 +172,19 @@ module.exports = (API) => ({
...
+                const state = getState();
+                const delay = delayAutoSearchSelector(state);
+                return Rx.Observable.timer(delay);
+            })
+            .switchMap(({text}) => {
+                const state = getState();
+                const pageSize = pageSizeSelector(state);
+                const {type, url} = selectedCatalogSelector(state);
+                return Rx.Observable.of(textSearch(type, url, 1, pageSize, text))
+                    .startWith(setLoading(true));
```
mbarto: Is this needed? It looks like textSearch already dispatches setLoading(true)

web/client/reducers/catalog.js
```diff
@@ -59,15 +60,17 @@ function catalog(state = {
 }, action) {
```
mbarto: I would add the new state property loading here too

mbarto: And also pageSize, maybe?

web/client/epics/catalog.js
MV88: yes, on a second thought this can be removed

web/client/reducers/catalog.js
```diff
@@ -54,6 +54,8 @@ function catalog(state = {
         selectedService: "",
         newService: {}
     },
+    delayAutoSearch: 1000,
```
mbarto: and loading?

web/client/epics/__tests__/catalog-test.js
```diff
@@ -36,38 +36,5 @@ describe('catalog Epics', () => {
         });
 
     });
-    it('autoSearchEpic', (done) => {
```
mbarto: Why this has been removed?

MV88: it was dispatching only a thunk that is tested in the actions/__tests__/catalog-test.js
one way to test it here is to transform the thunk into an epic, if you think is worth the shot, i'll do it

PR #4004, opened 2019-07-24
Fix 2749 search urls for cross layer filter are compared ignoring "dirty" chars

MV88:
## Description
There was a problem when comparing search url for cross layer filter functionality. With this pr it allows to ignore some small differences like ? or ?&

## Issues
 - #2749 

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix

**What is the current behavior?** (You can also link to an open issue here)
some layers are filtered out from the select menu in the cross layer filter form

**What is the new behavior?**
Layers that have similar search url besides a ? or a ?& are not filtered out from the list

**Does this PR introduce a breaking change?** (check one with "x", remove the other)

 - [ ] Yes, and I documented them in migration notes
 - [X] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:
If you think there are other urls to test just drop some examples, thanks.

Issue #2749, opened 2018-03-15
Smarter match of cross layer filtering allowed layer


offtherailz:
### Description
Now cross layer filtering is allowed only between layers that have exactly the same WFS URL. 
Sometimes layers are added from different services 
e.g. 
 - WMS getCapabilities
 - CSW Catalogs
 - manually edited
 - legacy layers from MapStore1.

Differnces can be: 
 - /geoserver/wfs
 - http://domain/geoserver
 - http://domain/geoserver:80
 - http://domain/geoserver:80/?
 - http://domain/geoserver:80/?&
 - ...

In those case the cross layer filtering tool may not recognize that 2 layers as coming from the same source. So some layer that can be used as filters may not be shown in the list of available layers for cross filtering. 

we could improve the tool to split the URL in parts (protocol, domain, port, path, queryparams) and compare them.

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
any

*Steps to reproduce*

- Add vector layers of the same source from WMS and CSW catalog and/or manually editing the map
- Try to inspect data using cross layer filtering

*Expected Result*

- All layers from the same source are listed as available for cross layer

*Current Result*

- Not all the layers may be listed

PR #4013, opened 2019-07-25
Fixed issues due to sameURL check in #4004

offtherailz:
## Description
The solution provided in #4004 discovered some issues that was not present with the original implementation: 
 - If one of the URL is null, false, undefined now it fails
 - If one of the URL is an array (sometimes coming from WMTS or other multiple URL services)
 - Even the solution implemented in #4004 there was a wrong comparison between the 2 URLs. 

The solution proposed compares query params one by one to check if the URL is the same. 

Also added tests to cover various use cases. 

## Issues
 - #2749 

 - [x] Bugfix
 
Sometimes Query panels can not be loaded. 
Random errors here and there for map loading.

**What is the new behavior?**
URLs are compared correctly, preventing exceptions. 

 - [x] No

### DIFF
--- a/web/client/utils/URLUtils.js
+++ b/web/client/utils/URLUtils.js
@@ -35,7 +35,7 @@ export const sameQueryParams = ( q1 = "", q2 = "") => {
     if (!q1 && !q2) {
         return true;
     }
-    const params1 = q1 ? q1.split('&') : [];
+    const params1 = q1 ? q1.split('&').filter(v => !!v) : [];
     const params2 = q2 ? q2.split('&').filter(v => !!v) : [];
     return isEqual(sortBy(params1), sortBy(params2));
 };
@@ -56,7 +56,7 @@ export const isSameUrl = (u1, u2) => {
     if (originalUrl === otherUrl) return true;
     if (!originalUrl || !otherUrl) return false; // if one is undefined they are not the same
     // check type before parsing to avoid parse exceptions
-    if (!isString(originalUrl || !isString(otherUrl))) return false;
+    if (!isString(originalUrl) || !isString(otherUrl)) return false;
     const urlParsed = Url.parse(originalUrl);
     const otherUrlParsed = Url.parse(otherUrl);
 
--- a/web/client/utils/URLUtils.js
+++ b/web/client/utils/URLUtils.js
@@ -7,6 +7,7 @@
 */
 
 import Url from "url";
+import { isArray, isString, isEqual, sortBy } from 'lodash';
 
 export const urlParts = (url) => {
     if (!url) return {};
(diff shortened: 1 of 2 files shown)


web/client/utils/URLUtils.js
```diff
@@ -26,29 +27,46 @@ export const urlParts = (url) => {
...
  */
-export const isSameUrl = (originalUrl, otherUrl) => {
+export const isSameUrl = (u1, u2) => {
+    // if array takes the first
+    const originalUrl = isArray(u1) ? u1[0] : u1;
+    const otherUrl = isArray(u2) ? u2[0] : u2;
     if (originalUrl === otherUrl) return true;
+    if (!originalUrl || !otherUrl) return false; // if one is undefined they are not the same
+    // check type before parsing to avoid parse exceptions
+    if (!isString(originalUrl || !isString(otherUrl))) return false;
```
mbarto: wrong parentheses

web/client/utils/URLUtils.js
```diff
@@ -26,29 +27,46 @@ export const urlParts = (url) => {
...
 
+export const sameQueryParams = ( q1 = "", q2 = "") => {
+    if (q1 === q2) {
+        return true;
+    }
+    // if both "", false or undefined, means they are both empty query strings
+    if (!q1 && !q2) {
+        return true;
+    }
+    const params1 = q1 ? q1.split('&') : [];
```
mbarto: why don't you filter also this one?

web/client/utils/URLUtils.js
```diff
@@ -26,29 +27,46 @@ export const urlParts = (url) => {
...
  * @memberof utils.URLUtils
- * @param  {string} originalUrl the original url
- * @param  {string} otherUrl url to compare to
+ * @param  {string|string[]} u1 the first URL to compare (or an array of URLs)
+ * @param  {string!string[]} u2 the second URL to compare with (or an array of URLs)
  * @return {boolean} true when urls are the same else false
  */
-export const isSameUrl = (originalUrl, otherUrl) => {
+export const isSameUrl = (u1, u2) => {
+    // if array takes the first
```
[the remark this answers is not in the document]
offtherailz: This comparison needs to understand if the 2 layers come from the same source. Even in the case of multiple URLs, for this purpose it should be enough.

web/client/utils/URLUtils.js
offtherailz: thanks

web/client/utils/URLUtils.js
offtherailz: because I'm blind

web/client/utils/URLUtils.js
```diff
@@ -26,29 +27,46 @@ export const urlParts = (url) => {
...
+    const originalUrl = isArray(u1) ? u1[0] : u1;
+    const otherUrl = isArray(u2) ? u2[0] : u2;
     if (originalUrl === otherUrl) return true;
+    if (!originalUrl || !otherUrl) return false; // if one is undefined they are not the same
+    // check type before parsing to avoid parse exceptions
+    if (!isString(originalUrl || !isString(otherUrl))) return false;
```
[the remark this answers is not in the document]
offtherailz: fixed

PR #4025, opened 2019-07-30
Close #3025 Add boundary for draggable dialog

### LINKED ISSUE #3025: Limit the zone where draggable dialogs can be dragged or change them
### Description
Opening the ABOUT window from the BURGER MENU' (menù options) you can move this window also outside the screen but after this you can't close the window using the X.

### In case of Bug (otherwise remove this paragraph):
*Browser Affected*
- [x] Internet Explorer
- [x] Chrome
- [x] Firefox
- [ ] Safari

*Steps to reproduce*
- Open a new map
- open the burger menù
- open the ABOUT option
- move the window outside the screen 
- try to close using the X

*Expected Result*
- you can't move the window outside the screen and you can close the window using the X 

*Current Result*
- you can move the window also outside the screen but after this you can't close the window using the X.

immagine

kasongoyo:
## Description
Limit the dialog dragging zone to be within the view port. It also includes changes to the followings dialogs;
- DownloadDialog
- GroupDialog
- UserDialog
- MeasureDialog
- ShareDialog
- Print
- SearchServicesConfig
- Settings
- About

## Issues
 - #3025 
 - ...

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see #3025 

**What is the new behavior?**
The draggable dialog are by default limited to browser view port

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/misc/Dialog.jsx
+++ b/web/client/components/misc/Dialog.jsx
@@ -12,7 +12,6 @@ const Draggable = require('react-draggable');
 const Spinner = require('react-spinkit');
 const assign = require('object-assign');
 const Message = require('../I18N/Message');
-const {withState, withHandlers, compose} = require('recompose');
 
 class Dialog extends React.Component {
     static propTypes = {
@@ -29,9 +28,7 @@ class Dialog extends React.Component {
         modal: PropTypes.bool,
         start: PropTypes.object,
         draggable: PropTypes.bool,
-        bounds: PropTypes.object,
-        noBounds: PropTypes.bool, // if true and dialog is draggable then the there will no no biundary for dragging
-        onDragStart: PropTypes.func
+        bounds: PropTypes.object
     };
 
     static defaultProps = {
@@ -39,7 +36,7 @@ class Dialog extends React.Component {
         backgroundStyle: {
             background: "rgba(0,0,0,.5)"
         },
-        start: {x: 0, y: 0},
+        start: {x: 0, y: 150},
         className: "modal-dialog modal-content",
         maskLoading: false,
         containerClassName: "",
@@ -47,7 +44,8 @@ class Dialog extends React.Component {
         bodyClassName: "modal-body",
         footerClassName: "modal-footer",
         modal: false,
-        draggable: true
+        draggable: true,
+        bounds: 'parent'
     };
 
     renderLoading = () => {
@@ -76,8 +74,7 @@ class Dialog extends React.Component {
     };
 
     render() {
-        const { bounds, noBounds, onDragStart } = this.props;
-        const body = (<div id={this.props.id} style={{zIndex: 3, ...this.props.style}} className={this.props.className + " modal-dialog-container"}>
+        const body = (<div id={this.props.id} style={{zIndex: 3, ...this.props.style}} className={`${this.props.draggable ? 'modal-dialog-draggable' : ''} ${this.props.className} modal-dialog-container`}>
             <div className={this.props.headerClassName + " draggable-header"}>
                 {this.renderRole('header')}
             </div>
@@ -89,7 +86,7 @@ class Dialog extends React.Component {
                 {this.renderRole('footer')}
             </div> : <span/>}
         </div>);
-        const dialog = this.props.draggable ? (<Draggable defaultPosition={this.props.start} bounds={!noBounds ? bounds : null} onStart={onDragStart} onStop={this.handleStop} handle=".draggable-header, .draggable-header *">
+        const dialog = this.props.draggable ? (<Draggable defaultPosition={this.props.start} bounds={this.props.bounds} handle=".draggable-header, .draggable-header *">
             {body}
         </Draggable>) : body;
         let containerStyle = assign({}, this.props.style.display ? {display: this.props.style.display} : {}, this.props.backgroundStyle);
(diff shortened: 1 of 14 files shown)


web/client/components/misc/Dialog.jsx
```diff
@@ -105,4 +110,30 @@ class Dialog extends React.Component {
     };
 }
 
-module.exports = Dialog;
+const enhance = compose(
```
offtherailz: This is too complicated and hardly maintainable.

kasongoyo:
@offtherailz But with your proposed soln, the draggable will never move off the view port. I thought user will need to drag the dialog off the view port for some percentage in order to see clearly what's behind the dialog?  That's what I was hardly trying to achieve in the first place otherwise the proposed soln is simple and less complicated.

offtherailz:
I see your point. Maybe windows like print or other big ones may need this. @mbarto what do you think?
By my side, I don't think is necessary to make the dialog exit, even partially, and it takes a lot in terms of maintainability and stability. I should use build-in library functionalities as much as possible.
Let's wait a comment from @mbarto

kasongoyo:
@offtherailz  If dialog partial exit is not a priority, I can even use `bounds=parent` as the default for dialog bounds and it will work in almost all places where dialog is used since almost everywhere dialog is contained in fixed position container.

offtherailz:
@kasongoyo I tried Please take into account the other dialogs.
And so most of the dialogs need the modal-dialog class to be removed, because otherwise they are not draggable up the `150px` margin give by that class (e.g. About) but removing this class will move the About on the top line, that's why I suggested to add `start: {x: 0,y: 150}`.

Then I tried another couple of dialogs and I see that download dialog (available from the feature grid) need to be limited in width or will have a bug. And share window had problems too.

So I suggested to fix them one by one.

tdipisa:
@kaso please review the size of the issue according to the @offtherailz suggestion so I can evaluate if we can proceed with that instead of fixing only the About dialog.

kasongoyo:
The dialog is currently used in 18 places. I can implement what @offtherailz suggested with some modification as the default  i.e removing `margin-top` from `modal-dialog` and set start to `{x:0, y:150}` but I suggest bounds to start as `parent` instead of `#selector`. I think this implementation will work on many dialogs(I will verify) and only few will need some changes. So @tdipisa  I estimate it to take minimum of *one day* task  and maximum of *two days*.

mbarto:
My opinion is that  bounds=parent should work (if a dialog needs to be put out of screen to see things, the user can close it directly, and open it again later)

tdipisa:
In my opinion and considering the involved functionalities and needed user interactions:

- InfoButton -> not draggable (already not draggable)
- DownloaDialog -> not draggable (now draggable)
- HelpTextPanel -> I think it is no longer used, we are pointing the online doc now
- GroupDialog -> not draggable (already not draggable)
- UserDialog -> not draggable (already not draggable)
- MeasureDialog -> draggable  (already draggable)
- SnapShotPanel -> I would say not draggable (I don't know if is draggable or not since this plugin is disabled)
- SharePanel -> draggable  (already draggable)
- Print -> draggable  (already draggable)
- SearchServicesConfig -> I would say not draggable (now draggable)
- Settings -> not draggable (now draggable)
- Styler -> not draggable (already not draggable)
- About -> not draggable (now draggable)

@offtherailz @mbarto @allyoucanmap  what do you think ?

mbarto:
@tdipisa why the SharePanel needs to be draggable? I agree with the other ones

offtherailz:
I tried SnapShotPanel and I noticed it's not working. This is the effect.
image

It should be draggable as well as the print, but it have to be fixed too. Agree with @mbarto about share panel. 

With not draggable, you mean even "modal" (that means dark background) ?

kasongoyo:
@tdipisa I will add modal to all non draggable dialog just to emphasize that user need to interact with the dialog at that moment.

tdipisa:
@mbarto @offtherailz 

>@tdipisa why the SharePanel needs to be draggable? I agree with the other ones

Because there is a functionality that updates the **bbox** URL parameter in the shared URL when you pan the map.

@offtherailz 

>It should be draggable as well as the print, but it have to be fixed too. Agree with @mbarto about share panel.

We have the shapshot tool working well in the c040 project (recently deployed in production) so I'm wondering why it should not work in the MS product as well, anyway check it better and open a new issue for this please. Thank you.

offtherailz:
## Notes for PM and other developers

Looking at the code I noticed these things: **Please** @tdipisa and @kasongoyo take a look at them.


I took a look at the code and I've seen that Dialog has 3 specialization (in misc directory)
- ResizableDialog: --> wrong name, it should be named ResponsiveDialog.
- ConfirmDialog --> Used In many places. I tested  them here and there and they seems to work
- StandardDialog --> It has been developed to be a simplification of the Dialog to allow common use cases. It's used for Map Import

I think we should in the future we should migrate everything StandardDialog (good, simple and documented interface, Dialog's one is dark and full of workarounds :smile: ), renaming it as "Dialog"
and the old Dialog should be renamed as "BaseDialog" and discouraged for direct usage. 

ConfirmDialog is ok. Resizable should be removed in favor of standard. 

@mbarto what do you think?

kasongoyo:
@offtherailz Concern Bugs, did you test it against my [last commit](https://github.com/geosolutions-it/MapStore2/pull/4025/commits/207647c3e73f7307f147817bb76d6ada985d7bfb)?

tdipisa:
Discuss [these topics](https://github.com/geosolutions-it/MapStore2/pull/4025#issuecomment-517631329) in the weekly DEV meeting and open a dedicated issue if needed, then keep me updated on the status so I can schedule the work properly.

offtherailz:
Ok, @kasongoyo Probably I reviewed it this morning, probably I pulled it just before your last commit. 

 I talked with Tobia, we should make the snapshot draggable, instead.

PR #4028, opened 2019-07-31
Geostory sections

offtherailz:
## Description
Add base structure for geostory scroll handlers and background setup (base support for immersive or cover backgrounds) .


offtherailz:
Going to close in favor @allyoucanmap one #4029

PR #4029, opened 2019-07-31
Geostory sections

allyoucanmap:
Add base structure for geostory scroll handlers and background setup (base support for immersive or cover backgrounds) .


web/client/themes/default/less/sidegrid.less
```diff
@@ -1,7 +1,7 @@
 .msSideGrid {
-    margin: 30px;
-    width: ~"calc(100% - 60px)";
-    position: absolute;
+    margin: 8px 16px;
```
offtherailz: Are you sure this doesn't affect other sideGrids?

PR #4075, opened 2019-08-08
Fix #2344 improve catalog loading spinner

### LINKED ISSUE #2344: Improve catalog loading spinner
We should improve the loading phase of catalog, the spinner on button is not visible and it's difficult to understand if the request has been sent.
Update:
Sometimes the capabilities are too heavy (e.g. wmts with gridsets) and the parsing phase is longer than expected, that causes the end of loading with the results page still empty.

kasongoyo:
## Description
Catalog loading spinner was small and placed in such a way that it was
difficult to see it. This commit make it bigger and easy to notice that
the catalog is loading.

Improve catalog loading spinner

## Issues
 - #2344 

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see #2344 

**What is the new behavior?** 

MapStore HomePage

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/catalog/Catalog.jsx
+++ b/web/client/components/catalog/Catalog.jsx
@@ -20,6 +20,7 @@ const Message = require("../I18N/Message");
 const OverlayTrigger = require('../misc/OverlayTrigger');
 const RecordGrid = require("./RecordGrid");
 const SwitchPanel = require("../misc/switch/SwitchPanel");
+const Loader = require('../misc/Loader');
 
 require('react-select/dist/react-select.css');
 require('react-quill/dist/quill.snow.css');
@@ -186,8 +187,9 @@ class Catalog extends React.Component {
     };
 
     renderLoading = () => {
-        return this.props.loading ? <Spinner spinnerName="circle" noFadeIn overrideSpinnerClassName="spinner"/> : null;
-    };
+        return (<div className="catalog-results loading"><Loader size="176" /></div>);
+    }
+
     renderSaving = () => {
         return this.props.saving ? <Spinner spinnerName="circle" noFadeIn overrideSpinnerClassName="spinner"/> : null;
     };
@@ -210,7 +212,6 @@ class Catalog extends React.Component {
                 onSelect={this.handlePage} />
             <div className="push-right">
                 <Message msgId="catalog.pageInfo" msgParams={{start, end: start + returned - 1, total}} />
-                {this.renderLoading()}
             </div>
         </div>);
         }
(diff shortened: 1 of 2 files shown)


web/client/components/catalog/Catalog.jsx
```diff
@@ -360,7 +362,7 @@ class Catalog extends React.Component {
                             </Form>)}
                     footer={this.renderPagination()}>
                             <div>
-                                {this.renderResult()}
+                                { this.props.loading ? <div className="catalog-results empty"><Spinner spinnerName="circle" noFadeIn overrideSpinnerClassName="spinner"/></div> : this.renderResult() }
```
allyoucanmap: Instead of Spinner we should use Loader from /misc folder to keep the style aligned with other loader.
I suggest to use size 176px similar to the main spinner of Loading MapStore.
Probably it's better to use a different class for the container that describe the loading action:
proposal -> `catalog-results-loading` instead of  `catalog-results empty`

MV88: @allyoucanmap what about removing the spinner inside the search button?

kasongoyo: I think @allyoucanmap has forgotten to remove it in the snapshot but we agreed it before that I should remove all other spinners in favour of a single visible large spinner

web/client/components/catalog/Catalog.jsx
```diff
@@ -360,7 +359,9 @@ class Catalog extends React.Component {
                             </Form>)}
                     footer={this.renderPagination()}>
                             <div>
-                                {this.renderResult()}
+                                { this.props.loading ? <div className="catalog-results loading"><Loader size="176" /><h4>
```
MV88: i would keep the new loading stuff in the renderLoading function

kasongoyo: Initial, it was in dedicated function because it was used in multiple places but currently it is used in only one place. Besides the `Loader` by itself is already a component and the surrounding container make sense in that place. What do you think @MV88

MV88: I know it is used only there, but i think is clearer to use a function if the condition of the ternary is true. it is just a matter of reability taste.

 { this.props.loading ? this.renderLoading() : this.renderResult() }
i think this is more readable then your solution. as I said is a matter of code style.

Also renderResult is used only once but we can decide to leave it there in order to shrink the render method

kasongoyo:
@MV88 I have attached the gif with the new behaviour in PR description page under new behaviour

tdipisa:
@kaso please, can you backport this to 2019.02.xx stable branch ?

PR #4098, opened 2019-08-20
Fix #3927 Dashboard error in console when exporting a chart with no data

### LINKED ISSUE #3927: Dashboard error in console when exporting a chart with no data
### Description
there is an error in console when exporting a chart with no data

the app continues to work though

*Browser Affected*
- [ ] Internet Explorer
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

*Steps to reproduce*

- open [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/dashboard/5728) dashboard as admin
- move the map where no us states are visible
- click on export of the poulation per region chart
image

*Expected Result*

- export of empty widgets does not provoke any error (even if harmless)
- probably the export should be disabled while no data are there (and the "download chart data" tool too)

*Current Result*

in console it appears this

kasongoyo:
## Description
Remove console error when exporting a chart with no data

## Issues
 - #3927 
 - ...

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see #3927 

**What is the new behavior?**
No console error

 - [ ] No

**Other information**:

### DIFF
--- a/web/client/components/widgets/enhancers/tools/exportableWidget.js
+++ b/web/client/components/widgets/enhancers/tools/exportableWidget.js
@@ -19,12 +19,14 @@ module.exports = () =>
                 glyphClassName: "exportCSV",
                 target: "menu",
                 textId: "widgets.widget.menu.downloadData",
+                visible: data && data.length,
                 onClick: () => exportCSV({ data, title })
             }, {
                 glyph: "download",
                 target: "menu",
                 glyphClassName: "exportImage",
                 textId: "widgets.widget.menu.exportImage",
+                visible: data && data.length,
                 // NOTE: the widget widget-chart-${id} must be the id of the div to export as image
                 onClick: () => exportImage({ widgetDivId: `widget-chart-${id}`, title })
             }
--- a/web/client/components/widgets/enhancers/tools/exportableWidget.js
+++ b/web/client/components/widgets/enhancers/tools/exportableWidget.js
@@ -19,14 +19,14 @@ module.exports = () =>
                 glyphClassName: "exportCSV",
                 target: "menu",
                 textId: "widgets.widget.menu.downloadData",
-                visible: data && data.length,
+                disabled: !data || !data.length,
                 onClick: () => exportCSV({ data, title })
             }, {
                 glyph: "download",
                 target: "menu",
                 glyphClassName: "exportImage",
                 textId: "widgets.widget.menu.exportImage",
-                visible: data && data.length,
+                disabled: !data || !data.length,
                 // NOTE: the widget widget-chart-${id} must be the id of the div to export as image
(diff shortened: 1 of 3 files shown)


web/client/components/widgets/enhancers/tools/exportableWidget.js
```diff
@@ -19,12 +19,14 @@ module.exports = () =>
                 glyphClassName: "exportCSV",
                 target: "menu",
                 textId: "widgets.widget.menu.downloadData",
+                visible: data && data.length,
```
MV88: Hi, with this you have found solution that avoids the problem which can be good, but have you investigated the cause?

can you explain why hiding is better than disabling?

kasongoyo: Because I discovered the problem was in widget epic ` exportWidgetImage` https://github.com/geosolutions-it/MapStore2/blob/c6e014a7c071c977e20b1385a25ad26de8256c73/web/client/epics/widgets.js#L144 
If you look at the epic code, you can notice that the epic logic is expecting the chart to be exported to be available in the document and it works well when that condition is fulfilled. One soln was to modify that logic to include the condition when the chart is not available in the dom or leave as it is and prevent it to be called when the chart is not available. I thought the latter was more appropriate since it solve the problem from the root. Why should someone export empty data/image? There is no such use case.

MV88: i see, maybe we could also add a guard before start exporting to avoid the error

one other aspect is to leave the buttons visible but disabled and for this point i would ask confirmation of @tdipisa

kasongoyo: @MV88 sorry, I need to confess that I didn't understand your qn in the first place regarding why hiding and not disabling. You can ignore my last reply because it was answering something else. Well, back to your original question regarding hiding or disabling, I think it's a matter of preference and both ways works although I see more value on disabling only since it will give user knowledge that there is a possibility of exporting but it's only valid when there is data.

tdipisa: @kasongoyo disabling is better than hiding the button but maybe a tooltip should be shown to inform the user why the button is disabled.

MV88: @tdipisa i think that a disabled button has no possibility to show the tooltip

tdipisa: ok

tdipisa:
@kasongoyo please, can you backport this to 2019.02.xx stable branch ?

PR #4100, opened 2019-08-22
Fix #3144 Different ajax requests for empty params using proxy

### LINKED ISSUE #3144: Different ajax requests for empty params using proxy
### Description
Performing ajax ( `ajax.js`) requests  with undefined `params` values causes different query strings using or not using proxy. 
If you use proxy, the params with undefined values will be added to the request this: 

`paramName1=&paramName2 `

This doesn't happen if the request is performed directly to the server. This is because of this intercept
https://github.com/geosolutions-it/MapStore2/blob/master/web/client/libs/ajax.js#L117

### Example: 

with : 
```
params : { CQL_FILTER: undefined}
```

You will have an additional parameter in queryString `...&CQL_FILTER=` (empty). If the request is performed by axios without this intercept ,the undefined value is not used as parameter. 

*Browser Affected*
any

*Steps to reproduce*

 - No steps for testers, developers only. Found working on this issue #3142

kasongoyo:
## Description
Fix proxied ajax request with undefined and null params

## Issues
 - #3144 
 - ...

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see #3144 

**What is the new behavior?**
undefined and null params will not be part of the proxy url as the result it default to axios default behaviour on undefined and null params

 - [x] No

**Other information**:

### DIFF
--- a/web/client/libs/__tests__/ajax-test.js
+++ b/web/client/libs/__tests__/ajax-test.js
@@ -155,14 +155,17 @@ describe('Tests ajax library', () => {
                 param8: {
                     a: 'a'
                 },
-                param9: new Date()
+                param9: new Date(),
+                param10: false,
+                param11: NaN,
+                param12: 0
             }})
             .then(() => {
                 done();
             })
             .catch((ex) => {
                 const decodedUrl = urlUtil.parse(decodeURIComponent(ex.config.url), true);
-                expect(decodedUrl.query).toNotContainKeys(['param3', 'param4']);
+                expect(decodedUrl.query).toNotContainKeys(['param3', 'param4', 'param11']);
                 done();
             });
     });
--- a/web/client/libs/__tests__/ajax-test.js
+++ b/web/client/libs/__tests__/ajax-test.js
@@ -141,6 +141,32 @@ describe('Tests ajax library', () => {
             });
     });
 
+    it('ignore undefined and null query params with custom proxy', (done) => {
+        axios.get('http://fakeexternaldomain.mapstore2', {
+            proxyUrl: '/proxy/?url=',
+            params: {
+                param1: 'param1',
+                param2: '',
+                param3: undefined,
+                param4: null,
+                param5: [],
+                param6: [1, 2, "3", ''],
+                param7: {},
+                param8: {
+                    a: 'a'
+                },
+                param9: new Date()
+            }})
+            .then(() => {
+                done();
+            })
+            .catch((ex) => {
+                const decodedUrl = urlUtil.parse(decodeURIComponent(ex.config.url), true);
+                expect(decodedUrl.query).toNotContainKeys(['param3', 'param4']);
+                done();
+            });
+    });
+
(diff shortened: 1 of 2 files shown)


web/client/libs/ajax.js
```diff
@@ -122,11 +122,17 @@ axios.interceptors.request.use(config => {
             const cannotUseCORS = corsDisabled.reduce((found, current) => found || uri.indexOf(current) === 0, false);
             if (!isCORS && (!autoDetectCORS || cannotUseCORS)) {
                 const parsedUri = urlUtil.parse(uri, true, true);
+                const params = Object.keys(config.params || {}).reduce((acc, key) => {
+                    if (config.params[key] !== undefined && config.params[key] !== null) {
```
MV88: Here you can use the isNil function from lodash
`if ( !isNil(config.params[key]) ) {`

or you can change it to a shorter solution as explained [here](https://stackoverflow.com/questions/30812765/how-to-remove-undefined-and-null-values-from-an-object-using-lodash)
`var params = omitBy(config.params, isNil);`

web/client/libs/__tests__/ajax-test.js
```diff
@@ -141,6 +141,32 @@ describe('Tests ajax library', () => {
...
+                    a: 'a'
+                },
+                param9: new Date()
+            }})
+            .then(() => {
+                done();
+            })
+            .catch((ex) => {
+                const decodedUrl = urlUtil.parse(decodeURIComponent(ex.config.url), true);
+                expect(decodedUrl.query).toNotContainKeys(['param3', 'param4']);
```
MV88: maybe you can move this expectations inside the previous test
adding a comment just before like this
```
// ignore undefined and null query params
 expect(decodedUrl.query).toNotContainKeys(['param3', 'param4']);
```

web/client/libs/__tests__/ajax-test.js
```diff
@@ -141,6 +141,32 @@ describe('Tests ajax library', () => {
...
+                param3: undefined,
+                param4: null,
+                param5: [],
+                param6: [1, 2, "3", ''],
+                param7: {},
+                param8: {
+                    a: 'a'
+                },
+                param9: new Date()
+            }})
```
MV88: please, add to the test 
param10 = false 
param11 = NaN
param12 = 0

web/client/libs/__tests__/ajax-test.js
kasongoyo: Because of uniqueness of this behaviour then I thought having its own test case could aid as part of the documentation of the behaviour. What do you think? @MV88

MV88: I think it's a choice, by adding a comment and integrating the two tests could save some line of code.. but is not the end of the world

tdipisa:
@kasongoyo please, can you please backport this to the stable branch 2019.02.xx ?

PR #4167, opened 2019-09-09
Fix #3842 Switching CRS, circle annotations change size

### LINKED ISSUE #3842: Switching CRS, circle annotations change size
### Description
When switching CRS using the CRS selector, **circle annotations** (if present) change their size.

*Browser Affected*
- [x] Internet Explorer
- [x] Chrome
- [x] Firefox
- [x] Safari

*Browser Version Affected*

Last

*Steps to reproduce*

- Open a map
- Add some layer
- Add a circle annotation
- switch CRS through the CRS selector

*Expected Result*

The circle annotation doesn't change size.

*Current Result*

The circle annotation changes size.

vlt1:
## Description
Circles are no longer transformed from Polygon to Circle to allow correct rendering
under projection changes

## Issues
 - #3842 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

Circle annotations don't change size upon switching projections

**What is the new behavior?**
Circle annotations transform correctly with projection changes

 - [x] No

### DIFF
--- a/web/client/components/map/openlayers/Feature.jsx
+++ b/web/client/components/map/openlayers/Feature.jsx
@@ -14,7 +14,6 @@ import find from 'lodash/find';
 import castArray from 'lodash/castArray';
 
 import { parseStyles } from './VectorStyle';
-import { transformPolygonToCircle } from '../../../utils/openlayers/DrawSupportUtils';
 import { createStylesAsync } from '../../../utils/VectorStyleUtils';
 import GeoJSON from 'ol/format/GeoJSON';
 
--- a/web/client/components/map/openlayers/Feature.jsx
+++ b/web/client/components/map/openlayers/Feature.jsx
@@ -85,15 +85,7 @@ export default class Feature extends React.Component {
                     // reproject features from featureCrs
                     dataProjection: props.featuresCrs
                 });
-            this._feature.map(f => {
-                let newF = f;
-                if (f.getProperties().isCircle) {
-                    newF = transformPolygonToCircle(f, props.crs || 'EPSG:3857', props.featuresCrs);
-                    newF.setGeometry(newF.getGeometry().transform(props.crs || 'EPSG:3857', props.featuresCrs));
-                }
-                return newF;
-            }).forEach(
-                (f) => f.getGeometry().transform(props.featuresCrs, props.crs || 'EPSG:3857'));
+            this._feature.forEach((f) => f.getGeometry().transform(props.featuresCrs, props.crs || 'EPSG:3857'));
 
             if (props.style && (props.style !== props.layerStyle)) {
                 this._feature.forEach((f) => {


mbarto:
I suppose circles were transformed to a polygon for a reason. Maybe @MV88 can explain us why and if this fix is reasonable or not?

MV88:
> I suppose circles were transformed to a polygon for a reason. Maybe @MV88 can explain us why and if this fix is reasonable or not?

It's the contrary, but the confusion is around the management of circles.

Polygons that are circles in reality (when drawn as circles) should be rendered as circles (ol has the Circle type of geometry). they are transformed and passed to the state as polygons in order to perform spatial search or GFI operations, but when rendered they should be circles.

@vlt1 have you tested it by playing with Query form spatial filter and annotations? i suppose that if they are not transformed back to circle they would be rendered as polygons and by zooming in you should start seeing it that is not a circle, but a polygon.

vlt1:
What should happen to circles when projection changes?

mbarto:
Looking at the animated gif in the original issue, I think the problem exists when you switch between a meters based projection (EPSG:3857) and a degrees based one (EPSG:4326). Maybe  the problem is that radius is not converted between meters and degrees?
From openlayers documentation I see that radius should be expressed in the unit of measure of the current projection, so we should probably also convert radius when we reproject circles.

vlt1:
That is exactly what was implemented, which is why the circle doesn't change it's size going from one projection to another. The fact of the matter is, when projection changes, what was a circle in one projection would become an ellipse in another one. So the question is, what should be the behaviour? Should an ellipse be drawn? Or should it be converted to circle in some way? If an ellipse should be drawn, how to do it? OpenLayers doesn't seem to be able to draw ellipses.

mbarto:
I think a circle is ok, no need to draw an ellipse. @MV88 what do we do in other circle use cases?

MV88:
> I think a circle is ok, no need to draw an ellipse. @MV88 what do we do in other circle use cases?

i think there are two problems related to circles and projections:
1. when che crs changes
1. when the crs does not change

the solution of course depends on the desired behaviour (@tdipisa we may need you here)
i think we need to go back a step and analyze a bit all the scenarios:

scenario: **circle from 3857 to 4326**
- i have drawn a circle which in the state is a polygon but is rendered as circle
- i switch to 4326, it should change to ellipse (in this case it will be polygon) because it should reflect original points which are used for GFI and spatial operations
- switching back it should be rendered again as circle

scenario: **circle from 4326 to 3857**
i think here is the same thing as the previous scenario

scenario: **circle drawn in 3857, no switch**
- i have drawn a circle as annotation
- i save it
- i want to see Circles rendered, not polygons.

@mbarto @offtherailz @tdipisa correct me if i'm wrong

Issue #4178, opened 2019-09-11
Add support for saving/loading geostory on external backend (geostore)


MV88:
### Description
Save and load of geostory

### Acceptance Criteria

- Geostory there should be the possibility to save geostory in geostore
  - Save (update current story)
  - Save As (create a new story)
- Also loading of geostories is required (the same way of maps)

the following can be moved to another issue:
- story properties like name and description and thumbnail should be changed
- stories should be managed in homepage (featured, content tab)

PR #4197, opened 2019-09-13
#4178 Save GeoStory 

offtherailz:
## Description
Save and SaveAs button for GeoStory 
Peek 13-09-2019 18-06

A few sentences describing the overall goals of the pull request' s commits.

## Issues
 - See issue #4178

 - [x] Feature

- [x] No


web/client/epics/feedbackMask.js
```diff
@@ -81,6 +83,25 @@ const updateDashboardVisibility = action$ =>
                     .takeUntil(action$.ofType(DETECTED_NEW_PAGE))
             );
         });
+/**
+ * Enabled/disabled mask based on geostory load feedback, in case of error enable feedbackMask.
+ * @param {Observable} action$ stream of actions. Manages `LOAD_GEOSTORY, `GEOSTORY_LOADED`, `GEOSTORY_LOAD_ERROR`, `LOGIN_SUCCESS`, `LOGOUT`, `LOCATION_CHANGE`
+ * @memberof epics.geostory
```
MV88:
```suggestion
 * @memberof epics.feedbackMask
```

web/client/epics/geostory.js
```diff
@@ -105,7 +124,37 @@ export const localizeTemplateEpic = (action$, store) =>
                 )
             );
     });
-
+/**
+ * Epic that handles the save story workflow. It uses persistance
```
MV88:
```suggestion
 * Epic that handles the save story workflow. It uses persistence
```

web/client/epics/geostory.js
```diff
@@ -105,7 +124,37 @@ export const localizeTemplateEpic = (action$, store) =>
...
+ * @param {Observable} action$ stream of redux action
+ */
+export const saveGeoStoryResource = action$ => action$
+        .ofType(SAVE)
+        // delay is for speed up tests, not part of the SAVE action
+        .exhaustMap(({ resource, delay = 1000 } = {}) =>
```
MV88: can you explain a bit this part?

web/client/epics/geostory.js
```diff
@@ -140,7 +189,7 @@ export const editMediaForBackgroundEpic = (action$, store) =>
             const state = store.getState();
             const resourceId = resourceIdSelectorCreator(path)(state);
             return Observable.of(
-                    show(owner), // open mediaEditor
+                    showMediaEditor(owner), // open mediaEditor
```
MV88:
```suggestion
                    showMediaEditor(owner),
```

web/client/epics/geostory.js
```diff
@@ -166,20 +215,26 @@ export const loadGeostoryEpic = (action$, {getState = () => {}}) => action$
...
+                        .then(({data}) => ({data}));
                 }
-                // TODO manage load process from external api
-                return axios.get(`configs/sampleStory.json`);
+                return getResource(id);
             })
-            .map(({ data }) => {
-                if (isObject(data)) {
-                    return setCurrentStory(data);
+            .filter(({data}) => {
```
MV88: Can you explain why the filter if you throw the error?  to avoid to launch following actions, but the throw should "jump" the flow tot the catch method, right?

web/client/epics/geostory.js
```diff
@@ -208,6 +263,17 @@ export const loadGeostoryEpic = (action$, {getState = () => {}}) => action$
                 }
             ))
         );
+/**
+ * Triggers reload of last loaded story when user login-logout
+ * @param {Observable} action$ the stream of redux actions
+ */
+export const reloadGeoStoryOnLoginLogout = (action$) =>
```
MV88: you do the reload for permissions checks?
and for rendering feedback mask ?

web/client/reducers/geostory.js
```diff
@@ -189,6 +205,24 @@ export default (state = INITIAL_STATE, action) => {
...
+            const {control, value} = action;
+            return set(`controls.${control}`, value, state);
+        }
+        /**
+         * **NOTE** this is the resource that contains the whole story (e.g. GeoStore).
+         * It contains permissions and so on. Don't confuse with story resources (media).
+         */
+        case SET_RESOURCE: {
+            const { resource } = action;
+            return set(`resource`, resource, state);
```
MV88: what about using storyResource,  backendResource, externalResource?

offtherailz: I wanted to follow the same name convention of dashboard. I documented this name conflict but I don't think is a problem by itself. The two resources concepts are very different, and the signature is different too.

The only problem may concern the action names, but I think is more correct to rename the story resources (media) actions instead.

I should postpone this when working more on story resources. What do you think?

web/client/epics/geostory.js
offtherailz: The story reload should trigger automatically the permission and feedback mask actions (so in case of missing permission, show do the correct actions to show the feedkback mask, prompt the login and so on). 

Did you tried it and you found it doesn't work in some case? If so, please notify me the bug. I'll fix it

web/client/epics/geostory.js
offtherailz: Yes, maybe a `do` method was more appropriate for this. I wanted a method that didn't affect the result, just for check.
I'll replace it with do
```suggestion
            .do(({data}) => {
```

PR #4246, opened 2019-09-24
Fix #4203 Widgets on protected data are not correctly reloaded

### LINKED ISSUE #4203: Widgets on protected data are not correctly reloaded
### Description
If a map contains a protected layer, widgets related to this layer are not correctly reloaded when an user with required rights (as ADMIN) login into MapStore.

states_permissions

*Browser Affected*
- [x] Internet Explorer
- [x] Chrome
- [x] Firefox
- [x] Safari

*Steps to reproduce*

- Open [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10828) map as anonymus user 
- Login as admin

*Expected Result*

- Widgets are updated according to the user rights

*Current Result*

- Widgets are updated only after a pan in the map

To fix this probably the user integration with GeoServer needs to be set

vlt1:
## Description
Added an epic to update widgets layer prop on LAYER_LOAD and various modifications
to ensure correct reloading of widgets.

## Issues
 - #4203 
 - ...

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

#4203 

**What is the new behavior?**
Widgets are reloaded correctly

 - [x] No


web/client/components/widgets/enhancers/dependenciesToFilter.js
```diff
@@ -29,6 +29,7 @@ module.exports = compose(
             mapSync !== nextProps.mapSync
             || dependencies.viewport !== (nextProps.dependencies && nextProps.dependencies.viewport)
             || geomProp !== nextProps.geomProp
+            || (nextProps.layer && layer.loadingError !== nextProps.layer.loadingError)
```
offtherailz: Can you add a test for mapSync change on loadingError change ?

web/client/epics/widgets.js
```diff
@@ -176,5 +176,19 @@ module.exports = {
...
+            }),
+
+    /**
+     * Triggers updates of the layer property of widgets on LAYER_LOAD
+     * @memberof epics.widgets
+     * @param {external:Observable} action$ manages `LAYER_LOAD`
+     * @return {external:Observable}
+     */
+    updateLayerOnLoad: (action$, store) =>
+        action$.ofType(LAYER_LOAD)
```
offtherailz: Layer load is called many times (one for each tile loaded, if I remember well). The only update needed is when error changed. Is there a way to avoid unnecessary actions flood ? 
I think a distinctUntilKeyChanged on error field could to the job. Please add this improvement and test.

vlt1: Sounds like a good idea.

web/client/components/widgets/enhancers/dependenciesToFilter.js
vlt1: Will do.

web/client/epics/widgets.js
vlt1: Actually the update is needed when error property is changed for a particular layerId, so distinctUntilKeyChanged would have to be both on layerId and error fields. Would that be particularly useful?

web/client/components/widgets/enhancers/dependenciesToFilter.js
vlt1: Wait, I'm not sure what you mean. mapSync change? Why would mapSync change?

offtherailz: Sorry I meant dependenciesToFilter

web/client/epics/widgets.js
offtherailz: note, i updated the example with take(1) after the first edit

offtherailz: Edited again. What I'm trying to do is to create a stream for each layerid that errored waiting for load (maybe other checks) to trigger once the updateLayertWidget. Did you get my point?

Anyway this is an alternative solution. The problem is to avoid unnecessary widget's update, even if they don't cause, as far as I can see, reloads or other kind of problems.

vlt1:
The update action will be performed anytime LAYER_ERROR or LAYER_LOAD action changes the value of a loadingError(which is every single time they are invoked), because in that case previousLoadingError !== loadingError. Also from the code it is evident , that LAYER_LOAD can both set and clear the error field, just as LAYER_ERROR, so I'm not sure that the last approach you've suggested would work. They also set them in different ways. Good point about switchMap though, I should probably groupBy by layerId and then switchMap on inner observables, as for a particular layer we care only about the latest change.
```javascript
action$.ofType(LAYER_LOAD, LAYER_ERROR)
    .groupBy(({layerId}) => layerId)
    .map(layerStream$ => layerStream$ 
        .switchMap(({layerId}) => {
            const flatLayer = getLayerFromId(state, layerId);
            return Rx.Observable.of(
                ...(flatLayer && flatLayer.previousLoadingError !== flatLayer.loadingError ?
                    [updateWidgetLayer(flatLayer)] :
                    [])
            );
        })
    ).mergeAll()
```

All in all I think that it is just safer to update widgets when loadingError changes, and it doesn't depend on any particular logic of setting that field. Maybe I'm missing something? I don't know. Sorry :(

offtherailz:
Yes, I checked and the `lastLayerError` goes away as well as you pan. I didn't want to go into the condition when lastLayerError remains "Error" and the other error flag becomes again "Error" so the update is not triggered. 

layer_load

Yes, the solution is not optimal but acceptable. My proposed solution didn't save any other error in the state. 
I'd like to mutate layer objects less as possible, because they are read by many other tools, and their continuous mutation could slow down the whole application. 

Let's discuss it a little.

offtherailz:
From my test seems to work, so please correct the switchMap issue and finalize the PR

tdipisa:
@vlt1 please, can you backport this to the stable branch 2019.02.01?

Issue #4216, opened 2019-09-17
Manage loading of maps from geostore in Media Editor


MV88:
### Description
- loading of maps is required in map section of media editor

- in this phase add support for generic persistence api adding browsing

### Acceptance Criteria
in the map section of media editor:
- a list of maps is loaded from geostore
- it should support pagination
- should be similar to the one present in widgets wizard

### Other useful information (optional):

part of https://github.com/geosolutions-it/MapStore2-C039/issues/28

PR #4258, opened 2019-09-26
4216 media maps

MV88:
## Description
This pr add possibility to add maps to the background via the media editor
some additions:
- media editor has a service menu to select which source needs to be used
- media editor has a map preview
- video tab temprary hidden
- there is a first implementation of map component

## Issues
 - #4216

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

**What is the new behavior?**

 - [x] No

**Other information**:
@offtherailz 
- Needs some more tests on epic and reducers, if current implementation is fine i'll add them
- there is some problem managin map ids, for example if you add a map and the reopen the media editor the ol-viewport is not visible because it has been moved away. 

@allyoucanmap probably some style changes are required in the media editor


web/client/reducers/mediaEditor.js
```diff
@@ -100,6 +102,13 @@ export default (state = DEFAULT_STATE, action) => {
         case SELECT_ITEM: {
             return set('selected', action.id, state);
         }
+        case SELECT_MAP: {
```
offtherailz: You should use loadMediaSuccess instead, storing params: "textSearch" and "resultData" with the items results. You can store only the map selected or all the maps from items. The second is more compatible. 
You can do it by inserting an ehancer that calls loadMediaSuccess when items change. 

Avoid to add specific type method when working with generic resources, in general.

web/client/actions/mediaEditor.js
```diff
@@ -18,8 +18,20 @@ export const SAVE_MEDIA_SUCCESS = "MEDIA_EDITOR:SAVE_MEDIA_SUCCESS";
 export const SET_MEDIA_TYPE = "MEDIA_EDITOR:SET_MEDIA_TYPE";
 export const SET_MEDIA_SERVICE = "MEDIA_EDITOR:SET_MEDIA_SERVICE";
 export const SELECT_ITEM = "MEDIA_EDITOR:SELECT_ITEM";
+export const SELECT_MAP = "MEDIA_EDITOR:SELECT_MAP";
 export const SHOW = "MEDIA_EDITOR:SHOW";
 
+// RESOURCE FORMAT DRAFT :
```
offtherailz:
```suggestion
// RESOURCE FORMAT :
```

web/client/actions/mediaEditor.js
```diff
@@ -18,8 +18,20 @@ export const SAVE_MEDIA_SUCCESS = "MEDIA_EDITOR:SAVE_MEDIA_SUCCESS";
...
+/*
+{
+    type: 'image'|'video'|'map'|'iframe'|'document' // (pdf)
+    source: 'id' // id of the source, just to identify it in a local context
```
offtherailz:
```suggestion
    source: 'id', // id of the source, just to identify it in a local context
    // other meta-information about the resource
```

web/client/reducers/mediaEditor.js
MV88: textSearch is not needed imo, there is a withFilere enhancers that manages it.

MV88: by using the second options, there is something to changes, for example actually the GetData request is performed when selecting an item, not before

web/client/actions/mediaEditor.js
```diff
@@ -66,26 +78,21 @@ export const saveMediaSuccess = ({ mediaType, source, data, id }) => ({ type: SA
...
-    data: {
-        //specific data for the source type
-    }
-}
-*/
+/**
+ * update item in media editor list
+ * @param {object} param.item
+ */
+export const updateItem = ({map}) => ({ type: UPDATE_ITEM, item: map});
```
offtherailz: this map object is unnecessary in the item
```suggestion
export const updateItem = (item => ({ type: UPDATE_ITEM, item});
```

web/client/reducers/__tests__/mediaEditor-test.js
```diff
@@ -113,4 +113,29 @@ describe('Test the mediaEditor reducer', () => {
...
+                    geostoreMap: {
+                        resultData: {
+                            resources: [{
+                                id: "resId"
+                            }]
+                        }
+                    }
+                }
+            }
+        }, updateItem({map}));
```
offtherailz: to be coherent with the action interface
```suggestion
        }, updateItem(map));
```

web/client/components/mediaEditor/map/MapList.jsx
```diff
@@ -0,0 +1,107 @@
...
+                            }
+                        });
+                    }
+                })
+                .ignoreElements() // don't want to emit props
+    )))(MapCatalogComp);
+const defaultPreview = <Icon glyph="geoserver" padding={20} />;
+
+
+const MapList = handleSelectEnhancer(
```
offtherailz: handleSelectEnhancer makes call onMapSelected this way
```
onMapSelected({map})
```
While you need to call updateItem(item);

add here an enhancer that transforms the handler (that is an adapter) instead of modelling all the other actions to match existing code.
Something like 
```javascript
withHandlers ({
    onMapSelected: ({onMapSelected = () => {}}) => ({map}) {
    // transform into a standard mediaEditor resource, if you need (I don't remember)
    onMapSelected(map);
};
})
```

web/client/components/misc/enhancers/withFilter.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import {
+    compose,
+    withState
+} from 'recompose';
+
+const withFilter = compose(
```
offtherailz: This is not a global enhancer but proper of the mediaEditor, please move this in the correct directory and document a little bit. (something like `holds the state of the text filter in the media editor`)

MV88: Why is not global? It can be applied to any filter anywhere, but ok.

web/client/components/mediaEditor/map/MapList.jsx
MV88: mmh, i think handle select should have generic names
when using it we should add an adapter that renames specific names into the generic one.
would this be good for you?

this would requrie a change to the components that are using this enhancer 
two actually: MapSelector (Widgets) and MapList (MediaEditor)

web/client/components/geostory/media/Map.jsx
```diff
@@ -34,7 +34,7 @@ export default compose(
     options = {},
     fit
 }) => {
-    const { layers = [], mapOptions, ...m} = map; // remove mapOptions to not override options
+    const { layers = [], mapOptions, ...m} = (map.data ? map.data : map); // remove mapOptions to not override options
```
offtherailz: maybe using only data is better

web/client/components/mediaEditor/map/MapPreview.jsx
```diff
@@ -15,7 +15,7 @@ import { isEmpty } from 'lodash';
 const Preview = ({
     selectedItem
 }) => {
-    const { layers = [], mapOptions, ...m} = selectedItem; // remove mapOptions to not override options
+    const { layers = [], mapOptions, ...m} = selectedItem.data ? selectedItem.data : selectedItem; // remove mapOptions to not override options
```
offtherailz: maybe using only data is better

web/client/components/geostory/media/Map.jsx
MV88: just map all data are present there

MV88:
closed in favor of #4278

PR #4288, opened 2019-10-03
Attempt to fix react grid layout build issues

offtherailz:
Couldn't reproduce locally the issue notified by travis.

By checking the log, this PR trie to fix the build by loading react-grid-layout with babel loader.


offtherailz:
From investigation I've found that, if package-lock.json is present, travis uses `npm ci` instead of `npm install`.

See doc [here](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/)
image

This means that the version used is the one in package-lock.json

Updating to a newer version of node, the problem of package-lock out of sync is revealed.

What should we do? We wanted to remove the package-lock.json, this could solve the problem.

Issue #4334, opened 2019-10-10
Catalog loading spinner changed style and dimension


francoperversi:
### Description
Loading spinner after react update:

![15](https://user-images.githubusercontent.com/54808782/66585063-7d30b580-eb86-11e9-9948-79f46a4843d2.png)

Loading spinner before react update:

![13](https://user-images.githubusercontent.com/54808782/66585333-ee706880-eb86-11e9-93d8-59adae645d57.png)

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [ ] Internet Explorer
- [X] Chrome
- [X] Firefox
- [X] Safari

*Browser Version Affected*

- Not browser related

*Steps to reproduce*

- Open mapstore
- Open a random map
- Open burger menu
- Open catalog

*Expected Result*

- The loading spinner is the same

*Current Result*

- The loading spinner is different

### Other useful information (optional):
See also #4075

PR #4362, opened 2019-10-15
Fix 4334 Catalog loading spinner changed style and dimension

kasongoyo:
## Description
Fix style of catalog loading spinner

## Issues
 - #4334 
 - ...

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see #4334 

**What is the new behavior?**
The loading spinner look restored

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/misc/Loader.jsx
+++ b/web/client/components/misc/Loader.jsx
@@ -6,13 +6,24 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+const React = require('react');
+const PropTypes = require('prop-types');
+
 const sizeToClass = size => size > 100
     ? 'full'
     : size > 40
         ? 'medium'
         : 'small';
-const React = require('react');
-module.exports = ({ size, style = {}, className}) => (<div className={className}
-    style={{ width: `${size}px`, height: `${size}px`, overflow: "hidden", ...style }}>
+
+const Loader = ({ size, style = {}, className}) => (<div className={className}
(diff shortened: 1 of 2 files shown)


web/client/components/misc/Loader.jsx
```diff
@@ -13,6 +13,6 @@ const sizeToClass = size => size > 100
         : 'small';
 const React = require('react');
 module.exports = ({ size, style = {}, className}) => (<div className={className}
-    style={{ width: size, height: size, overflow: "hidden", ...style }}>
+    style={{ width: `${size}px`, height: `${size}px`, overflow: "hidden", ...style }}>
```
allyoucanmap: Note: I removed this fix to test this.

I was checking how this fixed the loader issue and I noticed this in [Catalog.jsx](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/catalog/Catalog.jsx#L191) component:

loader use prop size as a string

```jsx
... <Loader size="176" /> ...
```

this probably is applied as is in the new react so it creates a invalid css style (eg: `width: 176;` )

While if we use a number for size prop:
```jsx
... <Loader size={176} /> ...
```

this will be converted to a string in the style with the correct unit of measure (eg: `width: 176px;` ).

I think we should change the prop in Catalog loader instead to change this component.

MV88: in addition to @allyoucanmap  i would also add propType validation for the loader comp

Issue #4447, opened 2019-11-06
Story Edit Button, visible for anonymous users


MV88:
### Description
edit button in view mode is visible for anonymous user while it should not
only user that as canEdit permission can see it

### In case of Bug (otherwise remove this paragraph)

- [x] not browser related 

*Steps to reproduce*

- open [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/11555) story that has everyone group permission permission

*Expected Result*

- to not see the edit button

*Current Result*

- you can switch to edit mode via edit button

### Other useful information (optional):
see https://github.com/geosolutions-it/MapStore2-C039/issues/115

Issue #4407, opened 2019-10-25
Builder - Global Tools - General settings - Basis & Navigation 


MV88:
### Description
A story needs a tool to configure general base settings in terms of:

- [ ] Story logo
- [ ] Story title
- [ ] Navigation

![general_settings](https://user-images.githubusercontent.com/1280027/66139192-43006a80-e600-11e9-88b8-525543e7837b.jpg)

## Title

The owner of the story can decide if the Title of the story must appear or not within the story itself. In that case a banner will be visible in view mode-only with the title on the left.

## Logo

The owner of the story can decide if a logo for the story must be visible or not. In that case a banner will be visible in view mode-only with the selected logo on the left. An edit button allows to open the media editor to select an image to use for the logo (by default no logo is set).

![general_settings2](https://user-images.githubusercontent.com/1280027/66140677-d3d84580-e602-11e9-9cbf-0fb7ffead116.jpg)

If Logo and/or Title are present, a top bar appears in view mode of the story to show them in the following order.
 
![bar](https://user-images.githubusercontent.com/1280027/66141002-54974180-e603-11e9-9431-d6651dd5a78b.jpg)

## Navigation
It must be possible to setup the navigation toolbar of the story:

- Enable/disable the navigation toolbar
- If enabled, configure which sections must be included: sections need to be in a list so the user can select the ones to include. the list appear only if the navigation toolbar is enabled

![nav_settings](https://user-images.githubusercontent.com/1280027/66201527-d9d23300-e6a3-11e9-8b62-51233fe86e01.jpg)

a list with checkboxes will be shown:
- all sections but immersive ones will be selectable, (only at section level

for immersive section:
- the section title will appear but without selection
- user will be able to select only Immersive Content (columns) only, not its contents

### Other useful information (optional):
see https://github.com/geosolutions-it/MapStore2-C039/issues/88
see https://github.com/geosolutions-it/MapStore2-C039/issues/89

PR #4477, opened 2019-11-11
Fix 4407 add story settings and story edit button for anonymous users

MV88:
## Description
This pr adds the settings panel to the story.

it will contain the following options:
- possibility to show a title in the navbar, default story title is used as default value if available
- possibility to show a logo in the navbar
- possibility to customize the navbar item list with a tree checkbox (this will not affect story content visibility)

**Note:** progress bar will be based on the story content

## Issues
 - #4407
 - #4447

 - [x] Feature

you cannot setup story settings

**What is the new behavior?**
you can

 - [x] No

**Other information**:

this is the new look and feel
image

Some points after meeting with @tdipisa 

- [x] Settings - Fix "back" tooltip in settings
- [x] Settings - Fix message in unsaved changes modal for settings
- [x] Settings - when opening the expanded nodes are the one with at least one immersive content checked
- [x] Settings - change message for thumbnail to 32px x 32px
- [x] Settings - change the label Navbar to Navigation bar
- [x] View mode - the title should not go beyond 50% of space
- [x] View mode - We agreed to have the navigation bar to the left, but also the edit tool should be moved to the left as first item, add the needed space between tools
- [x] View mode - make the height of navbar at least as the other bar (omnibar) around 53 px
- [x] View mode - add more space between logo and title  (maybe 20px)

- [x] If you add only the nav (without title) there is a  little misalignment: 
image
I suggest to review a little bit the style because also the pencil and the menu are not in line. I suggest also to move the pencil in a safest place


build/buildConfig.js
```diff
@@ -212,5 +212,5 @@ module.exports = (bundles, themeEntries, paths, extractThemesPlugin, prod, publi
         }
     },
 
-    devtool: !prod ? 'eval' : undefined
+    devtool: !prod ? 'cheap-eval-source-map' : undefined
```
offtherailz: don't commit this
```suggestion
    devtool: !prod ? 'eval' : undefined
```

web/client/reducers/geostory.js
```diff
@@ -208,6 +226,26 @@ export default (state = INITIAL_STATE, action) => {
...
         return set('isCollapsed', !state.isCollapsed, state);
     }
+    case TOGGLE_SETTINGS: {
+        const visibility = get(state, `currentStory.settings.${action.option}`);
+        return set(`currentStory.settings.${action.option}`, !visibility, state);
+    }
+    case TOGGLE_SETTINGS_PANEL: {
+
+        // when opening, the expanded items will be calculated based on checked ones
+        const newStatus = !state.isSettingsEnabled;
```
offtherailz: You shouldn't use selectors in reducers ( I see `selectedCardSelector` is already used, we should check if we can remove that too ) because: 
 - they may cache results and have unwanted side-effects
 - actually they are written at application scope (geostory.<state>) while the reducer doesn't know it's name. We should evaluate before to do this.
Expanded is a derived information from selected. therefore please create instead a selector outside the reducer to derive expanded from checked instead of saving this duplicated information in the state.

web/client/actions/geostory.js
```diff
@@ -134,6 +137,14 @@ export const setCurrentStory = (story) => ({ type: SET_CURRENT_STORY, story});
  * Turn on/off preview in cards.
 */
 export const toggleCardPreview = () => ({ type: TOGGLE_CARD_PREVIEW});
+/**
+ * Turn on/off settings options in panel, like logo etc.
+*/
+export const toggleSettings = (option) => ({ type: TOGGLE_SETTINGS, option});
```
offtherailz: it should be singular (it toggles a single setting)

web/client/actions/geostory.js
```diff
@@ -193,3 +202,10 @@ export const setFocusOnContent = (status, target, selector, hideContent, path) =
     hideContent,
     path
 });
+
+/**
+ * update settings for the story
+ * @param {string} prop prop name to update in the state
+ * @param {*} value value used to update the prop
+ */
+export const updateSettings = (prop, value) => ({type: UPDATE_SETTINGS, prop, value});
```
offtherailz: updateSetting (because the function updates a single setting

web/client/components/geostory/contents/Column.jsx
```diff
@@ -27,6 +27,7 @@ export default ({
     viewHeight,
     contents = [],
     mode,
+    updateCurrentColumn = () => {},
```
offtherailz: This doesn't exist, please remove
```suggestion
```

web/client/components/geostory/contents/Column.jsx
```diff
@@ -39,6 +40,7 @@ export default ({
         mode={mode}
         add={add}
         editMedia={editMedia}
+        updateCurrentColumn={updateCurrentColumn}
```
offtherailz:
```suggestion
```

web/client/plugins/GeoStory.jsx
```diff
@@ -14,7 +14,7 @@ import {compose, withPropsOnChange, withHandlers} from 'recompose';
 import {createPlugin} from '../utils/PluginsUtils';
 import { Modes } from '../utils/GeoStoryUtils';
 import { getMessageById } from '../utils/LocaleUtils';
-import { add, update, updateCurrentPage, remove } from '../actions/geostory';
+import { add, update, updateCurrentPage, updateCurrentColumn, remove } from '../actions/geostory';
```
offtherailz:
```suggestion
import { add, update, updateCurrentPage, remove } from '../actions/geostory';
```

web/client/plugins/GeoStory.jsx
```diff
@@ -56,6 +56,7 @@ export default createPlugin("GeoStory", {
                 add,
                 update,
                 updateCurrentPage,
+                updateCurrentColumn,
```
offtherailz:
```suggestion
```

web/client/components/geostory/contents/Contents.jsx
```diff
@@ -60,6 +61,7 @@ export default ({
                     viewWidth={viewWidth}
                     sectionType={sectionType}
                     viewHeight={viewHeight}
+                    updateCurrentColumn={updateCurrentColumn}
```
offtherailz:
```suggestion
```

web/client/reducers/geostory.js
```diff
@@ -125,6 +127,15 @@ const getIndexToInsert = (array, position) => {
  * }
  *
  */
+let INITIAL_STATE = {
```
offtherailz: moving the initial state here you associate the jsdoc of the reducer to the INITIAL_STATE... but the comment is related to the reducer.  Please move this in the original position.

web/client/reducers/geostory.js
```diff
@@ -218,7 +256,11 @@ export default (state = INITIAL_STATE, action) => {
         }
         return set(path, newElement, state);
     }
+    case UPDATE_SETTINGS: {
+        return set(`currentStory.settings.${action.prop}`, action.value, state);
+    }
     case UPDATE_CURRENT_PAGE: {
+        /** if column update state only if the column for the current immersive section has changed */
```
offtherailz:
```suggestion
        /* if the page update updates a column, update state only if the column for the current immersive section has changed.
        * This to avoid to select a column that doesn't belong to the current section (it could happen due to scroll events, in case of multiple immersive sections in the same viewport).
        */
```
This is a comment, not documentation

web/client/reducers/geostory.js
MV88: - i will remove the selectors form reducers
- expanded, i would suggest to edit the settingsSelector to add expanded info, because actually they are present in settings

web/client/actions/geostory.js
MV88: ok

web/client/reducers/geostory.js
MV88: i just wanted to have the initial state near to the reducer to avoid scrolling up..

offtherailz: I understand the purpose... Consider the documentation as a part of the function.

Issue #4299, opened 2019-10-04
Map used in geostory throws error if the projection is not supported


MV88:
### Description
When media editor for maps is available try to load terrestris map which has a custom projection (EPSG:31468)

since its proj definition is missing a red error page appears

Error is because it never runs [this](https://github.com/geosolutions-it/MapStore2/blob/3cef0ed82658babaab793e067e66dc38665462ac/web/client/components/map/openlayers/Map.jsx#L91) line

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [ ] Internet Explorer
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

- Indicate the browser version in which the issue has been found

*Steps to reproduce*

- load a story (newgeostory)
- open media editor
- select map
- change service to geostore DEV
- select terrestris map
- click apply

*Expected Result*

- everything works

*Current Result*

- error

### Other useful information (optional):

related to https://github.com/geosolutions-it/MapStore2-C039/issues/94

I think we should handle projection dynamically, @mbarto @offtherailz do you agree?

```
Map.jsx?31cc:314 Uncaught TypeError: Cannot read property 'degrees' of undefined
    at OpenlayersMap.eval [as getResolutions] (Map.jsx?31cc:314)
    at OpenlayersMap.UNSAFE_componentWillReceiveProps (Map.jsx?31cc:257)
    at callComponentWillReceiveProps (react-dom.development.js?61bb:14333)
    at updateClassInstance (react-dom.development.js?61bb:14546)
    at updateClassComponent (react-dom.development.js?61bb:18357)
    at beginWork$1 (react-dom.development.js?61bb:20108)
    at HTMLUnknownElement.callCallback (react-dom.development.js?61bb:362)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?61bb:411)
    at invokeGuardedCallback (react-dom.development.js?61bb:466)
    at beginWork$$1 (react-dom.development.js?61bb:25730)
    at performUnitOfWork (react-dom.development.js?61bb:24638)
    at workLoopSync (react-dom.development.js?61bb:24614)
    at performSyncWorkOnRoot (react-dom.development.js?61bb:24182)
    at eval (react-dom.development.js?61bb:12238)
    at unstable_runWithPriority (scheduler.development.js?3069:815)
[...]

PR #4479, opened 2019-11-11
Fix 4299 custom crs error

MV88:
## Description
The problem when a map is configured to use a custom projection not supported by default has been fixed.

[EDIT] Actually it prevents to render any map if the projection is not supported

For maps in the viewer it show a custom error for the projection (try to load [this](https://dev.mapstore.geo-solutions.it/mapstore/#/viewer/openlayers/10858) map )

## Issues
 - #4299

 - [x] Bugfix

red page error when selecting stories with map terrestis, dashboard with map terrestris  or the map terrestris itself which has EPSG:31468 as configured projection

**What is the new behavior?**
if in the projectionDefs does not contain EPSG:31468 then a mask appear
image
image

 - [x] No

**Other information**:

### DIFF
--- a/web/client/utils/openlayers/projUtils.js
+++ b/web/client/utils/openlayers/projUtils.js
@@ -7,19 +7,31 @@
  */
 
 import {addProjection, Projection} from 'ol/proj';
-export default {
+import ConfigUtils from '../ConfigUtils';
 
-    /**
-    * function needed in openlayer for adding new projection
-    */
-    addProjections: function(code, extent, worldExtent, axisOrientation, units) {
-        addProjection(new Projection({
-            code,
-            extent,
-            worldExtent,
-            axisOrientation,
-            units
-        })
-        );
-    }
+/**
+ * function needed in openlayers for adding new projection
+ */
+export const addProjections = function(code, extent, worldExtent, axisOrientation, units) {
+    addProjection(new Projection({
+        code,
+        extent,
+        worldExtent,
+        axisOrientation,
+        units
+    })
+    );
+};
+/**
+ * @returns {string} the default projection EPSG:3857 if no custom projectionDefs are defined
+ */
+export const fallbackToSupportedProjection = (projectionDefs = ConfigUtils.getConfigProp("projectionDefs") || [], projection) => {
+    const codes = (projectionDefs.length && projectionDefs.map(({code})  => code) || []).concat(["EPSG:4326", "EPSG:3857", "EPSG:900913"]);
+    return codes.filter(c => c === projection).length ? projection : "EPSG:3857";
+};
+
+
+export default {
+    addProjections,
+    fallbackToSupportedProjection
 };
--- a/web/client/components/errors/ResourceUnavailable.jsx
+++ b/web/client/components/errors/ResourceUnavailable.jsx
@@ -31,14 +31,14 @@ const HTML = require('../I18N/HTML');
 
 const ResourceUnavailable = emptyState(
     ({enabled, login, status, alwaysVisible}) => enabled && alwaysVisible || enabled && login || enabled && status !== 403,
-    ({status, mode = 'map', glyphs = {map: '1-map', dashboard: 'dashboard'}, errorMessage, showHomeButton, homeButton}) => ({
+    ({status, mode = 'map', glyphs = {map: '1-map', dashboard: 'dashboard'}, errorMessage, errorMessageParams, showHomeButton, homeButton}) => ({
         glyph: glyphs[mode] || '1-map',
         title: status === 403 && <Message msgId={`${mode}.errors.loading.notAccessible`} />
         || status === 404 && <Message msgId={`${mode}.errors.loading.notFound`} />
         || <Message msgId={`${mode}.errors.loading.title`} />,
         description: (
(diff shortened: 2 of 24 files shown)


web/client/utils/openlayers/projUtils.js
```diff
@@ -7,19 +7,29 @@
...
-            extent,
-            worldExtent,
-            axisOrientation,
-            units
-        })
-        );
-    }
+*/
+export const addProjections = function(code, extent, worldExtent, axisOrientation) {
+    addProjection(new Projection({
```
mbarto: You forgot units, I just added some days ago

MV88: good catch, i lost it when resolving the conflict

offtherailz:
This PR, can you please solve them?

MV88:
> Almost there
> 
> * [ ]  I'd like all things like this:
> 
> ```
> projectionDefs={ConfigUtils.getConfigProp("projectionDefs")}
> ```
> 
> to be hidden inside the enhancer. If the property is there, that it wins over ConfigProps. But, wherever not necessary (and i think it's never necessary), it should not be added.
> 
> * [ ]  Maybe MapView itself should contain that (mapview already have some enhancers or the internal component, don't remember..., just add it to the list), instead of applying the enhancer everytime
> 
> In general, nothing out from the map should be aware of projections definitions, if not strictly necessary.
> This should
> 
> * reduce a lot the number of the files you change
> * make the code more readable and maintainable, without redundancy
> * Reduce possible errors, because you don't have to remember thousands of settings for each map you create, but only the ones necessary to the it's usage.
>   Remember S.O.L.I.D.

In order to render correctly the map, we need to pass to the map component the projectionDefs because of [these ](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/map/openlayers/Map.jsx#L89-L93 )lines 

so i think this is necessary, i'll add to the MapView comp the two enhancers:
- one for fetching projectionDefs from props || config 
- one for emptyState if projection is not supported

web/client/components/geostory/media/Map.jsx
```diff
@@ -12,9 +12,9 @@ import uuid from "uuid";
 import MapView from '../../widgets/widget/MapView'; // TODO: use a external component
 import { applyDefaults } from '../../../utils/GeoStoryUtils';
 import {defaultLayerMapPreview} from '../../../utils/MediaEditorUtils';
-
 import connectMap from '../common/enhancers/map';
 
+
```
offtherailz:
```suggestion
```

web/client/components/geostory/media/Map.jsx
```diff
@@ -12,9 +12,9 @@ import uuid from "uuid";
 import MapView from '../../widgets/widget/MapView'; // TODO: use a external component
 import { applyDefaults } from '../../../utils/GeoStoryUtils';
 import {defaultLayerMapPreview} from '../../../utils/MediaEditorUtils';
-
 import connectMap from '../common/enhancers/map';
```
offtherailz:
```suggestion

import connectMap from '../common/enhancers/map';
```

web/client/actions/config.js
```diff
@@ -72,6 +72,7 @@ const mapSaveError = error => ({type: MAP_SAVE_ERROR, error});
 
 const mapSaved = () => ({type: MAP_SAVED});
 
+
```
offtherailz:
```suggestion
```

web/client/components/mediaEditor/MediaEditor.jsx
```diff
@@ -14,6 +14,7 @@ import localizedProps from '../misc/enhancers/localizedProps';
 import Toolbar from '../misc/toolbar/Toolbar';
 import MediaSelector from './MediaSelector';
 import MediaPreview from './MediaPreview';
+
```
offtherailz:
```suggestion
```

web/client/components/widgets/builder/wizard/map/PreviewMap.jsx
```diff
@@ -7,4 +7,5 @@
  */
 
 const previewMap = require('./enhancers/previewMap');
+
```
offtherailz:
```suggestion
```

web/client/plugins/GeoStory.jsx
```diff
@@ -19,6 +19,7 @@ import { editMedia } from '../actions/mediaEditor';
 import * as epics from '../epics/geostory';
 import { currentStorySelector, modeSelector, getFocusedContentSelector, isFocusOnContentSelector} from '../selectors/geostory';
 import { currentMessagesSelector } from '../selectors/locale';
+
```
offtherailz:
```suggestion
```

web/client/plugins/widgetbuilder/commons.js
```diff
@@ -22,10 +23,12 @@ const wizardStateToProps = ( stateProps = {}, dispatchProps = {}, ownProps = {})
     }, ownProps.target)
 });
 const wizardSelector = createSelector(
+    projectionDefsSelector,
```
offtherailz: Is this needed? doesn't uses MapView with projectionDefs?

Issue #4274, opened 2019-10-01
Annotations Circle has problem with radius


MV88:
### Description
Especially when changing the center only via click on map and via changing radius in the form

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [ ] Internet Explorer
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

*Browser Version Affected*

- Not browser related

*Steps to reproduce*

Common steps:
- Create a new circle annotation
- While being in edit mode of the circle (coordinate editor in the right panel)

<hr>

#### change center problem
- Click on another point on the map

<hr>

#### radius edit problem
- Change the radius (note the unit of measure)
<hr>

*Expected Result*

#### change center problem
Circle in map does not change, it just moves its center to the clicked point

#### radius edit problem
Radius changes should take into account UoM
Especially if meters and I change from 300000 to 300001 it should not "explode"

*Current Result*

#### change center problem
Circle resizes to a small point in map (wrong uom probably)

#### radius edit problem
The size of the circle is clearly wrong

### Other useful information (optional):
All tests were done in with default CRS i.e. EPSG:3857

francoperversi:
@kasongoyo the exploding circle is still present in 2 situations:

- Changing radius
- Changing center lat/long

PR #4131, opened 2019-09-02
Fix #3842 Switching CRS, circle annotations change size

vlt1:
## Description
* Added featureCrs param to transformPolygonToCircle and calculateRadius
for correct radius calculations
* Radius is recalculated on every call to transformPolygonToCircle

## Issues
 - #3842 

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)

 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)
#3842

**What is the new behavior?**
In description

**Does this PR introduce a breaking change?** (check one with "x", remove the other)

 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #4482, opened 2019-11-12
Fix 4274 Annotations circle has problem with radius

kasongoyo:
## Description
Fix annotations circle when changing radius or coordinates, a bug which caused by https://github.com/geosolutions-it/MapStore2/pull/4131. This PR restore some code which was removed by 4131 and apply fix which work for both 4274 and 3842

## Issues
 - #4274 
 - ...

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see #4274 

**What is the new behavior?**
The circle change smoothly

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/map/openlayers/DrawSupport.jsx
+++ b/web/client/components/map/openlayers/DrawSupport.jsx
@@ -421,7 +421,7 @@ export default class DrawSupport extends React.Component {
                     // TODO verify center is projected in 4326 and is an array
                     center = reproject(center, this.getMapCrs(), "EPSG:4326", false);
                     const originalId = newProps && newProps.features && newProps.features.length && newProps.features[0] && newProps.features[0].features && newProps.features[0].features.length && newProps.features[0].features.filter(f => f.properties.isDrawing)[0].properties.id || id;
-                    newFeature.setProperties({isCircle: true, radius, center: [center.x, center.y], id: originalId});
+                    newFeature.setProperties({isCircle: true, radius, center: [center.x, center.y], id: originalId, crs: this.getMapCrs()});
                 } else if (drawMethod === "Polygon") {
                     newDrawMethod = this.props.drawMethod;
                     let coordinates = drawnGeom.getCoordinates();
--- a/web/client/reducers/__tests__/annotations-test.js
+++ b/web/client/reducers/__tests__/annotations-test.js
@@ -1377,6 +1377,41 @@ describe('Test the annotations reducer', () => {
         expect(state.selected.geometry.coordinates[0]).toBe(1);
         expect(state.selected.geometry.coordinates[1]).toBe(1);
     });
+    it('changeSelected, changing coords of a Circle Feature', () => {
+        const selected = {
+            properties: {
+                canEdit: true,
+                center: [-6.576492309570317, 41.6007838467891],
+                id: "259d79d0-053e-11ea-b0b3-379d853a3ff4",
+                isCircle: true,
+                isValidFeature: true,
+                radius: 3567
+            },
+            geometry: {
+                type: "Circle",
+                coordinates: [-6.576492309570317, 41.6007838467891]
+            }
+        };
+        const featureColl = {
+            type: "FeatureCollection",
+            features: [selected],
+            tempFeatures: [],
+            properties: {
+                id: '1asdfads'
+            },
+            style: {}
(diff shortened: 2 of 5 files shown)


MV88:
@kasongoyo Please, pay more attention before submitting a PR because you have not fixed completely the original problems
4274annotation circle problems

- The circle should remain the same size when changing lat lon from the form
- and if you change 1 meter it should not grow
@offtherailz do you confirm this?

kasongoyo:
@MV88 
The value you saw in the radius form field is the one returned by geometry.getRadius() in drawend event. So do you mean that if projection is 4326 then that value should be in degree units instead of meters? I think all value returned by geometry.getRadius https://github.com/geosolutions-it/MapStore2/blob/dda8563b7a3bdd7b52fac7f9a3475d7f732bebab/web/client/components/map/openlayers/DrawSupport.jsx#L417 are in meters ?

offtherailz:
Actually the form shows the uom in meters, therefore you should show the number consistent with the UOM.

In the actual implementation the circle annotation is a circle in the projected system, therefore if I change the projection it should change it's shape. it doesn't happenm so it is a little inconstistent per se.
Peek 2019-11-15 15-30

For the moment convert the UOM accordingly for the selector. @MV88 please take note of this.

offtherailz:
any news about this PR?

offtherailz:
I'm reviewing this PR. Now I have to switch to another task. I report my observations. 
In fact the radius problem is also a problem of the Filter Panel (radius becomes in degrees, but the form indicates it's in meters. So I think we can open a separate issue to fix UOM indicated 

I'd like to clarify : 
 - [ ] Still didn't understand 1000 division or multiplication issue. @kasongoyo can you please explain, even in code, why that operation is present?
 - [ ] I found this issue, not sure is new moving the circle in invalid coordinates and then back to valid. Couldn't replicate in dev but couldn't investigate too much.  
Peek 2019-11-27 11-12

When these two points are clarified, we can merge (opening the issue for radius fix in both query panel and annotations panel) because the issue is solved. 
Any comment @MV88 ?

kasongoyo:
> Still didn't understand 1000 division or multiplication issue. @kasongoyo can you please explain, even in code, why that operation is present?

The unit parameter is missing in [turf/circle](https://www.npmjs.com/package/@turf/circle) library used to generate circle coordinates [here](https://github.com/geosolutions-it/MapStore2/blob/bfe9a4edb9f653b823a3c3ab6ed902d6fc08ad74/web/client/reducers/annotations.js#L96) so it means the turf/circle use default unit which is `kilometers` and map in `3857` projection report the distance in `meters` so there was a need to convert the reported radius in meter to kilometers to enable turf/circle to produce accurate coordinates. 

I agree with you @offtherailz  that we should open a new issue to address distance calculation and conversion when a map is in `4326` projection and actually, this is what I suggested to @tdipisa well the issue is when the map is in `4326` projection it's by default report distance in degrees, turf/circle support radius in a degree so the issue could be easily solved changing unit parameter to degrees when projection is in `4326` but the annotation form does not support degrees among the units of measurements and it's difficult to change between degrees to other units used in annotation form.

offtherailz:
Agree. Please : 
 - [ ] Please open an issue about UOM conversion with your observations and also the problem in filter panel (in 4326 it reports degrees but indicates `m` as unit of measure. When the specification about what we should do for radius in 4326 in annotations is clear. Some alternatives: 
    - convert implicitly in the UI from meeters to degrees (it may not work with every projection)
    - adding degrees to uoms and denying conversion in that case... it could be more general solution
    - other...
- [ ] Please comment the code to quote the reason of multiplication / division by 1000 
image

MV88:
@offtherailz i agree with you.
So, this pr can be merged and a new issue is needed for the other use cases mentioned by you.
@kasongoyo please follow Lorenzo's suggestions

offtherailz:
Just to have a complete panorama, now we have the same behavior of annotations "view" mode, as discussed here: 
https://github.com/geosolutions-it/MapStore2/pull/4167

Reprojecting circles (as ellipse, oval...) can be a  future task but not part of this issue.

tdipisa:
@offtherailz so can we merge this? If yes please approve

kasongoyo:
see #4577

Issue #4501, opened 2019-11-14
Map Inline Editing - Advanced


MV88:
# Description
When the Map inline editing is active, we should consider on having some controls to:
- [x] add tabs for each section: TOOLS, TOC
- [x] enable/disable the zoom in/out buttons and their position in the map
- [x] enable/disable the pan movement in the map (should disable both DragPan and KeyboardPan)
- [x] when inline editing is opened and zoom control is enabled, then MouseWheelZoom is enabled 
- [x] when zoom control is disabled then outside inline editing all the possible ways to zoom should be disabled (DoubleClickZoom, KeyboardZoom, PinchZoom, DragZoom, MouseWheelZoom)

NOTE: there is no support for disabling enabling interactions in openlayers map. part of the time will be used to add it.

see https://github.com/geosolutions-it/MapStore2-C039/issues/118

ElenaGallo:
Writing mistake in zoom in/out interactions dropdown menu. not 'Angolo in basse a destra' but 'angolo in basso a destra'. 
![issue 4501](https://user-images.githubusercontent.com/56537133/69791801-8ec92d80-11c5-11ea-92c3-1d6f4124b82b.JPG)

PR #4518, opened 2019-11-19
#4501 inline editing advanced

MV88:
## Description
Improved inline editing with some map configuration for zoom and pan interaction

Added also a tab panel where toc is the default one selected

## Issues
 - #4501

 - [x] Feature

see issue

**What is the new behavior?**
see description

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/geostory/media/Map.jsx
+++ b/web/client/components/geostory/media/Map.jsx
@@ -15,6 +15,7 @@ import {defaultLayerMapPreview} from '../../../utils/MediaEditorUtils';
 
 import connectMap from '../common/enhancers/map';
 
+
 // current implementation will update the map only if the movement
 // between 12 decimals in the reference system to avoid rounded value
 // changes due to float mathematic operations.
@@ -44,13 +45,14 @@ export default compose(
 )(({
     id,
     map = {layers: [defaultLayerMapPreview]},
-    options = {},
     fit,
+    isContentFocused,
     onMapViewChanges
 }) => {
-    const { layers = [], mapOptions, ...m} = (map.data ? map.data : map); // remove mapOptions to not override options
+    const { layers = [], mapOptions = {}, ...m} = (map.data ? map.data : map);
     return (<div
-        className="ms-media ms-media-map" style={{
+        className={`ms-media ms-media-map ${mapOptions.zoomPosition || ""}`}
+        style={{
             objectFit: fit
         }}>
         {/* BaseMap component overrides the MapView id with map's id */}
@@ -61,7 +63,18 @@ export default compose(
                 id: `media-${id}`
             }} // if map id is passed as number, the resource id, ol throws an error
             layers={layers}
-            options={applyDefaults(options)}
+            options={
+                // mouseWheelZoom is enabled only if inlineEditing is active and zoomControl too
+                applyDefaults(!isContentFocused ? {mapOptions} : {
+                    mapOptions: {
+                        ...mapOptions,
+                        interactions: {
+                            ...mapOptions.interactions,
+                            mouseWheelZoom: m.zoomControl
+                        }
+
+                    }
+                })}
         />
     </div>);
 });
--- a/web/client/components/geostory/media/Map.jsx
+++ b/web/client/components/geostory/media/Map.jsx
@@ -46,7 +46,7 @@ export default compose(
     id,
     map = {layers: [defaultLayerMapPreview]},
     fit,
-    isContentFocused,
+    editMap = false,
     onMapViewChanges
 }) => {
     const { layers = [], mapOptions = {}, ...m} = (map.data ? map.data : map);
@@ -65,7 +65,7 @@ export default compose(
             layers={layers}
             options={
                 // mouseWheelZoom is enabled only if inlineEditing is active and zoomControl too
-                applyDefaults(!isContentFocused ? {mapOptions} : {
+                applyDefaults(!editMap ? {mapOptions} : {
                     mapOptions: {
                         ...mapOptions,
                         interactions: {
--- a/web/client/components/map/openlayers/Map.jsx
+++ b/web/client/components/map/openlayers/Map.jsx
@@ -245,6 +245,9 @@ class OpenlayersMap extends React.Component {
         /*
          * Manage interactions programmatically.
          * map interactions may change, i.e. becoming enabled or disabled
+         * TODO: with re-generation of mapOptions the application could do this operation
+         * on every render. We should prevent it with something like isEqual if this becomes
+         * a performance problem
          */
         if (this.map && (this.props.mapOptions && this.props.mapOptions.interactions) !== (newProps.mapOptions && newProps.mapOptions.interactions)) {
             const newInteractions = newProps.mapOptions.interactions || {};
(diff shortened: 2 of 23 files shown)


web/client/components/geostory/media/Map.jsx
```diff
@@ -61,7 +63,18 @@ export default compose(
                 id: `media-${id}`
             }} // if map id is passed as number, the resource id, ol throws an error
             layers={layers}
-            options={applyDefaults(options)}
+            options={
+                // mouseWheelZoom is enabled only if inlineEditing is active and zoomControl too
```
offtherailz: You should never activate mouseWheel interaction, even in edit mode. you could activate it using a switch, at least.

web/client/components/geostory/media/Map.jsx
```diff
@@ -61,7 +63,18 @@ export default compose(
...
+                applyDefaults(!isContentFocused ? {mapOptions} : {
+                    mapOptions: {
+                        ...mapOptions,
+                        interactions: {
```
offtherailz: You should not activate zoom wheel when editing. make it a normal option like the others:
 it simplifies a lot and give a feature more to the user.

web/client/components/map/openlayers/Map.jsx
```diff
@@ -228,14 +261,36 @@ export default class OpenlayersMap extends React.Component {
...
+        /*
+         * Manage interactions programmatically.
+         * map interactions may change, i.e. becoming enabled or disabled
+         */
+        if (this.map && (this.props.mapOptions && this.props.mapOptions.interactions) !== (newProps.mapOptions && newProps.mapOptions.interactions)) {
+            const newInteractions = newProps.mapOptions.interactions || {};
+            Object.keys(newInteractions).forEach(newInteraction => {
+                this.map.getInteractions().getArray().forEach((mapInteraction) => {
+                    if (interactionsDefaults[newInteraction] && mapInteraction instanceof interactionsDefaults[newInteraction].Instance) {
+                        this[`${newInteraction}Interaction`] = mapInteraction;
```
offtherailz: Use a local variable. 
Using this could cause problems if some interactions are removed, they may be still in "this" object
This is a simple "find interaction". You can event use a find from lodash on the array

web/client/components/misc/enhancers/localizedProps.js
```diff
@@ -33,12 +42,12 @@ const accumulate = (props, messages) => (acc = {}, propName) => ({
  * //...
  * <Input placeholder="path.to.placeholder.message" />
  */
-module.exports = (propNames) => compose(
+module.exports = (propNames, labelName) => compose(
```
offtherailz: - call this localizedKey
- document the new prop and how to use in case of array

web/client/components/map/openlayers/Map.jsx
```diff
@@ -228,14 +261,36 @@ export default class OpenlayersMap extends React.Component {
...
                 this.map.addControl(new Zoom());
             } else {
                 this.map.removeControl(this.map.getControls().getArray().filter((ctl) => ctl instanceof Zoom)[0]);
             }
         }
+        /*
+         * Manage interactions programmatically.
+         * map interactions may change, i.e. becoming enabled or disabled
+         */
+        if (this.map && (this.props.mapOptions && this.props.mapOptions.interactions) !== (newProps.mapOptions && newProps.mapOptions.interactions)) {
```
offtherailz: This part need is very important, need unit tests.

web/client/components/map/openlayers/Map.jsx
```diff
@@ -228,14 +261,36 @@ export default class OpenlayersMap extends React.Component {
...
+                });
+                if (!this[`${newInteraction}Interaction`]) {
+                    // if the requested interaction does not exist in the map then add it
+                    this[`${newInteraction}Interaction`] = new interactionsDefaults[newInteraction].Instance(interactionsDefaults[newInteraction].options);
+                    this.map.addInteraction(this[`${newInteraction}Interaction`]);
+                } else {
+                    // otherwise use existing interaction and enable or disable it based on newProps values
+                    this[`${newInteraction}Interaction`].setActive(newInteractions[newInteraction]);
+                }
+            });
```
offtherailz: strange algorithm 
 -  create new interaction and add if it wasn't there
 - if at least once in the object lifetime (see the comment before) it existed, set active, if the value is true or false. 
Can you make it simpler? Please take even into account that these interactions can have options, so don't wire the solution to your use case.

web/client/components/geostory/media/Map.jsx
MV88: as written and requested in the issue #4501, when inline editing is activated and zoomContorl is enabled then activate also the MouseWheelZoom interaction.

probably i just need to change the actual condition where it does it

web/client/components/geostory/media/Map.jsx
MV88: @tdipisa what about this?

web/client/components/geostory/media/Map.jsx
offtherailz: Ok, so this becomes a suggestion

web/client/components/map/openlayers/Map.jsx
MV88: the problem is that we should not duplicate existing interactions. so :
if (already exists) then 
   activate / deactivate  (to do this i need to fetch the existing interaction, and probably i can 
   avoid to store it in the this object)
else
  create it and addit to the map

side note: regarding the options they are mostly added with defaults and a few of them have some customization. btw i am not removing existing interactions

web/client/components/map/openlayers/Map.jsx
MV88: the problem is that when [this](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/map/openlayers/Map.jsx#L109) line runs it creates a bunch of interactions by default. 

we should avoid to duplicate them. i'll improve this with a find function and then adding if it not exists.

if i find the interaction then i'll have just to activate/ deactivate it

web/client/components/geostory/media/Map.jsx
tdipisa: @MV88 @offtherailz no because if you include the mouse wheel in settings then the user expect to have it also in the story where the wheel must be disabled for scrolling purposes.

offtherailz:
please solve conflicts

web/client/components/map/openlayers/Map.jsx
```diff
@@ -229,14 +235,36 @@ class OpenlayersMap extends React.Component {
...
         if (newProps.zoomControl !== this.props.zoomControl) {
             if (newProps.zoomControl) {
                 this.map.addControl(new Zoom());
             } else {
                 this.map.removeControl(this.map.getControls().getArray().filter((ctl) => ctl instanceof Zoom)[0]);
             }
         }
+        /*
+         * Manage interactions programmatically.
+         * map interactions may change, i.e. becoming enabled or disabled
```
offtherailz:
```suggestion
         * map interactions may change, i.e. becoming enabled or disabled
         * TODO: with re-generation of mapOptions the application could do this operation
         * on every render. We should prevent it with something like isEqual if this becomes
         * a performance problem
```

Issue #4424, opened 2019-10-30
Metadata no more displayed in catalog preview


francoperversi:
### Description
When adding metadata to a Service, the layer's preview in catalog should contains metadata, but it doesn't

### In case of Bug (otherwise remove this paragraph)

*Browser Affected*
(use this site: https://www.whatsmybrowser.org/ for non expert users)
- [ ] Internet Explorer
- [X] Chrome
- [X] Firefox
- [X] Safari

*Browser Version Affected*

- Not browser related

*Steps to reproduce*

- Open [this](https://dev.mapstore.geo-solutions.it/mapstore/#/viewer/openlayers/11803) map 
- Open burger menu
- Open catalog
- Edit CSW service
- Verify that the metadata template is enabled and that the metadata properties are set
- Click on Save

*Expected Result*

![metadata-right](https://user-images.githubusercontent.com/54808782/67874475-428ebd00-fb35-11e9-9757-03141a17598a.gif)

*Current Result*

![metadata-wrong](https://user-images.githubusercontent.com/54808782/67874503-4b7f8e80-fb35-11e9-87f8-33a478fd92ef.gif)

PR #4526, opened 2019-11-19
Fix 4424 regression for catalog records using metadata template

MV88:
## Description
Fix regression on catalog records

the problem was property hiding introduced [here](https://github.com/geosolutions-it/MapStore2/blame/master/web/client/components/catalog/Catalog.jsx#L287)

## Issues
 - #4424

 - [x] Bugfix

see issue

**What is the new behavior?**
now metadata template is applied correctly (as before)

 - [x] No

**Other information**:


MV88:
> * [ ]  A unit text?
> * [ ]  Did you tested background selector?
> * [ ]  Can you please check with @vlt1 if this change somehow backgrounds records, and how to support both possibily?

- test added,
- for background it is not affected since the metadata template cannot be defined via UI and in configuration they are not defined also. should we add a more robust check for this use case?

PR #4538, opened 2019-11-21
#4378 Groups with same name cause issues

### LINKED ISSUE #4378: Groups with same name cause issues
### Description
If you try to create groups of the same name in the TOC, there are some problems while dragging and moving.

The problem is that the name of the group is used as id (`name` is the id a group, i think). 

Maybe we should generate a unique id (name) for the group, anyway different from any other group in the TOC, and use the `title` (there is a default title for the localization) to allow the user to ignore this issue and create groups with the same title.

Another solution can be to deny creation 2 groups with same name, but because of nesting, I think It's not the best solution. Users may need to do subgroups like this.

```
Public Buildings:
 - Type 1
 - Type 2
PrivateBuildings: 
 - Type 1
 - Type 2
```

*Browser Affected*
any

*Steps to reproduce*

- create 2 groups with the name "test" in the root (without selecting target group)
- try to drag one of the groups in the default group

*Expected Result*

-  the group is dragged 

*Current Result*

- one of the layers disappears

Peek 18-10-2019 11-34

vlt1:
## Description
* When a group is added unique id is generated
* title for groups is preserved now on map save
* added tests

## Issues
 - #4378

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

#4378 

**What is the new behavior?**
In the description

 - [x] No

### DIFF
--- a/web/client/plugins/__tests__/AddGroup-test.jsx
+++ b/web/client/plugins/__tests__/AddGroup-test.jsx
@@ -73,8 +73,6 @@ describe('AddGroup Plugin', () => {
         const btn = document.getElementsByTagName('button')[1];
         ReactTestUtils.Simulate.click(btn);
         expect(store.getState().layers.groups.length).toBe(2);
-        expect(store.getState().layers.groups[1].name).toBe('newgroup');
-        expect(store.getState().layers.groups[1].id).toBe('newgroup');
         expect(store.getState().layers.groups[1].title).toBe('newgroup');
         expect(store.getState().layers.groups[1].nodes.length).toBe(0);
     });
@@ -105,8 +103,6 @@ describe('AddGroup Plugin', () => {
         expect(store.getState().layers.groups[0].nodes.length).toBe(1);
         expect(store.getState().layers.groups[0].nodes[0].nodes.length).toBe(1);
         const newgroup = store.getState().layers.groups[0].nodes[0].nodes[0];
-        expect(newgroup.name).toBe('newgroup');
-        expect(newgroup.id).toBe('group1.group2.newgroup');
         expect(newgroup.title).toBe('newgroup');
         expect(newgroup.nodes.length).toBe(0);
     });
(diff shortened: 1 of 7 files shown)


web/client/plugins/__tests__/AddGroup-test.jsx
```diff
@@ -73,8 +73,6 @@ describe('AddGroup Plugin', () => {
         const btn = document.getElementsByTagName('button')[1];
         ReactTestUtils.Simulate.click(btn);
         expect(store.getState().layers.groups.length).toBe(2);
-        expect(store.getState().layers.groups[1].name).toBe('newgroup');
```
MV88: Hi, if you are removing bcause the id is generated each time maybe you can test the length of it.
like if using uuid is 32 chars

web/client/plugins/__tests__/AddGroup-test.jsx
```diff
@@ -105,8 +103,6 @@ describe('AddGroup Plugin', () => {
         expect(store.getState().layers.groups[0].nodes.length).toBe(1);
         expect(store.getState().layers.groups[0].nodes[0].nodes.length).toBe(1);
         const newgroup = store.getState().layers.groups[0].nodes[0].nodes[0];
-        expect(newgroup.name).toBe('newgroup');
```
MV88: same as above

web/client/reducers/__tests__/layers-test.js
```diff
@@ -845,10 +845,87 @@ describe('Test the layers reducer', () => {
...
+        expect(state).toExist();
+        expect(state.groups.length).toBe(1);
+        expect(state.groups[0].nodes.length).toBe(1);
+        expect(state.groups[0].nodes[0].nodes.length).toBe(3);
+        const newgroup1 = state.groups[0].nodes[0].nodes[1];
+        const newgroup2 = state.groups[0].nodes[0].nodes[2];
+        expect(newgroup1.title).toBe('newgroup');
+        expect(newgroup2.title).toBe('newgroup');
+        expect(newgroup1.name).toNotBe(newgroup2.name);
+        expect(newgroup1.id).toNotBe(newgroup2.id);
```
MV88: same as above, add a check for length of id

Issue #4537, opened 2019-11-21
Modal windows style diversity when exit from edit properties without save changes


francoperversi:
### Description
The confirm modal window has different style when exit from the edit properties window without save changes in various situation:

* Cards **Edit properties** button
   - For maps see image 1 in 'Other useful information'
   - For dashboards see image 2 in 'Other useful information'
   - For stories same as dashboards
   - For contexts same as dashboards
   - Editing existing users any confirm modal windows are present
   - Editing existing groups any confirm modal windows are present

* Burger menu **Save/Save as** buttons
   - For maps same as image 2 in 'Other useful information'
   - For dashboards same as maps
   - For stories same as maps
   - For contexts can't text yet
   - Creating new user any confirm modal windows are present
   - Creating new group any confirm modal windows are present

**In conclusion, the propose is the following:**
In all the listed situation, when the user exit from a settings window without saving some changes, the same modal window should appears: image 2 in 'Other useful information'

### Other useful information (optional):
1.
![map_ep](https://user-images.githubusercontent.com/54808782/69360298-969b4600-0c8a-11ea-85d7-ef2074f37515.jpg)
2.
![maps_save](https://user-images.githubusercontent.com/54808782/69360645-36f16a80-0c8b-11ea-9a0f-447f946112c3.jpg)

Issue #2908, opened 2018-05-16
Related resources saving flow issues due to permission updates - Reuse dashboard code


offtherailz:
### Description
**note** the resources linked in this epic do not complete the epic yet. 
When you save a map with detail card and/or thumbnails, related resources are not updated in terms of permissions. We should reuse both the resource form and `observables/geostore.js` used in dashboard (that are not afflicted by this bugs) to update the permission correctly to both related resources and main resource.
The metadata form should be reused also in the map viewer context, to allow the editor to change metadata and permission without having to go to the home page and re- search for the map. 

Pros: 
 - No orphan resources (thumbnails and detail cards)
 - No issues with permission of detail cards and thumbnails
 - Can Full Map metadata form inside map (as for dashboards)
 - Future maintenance (code shared)
 - Future similar resources can be managed in the same way
- Users that can edit a map can also edit permission [TODO: avoid to remove own permission]

Draft Tasks: 
 - [ ] Update featured maps and maps plugins to work like `Dashboards` plugin
 - [ ] Implement detail card support for dashboard (and maps)
 - [x] Reuse modal form in the map context (Save/ Save as plugins) 
 - [ ] Decouple currentmap reducer and reuse the logic of dashboard

*Browser Affected*
any

*Steps to reproduce*
- Create a map with thumbnail or detail card
- Save it
- Update permission to grant edit access to a group (from home)
- Access as a user of the group
- Open developer tools
- Try to change the thumbnail

*Expected Result*

- The thumbnail is updated

*Current Result*

- You will not be able to delete the old thumbnail (error from dev tools) you will create another resource instead, but the current is lost. Sometimes new thumbs or detail cards are not accessible from other users.

offtherailz:
going to close this issue in favor of #5772

PR #4560, opened 2019-11-25
Fix 4537 adding unsaved changes modal to map edit properties, user and group dialog

MV88:
## Description
This pr add unsaved changes dialog to map metadata modal, user dialog and group dialog

## Issues
 - #4537

 - [x] Feature

see issue

**What is the new behavior?**
now if something has changes for mep edit properties or user dialog or group dialog, when clicking on close it shows the unsaved changes modal

 - [x] No

**Other information**:


MV88:
> The code seems good, I didn't tested this.
> @MV88 please notify you added some new feature also in user and groups modals, so he can test them. Maybe is a good thing if, before merging, you open a separate issue dedicated to these improvements.

The changes to user dialog and group dialog has been notified in the **What is the new behavior?** section

> @MV88 please note that we want to unify all the resource handling (now only map is different). See this epic #2908. So in fact we only need to add support to detail cart do the generic resource window.
> This make this change (that anyway is small) a little unuseful, because this difference will be removed soon by using the same component.

I know this will be removed soon, but the alternative, for the moment, is inconsistency between map and other resources. @tdipisa what do you think regarding this point?

Issue #4535, opened 2019-11-21
Warning message problems saving resources with same name


francoperversi:
### Description
Actually the behavior of the warning message when saving resources with the same name is quite various, depending of resource type:
1. **Renaming a resource with an existing  resource name from the Edit properties button**
    - For maps the following warning message appears : `A thumbnail with this name already exist` (see image 1. in 'Other useful information')
    - For dashboards no warning message appears, when you try to save nothing happens (see image 2. in 'Other useful information')
    - For stories same than dashboards
    - For contexts same than dashboards
2. **Naming a resource with an existing resource name from Save/Save as button**
     - For maps the following warning message appears : `A resource with this name already exist, try 'resourcename - 2'` and a warning window on background appears with the following message: `Cannot save current map - A map with this name already exist` (see image 3. in 'Other useful information')
    - For dashboards only a warning message appears: `A resource with this name already exist, try 'resourcename - 2'` (see image 4. in 'Other useful information')
    - For stories same than dashboards
    - For contexts same than the others

I propose to uniform the behavior that has to be the same for all the resources (maps, dashboards, stories and contexts), in this way:
1. **Renaming a resource with an existing resource name from the Edit properties button**
    - The following warning message appears: `A resource with this name already exist`
2. **Naming a resource with an existing resource name from Save/Save as button**
    - The following warning message appears: `A resource with this name already exist`
    - The following warning window appears in background: `Cannot save current resource - A 
resource with this name already exist`

Moreover, a little fix is needed, only for maps and stories:
When you rename a resource with an existing resource name from the Save button, if you close the Save and you open the Save as (or again the Save), the warning message is still there. All the way around, when you rename a resource with an existing resource name from the Save as button, if you close the Save as and you open the Save (or again the Save as), the warning message is still there (see image 5. in 'Other useful information')

### Other useful information (optional):
1. 
![maps_ep](https://user-images.githubusercontent.com/54808782/69352020-7795b780-0c7c-11ea-9fb5-fb849096de7b.jpg)
2.
![dash_ep](https://user-images.githubusercontent.com/54808782/69352037-7e242f00-0c7c-11ea-91a3-d713fdb14d86.gif)
3.
![maps_s](https://user-images.githubusercontent.com/54808782/69352048-82e8e300-0c7c-11ea-8dae-833ae70b6c12.jpg)
4.
[...]

PR #4567, opened 2019-11-26
Fix 4535 improved and uniformed feedback for resources

MV88:
## Description
Fixes some problems on edit properties modal for gestore resources (maps, dashboards, stories)

## Issues
 - #4535

 - [x] Bugfix
 - [x] Feature

see issue

**What is the new behavior?**
- Error messages has ben generalized cor maps, dashboards and stories
- a successful notification has been added for the three above
- in homepage if you save and use an existing name, the error will appear, if you close and reopen the same card the error is gone
- save and save as in map no longer keeps the error message

 - [x] No

**Other information**:

@offtherailz please note that some of the code has been done again on the map component actually being used


web/client/plugins/Save.jsx
```diff
@@ -26,7 +27,8 @@ export default createPlugin('Save', {
             mapInfoSelector,
             (show, resource) => ({show, resource})),
         {
-            onClose: toggleControl.bind(null, 'mapSave', false)
+            onClose: toggleControl.bind(null, 'mapSave', false),
+            onMapSaveError: mapSaved
```
offtherailz: This looks a little strange to me? mapSaved bind to mapSaveError?

web/client/actions/dashboards.js
```diff
@@ -14,7 +14,11 @@ const DASHBOARD_DELETED = "DASHBOARDS:DASHBOARD_DELETED";
 const RELOAD = "DASHBOARDS:RELOAD_DASHBOARDS";
 const LOADING = "DASHBOARDS:LOADING";
 
+const { success: successNotification } = require('./notifications');
+
+
 module.exports = {
+    showSuccessNotification: () => successNotification({ title: "success", message: "resources.successSaved" }),
```
offtherailz: This is replicated 3 times in 3 files and then again in the files. 
This is not necessary. There is a little of duplication for these tools that have to be put in common. Please include these declarations in the plugins.

web/client/plugins/Save.jsx
MV88: yes, it is used to [reset errors](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/reducers/config.js#L118) when edit properties modal is closed..

going to change the name to
```suggestion
            onResetMapSaveError: mapSaved
```

offtherailz: Create a specific action for it, otherwise is confusing

offtherailz: Or you create a RESET_ERRORS action or you explain that you are using the mapSaved action to reset error. I'm not sure if you pass the map or not. Please explain and find out a better name.

onResetMapSaveError doesn't make to much sense. what it means? "on Reset, MapSaveError"/ on reset map, save error"...
handlers name with "on" means that they are called in sequence of an event. 
onReset means that is called when button "Reset" is clicked, or a "reset" action is performed by the user, anyway,
onPan means the handler is called when the map pans.

Instead, if the property is instead a function that have to be called to have some effect, you can call it "reset". 
"resetError" will reset the error.

MV88: i could create another constant in config.js action file that does the same thing as MAP_SAVED
i wanted to reuse it, gonna create it anyway and change the name of the action passed to the plugin to resetErrors

MV88: *correction gonna call it **clearErrors**

MV88:
closing this in favor of https://github.com/geosolutions-it/MapStore2/pull/4592

Issue #4566, opened 2019-11-26
Share tool problems


francoperversi:
### Description
The following fixes have to performed for the share tools:

**Opening the share tool from the Homepage:**

- [ ] Add `/mapstore/` in link for maps, dashboards, stories and contexts (an example of correct link is `https://dev.mapstore.geo-solutions.it/mapstore/#/viewer/openlayers/356`)

- [ ] Remove the embed tab in dashboards and stories

- [ ] Add `/mapstore/` in embed iframe for maps, and contexts 

- [ ] Remove the _Show TOC_ option in embed section for contexts 

- [ ] Fix the maps API that is not working (in order to fix this, see the API from the share in burger menu that works)

**Opening the share tool from the Burger Menu:**

- [ ] Remove the _Show TOC_ option in embed section for contexts

francoperversi:
@offtherailz after the test I report the following observations:

- [ ] Opening the share tool from burger menu in contexts, the `Show TOC` option is still present

- [ ] Opening the share tool from featured maps in homepage, the embed tab is not present

tdipisa:
> @offtherailz after the test I report the following observations:
> 
> * [ ]  Opening the share tool from burger menu in contexts, the `Show TOC` option is still present

@francoperversi Contexts will not be released now, so please put a note for this in [this](https://github.com/georchestra/mapstore2-georchestra/issues/22) issue. 

> * [ ]  Opening the share tool from featured maps in homepage, the embed tab is not present

@francoperversi we know about that, please open a new issue for this and put it on 2019.03.00

PR #4575, opened 2019-11-27
#4566 Fix issues with application path

offtherailz:
## Description
This PR solves the issue as in 4566 getting the pathname from the window.location with the origin. 
This way the path should be generated using 
http://some.doman/path-to-app/....

We have to test all the use case (embedded.html for map is added correcty etc...)

I've added tests for the base use cases. Please ensure that everything works also when deployied.

## Issues
 - #4566

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

**What is the new behavior?**

 - [ ] No

**Other information**:

### DIFF
--- a/web/client/components/resources/enhancers/__tests__/withShareTool-test.js
+++ b/web/client/components/resources/enhancers/__tests__/withShareTool-test.js
@@ -75,7 +75,32 @@ describe('withShareTool enhancer', () => {
             const iframeCode = codeBlock.innerText;
             // normal behaviour has the URL (without embedded)
             expect(iframeCode.indexOf(FAKE_RESOURCE_URL)).toBeGreaterThan(0);
-
+        });
+        it('Check shareOptions (embedOptions.showTOCToggle)', () => {
+            const Sink = addSharePanel(createSink(() => { }));
+            ReactDOM.render(<Sink
+                showShareModal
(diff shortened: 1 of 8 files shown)


offtherailz:
I removed the embedded from the share tool of dashboards because it has the home button and other tools that should not be shown in an embedded context

offtherailz:
From testing with the war some minor things to check:
- [ ]  l'API doesn't seems to work, but maybe a local problem. Need to check

Other observations: 
* Share tool in burger menu is not consistent with the one in app --> the tool is not special itself, it will be configured as the other plugins.
* Share tool probably should not have an embed tab, because it doesn't have minimal tools, but a full MapStore --> good observation, I should remove it

Observation by me: 
- Share maps with context is a future task, (So the link of the map created within a context is linked with the context (ì/context/<contextName>/<mapId>`). isn't it?

@tdipisa please confirm these 2+1 observations . In meanwhile I'llcheck the API issue

offtherailz:
I verified the specification. Everything is already as expected: 
 - Maps ( as usual. The API points to the correct JS, so deployed it should work
 - Stories and Dashboard have only link and social
 - Context has only iframe in Embeded tab, no show toc.

Issue #4034, opened 2019-08-02
Confirmation dialog in map on user logout 


tdipisa:
### Description
When you logout from a new map containing unsaved changes, MS should shows a confirmation dialog to proceed or not. The redirect to the home page should be performed if the user choose to discard changes.

The same behavior should be adopted in case of existing maps with pending changes on them and must be applied also when the user click on the Home button (not just when disconnecting).
Obviously, the confirmation dialog must be shown only if the user has write permissions.

Modifications on map center can be excluded from the check but not the zoom level.

### Other useful information:
This warning must be enabled by configuration (by default is enabled)

ElenaGallo:
@kboczkowski-clurgo When you logout from a new map containing unsaved changes, and click on yes on the unsaved changes modal.You are redirect to the login modal window.
![test34](https://user-images.githubusercontent.com/56537133/71190088-31844180-2284-11ea-8547-8ed07eede34e.gif)

*How to reproduce*

- Open MapStore
- Login as admin
- Open New Map
- Add a layer or change zoom level
- Log out 
- Click on yes on the unsaved changes modal

*current:*

- The login modal opens

*expected:*

- Redirect to the Home Page

ElenaGallo:
@kboczkowski-clurgo When disconnecting from a new map that contains no changes. You are redirect to the login modal window.
![issue 4034](https://user-images.githubusercontent.com/56537133/71983621-7d912900-3227-11ea-87e8-9e5a0d69a79a.gif)


*How to reproduce*

- Open MapStore
- Login as admin
- Open New Map
- Log out 

*current:*

- The login modal opens

*expected:*

- Redirect to the Home Page

PR #4581, opened 2019-11-28
#4034 Confirmation dialog in map on user logout

kboczkowski-clurgo:
## Description
In this PR I've added before leave popup with map changes validation

## Issues
 - #4034
 - ...

 - [x] Feature

**What is the new behavior?**

 - [x] No

**Other information**:


kboczkowski-clurgo:
@MV88 I've found problem with changes recognition. I've prepared method for cleaning objects right before update. In single layer some field appears: `apiKey`, which doesn't comes from API. Now should work properly

MV88:
@kboczkowski-clurgo what about other changes requested?

tdipisa:
@kboczkowski-clurgo please can you fix conflicts on this?

web/client/reducers/searchconfig.js
```diff
@@ -19,7 +19,7 @@ function searchconfig(state = null, action) {
             [action.property]: action.value
         });
     case MAP_CONFIG_LOADED: {
-        const textSearchConfig = action.config.map.text_serch_config;
+        const textSearchConfig = action.config.map.text_search_config;
```
offtherailz: probably some resources has been saved as text_serch_config :(. 
Maybe is a good Idea if we make old saved maps working. (you should test an old config, but putting (and adding a test)  here, something like : 
```suggestion
        const textSearchConfig = action.config.map.text_search_config || action.config.map.text_serch_config ;
```
may do the job

web/client/stores/StandardStore.js
```diff
@@ -68,6 +69,7 @@ module.exports = (initialState = {defaultState: {}, mobile: {}}, appReducers = {
                 loadingError: mapState.loadingError,
                 mapId: mapState.loadingError.mapId
             } || null,
+            mapConfigRawData: mapState && mapState.mapConfigRawData || null,
```
offtherailz: It think this have to be removed

kboczkowski-clurgo: @offtherailz without this i can't store in this field in redux anything, even when i made proper action and dispatch it

web/client/utils/MapUtils.js
```diff
@@ -411,6 +412,38 @@ const parseLayoutValue = (value, size = 0) => {
...
+            delete obj[key];
+        }
+    });
+};
+
+const compareMapChanges = (map1 = {}, map2 = {}) => {
+    const pickedFields = [
+        'map.layers',
+        'map.backgrounds',
+        'map.text_search_config',
```
offtherailz: I noticed that I can still not leave this map /#/viewer/openlayers/2464 trying to open and exiting, the map asks me to confirm This is because of the name mismatch between old search configuration name and the new...
Can we do something to pick also old configs for compare?

kboczkowski-clurgo: done

Issue #4514, opened 2019-11-18
Builder - Content Editor - Web page


kboczkowski-clurgo:
### Description
See [geosolutions-it/MapStore2-C039/issues/32](https://github.com/geosolutions-it/MapStore2-C039/issues/32)

ElenaGallo:
@kboczkowski-clurgo 

- [ ] The drag and drop, works, but the modal appears
-  open a web page section
- add a web page
- drag and drop the web page

*current result*
the modal appears

![32 1](https://user-images.githubusercontent.com/56537133/70799378-c5986980-1da9-11ea-8ed1-62a1cde29d1f.gif)

- [ ] The web page is not more editable in immersive section.
- Open an immersive section
- Add web page content 
- Click on Edit we page URL

*current result*
the page is not editable 
![32 3](https://user-images.githubusercontent.com/56537133/70802524-f7adc980-1db1-11ea-821e-34e9e805ffdd.gif)

ElenaGallo:
@kboczkowski-clurgo 
- The drag and drop described [here](https://github.com/geosolutions-it/MapStore2/issues/4514#issuecomment-565435794) still dose not work propertly.

- The WebPage content creation has a different icon button after an immersive section. 
*Expected Result*
![test14si](https://user-images.githubusercontent.com/56537133/71079594-da9b4100-218b-11ea-8fa8-919d6af238da.JPG)
*Current Result*
![test14](https://user-images.githubusercontent.com/56537133/71079642-ed157a80-218b-11ea-8951-ab5a83cbd953.JPG)

PR #4600, opened 2019-12-04
Builder - Content Editor - Web page (C039 - #32)

kboczkowski-clurgo:
## Description
In this PR I've addeed create new geostory component mechanism - WebPage

## Issues
 - #4514

 - [x] Feature

**What is the new behavior?**

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/geostory/contents/WebPageWrapper.jsx
+++ b/web/client/components/geostory/contents/WebPageWrapper.jsx
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2019, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import Dialog from '../../misc/StandardDialog';
+import { FormControl, ControlLabel, FormGroup, Button } from 'react-bootstrap';
+import { connect } from 'react-redux';
+import { setWebPageUrl } from '../../../actions/geostory';
+import Message from '../../I18N/Message';
+import Portal from '../../misc/Portal';
+import PropTypes from 'prop-types';
+import WebPage from './WebPage';
+import { compose, withHandlers } from 'recompose';
+
+export class WebPageWrapper extends React.PureComponent {
+    static propTypes = {
+        editURL: PropTypes.bool,
+        onClose: PropTypes.func,
+        onChange: PropTypes.func,
+        src: PropTypes.string
+    }
+
+    state = {
+        url: this.props.src || ''
+    }
+
+    render() {
+        const { onChange, onClose, editURL } = this.props;
+        const { url } = this.state;
+
+        return (
+            <React.Fragment>
+                <Portal>
+                    <Dialog
+                        modal
+                        show={editURL}
+                        title={<Message msgId="geostory.webPageCreator.title" />}
+                        onClose={onClose}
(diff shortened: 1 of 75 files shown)


tdipisa:
@kboczkowski-clurgo please fix the issues identified by @MV88, the Web Page content must be added not only as a additional content of an existing section (immersive or not) but also as a section per se like the other ones. The requirement in the [issue](https://github.com/geosolutions-it/MapStore2-C039/issues/32) description is clear about that.

tdipisa:
>@tdipisa I noticed that you cannot add directly a web page content, you have to add a media section or paragraph

@kboczkowski-clurgo looking at the last commit message I suppose that latest updates don't fix this problem but only the consol error. Isn't it?

kboczkowski-clurgo:
@tdipisa actualy I've fixed problem with editing which @MV88 discover. Now I'm working on adding this component via section directly.

kboczkowski-clurgo:
https://github.com/geosolutions-it/MapStore2/pull/4600#pullrequestreview-328832449
It's done

tdipisa:
> Hi, i was not able to load these urls:
> 
> * http://localhost:8081/?debug=true#/geostory/newgeostory/
> * http://localhost:8081/#/geostory/newgeostory/

This doesn't make sens to me, why we should support these kind of URLs? 

Maybe what we can do here is just to make the regex configurable from plugin config 

@offtherailz what do you think?

offtherailz:
the only problem concerns "localhost", because the regex denies domain names without a point. This may fail with host files or local DNS. 

Because the system may need to satisfy different requirements, as @tdipisa suggested, I should allow to configure the regex (2nd param of isValidURL may be the regex, by default the one currently used, maybe without the `.` check, but this is not necessary, if configurable. 
From outside, (where isValidURL is called) you can get it from ConfigUtils.getConfigProp("GeoStoryValidIframeURLRegex")  and pass it as 2nd param. I think it's enough.

kboczkowski-clurgo:
@MV88 @tdipisa @offtherailz PR is updated with changes suggested by @offtherailz. Now validation util has optional parameter to pass custom regexp.

web/client/components/geostory/contents/WebPageWrapper.jsx
```diff
@@ -0,0 +1,101 @@
...
+        );
+    }
+
+    updateURL = ({ target: { value: url }}) => {
+        this.setState({ url });
+    }
+
+    save = () => {
+        const { url } = this.state;
+        const error = !isValidURL(url);
```
MV88: can you pass ConfigUtils.getConfigProp("GeoStoryValidIframeURLRegex") as second argument? 
in that case we can override it anytime without changing the code

Verify that if the property is missing in config, then it uses the default one

web/client/components/geostory/contents/WebPageWrapper.jsx
```diff
@@ -0,0 +1,101 @@
...
+        this.setState({ error: !isValidURL(url) });
```
MV88:
```suggestion
        this.setState({ error });
```

web/client/components/geostory/contents/WebPageWrapper.jsx
kboczkowski-clurgo: ok

kboczkowski-clurgo: @MV88 done

Issue #4593, opened 2019-12-03
Permission editor available for unauthorized users


francoperversi:
### Description
The permission editor in a resource should be available only for the owner of that resource (the one who created it) and the administrator 
Actually the permission editor is available also for normal users with editing permission and this thing needs to be fixed. In particular the permission editor should be removed from the following situations:

### FOR MAPS, DASHBOARDS AND STORIES

* Login as admin
* Create a new map/dash/story
* Save it assigning the following permission: `Testers` - `can edit`
* Go back to the homepage
* Logout from admin and login as a tester
* Open the map/dash/story just created
* Click on Burger menu
* Click on save

**Problem:** The permission editor should not be present

### FOR DASHBOARDS AND STORIES

* Login as admin
* Create a new dash/story
* Save it assigning the following permission: `Testers` - `can edit`
* Go back to the homepage
* Logout from admin and login as a tester
* In homepage, click on edit properties on the dash/story just created

**Problem:** The permission editor should not be present

### Other useful information (optional):
The only situation where we have the ideal situation is the following:
* Login as admin
* Create a new map
* Save it assigning the following permission: `Testers` - `can edit`
* Go back to the homepage
* Logout from admin and login as a tester
* In homepage, click on edit properties on the map just created
* The permission editor is not present

ElenaGallo:
- [ ] now for the **first situation** the permission is forbidden but the permission editor is still present 

FOR MAPS, DASHBOARDS AND STORIES

- Login as admin
- Create a new map/dash/story
- Save it assigning the following permission: Testers - can edit
- Go back to the homepage
- Logout from admin and login as a tester
- Open the map/dash/story just created
- Click on Burger menu
- Click on save

*Current Result*: The permission editor is Forbidden but the options to change the permission are present on the Eidt properties window
![45](https://user-images.githubusercontent.com/56537133/70425597-cc557280-1a71-11ea-8723-4c4d8dd399f7.JPG)

*Expected Result*: The permission editor should not be present

- [ ] the **second situacion** has not been fixed 

FOR DASHBOARDS AND STORIES
- Login as admin
- Create a new dash/story
- Save it assigning the following permission: Testers - can edit
- Go back to the homepage
- Logout from admin and login as a tester
- In homepage, click on edit properties on the dash/story just created

*Current Result*: The permission editor is availabe 
*Expected Result*: The permission editor should not be present

tdipisa:
[...]

PR #4605, opened 2019-12-04
Fix 4593 Permission editor available for unauthorized users

kasongoyo:
## Description
see title

## Issues
 - #4593 
 - ...

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

see #4593

**What is the new behavior?**
Permission editor became unavailable for unauthorized users

 - [x] No

**Other information**:

### DIFF
--- a/web/client/components/resources/modals/Save.jsx
+++ b/web/client/components/resources/modals/Save.jsx
@@ -84,8 +84,8 @@ class SaveModal extends React.Component {
      * @return the modal for unsaved changes
     */
     render() {
-        const canEditResourcePermission = this.props.user.role === 'ADMIN' ||
-        this.props.resource.attributes && this.props.resource.attributes.owner === this.props.user.name;
+        const canEditResourcePermission = this.props.user && this.props.user.role === 'ADMIN' ||
+        this.props.resource && this.props.resource.attributes && this.props.resource.attributes.owner === this.props.user.name;
         return (<Portal key="saveDialog">
             {<ResizableModal
                 loading={this.props.loading}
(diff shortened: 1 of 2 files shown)


web/client/components/resources/modals/Save.jsx
```diff
@@ -82,6 +84,8 @@ class SaveModal extends React.Component {
      * @return the modal for unsaved changes
     */
     render() {
+        const canEditResourcePermission = this.props.user.role === 'ADMIN' ||
```
MV88:
```suggestion
        const canEditResourcePermission = this.props.user && this.props.user.role === 'ADMIN' ||
```

web/client/components/resources/modals/Save.jsx
```diff
@@ -82,6 +84,8 @@ class SaveModal extends React.Component {
...
+        this.props.resource.attributes && this.props.resource.attributes.owner === this.props.user.name;
```
MV88:
```suggestion
        this.props.resource && this.props.resource.attributes && this.props.resource.attributes.owner === this.props.user.name;
```

web/client/components/resources/modals/Save.jsx
kasongoyo: I have initialized user in default prop. @MV88

web/client/components/resources/modals/Save.jsx
kasongoyo: Resource is also initialized in default prop.

web/client/components/resources/modals/Save.jsx
MV88: user can be null and the default props, works only with undefined values. please update pr

web/client/components/resources/modals/Save.jsx
MV88: not sure if resource can be null, but i would do the fix anyway

PR #4639, opened 2019-12-11
Application Context Plugins

vlt1:
## Description
Adds an ability to add and configure plugins in application contexts

## Issues
 - [mapstore2-georchestra/#20](https://github.com/georchestra/mapstore2-georchestra/issues/20)
 - [mapstore2-georchestra/#22](https://github.com/georchestra/mapstore2-georchestra/issues/22)

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

 - [x] No


web/client/localConfig.json
```diff
@@ -345,7 +345,10 @@
                 }
             },
             "AddGroup",
-            "TOCItemsSettings",
+            {
+                "name": "TOCItemsSettings",
+                "parentPlugin": "TOC"
```
mbarto: I don't like this: we already have containement information through the plugin containers configuration. Can you try to use that instead?

web/client/components/plugins/PluginsContainer.jsx
```diff
@@ -115,7 +115,7 @@ class PluginsContainer extends React.Component {
     };
 
     renderPlugins = (plugins) => {
-        return plugins
+        return plugins && plugins
```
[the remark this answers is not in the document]
vlt1: Yes, it's not a good idea and technically on real data it's an unreached state.
Thanks, I think default params and check on array would safer.

web/client/components/misc/DoubleColumnTransfer.jsx
```diff
@@ -0,0 +1,152 @@
...
+                items.length > 0 && filterText.length > 0 ? emptyStateSearchProps : emptyStateProps)}
+            side={side}
+            selectedItems={selectedItems}
+            selectedSide={selectedSide}
+            allowCtrlMultiSelect={allowCtrlMultiSelect}
+            onSelect={onSelect}/>
+    </div>
+);
+
+const DoubleColumnTransfer = ({
```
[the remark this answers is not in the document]
vlt1: @mbarto hm, I have another point of view :) 
https://ant.design/components/transfer/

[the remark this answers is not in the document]
mbarto: But... in that case, the component is simply called "Transfer". I would simply use a shorter name for this general purpose components. Either Transfer or TransferList are valid in my opinion.

web/client/localConfig.json
offtherailz: I think @vlt1 need it to create this kind of structure in the mockup, isn't it?
image

In this case LayerSetting(TOCItemsSettings), as well as AttributeTable (FeatureEditor) is not related by plugins settings to the TOC. Anyway I suggest to create a separate configuration that defines the structure of the UI, strictly related to the ContextEditor, that defines parent-child relationship. 
Also the different labels (Layer Settings for TOCItemsSettings, Attribute Table for FeatureEditor and so on) should be stored somewhere.

offtherailz:
After a small discussion with @tdipisa we decided to provide the missing functionalities defined today in the context of this PR. 

So please take into account the format described [here](https://github.com/georchestra/mapstore2-georchestra/issues/15#issuecomment-564974270) to add these functionalties:
- ### Mandatory plugins (that can't be moved, are on right column by default
- **Hidden**: not shown in UI (typically mandatory, dependencies or children
- **children**: support to indicate which widgets are children of 
- **dependencies** the plugins as dependencies must be included if at least one of the  plugins defined as dependent (has its name in the dependencies array) is present.
- **title, description, symbology** (an icon at the left of the plugin, see [here](https://demo.geo-solutions.it/mockups/mapstore2/georchestra-templates-extensions/#/) --> burger menu/mapExtension ) for a better look and feel

Moreover:
 - [ ] MapType should be specified explicitely by Map configuration (openlayers by default)

This configuration described at the end of the project will be a mix of : 
 - Plugins definitions
 - app configuration
 - Extensions coming from the server

For the moment you can store it statically in a separated json file.

web/client/epics/contextcreator.js
```diff
@@ -167,44 +222,228 @@ export const mapViewerReload = (action$, store) => action$
...
+
+        return Rx.Observable.of(enablePlugins(plugins.filter(plugin => plugin.mandatory).map(plugin => plugin.name)));
+    });
+
+/**
+ * Handles plugin enabling
+ * @param {observable} action$ manages `ENABLE_PLUGINS`
+ * @param {object} store
+ */
+export const enablePluginsEpic = (action$, store) => action$
```
offtherailz: You should test this epic and describe a little bit more its behavior in doc

web/client/epics/contextcreator.js
```diff
@@ -167,44 +222,228 @@ export const mapViewerReload = (action$, store) => action$
...
+            ))
+        );
+    });
+
+/**
+ * Handles plugin disabling
+ * @param {observable} action$ manages `DISABLE_PLUGINS`
+ * @param {object} store
+ */
+export const disablePluginsEpic = (action$, store) => action$
```
offtherailz: same for this, test and describe

web/client/translations/data.it-IT.json
```diff
@@ -2110,6 +2110,12 @@
             "back": "Indietro",
             "next": "Avanti"
         },
+        "plugins": {
+            "TOC": {
+                "title": "Elenco di livelli",
```
offtherailz:
```suggestion
                "title": "Elenco dei livelli",
```

web/client/translations/data.it-IT.json
```diff
@@ -2110,6 +2110,12 @@
...
+                "description": "Fornisce un elenco di livelli con la possibilità di ordinarli"
```
offtherailz:
```suggestion
                "description": "Fornisce l'elenco dei livelli con la possibilità di ordinarli"
```

web/client/plugins/Map.jsx
```diff
@@ -182,6 +183,7 @@ let plugins;
 class MapPlugin extends React.Component {
     static propTypes = {
         mapType: PropTypes.string,
+        mapTypeOverride: PropTypes.string,
```
mbarto: Can we implement this behabiour without duplicating mapType ?
mapType is a "global" setting, impacting different plugins. Maybe we should add an option in ContextCreator for it instead of using a Map plugin property.
We already have a UI component (a combobox) to select mapType. We can eventually put that in one of the contextcreator steps.
What do you think?

vlt1: @mbarto good point of view, but it requires not estimated time for this improvement.

offtherailz: mmm... yes, can we isolate it in a separate enhancer (with a comment) so we can remove it as far as we found a better solution?
I agree with @mbarto that this plugin, because of it's general purpose, should be as much clear as possible. 
Maybe before connect, so we can use mapType property.



something like 
```javascript
/**
 * If mapType coming as prop ( e.g. from a custom context, it updates the global maptype)
 */
const handleMapType = compose(
    connect(() => ({}), { changeMapType },
    lifecycle({
       componentDidMount()  => { // use this.props.mapType }
   })
)
//...
MapPlugin: compose( 
  handleMapType,
  connect(selector, {
        onFontError: errorLoadingFont,
        onResolutionsChange: setMapResolutions
        onResolutionsChange: setMapResolutions,
        changeMapType
    })(MapPlugin),
```

 What do you think, @mbarto, can this improve a little bit? I know a good solution should be to setup state to load.

In addition I notice that this doesn't stash maptype, so after this the map type will change for all. This is not strictly related to this issue and happens also during navigation, if you manually change the map type from the URL, but anyway we should evaluate and maybe open an issue about it

mbarto: ok for me

PR #4820, opened 2020-01-30
Map templates - WMC

vlt1:
## Description
Very basic implementation of WMC loading

P.S. contains maptemplates branch as it depends on it

How conversion happens:

It gets maxExtent from BoundingBox tag or, if available, from openlayers maxExtent tag from extensions. SRS of the BoundingBox is assumed to be the projection
If generateLayersGroup is true, then all layers are going to be put in the separate group with the title of the context(Title tag inside General), otherwise group for layers is going to be undefined therefore they are going to be in the Default group.
Generated layers correspond to Layer elements inside LayerList
Each layer is going to have the following props:

* **id** - uses generated uuid
* **visibility** - according to the "hidden" attribute of Layer tag
* **type** - according to "service" attribute of Server tag. Currently recognizes only 'OGC:WMS' mapping it to 'wms'
* **url** - extracted from OnlineResource inside Server
* **name** - according to Name tag
* **title** - according to Title tag
* **format** - extracted from FormatList Format elements(takes the first one that has its "current" attribute set to true
* **style** - extracted from StyleList similarly to FormatList
* **bbox** - if ol:maxExtent from Extension is present put it in bbox.bounds
* **group** - depends on generateLayersGroup as described above

Other features of WMC that need to be considered:
* FormatList and StyleList are lists of all possible formats and styles. Also styles can be in SLD(Style Layer Descriptor) which is represented in a wmc file by a link to an SLD document.
* Styles can have legendURL that points to the location of a map legend describing current style(extracted from Capabilities by context document creator). This element is optional.
* WMC can have sld:MinScaleDenominator and sld:MaxScaleDenominator that define min/max scale at which the particular layer should be visible.
* Layers also are **required** to have a "queryable" attribute that denotes whether the layer can be queried
* The sample contexts also have a bunch of openlayer tag elements inside Extension(for example, ol:maxExtent). According to the spec, these are not a part of interoperable WMC format and basically can have absolutely anything. Not sure if they should be processed and how, or simply ignored.
* Layers can also have DataURL and MetadataURL that point to data or descriptive metadata respectively, corresponding to the layer. Both are optional

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
[mapstore2-georchestra/#151](https://github.com/georchestra/mapstore2-georchestra/issues/151)

## Breaking change
 - [x] No


offtherailz:
thanks. I'm AFK all the day.
@vlt1 can you write (in the PR) how you're mapping values from one format to another please, to allow me to read the code without doing reverse-engineering.

With some samples of conversions (unit tests should be useful too)

PR #4718, opened 2020-01-14
Change UI of the import vector files functionality. Remove the widget…

UMarkau:
…s button from the TOC toolbar for imported files.

Add translations for needed phrases. Modify tests to fit new functionality.

## Description
Change UI of the import vector files functionality. Remove the widgets button from the TOC toolbar for imported files.

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix

## Issue

**What is the current behavior?**
#4678 

**What is the new behavior?**
UI of the import vector files functionality has been changed. Removed the widgets button from the TOC toolbar for imported files.

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #4856, opened 2020-02-06
Application Context Plugins various enhancements

vlt1:
## Description
This is the MapStore part of [the issue](https://github.com/georchestra/mapstore2-georchestra/issues/155). The GeOrchestra part is [here](https://github.com/georchestra/mapstore2-georchestra/pull/159).

Configure Plugins step of context creator was improved. It was identified that dependencies were not resolved correctly if not on the same level in plugin tree, that was fixed. Due to new requirements autoEnableChildren work not only if enabled directly but when enabled through a dependency as well, also now children plugins can have mandatory flag on them, that will make that plugin mandatory, if it's parent is enabled.

Comments on some of the points:

> The Query panel plugin should probably be hidden from plugin list, and should be enabled by default when others plugins that use the Query panel are added (like Attribute Table, layer filter, Widgets). There's probably a problem related to Query panel plugin because sometimes it enables adding the List of Layers plugin and sometimes does not). Moreover I can add this plugin alone also if it depends from other plugins.

As @offtherailz pointed out it is configurable so it should stay listed. Also I didn't get the "There's probably a problem related to Query panel plugin because sometimes it enables adding the List of Layers plugin and sometimes does not)" part. This plugin can be added alone because it has no dependencies.

> Style Editor plugin what is it for? Seems that nothing change adding or not this plugin (if this plugin serves to create new styles it doesn't work). Moreover this plugin can be added alone, also if it probably depends from TOC and from layer settings

Style Editor is a style editor in TOC layer settings. It seems to be directly handled by TOCItemsSettings, and so is always enabled when layer settings button is present. I've made StyleEditor a child of TOCItemsSettings plugin and made it mandatory.

> Widgets plugin should depends from List of Layers plugin, but I can actually add it alone. If you add both Widgets plugin and List of Layers plugin, the widgets button is not present in toc selecting a layer

The widgets button not being present in toc is not a context specific issue. It seems it was introduced with [this change](https://github.com/geosolutions-it/MapStore2/pull/4718/files#diff-fce06684499b5cd749478bbc3c2bd4a9). It effectively disabled the create widget button for all layers in TOC. That was rolled back.

> The show metadata option should be disabled by default, so the related button in TOC toolbar does not appears (the button is automatically enabled when you ad a layer from a CSW source)

I've added "activateMetedataTool: false" to TOC defaultConfig in pluginsConfig.json. Is this correct?

> Adding "Save" plugin, the Burger Menu is added together but after the save option is not present in the Menu

The Save plugin is not present by design. It is hiddent, if there is no id in mapInfo, which there isn't, because the map is loaded from context, not from a MAP resource and so mapId is null. In any case Save should be either hidden from the list, or it's functionality should be expanded to deal with contexts.

> The Download data from attribute table plugin can be added alone, but should be related to Attribute Table plugin

What is an "Attribute Table" plugin? Is it FeatureEditor? And what is "Download data"???

Also testEpic helper function was expanded a little to allow to use in a way, so that when expect fails inside a callback it actually prints out the error instead of a test just timing out.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
[mapstore2-georchestra/#155](https://github.com/georchestra/mapstore2-georchestra/issues/155)

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/TOC/Toolbar.jsx
+++ b/web/client/components/TOC/Toolbar.jsx
@@ -282,7 +282,7 @@ class Toolbar extends React.Component {
                             </Button>
                         </OverlayTrigger>
                         : null}
-                    {this.props.activateTool.activateWidgetTool && (status === 'LAYER') && !this.props.selectedLayers.length && !this.props.settings.expanded && !this.props.layerMetadata.expanded && !this.props.wfsdownload.expanded ?
+                    {this.props.activateTool.activateWidgetTool && (status === 'LAYER') && this.props.selectedLayers.length === 1 && !this.props.settings.expanded && !this.props.layerMetadata.expanded && !this.props.wfsdownload.expanded ?
                         <OverlayTrigger
                             key="widgets"
                             placement="top"
--- a/web/client/components/contextcreator/ConfigureMapStep.jsx
+++ b/web/client/components/contextcreator/ConfigureMapStep.jsx
@@ -24,6 +24,7 @@ export default ({
 }) => (
     <React.Fragment>
         <ConfirmDialog
+            draggable={false}
             modal
             show={showConfirm}
             onClose={() => {
@@ -34,6 +35,7 @@ export default ({
                 onMapViewerReload();
             }}
             confirmButtonBSStyle="default"
+            closeGlyph="1-close"
             confirmButtonContent={<Message msgId="confirm"/>}
             closeText={<Message msgId="cancel"/>}
         >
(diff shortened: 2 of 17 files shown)


offtherailz:
Hi, about query panel, they missed the sequence of step to reproduce: 
If you add List of layers TOC then QP is added automatically and can not be removed
If you reset and add first the query panel first, then the TOC, the QP can be removed (and it shouldn't). 

@tdipisa what if : 
- Add the QP
- Add the TOC
- Remove the TOC

The QP should go away ? I think the issue is that QP can be added manually, why, maybe, this should be denied, to avoid any problem of this kind.

vlt1:
@offtherailz QueryPanel is a dependency of the two of the four child plugins of TOC. When these plugins are disabled, if QueryPanel has no enabled dependencies, it is automatically disabled. So, when TOC is disabled, it's children are disabled too, so QueryPanel would be disabled automatically in this case. Also, the issue that you described was present due to incorrect handling of dependencies that are on different levels of the plugin tree. As stated in the PR comment, that's been fixed.

PR #4890, opened 2020-02-11
#4876. Introduced noRoot option to skip root rendering playback

offtherailz:
## Description
 Introduced noRoot option to skip root rendering of plugins that doesn't make sense in a stand-alone context. This solves also the problem notified in #4876 setting `noRoot` option for the Playback plugin. 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#4876

**What is the new behavior?**
Playback is rendered only in timeline. 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/plugins/PluginsContainer.jsx
+++ b/web/client/components/plugins/PluginsContainer.jsx
@@ -180,8 +180,8 @@ class PluginsContainer extends React.Component {
             0
         );
         // render on root if container is root (plugin === null || plugin.impl === null) or plugin explicitly wants to render on root (doNotHide = true)
-        // in addition to the container
-        return !container.plugin || !container.plugin.impl || container.plugin.impl.doNotHide;
+        // in addition to the container. If the plugin impl has option "noRoot", rendering is skipped for root by default.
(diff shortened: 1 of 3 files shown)


offtherailz:
@mbarto 
Good point. No, it can't. It have been developed to separate concerns of playback from the specific timeline implementation. Anyway, as you say, it's only a part of the timeline. 
Apart from it (maybe some future refactor can join the two plugin, `noRoot` make sense per-se, right? It is not a duplication of some existing option that I ignore, isn't it?

mbarto:
Time of refactor is now, in my opinion, because this solves the original issue, and does not introduce more complexity (new configuration options)

offtherailz:
A refactor like this is 0 gain, can introduce many bugs a requires a lot of work. And the result may not surely be better then the previous. 
Adding one (optional) configuration option, that makes sense in any case, ... well ... I think is a good compromise, in terms of cost-benefits. It is fully covered by tests, 1 line of code, so maintainance should not cost that much ( I think nothing )

mbarto:
@offtherailz if you prefer I can do the refactor work by myself

offtherailz:
I see this is the only point. I'm going to merge, we need this now.

PR #4891, opened 2020-02-11
#4735 add background visible only with catalog

offtherailz:
## Description
See #4735. With this changes background selector shows add button only if catalog is present. 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#4735

**What is the new behavior?**
Now background add button is hidden when catalog is not present.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/background/BackgroundSelector.jsx
+++ b/web/client/components/background/BackgroundSelector.jsx
@@ -39,6 +39,8 @@ class BackgroundSelector extends React.Component {
         onLayerChange: PropTypes.func,
         onStartChange: PropTypes.func,
         onAdd: PropTypes.func,
+        hasCatalog: PropTypes.bool,
+        alwaysVisible: PropTypes.bool, // useful for tests
         enabledCatalog: PropTypes.bool,
         onRemove: PropTypes.func,
         onBackgroundEdit: PropTypes.func,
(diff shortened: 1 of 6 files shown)


web/client/plugins/BackgroundSelector.jsx
```diff
@@ -88,30 +92,36 @@ const backgroundSelector = createSelector([
...
-}, (stateProps, dispatchProps, ownProps) => ({
-    ...stateProps,
-    ...dispatchProps,
-    ...ownProps,
-    thumbs: {
-        ...thumbs,
-        ...ownProps.thumbs
-    }
-})
+const BackgroundSelectorPlugin =
```
MV88: Hi, i think this little refactor to recompose utilities was not so necessary.

web/client/components/background/BackgroundSelector.jsx
```diff
@@ -39,6 +39,8 @@ class BackgroundSelector extends React.Component {
         onLayerChange: PropTypes.func,
         onStartChange: PropTypes.func,
         onAdd: PropTypes.func,
+        hasCatalog: PropTypes.bool,
```
MV88: I have a question, With this we are binding the BackgroundSelector component to something else like a Catalog.
Haven't we always said that this should be something hidden from the Receiver perspective.
I know Stefano has did something similar where pluginSender inject tools in pluginReceiver.
My opinion is to discuss this in the next dev meeting

offtherailz: sure, this has been done to avoid refactors of complex nested components inside these elements. I added a TODO for this. 
I'll also open an issue about these improvements for the future.

web/client/plugins/BackgroundSelector.jsx
offtherailz: I only added hasCatalog property calculation from props.

Issue #4865, opened 2020-02-07
Investigate on Styler pluggability


offtherailz:
Style actually do not switch on/off using User Extensions tool. Investigate feasibility of this or, as B plan, disable possibility to add it to user extensions

PR #4907, opened 2020-02-18
#4865: Improved styler pluggability

offtherailz:
## Description
This PR allows to plug/unplug style editor using user extensions.
It allows also to keep ThematicLayers plugin compatible and pluggable too (it has priority over style editor at the moment, if present. If we want to make the 2 tools alternative, we need to restructure the UI).

To test locally thematicLayers pluggability, use this context http://localhost:8081/#/context-creator/21352 I created it by adding to pluginsConfig thematic plugin and using GUI to add it as user extension, together with Style editor. Thematic maps doesn't work at all, because the SLD service is missing, but it will plug and unplug correctly.

 - [x] Feature

## Issue

#4865 the style editor couldn't be used as user extension

**What is the new behavior?**
Now the style editor plugs and unplugs using user extensions

## Breaking change
 - [x] No

## Other useful information
**For DEVs/Reviewer**
This PR doesn't solve at all TOCItemsSettings pluggability problems (we should work a little bit more for this) but externalize some of the behaviours.

### DIFF
--- a/web/client/components/TOC/TOCItemsSettings.jsx
+++ b/web/client/components/TOC/TOCItemsSettings.jsx
@@ -29,7 +29,7 @@ const Message = require('../I18N/Message');
  * @prop {string} className additional calss name
  */
 
-const TOCItemSettings = (props, context) => {
+const TOCItemSettings = (props) => {
     const {
         className = '',
         activeTab = 'general',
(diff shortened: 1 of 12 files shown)


offtherailz:
I noticed that loading spinner doesn't show anymore. I tried to straightly apply the enhancer but it doesn't work, because it triggers capabilities load continuously. It may need to be applied after the capabilities trigger, or make it conditional.
Going to investigate.

Issue #4460, opened 2019-11-07
Application Context - MS Landing Page


tdipisa:
### Description
Some updates are needed in MS home page to manage context maps.

1 - Possibility to filter maps by context to show resources belonging to a specific context. It should be possible to enable/disable the context filter by configuration file (localConfig). 

**Note:** This filter tool could be improved in the future to host additional filtering capabilities (like data, keywords etc).

The mockup for this is available [here](https://demo.geo-solutions.it/mockups/mapstore2/georchestra-search-filter/#/):

![image](https://user-images.githubusercontent.com/1280027/72161143-d98eb580-33bf-11ea-9047-ca6a46e24e00.png)

2 - Possibility to create maps starting from a created application context instead of the default one: clicking on "New Map" button a modal opens to show the list of app context the user can access to. If one of them is clicked the viewer opens and the user can create the map as usual.

3- If you have not access to the context you now see the full screen message that explain that the resource is not accessible. You have to add the possibility (this functionality can be turned off via configuration) to redirect the user to the default viewer, sending a warning message:
`The context of this map is not accessible or has been deleted, you have been redirected to the default viewer`

### Other useful information (optional):
Generally speaking Users that are able to access an application context defined in MS, should be able to create maps from it, this means maps with a final URL like the one below:

https://dev.mapstore2.geo-solutions.it/mapstore/#/context/{ContextName}/{MapId}

vlt1:
When context is deleted, do we need to find maps that belong to this context and delete them as well?

offtherailz:
@vlt1 I opened an issue about this on geostore, 

In order to batch update all the resources.
[see this](https://github.com/geosolutions-it/geostore/issues/205)
We can move the task of "updating maps when context has been updated in another issue

offtherailz:
@vlt1 for the moment the only thing you need to add to this task is, in case of missing or not accessible context, redirect to the main viewer to see the map. A warnig message will notify the user that the context is not accessible, so he has been redirected to the default viewer. This behaviour should be turned on or off via configuration.

ElenaGallo:
@tdipisa @vlt1 

1_When log out from MS, the filter should be disabled and the pop up closed
![filtercontext](https://user-images.githubusercontent.com/56537133/76546189-56efa880-648b-11ea-8222-d512df2e200c.gif)

**How to reproduce**
- open MapStore
- Login as admin
- filter a context from search bar
- log out 

*Current Result*
[...]

PR #4934, opened 2020-02-25
Application Context - MS Landing Page

vlt1:
## Description
Adds advanced search filter and modifies new map

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#4460 

## Breaking change
 - [x] No


web/client/epics/maps.js
```diff
@@ -210,17 +218,148 @@ const reloadMapsEpic = (action$, { getState = () => { } }) =>
...
+                    ...(opts.params || {}),
+                    includeAttributes: true
+                }
+            })
+                .switchMap(({results, totalCount}) => {
+                    const maps = {
+                        results: results.map(result => ({
+                            ...omit(result, 'attributes', 'permissions'),
+                            ...pick(result.attributes, 'thumbnail', 'context'),
+                            canCopy: userRole === 'ADMIN',
```
offtherailz: Correct me if I'm wrong. You added this to results because the remove service doesn't provide this information (canCopy, canDelete, canEdit), isn't it? 
We can not assume that edit/delete are allowed only to admin user. Maybe we need to extend the entry point.

web/client/components/mapcontrols/search/SearchBar.jsx
```diff
@@ -13,6 +13,7 @@ const LocaleUtils = require('../../../utils/LocaleUtils');
 const DropdownToolbarOptions = require('../../misc/toolbar/DropdownToolbarOptions');
 const CoordinateEntry = require('../../misc/coordinateeditors/CoordinateEntry');
 const Toolbar = require('../../misc/toolbar/Toolbar');
+const AdvancedSearch = require('./AdvancedSearch').default;
```
offtherailz: SearchBar is used both for map search and for GeoCoder search. I think this is wrong. I suggest to generalize this or separate the two components.

web/client/epics/maps.js
vlt1: I've extended the searchListByAttributes handling function to handle option "withPermissions" which when true, gets permissions for every returned resource with additional getResourcePermissions calls. It returns read/write permission though. canDelete and canEdit could be set to true if the user can write and canCopy if the user can read or write. Should I do this?

web/client/components/mapcontrols/search/SearchBar.jsx
vlt1: Generalize in what way? AdvancedSearch tools and SearchBar parts are controlled with props. Both can enable the parts of the component that are appropriate for them. SearchBar is also used for managers(user, group, context)

web/client/epics/maps.js
offtherailz: You should read the full permission list and guess if you have permission or not. I prefer if this information is provided by the server and anyway it requires a request for each result. I prefer if the server returns these information directly

vlt1: > I prefer if the server returns these information directly

That would be ideal, yes.

web/client/components/mapcontrols/search/SearchBar.jsx
offtherailz: I should generalize the SearchBar with buttons (other tools) and a panel to render as children. Then I should provide one implementation with the buttons and the tools  of the map and another with the buttons of the home page and with the advancedsearch as children

offtherailz: sorry but yesterday github had some issues and you didn't received this comment

vlt1:
> Great PagedSelect component, very useful! We had PagedCombobox but it had some graphic glitches that this haven't. It could be a great thing if we could only one component for this. IS it possible

You want me to change code that uses PagedCombobox to use PagedSelect?

offtherailz:
@vlt1 mmm... Maybe PagedSelect and PagedCombobox have 2 different scopes, and unifying them at this stage is a little out of scope. Please continue with yours and ignore PagedCombobox

offtherailz:
All the important tasks seems to be checked, did you finished this task?
If so please solve the conflicts and ask for review. Thank you :+1:

vlt1:
@offtherailz what do you mean searchbar takes more time to load? which searchbar?

Issue #4954, opened 2020-02-28
Introduce user session base support


mbarto:
## Description
Needed by https://github.com/georchestra/mapstore2-georchestra/issues/36.

We want the ability to periodically persist the user session as a geostore resource.
The session details will be configurable by project (no default):
 * session identifier
 * session content (state fragments to include)
 
## Acceptance criteria

- [x] epic creator to save the user session, triggered by a specific action
- [x] epic creator to schedule periodic persistence of user session
- [x] epic creator to load user session at request
- [x] epic creator to remove user session
- [x] actions and reducers to manage current user session status

tdipisa:
We leave this in progress untill the georchestra work is completed (#36 and #20) then we will open a new issue for the enablement in MS (using probably localStorage).

tdipisa:
@mbarto moved back in ready as reported [here](https://github.com/georchestra/mapstore2-georchestra/issues/36#issuecomment-625222957). Can you please provide the fix on Monday morning?

PR #4955, opened 2020-02-28
#4954: basic support for user session persistence

mbarto:
## Description

Basic infrastructure for user session persistence.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#4954

### DIFF
--- a/web/client/epics/usersession.js
+++ b/web/client/epics/usersession.js
@@ -26,7 +26,7 @@ const saveUserSessionErrorStatusToMessage = (status) => {
  *
  * @param {function} nameSelector selector that builds the session identifier
  * @param {*} sessionSelector selector that builds the session object to save
- * @param {*} idSelecotr selector of the identifier for the current session (if it already exists)
+ * @param {*} idSelector selector of the identifier for the current session (if it already exists)
  * @example
  *
  * const nameSelector = (state) => ({state.context.name + "." + state.security.user.name})
--- a/web/client/epics/usersession.js
+++ b/web/client/epics/usersession.js
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2020, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import Rx from 'rxjs';
+import { error } from '../actions/notifications';
+import { SAVE_USER_SESSION, userSessionSaved, loading, saveUserSession } from "../actions/usersession";
+import { createResource, createCategory, updateResource } from '../api/persistence';
+import {userSelector} from '../selectors/security';
+import { wrapStartStop } from '../observables/epics';
+import isString from "lodash/isString";
+
+const saveUserSessionErrorStatusToMessage = (status) => {
+    switch (status) {
+    default:
+        return 'userSession.saveErrorNotification.defaultMessage';
+    }
+};
+
+/**
+ * Returns a user session save epic.
+ * The epic triggers on a SAVE_USER_SESSION action.
+ *
+ * @param {function} nameSelector selector that builds the session identifier
+ * @param {*} sessionSelector selector that builds the session object to save
+ * @param {*} idSelecotr selector of the identifier for the current session (if it already exists)
+ * @example
+ *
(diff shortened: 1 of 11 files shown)


tdipisa:
@mbarto as we agreed in the call, I think that an update to the DEV guide is needed to properly understand how to enable this support and setup it. Thank you.

web/client/epics/usersession.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+};
+
+/**
+ * Returns a user session save epic.
+ * The epic triggers on a SAVE_USER_SESSION action.
+ *
+ * @param {function} nameSelector selector that builds the session identifier
+ * @param {*} sessionSelector selector that builds the session object to save
+ * @param {*} idSelecotr selector of the identifier for the current session (if it already exists)
```
offtherailz:
```suggestion
 * @param {*} idSelector selector of the identifier for the current session (if it already exists)
```
Is not clear to me the role of the session name and the role of the session id. Can you clarify in JSDoc (maybe the right place is the reducer)

web/client/epics/usersession.js
```diff
@@ -0,0 +1,94 @@
...
+ * @example
+ *
+ * const nameSelector = (state) => ({state.context.name + "." + state.security.user.name})
+ * const sessionSelector = (state) => ({state.map.present})
+ * const idSelector = (state) => ({state.usersession.id})
+ * const epic = saveUserSessionEpic(nameSelector, sessionSelector, idSelector)
+ * setInterval(() => dispatch({type: SAVE_USER_SESSION}), 60 * 1000)
+ */
+export const saveUserSessionEpic = (nameSelector, sessionSelector, idSelector = () => {}) => (action$, store) => action$
```
offtherailz: because they are not epics, maybe a different naming, like `saveSessionEpicCreator` is more descriptive... or any better idea, if you can think any

web/client/epics/usersession.js
```diff
@@ -0,0 +1,94 @@
...
+                        autoDismiss: 5,
+                        values: {
+                            data
+                        }
+                    }));
+                }
+            ));
+    });
+
+export const startStopUserSessionSaveEpic = (startAction, endAction, frequency) => (action$) => action$
```
offtherailz: The same here. We are not forced to have unique names for these epicCreators, therefore we can simple call it something like `autoSaveEpicCreator` or `autoSaveSessionEpicCreator`

web/client/reducers/usersession.js
```diff
@@ -0,0 +1,22 @@
+import { USER_SESSION_SAVED, USER_SESSION_LOADING } from "../actions/usersession";
```
offtherailz:
```suggestion
/*
 * Copyright 2020, GeoSolutions Sas.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree.
 */
import { USER_SESSION_SAVED, USER_SESSION_LOADING } from "../actions/usersession";
```

web/client/selectors/usersession.js
```diff
@@ -0,0 +1,2 @@
+export const userSessionIdSelector = (state) => state.usersession && state.usersession.id || null;
```
offtherailz:
```suggestion
/*
 * Copyright 2020, GeoSolutions Sas.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree.
 */
export const userSessionIdSelector = (state) => state.usersession && state.usersession.id || null;
```

web/client/epics/usersession.js
mbarto: saveSessionEpicCreator is fine

web/client/epics/usersession.js
mbarto: fine

mbarto:
> Cool 👍
> only a typo and some name change or doc proposal.
> 
> I didn't figure out well the reason of the name of the session, maybe an implementation can clarify the reason of this ( and epics creators )

the name of the resource for the session is configurable, so that we can decide to save global user sessions (using only the user id as the name) or by context / map (and then we need to have a more long name to store / retrieve the session)

mbarto:
@offtherailz just made a new commit

Issue #4910, opened 2020-02-18
Contents - Map - Popups


kappu72:
@tdipisa commented on [Tue Feb 19 2019](https://github.com/geosolutions-it/MapStore2-C039/issues/52)

Actually MS has not popups. We need to design how to display data for clicked points. 

Investigation:

- [x] Evaluate possible solution to implement a Popup component as an alternative for the Identify tool (the Popup must be available for both OL and Leaflet). As an implementation suggestion it should be implemented as a support component like the DrawSupport and MeasurementSupport. Provide the possibility to have the popup as an alternative visualizzation of the Identify tool, this must be configured in plugin config.

Write a small document with the results of the investigation and the proposed solution.

Then:

- [ ] The identify tool needs to be available for layers/vector features in the map with a popup as the Identify tool currently does
- [ ] The identify can be activated by the Map Inline Editor to show or not the info popup if the user click on the map (by default disabled)

---

@ghost commented on [Mon Dec 02 2019](https://github.com/geosolutions-it/MapStore2-C039/issues/52#issuecomment-560400155)

Both tools (Openlayers and Leaflet) provides API for popup handling.  
 
Openlayers  provides API to create popup instance:
https://openlayers.org/en/latest/apidoc/module-ol_Overlay-Overlay.html
Overlay should be added via `addOverlay` method on map instance:
https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#addOverlay
Example of working popup:
https://openlayers.org/en/latest/examples/popup.html?q=popup

In leaflet there is also specific API for create popup instance:
https://leafletjs.com/reference-1.6.0.html#popup
But in this case it can be provided via marker/map feature or adding as layer

Based on MapStore2 code main element of popup which will be displayed can be created as single component. It can be used in both solutions (OpenLayers and Leaflet). For adding support of popup it can be managed like DrawSupport or MeasurementSupport written separately for each library.

---

@tdipisa commented on [Mon Dec 02 2019](https://github.com/geosolutions-it/MapStore2-C039/issues/52#issuecomment-560413168)

@offtherailz can you please provide your opinion on the above comment?

---

@offtherailz commented on [Fri Dec 06 2019](https://github.com/geosolutions-it/MapStore2-C039/issues/52#issuecomment-562194584)

sure, this is more or less the thinked solution. I'd like to have a prototype of this <PopupSupport> so it can be used in our maps.
[...]

PR #4973, opened 2020-03-04
Fix 4910 Contents - Map - Popups

kappu72:
## Description
Adds popup support to Ol and leaflet. To ad a popup just create one with an appPopup action,
passing point coordinates and a content (text or html) or a component (React component or element)
To show the mapInfo response in a popup just add "showInMapPopup": true in identify cfg.
This PR adds also identify with popup support to geostory.

 - [X] Feature

## Issue
#4910 

#4910

**What is the new behavior?**

#4910
## Breaking change
 - [x] No


web/client/plugins/Identify.jsx
```diff
@@ -189,7 +190,12 @@ const IdentifyPlugin = compose(
         showRevGeocode: showMapinfoRevGeocode,
         hideRevGeocode: hideMapinfoRevGeocode,
         onEnableCenterToMarker: updateCenterToMarker.bind(null, 'enabled')
-    }),
+    }, (stateProps, dispatchProps, ownProps) => ({
+        ...ownProps,
+        ...stateProps,
+        ...dispatchProps,
+        enabled: stateProps.enabled && (stateProps.isCesium || !ownProps.showInMapPopup)
```
offtherailz: So this is the property? `showInMapPopup` should be documented.

web/client/components/geostory/common/map/Controls.jsx
```diff
@@ -86,7 +86,7 @@ export const Controls = ({
             />
             <FeatureInfoFormatSelector
                 disabled={!options.mapInfoControl}
-                infoFormat={options.mapOptions && options.mapOptions.mapInfoFormat || "text/plain"}
+                infoFormat={options.mapOptions && options.mapOptions.mapInfoFormat || "application/json"}
```
offtherailz: why did you changed the default format? 
`text/plain` is compatible with all servers. application/json is available only on GeoServer, so in some cases it may not work.

Issue #5013, opened 2020-03-12
Improve testing technologies


mbarto:
## Description

We would like to improve our automatic testing infrastructure.
These are some things we could do:
 - [ ] upgrade expect library to a newer version
 - [ ] investigate on switching to jest instead of mochajs

PR #5014, opened 2020-03-12
#5013: upgrade expect to latest version

mbarto:
## Description

Upgrade of expect library to:
 * expect 25.1.0 (latest from jest)
 * jest-mock 25.1.0 (contains the mocking / stubbing / spying functionalities of the old jest)

 - [x] Other... Please describe: library upgrade

## Breaking change
 - [ ] No

Tests need to be updated to some syntax changes in newest version.
The most important ones:
 * toExist -> toBeTruthy
 * toNotExist -> toBeFalsy
 * toNotSomething -> .not.toSomething
 * spy.calls -> spy.mock.calls
 * spy.calls[n].arguments -> spy.mock.calls[n]
 * spies always call the original function, you can mock it using mockImplementation


build/expect.customize.js
```diff
@@ -0,0 +1,41 @@
+import expect from 'expect';
+import jest from "jest-mock";
```
MV88: Hi, i would call this JestMock, what do you think?

web/client/utils/StateUtils.js
```diff
@@ -125,7 +125,7 @@ const fetchEpic = (storeName = PERSISTED_STORE_NAME, name = 'rootEpic') => {
  * @param {string} name optional name (if you want to persist more than one store)
  */
 export const getState = (name) => {
-    return getStore(name) && getStore(name).getState() || {};
+    return getStore(name)?.getState?.() ?? {};
```
MV88: i think other falsy values are useful here, don't they?
for example if getState returns for a strange reason "" or 0, would be better to return a {} right instead of them.

i know i know, probably i'm doing too much mental saws

web/client/components/misc/__tests__/Filter.jsx
```diff
@@ -36,7 +36,7 @@ describe('TOC Filter', () => {
         const cmp = ReactDOM.render(<Filter onFilter={actions.onFilter}
             onFocus={actions.onFocus}/>, document.getElementById("container"));
         const el = ReactDOM.findDOMNode(cmp);
-        expect(el).toExist();
+        expect(el).toBeTruthy();
```
MV88: this is msssing -test  in the filename

build/expect.customize.js
MV88: Is this a temporary solution that we are going to remove if/when we migrate to jest ??

mbarto: I tried to mantain compability where it made sense: spyOn is now part of jest, not of expect anymore, but for now I would not make jest presence explicit.
About the name, yes, np.

web/client/utils/StateUtils.js
mbarto: The purpose of ?? is to avoid null and undefined. 
Store getState returns an object, not a number or a string, so I think ?? is the right choice here.

offtherailz:
As well as selected the milestone, we should add the notes in the PR to migration notes (also for depending projects):

```
toExist -> toBeTruthy
toNotExist -> toBeFalsy
toNotSomething -> .not.toSomething
spy.calls -> spy.mock.calls
spy.calls[n].arguments -> spy.mock.calls[n]
spies always call the original function, you can mock it using mockImplementation
```

PR #5061, opened 2020-03-24
Update PluginsContainer.jsx

offtherailz:
## Description
Instead of using componentWillReceiveProps (that is applied everytime) we should use componentWillUpdate that is applied only if shouldComponentUpdate returns true (at least until rewrite this with hooks). 
This seems to speed up performances a lot. If it doesn't affect too much the dynamic plugin load system, i should propose this changes.

What do you thinj @mbarto 

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#<issue>

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/plugins/PluginsContainer.jsx
+++ b/web/client/components/plugins/PluginsContainer.jsx
@@ -89,21 +89,20 @@ class PluginsContainer extends React.Component {
         this.loadPlugins(this.props.pluginsState, this.props);
     }
 
-    UNSAFE_componentWillReceiveProps(newProps) {
-        this.loadPlugins(newProps.pluginsState, newProps);
-    }
-
     shouldComponentUpdate(nextProps, nextState) {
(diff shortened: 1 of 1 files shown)


mbarto:
@offtherailz do you have metrics about the performance improvement or just impressions?
Not against the change, just curious

offtherailz:
Only in test env, while digiting i noticed slowdown, so I measured and  I noticed that PluginsContainer.componentWillReceiveProps called everytime
image
Here with my changes
image

I did a round trip and I had the impression that map load, tile load and general reactivity was increased, but only in dev env. For this reason I did a Draft PR, to get your impression if possible. 
I didn't finalized the analysis yet. 'll restart working on it probably in these days.

mbarto:
@offtherailz no concerns from me

PR #5092, opened 2020-03-27
Advanced measurement tool

vlt1:
## Description
Contains initial advanced measurements tool implementation. The leaflet version does not currently have the necessary changes, so I've disabled the Measure tool for leaflet for the time being. Also the Bearing tool has an issue with the adding as a layer, that I haven't resolved yet, so I disabled it for the time being as well.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
[mapstore2-georchestra/#51](https://github.com/georchestra/mapstore2-georchestra/issues/51)

## Breaking change
 - [x] No

### DIFF
--- a/web/client/actions/measurement.js
+++ b/web/client/actions/measurement.js
@@ -9,7 +9,6 @@ const CHANGE_MEASUREMENT_TOOL = 'CHANGE_MEASUREMENT_TOOL';
 const CHANGE_MEASUREMENT_STATE = 'CHANGE_MEASUREMENT_STATE';
 const CHANGE_UOM = 'MEASUREMENT:CHANGE_UOM';
 const CHANGED_GEOMETRY = 'MEASUREMENT:CHANGED_GEOMETRY';
-const SET_STYLE = 'MEASUREMENT:SET_STYLE';
 const SET_TEXT_LABELS = 'MEASUREMENT:SET_TEXT_LABELS';
 const ADD_AS_LAYER = 'MEASUREMENT:ADD_AS_LAYER';
 const RESET_GEOMETRY = 'MEASUREMENT:RESET_GEOMETRY';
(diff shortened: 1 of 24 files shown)


web/client/actions/measurement.js
```diff
@@ -57,12 +59,33 @@ function changeUom(uom, value, previousUom) {
...
 
-function changeGeometry(feature) {
+function changeGeometry(features) {
     return {
         type: CHANGED_GEOMETRY,
-        feature
+        features
+    };
+}
+function setStyle(style) {
```
offtherailz: can you add some doc about these actions ?

tdipisa:
> @tdipisa leaflet implementation is not ready yet, so @vlt1 disabled measure tool for leaflet. Is that acceptable or we should provide a PR with the full set of changes?

Why should we have a PR that  downgrade the Leaflet part? The Measurement tool is now available  in leaflet for desktop so it must be maintained.

As reported in the issue requirements:

>The MS's Measurement tool need to be improved to provide the above enhancements.

Yes, the Bearing must be maintained.

>We should consider to have measurements displayed in the map following the respective line element.

This part is missing in the @vlt1 implementation as far as I can see from the screenshot provided by @offtherailz. Can we do that ?

> We should increase readability

Yes, maybe using the same color of the line segment for measurements or, better a color in contrast with some halo around. 

I agree with the other points highlighted by @offtherailz.

tdipisa:
@offtherailz, @vlt1 see below please, thank you.

>Maybe indicating Area: and Full Length (for length measure) or perimeter (for area measures) as labels may help to ditinguish between the various ballons

Not needed here, too much things to show that make the measurements un-readable: the unit of measure and the position of the measurements are enough to understand them. Obviously **if we use balloons** for measurement values along the segments it makes the measurement in general not really clear: **here we can use balloons only for area and total length measurements and simple text with a readable color and halo for line segments measurements.** (this is not an option for me)

>Maybe allow to disable segments measures via UI (and set it enabled/disabled by default in the plugin configuration) could help a lot readability in case you're don't need all that information.

I agree, it could be useful.

Issue #4912, opened 2020-02-19
Import/export of WMC


tdipisa:
## Description

https://github.com/georchestra/mapstore2-georchestra/issues/60

## Acceptance criteria

- https://github.com/georchestra/mapstore2-georchestra/issues/60

francoperversi:
@vlt1 An example: exporting the following map in WMC:

![export](https://user-images.githubusercontent.com/54808782/75964473-90db1080-5ec7-11ea-8853-3bb559b10f0b.jpg)

If I import it in a map or in a context the result is the following:

![import](https://user-images.githubusercontent.com/54808782/75964552-aea87580-5ec7-11ea-8105-ea3e2af392cf.jpg)

In particular, the zoom is not set properly and the the background selector contains only an empty background

@tdipisa I noticed that the WMC export format is available only for maps and not for context, is that something chosen?

tdipisa:
@francoperversi 

>@tdipisa I noticed that the WMC export format is available only for maps and not for context, is that something chosen?

WMC export should be present also for contexts, @offtherailz is there a missing configuration for that?

offtherailz:
By default (without configuratin) now MapExport exposes only MapStore format.

MapStore is configured in localConfig,json to use also wmc. 

This configuration has not been copied in pluginConfig.json (that contains all the initial configurations for plugins), so you have to add it explicitly.

Try to put this in the MapExport plugin cfg in the UI.
```
"enabledFormats": ["mapstore2", "wmc"]
```
and it should work.

I see that it looks a little confusing...

If you want always to provide wmc, we may modify code to provide by default (without configuration) both formats and avoid to explicitly configure localConfig and/or plugin.

vlt1:
@francoxyz @offtherailz unless the background is a WMS layer, it won't be present in an exported WMC context, as this format supports only WMS layers. Zoom is also not preserved, instead when importing WMC context, the map zooms to max extent.

tdipisa:
@offtherailz and @vlt1 I cannot longer export maps in WMC, it looks like a regression intruduced in the last couple of days

@offtherailz the following sound like something to better clarify and review
>@francoxyz @offtherailz unless the background is a WMS layer, it won't be present in an exported WMC context, as this format supports only WMS layers. Zoom is also not preserved, instead when importing WMC context, the map zooms to max extent.

@offtherailz the WMC export option must be available by default **everywhere** for the export tool
[...]

PR #5104, opened 2020-03-31
Add catalog services to WMC import/export

vlt1:
## Description
Adds catalog services to WMC import/export, solves problems related to catalog services, as discussed in #4912.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#4912 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/test-resources/wmc/context.wmc
+++ b/web/client/test-resources/wmc/context.wmc
@@ -23,8 +23,23 @@
         <ms:Group id="groupid3" title="Group3" expanded="true"/>
       </ms:GroupList>
       <ms:CatalogServices xmlns:ms="http://geo-solutions.it/mapstore/context" selectedService="testservice">
-        <ms:Service url="https://testserver/csw" type="csw" title="Demo CSW Service" autoload="1" oldService="old_service" serviceName="Demo CSW Service"/>
-        <ms:Service url="https://testserver/testservice/csw" type="csw" title="testservice" autoload="0" showAdvancedSettings="1" showTemplate="1" hideThumbnail="0" metadataTemplate="&lt;p>${description}&lt;/p>" format="image/jpeg" serviceName="testservices"/>
+        <ms:Service serviceName="Demo CSW Service">
+          <ms:Attribute name="url" type="string">https://testserver/csw</ms:Attribute>
+          <ms:Attribute name="type" type="string">csw</ms:Attribute>
+          <ms:Attribute name="title" type="string">Demo CSW Service</ms:Attribute>
+          <ms:Attribute name="autoload" type="boolean">true</ms:Attribute>
+          <ms:Attribute name="oldService" type="string">old_service</ms:Attribute>
+        </ms:Service>
+        <ms:Service serviceName="testservices">
+          <ms:Attribute name="url" type="string">https://testserver/testservice/csw</ms:Attribute>
+          <ms:Attribute name="type" type="string">csw</ms:Attribute>
+          <ms:Attribute name="title" type="string">testservice</ms:Attribute>
+          <ms:Attribute name="autoload" type="boolean">false</ms:Attribute>
+          <ms:Attribute name="showAdvancedSettings" type="boolean">true</ms:Attribute>
+          <ms:Attribute name="showTemplate" type="boolean">true</ms:Attribute>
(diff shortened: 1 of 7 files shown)


web/client/utils/ogc/WMC/index.js
```diff
@@ -276,8 +276,44 @@ export const toMapConfig = (wmcString, generateLayersGroup = false) => {
...
+                    'metadataTemplate',
+                    'oldService',
+                    'showAdvancedSettings',
+                    'showTemplate',
+                    'hideThumbnail'
+                ))
+                .reduce((resultObj, pickedService) => ({
+                    ...resultObj,
+                    [pickedService.serviceName]: {
+                        autoload: parseBoolean(pickedService.autoload),
```
offtherailz, line 299: The attributes depend on the service. More services will be added with different attributes in the future. To avoid to need to maintain this conversion for every attribute, I suggest to find a way to guess and save attributes types. Sometimes also JSON objects.
I suggest to insert all the additional attributes other than standard ones inside the tag, using tags with xsd convension. E.g.
```
 <attribute name="myNumber" type="xsd:number">2</attibute>
<attribute name="options" type="object"><![CDATA[{...}]]>
<attribute name="metadataTemplate" type="string"><![CDATA[<p>My Template</p>]]>
```
This should allow to automate conversion

web/client/utils/ogc/WMC/index.js
```diff
@@ -276,8 +276,44 @@ export const toMapConfig = (wmcString, generateLayersGroup = false) => {
...
+                        title: pickedService.title,
+                        type: pickedService.type,
+                        url: pickedService.url,
+                        format: pickedService.format,
+                        metadataTemplate: pickedService.metadataTemplate,
+                        oldService: pickedService.oldService,
```
offtherailz, line 305: oldService is going to be removed, it is unuseful and so we should avoid to save it.

web/client/test-resources/wmc/context.wmc
```diff
@@ -23,8 +23,23 @@
...
+          <ms:Attribute name="oldService" type="string">old_service</ms:Attribute>
+        </ms:Service>
+        <ms:Service serviceName="testservices">
+          <ms:Attribute name="url" type="string">https://testserver/testservice/csw</ms:Attribute>
+          <ms:Attribute name="type" type="string">csw</ms:Attribute>
+          <ms:Attribute name="title" type="string">testservice</ms:Attribute>
+          <ms:Attribute name="autoload" type="boolean">false</ms:Attribute>
+          <ms:Attribute name="showAdvancedSettings" type="boolean">true</ms:Attribute>
+          <ms:Attribute name="showTemplate" type="boolean">true</ms:Attribute>
+          <ms:Attribute name="metadataTemplate" type="string">&lt;p>${description}&lt;/p></ms:Attribute>
```
offtherailz, line 40: you should use `<![CDATA[<p>${description}</p>]]>` for strings, so you can avoid to take care about xml encoding, ant it's useful also for JSON.
Please add also a test for JSON object (options attribute).

web/client/utils/ogc/__tests__/WMC-test.js
```diff
@@ -17,7 +17,27 @@ describe('WMC tests', () => {
...
+                },
+                "testservices": {
+                    autoload: false,
+                    type: "csw",
+                    url: "https://testserver/testservice/csw",
+                    title: "testservice",
+                    showAdvancedSettings: true,
+                    showTemplate: true,
+                    metadataTemplate: "<p>${description}</p>",
+                    format: "image/jpeg"
```
offtherailz, line 38: add a test for JSON object

web/client/test-resources/wmc/context.wmc
vlt1: >  I seed you needed to add it, but it can become problematic quickly

What do you mean? All needed characters are escaped by default.

offtherailz: Yes but are you sure you're covering all the use cases?
Expecially for unicode chars and so on. Please try to export and re-import also a metadata template (you can use the UI in new catalog ) with special chars like àèìòù. CDATA is more secure for unicode.

vlt1: Only control chars need to be escaped, these are well defined.

Issue #5159, opened 2020-04-20
Update of current styles - ENV parameter


gitcontributor:
## Description
https://github.com/geosolutions-it/MapStore2-C028/issues/91

## How to reproduce
https://github.com/geosolutions-it/MapStore2-C028/issues/91

PR #5165, opened 2020-04-21
5159 getMap request env param added

gitcontributor:
## Description
Update of current styles - ENV parameter

 - [x] Feature

## Issue

#5159 

**What is the new behavior?**
getMap request env param added

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/map/openlayers/Layer.jsx
+++ b/web/client/components/map/openlayers/Layer.jsx
@@ -32,7 +32,9 @@ export default class OpenlayersLayer extends React.Component {
         onLayerLoad: PropTypes.func,
         position: PropTypes.number,
         observables: PropTypes.array,
-        securityToken: PropTypes.string
+        securityToken: PropTypes.string,
+        localizedLayerStyles: PropTypes.string,
+        currentLocale: PropTypes.string
     };
 
     static defaultProps = {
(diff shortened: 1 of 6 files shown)


web/client/components/map/openlayers/Layer.jsx
```diff
@@ -49,7 +51,14 @@ export default class OpenlayersLayer extends React.Component {
         this.valid = true;
         this.tilestoload = 0;
         this.imagestoload = 0;
-        this.createLayer(this.props.type, this.props.options, this.props.position, this.props.securityToken);
+        this.createLayer(
```
mbarto, line 54: If possible I would have added a generic ENV parameter here, and handled the env params for localization at an higher level. What happens if we want to use ENV for other purposes?

gitcontributor: @mbarto so if I understood you correctly you want createLayer functions tolook like this
```
this.createLayer(
            this.props.type,
            this.props.options,
            this.props.position,
            this.props.securityToken,
            this.props.env
        );
```
so the last to params will be combined and formed on the upper level. Am I right?

mbarto: Yes, exactly, this would leave the Layer open to accept env for other goals

PR #5167, opened 2020-04-21
Identify tool - Edit mode

vlt1:
## Description
Adds edit buttons to Identify tool that open Attribute Table

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#5185
[mapstore2-georchestra/#45](https://github.com/georchestra/mapstore2-georchestra/issues/45)

## Breaking change
 - [x] No


tdipisa:
@offtherailz I think the edit button should stay in the top toolbar

offtherailz:
@tdipisa yes, this is the sample of "edit single feature" (that have to be removed) . I provided this example in animation because it was the only one that was filtering by point, in the current PR. 
In global version (present for text/html only format, but it will be only one in the final version) the button is in the right place.

web/client/actions/wfsquery.js
```diff
@@ -139,12 +141,13 @@ function createQuery(searchUrl, filterObj) {
     };
 }
 
-function query(searchUrl, filterObj, queryOptions) {
+function query(searchUrl, filterObj, queryOptions, reason) {
```
offtherailz, line 144: I understand why you need this additional field. Action sequence is not enough to understand if activate or not autoselect? like geometryFilterActive + updateFilter? It should work when you select from the feature grid and when you start from the Feature Info
If it is not, please document well this method and the point you use to intercept this.

web/client/actions/featuregrid.js
```diff
@@ -204,11 +204,11 @@ function setLayer(id) {
...
     };
 }
-function updateFilter(update) {
+function updateFilter(update, dontUpdateQuery) {
     return {
         type: UPDATE_FILTER,
-        update
-
+        update,
+        dontUpdateQuery
```
offtherailz, line 211: Please let me know if we can move this logic  inside the epic and avoid this flag to be passed through components

web/client/epics/featuregrid.js
```diff
@@ -242,29 +258,23 @@ module.exports = {
                         .take(1)
                     )
             ),
+    featureGridUpdateGeometryFilter: (action$, store) =>
```
offtherailz, line 261: I think you should put this logic in the `featureGridUpdateFilter` and/or document it. This way the update logic is all together and documented. Splitting it into different epics decrese readability, if you don't notice that the two filters are mutually exclusive.

Issue #5186, opened 2020-04-29
Enable/Disable localized layer styles through Catalog tool and Layer Settings


gitcontributor:
## Description
https://github.com/geosolutions-it/MapStore2-C028/issues/103

## How to reproduce
https://github.com/geosolutions-it/MapStore2-C028/issues/103

PR #5196, opened 2020-05-01
Enable/Disable localized layer styles through Catalog tool and Layer Settings

gitcontributor:
## Description

We need to add possibility to setup localized layer styles param for each WMS layer separately. To achieve that relative settings need to be added to Catalog tool and TOC Item Settings.

 - [x] Bugfix
 - [x] Feature

## Issue

#5186 

**What is the new behavior?**

If localized layer styles param was enabled through localConfig in Catalog tool user will be able to enable/disable ENV param by default when new layers are added. Even after that user is able to customize ENV param usage through display settings of certain WMS layer.
_Also ENV param was removed from background layers as bugfix._

## Breaking change
 - [x] No


gitcontributor:
@tdipisa
regarding previous discussion.
Finally I decided to add Enable localized layer styles option to display tab as I believe that option is more related to cache and display settings. But that could be aligned if needed and moved to general tab as discussed previously.

image

web/client/utils/LayersUtils.js
```diff
@@ -497,7 +497,8 @@ const LayersUtils = {
             legendOptions: layer.legendOptions
         },
         layer.params ? { params: layer.params } : {},
-        layer.credits ? { credits: layer.credits } : {});
+        layer.credits ? { credits: layer.credits } : {},
+        layer.localizedLayerStyles ? { localizedLayerStyles: layer.localizedLayerStyles } : {});
```
gitcontributor, line 501: If `true` will be added and saved.

web/client/utils/CatalogUtils.js
```diff
@@ -524,7 +524,8 @@ const CatalogUtils = {
             allowedSRS: allowedSRS,
             catalogURL,
             ...baseConfig,
-            ...record.layerOptions
+            ...record.layerOptions,
+            localizedLayerStyles: !isNil(localizedLayerStyles) ? localizedLayerStyles : undefined
```
gitcontributor, line 528: Will be defined is service has that option defined too.

gitcontributor:
@offtherailz just to clarify
```
when I click on checkbox in display panel, the map updates but the legend preview doesn't update. If I switch language instead it works.
```
currently localized local styles checkbox disables/enables localization only for GetMap requests.
Are you talking about GetLegendGraphic request  LANGUAGE param?
(Initially when I was talking with @tdipisa  the idea of that checkbox was only to control GetMap requests. GetLegendGraphic LANGUAGE param is controlled bu localConfig property.)

gitcontributor:
@offtherailz regarding 
```
when I click on checkbox in display panel, the map updates but the legend preview doesn't update. If I switch language instead it works.
```
Please checkout last commit. If I understood you correctly this is the fix.
Now when you toggle 'Enable localized styles' LANGUAGE param is added dynamically and legend in preview changes.

PR #5249, opened 2020-05-08
 #5236 Edit button in feature info doesn't work at first time

### LINKED ISSUE #5236: Edit button in feature info doesn't work at first time
## Description
If you save a map and reload, then you try to use the edit button, it doesn't work. I guess because describeFeatureType has not been retrieved yet, so the geometry field name is not present. Probably we should retrieve that information on map load or on demand.

## How to reproduce

-  Open an existing map with vector WMS layer as admin
- Click on a feature and click on edit button

*Expected Result*
The feature grid is filtered on the clicked point

*Current Result*
The feature grid is not filtered. A second try is successful (probably because describe feature type has been recovered).

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information
**I suggest,** when the user clicks on edit button, to check if the describeFeatureType is present, if not, trigger it's download before to open feature grid, instead of using spatialFieldSelector that gets the current (last opened) spatial field name from the feature grid)

Other then this, you should calculate the effective point, the radius is not correct

vlt1:
## Description
Fixes problems with geometry filter in Attribute Table and Identify edit. Radius of the filter in Identify edit is calculated dynamically.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#5236


web/client/utils/CoordinatesUtils.js
```diff
@@ -844,6 +844,21 @@ const CoordinatesUtils = {
             crs: 'EPSG:4326'
         };
     },
+    calclateCircleRadiusFromPixel: (coordinatesFromPixelConverter, projection, pixel, latlng, pixelRadius, defaultRadius = 0.01) => {
```
offtherailz, line 847: calclateCircleRadiusFromPixel <-- typo

web/client/actions/mapInfo.js
```diff
@@ -259,6 +260,11 @@ const setEditFeatureQuery = (query) => ({
     query
 });
 
+const setCurrentEditFeatureQuery = (query) => ({
```
offtherailz, line 263:
```suggestion
/**
 * Set a geometry filter of the current query that can be used to emulate feature info filter
 * in the WFS context. Useful to switch from WMS GFI to feature grid (e.g. to edit feature).
 */
const setCurrentEditFeatureQuery = (query) => ({
```

web/client/utils/CoordinatesUtils.js
```diff
@@ -844,6 +844,28 @@ const CoordinatesUtils = {
...
+        if (isNumber(lat) && !isNaN(lat) &&
+            isNumber(lng) && !isNaN(lng) &&
+            isNumber(pixel.x) && !isNaN(pixel.x) &&
+            isNumber(pixel.y) && !isNaN(pixel.y)) {
+            const radiusA = [lng, lat];
+            const pixelCoords = isFunction(coordinatesFromPixelConverter) ? coordinatesFromPixelConverter([
+                pixel.x,
+                pixel.y >= pixelRadius ? pixel.y - pixelRadius : pixel.y + pixelRadius
+            ]) : null;
+            const radiusB = pixelCoords &&
```
offtherailz, line 859: You should draw the circle in the current map projection, so it remains a circle and the query looks similar (and you have also a chance that the server supports the CRS you're using). In this case the point is near to both the states 
image
But the query will be something like this
image
So this is the result, that is wrong. 
image

Convert instead the latLong in the current map projection and draw the circle in the projected space

web/client/epics/identify.js
```diff
@@ -198,10 +200,40 @@ export default {
...
+                const hook = getHook(GET_COORDINATES_FROM_PIXEL_HOOK);
+                const radius = calculateCircleRadiusFromPixel(
+                    hook,
+                    projectionSelector(store.getState()),
+                    get(point, 'pixel'),
+                    get(point, 'latlng'),
+                    10
+                );
+
+                return Rx.Observable.of(setEditFeatureQuery({
```
offtherailz, line 216: Can you add this to the point object to the feature info click, so it can be reused later without handling a separate state ?

Issue #5220, opened 2020-05-05
Add Feature Info and Leaflet support to WFS layer


offtherailz:
## Description
We need information on map click also for WFS layer, and finalize support in leaflet layers. Contextually we can complete some additional task for stability.

## Acceptance criteria

Improvements for WFS layer: 
 - [ ] Feature Info
 - [ ] Support for leaflet
 - [ ] add maxFeatures or other types of control to limit the download / rendering of too big data.
 - [ ] Improve loading and error handling

## Other useful information
This completes https://github.com/georchestra/mapstore2-georchestra/issues/30

ElenaGallo:
@offtherailz here some observation after testing this issue: 

- [ ] Changing the coordinate in the coordinate editor the layer informations does not change accordinatly 
![coordin](https://user-images.githubusercontent.com/56537133/81705130-f969df00-946e-11ea-91ff-131b931b18eb.gif)

*How to reproduce*

- Open a WFS layer in a map
- Click on an feature 
- Click on Show Coordinate editor
- Change coordinate


- [ ] By changing the format to TEXT or HTML, GetFeatureInfo remains with the PROPERTIES format. 

*How to reproduce*

- Open a WFS layer in a map
- Click on Burger Menu
- Click on Settings
- Change format to HTML
- Click on an feature 

*Current Result*

The GFI format is PROPERTIES

offtherailz:
The second point is not feasable, becuase the WFS feature info is an emulation and the real format is only in properties format from vector data (as well as local shape files imported).

PR #5254, opened 2020-05-08
WFS GetFeatureInfo and Leaflet implementation

offtherailz:
## Description
Implementation of emulated GFI for WFS layer and leaflet implementation of WFS layer. Some minor fixes.

Need some work from #5249 to be completed (and unit tests)

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#5220 

## Breaking change
 - [x] No


web/client/components/geostory/common/enhancers/withIdentifyPopup.jsx
```diff
@@ -59,7 +59,7 @@ export const withIdentifyRequest  = mapPropsStream(props$ => {
                     const appParams = filterRequestParams(layer, includeOptions, excludeParams);
                     const param = { ...appParams, ...requestParams };
                     const reqId = uuidv1();
-                    return getFeatureInfo(basePath, param, false)
+                    return getFeatureInfo(basePath, param, false, undefined, layer)
```
mbarto, line 62: Maybe you can reverse the last two parameters in getFeatureInfo so you can avoid remove here.

web/client/components/map/leaflet/plugins/WFSLayer.jsx
```diff
@@ -6,37 +6,36 @@
...
     const onError = () => {
+        layer.fireEvent('loadError');
         // // TODO: notify error
     };
     return getFeature(options.url, options.name, {
         // bbox: extent.join(',') + ',' + proj,
         outputFormat: "application/json",
-        srsname: CoordinatesUtils.normalizeSRS(options.srs || 'EPSG:3857', options.allowedSRS),
+        maxFeatures: 1000,
+        srsname: CoordinatesUtils.normalizeSRS( 'EPSG:4326'),
```
mbarto, line 32: What happens if EPSG:4326 is not supported on the server?

web/client/utils/IdentifyUtils.js
```diff
@@ -13,9 +13,14 @@ import FeatureInfoUtils from './FeatureInfoUtils';
 export const getFormatForResponse = (res, props) => {
     const {format, queryParams} = res;
     let infoFormat;
+    // handle WMS/WMTS..,
```
mbarto, line 16: Possible simplification:

const infoFormat = queryParams?.info_format || queryParams?.outputFormat;

web/client/utils/mapinfo/wfs.js
```diff
@@ -0,0 +1,89 @@
...
+            service: 'WFS',
+            version: '1.1.1',
+            request: 'GetFeature',
+            outputFormat: 'application/json',
+            exceptions: 'application/json',
+            id: layer.id,
+            typeName: layer.name,
+            srs: CoordinatesUtils.normalizeSRS(map.projection) || 'EPSG:4326',
+            feature_count: maxItems,
+            info_format: infoFormat,
```
mbarto, line 46: outputFormat ?

web/client/components/map/leaflet/plugins/WFSLayer.jsx
offtherailz: Probably an error in the layer (not tested). 
I though for leaflet it was a minimal requirement. Do you think it is not?

mbarto: The problem is not leaflet here, it is the server. What if I have a WFS service that only supports EPSG:3857 but not EPSG:4326?

offtherailz: I said what happens, an error notified on the layer, but I didn't get what you suggest. 
- Do you want I handle the specific error, maybe with a clear message ?
- Also the opposite is technically possible, support 4326 and not 3857. 
    - Do you think we should handle something like "allowedSRS" from getCapabilities (catalog) to the layer? in this case it should be handled also by the openlayers implementation.
    - Should we assume that 3857 is supported 
    - should we do a getCapabilities to WFS to get allowedSRS (if not present)

I think this minimal implementation for leaflet was sufficient. Can you be more explicit about what you think is wrong or can improved?

mbarto: What I would expect is to use the same srs of the current map, so that we avoid client side reprojection if not needed.

offtherailz: For this reason I asked 4326, because leaflet works in 4326 coordinates internally, so I don't have to do any reprojection.
I though asking features in 3857 (the only possible projection of the map) and convert in 4326 was a unuseful complication (if we don't handle allowedSRS).
What do you think?

mbarto: ok, that makes sense, let's add allowedSRS use for a future improvement

PR #5285, opened 2020-05-14
C125-157: Length Measurement - Enhancement

dsuren1:
## Description
Length measurement enhancement and text labels display with length and bearing info with the option to configure it via localconfig

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#5746
The label on the line segment was displayed only with length measurement info

**What is the new behavior?**
https://github.com/geosolutions-it/austrocontrol-C125/issues/157

## Breaking change
 - [x] No


dsuren1:
Hi @offtherailz 
- For now, I have fixed and updated the second and third review comments. 
- Based on your suggestion of discussion with @tdipisa , I will proceed with the label text angle correction comment.
Kindly advise if you have an alternative suggestion. Thanks!

dsuren1:
> If I can propose, the simplest solution for me is to avoid rotation at all (calculate the effective tangent to the line can become complicated) and support this feature when it is converted to annotation.

@offtherailz 
Current version of Mapstore supports text rotation via annotation. 
It is safe to say that rotation can be handled as you have suggested above. In that case, shall I go ahead and remove the rotation while drawing line then?

offtherailz:
Yes @dsuren1 we can proceed removing rotation and allowing to handle it via annotation.

offtherailz:
Now there is a conflict, can you please fix it, so I can test it? thanks

offtherailz (approved):
I only noticed than when I export measure as annotation I see the line element as point, but this is a pre-existing bug (the same happens on the current stable, with old design, si I'll open a separate issue for that.

Issue #5346, opened 2020-05-27
Include security reducer in JS API


mbarto:
## Description

This would allow integration of the API in a secured environment.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #5457, opened 2020-06-11
[2020.02.xx] #5346: include security reducer in JS API (#5456)

mbarto:
Backports the following commits to 2020.02.xx:
 - #5346: include security reducer in JS API (#5456)


MV88:
this is not needed now, it has moved to milestone 2020.02.01

PR #5565, opened 2020-06-30
#5511 - Misplaced layers in TOC while adding a Map Template

### LINKED ISSUE #5511: Misplaced layers in TOC while adding a Map Template
## Description

Sometime happens that loading a Map Template, some layers in TOC change position even though that shouldn't change.

## How to reproduce

-  Open [plui_advanced](https://dev.mapstore2.geo-solutions.it/mapstore/#/context/plui_advanced) context
-  Open the Map Template tool from the Option menu
-  Add the Cadastre Rennes on the map and move the group on top 
-  Add the Administrative Boundaries template

toc_template

*Expected Result*

The position of the **Cadastre Rennes Métropole** group should not change.

*Current Result*

The position of the **Cadastre Rennes Métropole** changes but it shouldn't.

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

jstrR:
## Description

compares for objects in created groups and mapState.groups. puts object in created groups in the appropriate place

 - [x] Bugfix

## Issue

#5511

**What is the new behavior?**

Puts nested group in the appropriate place
## Breaking change
 - [x] No

## Other useful information
I don't understand how, but initially somehow in the created groups it adds an object with an existing nested group as the last element of the array. I do not see it in Devtools or in the reduce method. If other nested groups exist and are placed between layers (not at the top) before adding a new template, in this case an object with an existing nested group as the first element of the array is added to the created groups.


web/client/utils/LayersUtils.js
```diff
@@ -125,6 +125,52 @@ const getGroupNodes = (node) => {
 };
 
 
+const sortNestedGroups = (groups, configGroups) => {
```
MV88, line 128: please add jsdoc for this function and also we need unit test for this.
i suggest to write the test using tdd
- so you simply comment out your changes
- write the test that will fail
- reapply your changes
- verify now the test is successful

MV88:
@jstrR can you please solve the conflict for the pr?

web/client/utils/MapUtils.js
```diff
@@ -380,7 +380,7 @@ function saveMapConfiguration(currentMap, currentLayers, currentGroups, currentB
     return {
         version: 2,
         // layers are defined inside the map object
-        map: assign({}, map, {layers: formattedLayers, groups, backgrounds, text_search_config: textSearchConfig, bookmark_search_config: bookmarkSearchConfig},
+        map: assign({}, map, {layers: formattedLayers, groups, backgrounds, text_search_config: textSearchConfig, bookmark_search_config: bookmarkSearchConfig, previousGroups: currentGroups},
```
MV88, line 383: Hi @jstrR why you need this previousGroups to be saved in the map configuration?

jstrR: @MV88 to get the previous position of nested groups inside heading group

jstrR:
@ MV88 I tried to reproduce your case but failed. Pls provide more information

offtherailz:
Hi @jstrR , I've seen the solution and I found out that it looks to save previousGroups in the map. Moreover it adds some code specific code that complicates the layer management system (that is already too complicated) to add a fix to a bug that looks, to me, a problem more related to the MapTemplate plugin.

Can I suggest, as alternative solution, to solve this problem in the map templates plugin? 
If I understood correctly, the issue is somehow related to the "add" functionality, that, because of layer's IDs that conflict, has problem solving the nested layers. Isn't it?

In this case we could simply solve it by assigning some random IDs in case of add. What do you think about it?

MV88:
please @jstrR backport this to 2020.02.xx

Issue #5504, opened 2020-06-18
Move the default identify settings at plugin level 


tdipisa:
## Description

At the moment you have to configure the default Identify settings in [localConfig](https://dev.mapstore.geo-solutions.it/mapstore/docs/api/plugins#plugins.Identify) (initialState.defaultState) but it should be moved at plugin level and removed as a general config option: this could help the app contexts configuration.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

Gaetanbrl:
georchestra/mapstore2-georchestra#505

PR #5578, opened 2020-07-02
#5504-add-support-default-options-identify-plugin

jstrR:
## Description

 - [x] Feature

## Issue

#5504 

**What is the new behavior?**

Default options are configured from the plugin configuration. If a plugin is presented in localConfig, it is enabled by default.
## Breaking change
 - [x] No

### DIFF
--- a/web/client/actions/__tests__/mapInfo-test.js
+++ b/web/client/actions/__tests__/mapInfo-test.js
@@ -6,8 +6,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-var expect = require('expect');
-var {
+import expect from 'expect';
+import {
     CHANGE_MAPINFO_STATE,
     NEW_MAPINFO_REQUEST,
     PURGE_MAPINFO_RESULTS,
@@ -31,8 +31,9 @@ var {
     TOGGLE_SHOW_COORD_EDITOR, toggleShowCoordinateEditor,
     CHANGE_FORMAT, changeFormat,
     CHANGE_PAGE, changePage,
-    TOGGLE_HIGHLIGHT_FEATURE, toggleHighlightFeature
-} = require('../mapInfo');
+    TOGGLE_HIGHLIGHT_FEATURE, toggleHighlightFeature,
+    SET_DEFAULT_IDENTIFY, setDefaultIdentify
+} from '../mapInfo';
 
 describe('Test correctness of the map actions', () => {
 
@@ -151,5 +152,17 @@ describe('Test correctness of the map actions', () => {
         expect(retVal.type).toBe(CHANGE_PAGE);
         expect(changePage().index).toBeFalsy();
         expect(changePage(1).index).toBe(1);
+    });
+    it('setDefaultIdentify', () => {
+        const defaultConfig = {
+            enabled: true,
+            defaultConfiguration: {
+                infoFormat: 'test'
+            }
+        };
+        const retVal = setDefaultIdentify(defaultConfig);
+        expect(retVal).toExist();
+        expect(retVal.type).toBe(SET_DEFAULT_IDENTIFY);
+        expect(retVal.cfg).toBe(defaultConfig);
     });
 });
--- a/web/client/components/data/identify/IdentifyContainer.jsx
+++ b/web/client/components/data/identify/IdentifyContainer.jsx
@@ -8,7 +8,8 @@
 
 const React = require('react');
 const {Row, Col} = require('react-bootstrap');
-const {get} = require('lodash');
+const {get, isNil} = require('lodash');
+const PropTypes = require('prop-types');
 const Toolbar = require('../../misc/toolbar/Toolbar');
 const Message = require('../../I18N/Message');
 const DockablePanel = require('../../misc/panels/DockablePanel');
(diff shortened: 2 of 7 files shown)


web/client/reducers/mapInfo.js
```diff
@@ -344,7 +345,10 @@ function mapInfo(state = initState, action) {
     case MAP_CONFIG_LOADED: {
         return {
             ...state,
-            configuration: action.config.mapInfoConfiguration || state.configuration || {}
+            configuration: {
+                ...state.configuration,
+                ...action.config.mapInfoConfiguration
+            }
```
MV88, line 351: i think here you are changing a little the behaviour.
can you explain me which use case the new code covers?

MAP_CONFIG_LOADED is used when the configuration of the map is loaded and I think it should replace the whole current state of the mapinfo.configuration @offtherailz do you confirm?

web/client/reducers/mapInfo.js
```diff
@@ -55,7 +57,6 @@ function receiveResponse(state, action, type) {
     return state;
 }
 const initState = {
-    enabled: true,
```
MV88, line 58: i think this can stay, please revert

web/client/actions/__tests__/mapInfo-test.js
```diff
@@ -6,8 +6,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-var expect = require('expect');
-var {
+import expect from 'expect';
```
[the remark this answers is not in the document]
jstrR: @MV88 this is a test file and we are not exporting anything, so I don’t understand what you mean

web/client/components/data/identify/IdentifyContainer.jsx
```diff
@@ -28,160 +29,215 @@ const {responseValidForEdit} = require('../../../utils/IdentifyUtils');
...
-        draggable,
-        setIndex,
-        warning,
-        clearWarning,
-        zIndex,
-        showEmptyMessageGFI,
-        showEdit,
-        isEditingAllowed,
-        onEdit = () => {},
+class IdentifyContainer extends React.Component {
```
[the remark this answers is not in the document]
jstrR: This component is called inside the compose function, so I cannot use hooks because they are not in the upper body.

web/client/plugins/Identify.jsx
```diff
@@ -194,12 +196,25 @@ const IdentifyPlugin = compose(
...
+            && true
+        )
+        || (
+            isNil(ownProps.enabled)
+            && (stateProps.enabled
+            && (stateProps.isCesium
+            || !ownProps.showInMapPopup)
+            && !stateProps.floatingIdentifyEnabled)
+        )
+        || ownProps.enabled
```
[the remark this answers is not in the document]
jstrR: it must get value from the plugin configuration and from the state

web/client/reducers/mapInfo.js
```diff
@@ -5,6 +5,7 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
+const {isNil} = require('lodash');
```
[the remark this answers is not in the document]
jstrR: I can't. if i convert to es6, then, for some reason, this reducer slice is not added to the store

web/client/reducers/mapInfo.js
jstrR: it should be here because default value is configured through the plugin configuration

web/client/reducers/mapInfo.js
jstrR: When we have properties in mapInfo, and then action.config.mapInfoConfguration comes in, in this case all previous properties will be deleted. So at the moment, when action.config.mapInfoConfguration comes, it merges with the previous properties

web/client/plugins/Identify.jsx
```diff
@@ -38,7 +40,7 @@ const Message = require('./locale/Message');
 require('./identify/identify.css');
 
 const selector = createStructuredSelector({
-    enabled: (state) => state.mapInfo && state.mapInfo.enabled || state.controls && state.controls.info && state.controls.info.enabled || false,
+    enabled: (state) => (state.mapInfo && state.mapInfo.enabled || state.controls && state.controls.info && state.controls.info.enabled) || state.mapInfo && state.mapInfo.enabled,
```
offtherailz, line 43: This change does not seems to make sense.
(A OR B ) OR A = A OR B = A OR B OR false 

At least becomes undefined if (state.mapInfo === undefined && state.controls === undefined)
```suggestion
    enabled: (state) => (state.mapInfo && state.mapInfo.enabled || state.controls && state.controls.info && state.controls.info.enabled) || false,
```

web/client/reducers/mapInfo.js
offtherailz: It makes sense.

web/client/plugins/Identify.jsx
```diff
@@ -164,6 +154,11 @@ const identifyDefaultProps = defaultProps({
  * @prop cfg.disableCenterToMarker {bool} disable zoom to marker action
  * @prop cfg.zIndex {number} component z index order
  * @prop cfg.showInMapPopup {boolean} if true show the identify as popup
+ * @prop cfg.enabled {boolean} // enabled by default
```
offtherailz, line 157: Correct me if I'm wrong. I think you encountered issues using this variable to set up configuration. This is the reason of a lot of configuration checks, here and there, and attempts to workaround, with this variable.

The problem is that enabled is **always** overridden by the selector in connect. Same for configuration and other properties.
So there is no way, at this stage, to get the enabled value from cfg. (you could if you get it before 'connect', but you can't). Moreover the info `enabled` state need to be always updated in the global state, to sync with the other tools. 

For this reason I suggest you to use defaultConfiguration, that is a new property, not controlled by `connect` as a containers for all the set-up you need to configure on start-up enabled, disableAlwaysOn and configuration. 

Something like
```
defaultConfiguration: {
 enabled: ...
 ....
configuration : { ... } 
}
This allows you to remove all the unnecessary changes (that are wrong or non-sense) in this PR.

File: web/client/plugins/Identify.jsx [hunk 9]
```diff
@@ -194,12 +193,22 @@ const IdentifyPlugin = compose(
...
+        enabled: (
+            isNil(ownProps.enabled)
+            && isNil((stateProps.enabled && (stateProps.isCesium || !ownProps.showInMapPopup) && !stateProps.floatingIdentifyEnabled))
+            && true
+        )
+        || (stateProps.enabled
+            && (stateProps.isCesium
+            || !ownProps.showInMapPopup)
+            && !stateProps.floatingIdentifyEnabled)
+        || isNil(stateProps.enabled) && ownProps.enabled
```
offtherailz, line 211: This is hard to undestand for me. I think it is because of the issues you encountered with enabled flag (see my other comments), so with the proposed solution, you can remove it

web/client/components/data/identify/enhancers/identify.js
```diff
@@ -70,13 +70,25 @@ const identifyLifecycle = compose(
...
                 changeMousePointer('pointer');
             }
 
+            identifyConfigureDefault({
+                enabled: isNil(enabled) && defaultConfiguration?.enabled
+                || enabled
+                || (isNil(defaultConfiguration?.enabled) && isNil(defaultConfiguration?.enabled) && true)
+                || false,
+                ...(!isNil(disabledAlwaysOn) && {disabledAlwaysOn: disabledAlwaysOn}),
+                defaultConfiguration: {...defaultConfiguration}
```
offtherailz, line 89: This part hard to understand. It is a result of a try and error? 
If so, I think you should refactor it when you understood at all the logic behind.

cfg are automatically passed to plugins as props.

These props are overridden by the `cfg` from "connect" . 
So you have to use one properly to read the configuration from `cfg` to set it the proper "identify.configuration" property, that has the real control of the tool.

To skip this overriding you can;
- use a different name(`defaultConfiguration` i suggested this because maybe it was the easier way, but maybe I confused you more with this
- do it in a wrapper before the connect, that get's the configuration, not yet overridden and trigger the action.
- use connects `mergeProps` to rename it or whatever you prefer.

The important thing is that the action should trigger the default values and get from the prop only the overrides, so the Identify tool is always in a known predictable state (without configurations from pervious mount event, and without having the necessity to specify the whole configuration every time). 

So if this part is confusing you, you can isolate the enhancer outside the connect and do it clean.

jstrR: Hi @offtherailz, i made a small changes, so now, default values only sets if the configs empty or enabled is not persist in redux store. So it didn't update store vie renders and routing (obviuosly). 

- use a different name(defaultConfiguration i suggested this because maybe it was the easier way, but maybe I confused you more with this -- it's already stored with this name in redux, but in many places the data does not go thru the selectors, making it difficult to update.
- do it in a wrapper before the connect, that get's the configuration, not yet overridden and trigger the action. -- It will have same effect and +1 more component wrapped with HOC.
- use connects mergeProps to rename it or whatever you prefer. -- wont make any impact

PR #5266, opened 2020-05-12
#3688: json-patch format overrides support for dynamically loaded con…

mbarto:
…figuration files

## Description

Introduces support for json-patch format files (http://jsonpatch.com/) to allow building configuration  files (localConfig, extensions, pluginsConfig.json etc.) from an application config (e.g. localConfig.json) and a patch file (e.g. localConfig.json.patch).

This requires:

 * datadir to be enabled and configured
 * loading configuration files through the mapstore backend config service

Have a look here for more information: https://mapstore.readthedocs.io/en/latest/developer-guide/externalized-configuration/

In addition, when a datadir is configured, the extensions upload service now uses a json-patch file to store the context creator configuration of extensions, so that this one is separated from built-in plugins configuration.

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [x] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue
#3688 

**What is the current behavior?**

#3688 

**What is the new behavior?**

I already described it above, why do you need this duplicate?

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

Issue #3688, opened 2019-04-12
Externalize overrides for localConfig.json


offtherailz:
### Description
Actually MapStore has no way to externalize configurations for localConfig.json. 
The server where MapStore is installed may need specific configurations (e.g. auth rules, catalog services, cors settings, specific URLs).
To do this setting the localConfig.json must be edited. This has a series of cons: 
 - Any update of the application will reset the changes
 - versioning of localConfig.json should be platform indipendent.

To allow externalization/customization we (me and @mbarto ) propose the following: 
 - MapStore should load 2 files instead of one: 
    - `localConfig.json`
    - `configOverrides.json`
 - `configOverrides.json` contains the changes to apply to `localConfig.json` and is added to the project, empty by default. 
 - The `configOverrides.json` file path can be intercepted by the MS back-end, if a env  `MS_CONFIG_OVERRIDES_FILE` is defined. If so, the back-end will provide the external file instead of the local one. 

#### configOverrides.json definition
This file should allow to apply simple changes to the localConfig.json.

Typical attributes to override are: 
 - root variables (printUrl, geoStoreUrl, bingApiKey, mapquestApiKey....)
 - `authenticationRules` array of rules.
 - `initialState` allow to configure the initial state of the application (e.g. `initialState.defaultState.catalog.default.Services` 
 - `plugins`: arrays of plugin configs (strings or object)
```
plugins: {
desktop: ["PluginName1", {name:"PluginName2", cfg: {someattribute: "someValue}]
}
```
`configOverrides.json` entries should allow to override all the properties. 

One of the existing languages is: 
##### JSON Patch
One way to express the changes can be [JSON Patch](http://jsonpatch.com/) [RFC6902](https://tools.ietf.org/html/rfc6902). There are many implementation in many languages (now we are interested only to JavaScript of course). 
This language allow to do insert, update, delete, replace operations on a JSON.

###### Limitations of JSON Patch
Typically the overrides could be some changes to the plugins configuration.
The current plugin configuration look like this:

This mean that we may need to find the plugin (that can be expressed  as string or object) and replace/add some configuration). The index of the element in the array may cange, and I didn't find any kind of "find and replace" functionality in jsonpatch.

So we may need to do a custom technique to replace configurations in array that supports some kind of search and replace rule. 
An alternative is to add a series of "test" rules before to override elements, to prevent issues due to array shift (insert of new plugins). If test fail, the client should send a message to notify the problem.

[...]

PR #5609, opened 2020-07-08
[2020.02.xx] #3688: json-patch format overrides support for dynamically loaded configuration files (#5266)

mbarto:
Backports the following commits to 2020.02.xx:
 - #3688: json-patch format overrides support for dynamically loaded configuration files (#5266)


mbarto:
@MV88 yes, it is, most of the patch was already backported, this is the only missing fix

Issue #5624, opened 2020-07-09
Sorting rules in visual style editor


allyoucanmap:
## Description

Add the possibility to sort the rule order of rule components.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

PR #5680, opened 2020-07-22
Fix 5624 Sorting rules in visual style editor

allyoucanmap:
## Description

This PR adds following functionalities:
- drag and drop to style editor rules
- reverse order of rules in the ui to make them consistent with the visibility order of TOC, style rule on top cover the bottom ones
- new icon for grab handler image

sort-style-rules

 - [x] Enhancement

## Issue

#5624

**What is the new behavior?**

Now it's possible to order the style rules in visual style editor with drag and drop interactions

## Breaking change
 - [x] No


giohappy:
@offtherailz 

> drag and drop is hard, expacially for big blocs. There should be a way to collapse/expand rules (to have a more general view of the rules I have and to make easier to drag and drop and browse them). Also a button collapse all/expand all maybe useful

I would open a distinct issue for this. This is an enhancmente but the sorting works as expected here.

> Because geoserver put labels on top, sorting them maybe misleading (but even when you edit the SLD you have the same effect). So we should decide if put the text rules on top, undraggable, or not

I'd rather use an info message (hovering on an info icon, or something like that) that will appear and tell the user what is the logic that Geoserver applies in case a rule with labels is placed under other geometry symbolizers.
Implementing a logic that enables / disables, and blocks rules depending if it has a label, or is under / over other rules, could be complex and not so clear for the end user.

allyoucanmap:
> I'd rather use an info message (hovering on an info icon, or something like that) that will appear and tell the user what is the logic that Geoserver applies in case a rule with labels is placed under other geometry symbolizers.
> Implementing a logic that enables / disables, and blocks rules depending if it has a label, or is under / over other rules, could be complex and not so clear for the end user.

@offtherailz I updated the PR with the info message as suggested by @giohappy 

image

offtherailz:
Great job :+1: when collapse items will be present, it will be perfect :rocket: 
Please solve conflicts so we can merge this PR

allyoucanmap:
> I'd rather use an info message (hovering on an info icon, or something like that) that will appear and tell the user what is the logic that Geoserver applies in case a rule with labels is placed under other geometry symbolizers.
> Implementing a logic that enables / disables, and blocks rules depending if it has a label, or is under / over other rules, could be complex and not so clear for the end user.

@giohappy @offtherailz 
opened a separated issue for collapse/expand improvements

see https://github.com/geosolutions-it/MapStore2/issues/5810

Issue #5569, opened 2020-07-01
Story section URL


tdipisa:
## Description

Related to issue 
https://github.com/geosolutions-it/MapStore2-C039/issues/190

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe: some updates to interlink for stories

ElenaGallo:
@jstrR here are some annotations after testing this issue:
- [ ]  It is no longer possible to share the story by copying the link from the Options button
![share ](https://user-images.githubusercontent.com/56537133/88056014-2964eb00-cb60-11ea-861f-f50c6da5c956.gif)

*How to reproduce*
- Open [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/11555/title_section_id1) story 
- Click on Options 
- Click on Share
- Copy the Story link

*Expeted Result*
The story link should be https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/shared/11555

*Current Result*
The link of the story is: https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/shared/

- [ ] After scrolling down the story, the sharing link became the URL with the section id parameter even if the scrolling position is not activated
![2share](https://user-images.githubusercontent.com/56537133/88058839-53b8a780-cb64-11ea-81e1-f18a0d25eaf1.gif)

*How to reproduce*
- Open [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/11555/title_section_id1) story
- Scroll down the story 
- Click on Options 
- Click on Share
- Copy the Story link

*Expeted Result*
The story link should be https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/shared/11555

*Current Result*
The link of the story is: https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/shared/11555/7eca906c-5dae-431c-b313-80721d4af301/7ee2c8a2-e960-45aa-b008-f75589606d6b

tdipisa:
@jstrR, in addition to the above points mentioned by @ElenaGallo, I have to add also the following one.

- [ ] the first time you open the story from home page, the URL is the following one:
https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/11555/title_section_id1

[...]

PR #5683, opened 2020-07-22
#5569. shareUrl parsing fixed

jstrR:
## Description

 - [x] Bugfix

## Issue

#5569 https://github.com/geosolutions-it/mapstore2/issues/5569#issuecomment-661852246

**What is the new behavior?**

Fixes https://github.com/geosolutions-it/mapstore2/issues/5569#issuecomment-661852246
## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/share/SharePanel.jsx
+++ b/web/client/components/share/SharePanel.jsx
@@ -26,7 +26,7 @@ import {
     Tooltip
 } from 'react-bootstrap';
 import Message from '../../components/I18N/Message';
-import { join, isNil, isEqual, inRange, replace } from 'lodash';
+import { join, isNil, isEqual, inRange } from 'lodash';
 import { removeQueryFromUrl, getSharedGeostoryUrl } from '../../utils/ShareUtils';
 import SwitchPanel from '../misc/switch/SwitchPanel';
 import Editor from '../data/identify/coordinates/Editor';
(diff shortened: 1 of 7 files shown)


web/client/components/share/SharePanel.jsx
```diff
@@ -168,9 +168,9 @@ class SharePanel extends React.Component {
             shareUrl = `${shareUrl}${settings.markerEnabled ? "?marker=" : "?center="}${this.state.coordinate}&zoom=${this.state.zoom}`;
         }
         if (!settings.showSectionId && advancedSettings && advancedSettings.sectionId) {
-            const parsedUrl = shareUrl.split('/');
-            if (parsedUrl.length === 8) shareUrl = replace(shareUrl, parsedUrl[parsedUrl.length - 1], '');
-            if (parsedUrl.length === 9) shareUrl = replace(shareUrl, `${parsedUrl[parsedUrl.length - 2]}/${parsedUrl[parsedUrl.length - 1]}`, '');
+            const parsedUrl = shareUrl.split('#')[1]?.split('/');
+            if (parsedUrl.length === 5 && parsedUrl.includes('shared')) shareUrl = replace(shareUrl, parsedUrl[parsedUrl.length - 1], '');
+            if (parsedUrl.length === 6 && parsedUrl.includes('shared')) shareUrl = replace(shareUrl, `${parsedUrl[parsedUrl.length - 2]}/${parsedUrl[parsedUrl.length - 1]}`, '');
```
allyoucanmap, line 173: The issue with the previus code was about the different path in the url between the local and the test enviroments. Now this changes ensure that the shareUrl is split on the hash before count the paresd parts of the path. This solutions seems to work but I would like to move this logic inside the [getSharedGeostoryUrl](https://github.com/geosolutions-it/MapStore2/blob/4ee11de7be140b78888fa304b6992a76e8feb5e2/web/client/utils/ShareUtils.js#L64-L68) function used some lines above https://github.com/geosolutions-it/MapStore2/blob/4ee11de7be140b78888fa304b6992a76e8feb5e2/web/client/components/share/SharePanel.jsx#L117

this is an example I did not test the code

in  ShareUtils.js
```js
getSharedGeostoryUrl: (url = '', removeSectionId) => {
    // check if the url is from geostory page
   // bacause share is used also in other page (eg: map viewer, dashboard)
    if (url.match(/\#\/(geostory)/)) {
        let geostoryUrl = url.match(/\/(geostory)\/((shared)|(newgeostory))/)
            ? url
            : url.replace('/geostory/', '/geostory/shared/');
        
        if (removeSectionId) {
            // implementation
        }
        return geostoryUrl;
    }

    return url;
}
```

in SharePanel.jsx

```js
const removeSectionId = !settings.showSectionId && advancedSettings && advancedSettings.sectionId;
let shareUrl = getSharedGeostoryUrl(removeQueryFromUrl(this.props.shareUrl), removeSectionId);

```

in this way we have all the logic for the geostory url in one function and we can add some tests.

web/client/epics/geostory.js
```diff
@@ -283,7 +284,19 @@ export const loadGeostoryEpic = (action$, {getState = () => {}}) => action$
                 return axios.get(`configs/${id}.json`)
                     // not return anything else that data in this case
                     // to match with data/resource object structure of getResource
-                    .then(({data}) => ({ data, isStatic: true, canEdit: true }));
+                    .then(({data}) => {
+                        // generates random id for section
+                        const defaultSection = data.sections[0];
```
allyoucanmap, line 289: This section id replacement is specific for [newgeostory.json](https://github.com/geosolutions-it/MapStore2/blob/4ee11de7be140b78888fa304b6992a76e8feb5e2/web/client/configs/newgeostory.json) configuration but the string `id` could be different from `newgeostory` and it contain a diffferent structure of the configuration. could we add a control to add the uuid only if the `id` is equal to `newgeostory`?
What do you think @offtherailz ?

web/client/utils/GeoStoryUtils.js
```diff
@@ -454,7 +454,7 @@ export const parseHashUrlScrollUpdate = (url, hash = '', storyId, sectionId, col
             return replace(url, `${storyIds[1]}`, `${sectionId}`);
         }
         if (hash.includes('shared')) {
-            return `${url}${sectionId}`;
+            return `${url}/${sectionId}`;
```
allyoucanmap, line 457: While testing other functionality I removed manually the section id from the url so I got the path with the trailing slash `#/geostory/:gid/` this caused a wrong parsing of the path stacking new id on the same path.

see the url in this gif
trailing-slash-1

I'm not sure the issue is related on this function anyway if the url argument is expected without trailing slash we should ensure is it removed from it in this parsing function before to add new parts to the path.

offtherailz:
please @jstrR provide a backport for 2020.02.xx

Issue #5564, opened 2020-06-30
MapTemplate plugin should be usable in localConfig


offtherailz:
## Description
Actually MapTemplates plugin can be configured by setting the templates available for the context in the UI. This templates list is saved in the root of the context object saved on geostore instead of being a part of the MapTemplate cfg.
This causes an if you try to use MapTemplates in the localConfig, because it is looking for a list of templates that is not available (we are not in a context but in the default viewer).

We should improve the MapTemplates plugin to receive templates also via configuration, and not only from the context object. 
Moreover I think that the templates should always be stored in the plugins configuration also for context, and not in the root. 
We will need anyway the retro-compatibility getting the from the context, for old context created.

## How to reproduce
-  Try to add the MapTemplates to localConfig plugins --> desktop
- Open the map viewer

*Expected Result*
The plugin is present

*Current Result*
You have an error. 

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #5689, opened 2020-07-23
MapTemplate plugin should be usable in localConfig (#5564)

AGMETEOR:
## Description
MapTemplate plugin should be usable in localConfig (#5564). This PR enables the MapTemplates plugin to load templates both from context object stored on geostore and localConfig i.e. 
```json
{
"name": "MapTemplates",
"cfg": {
"templates": [{"id": <template resource id>}]
}
}
```

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#5564 

**What is the new behavior?**
MapTemplates plugin can load templates from both context object and localConfig

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/MapTemplates.jsx
+++ b/web/client/plugins/MapTemplates.jsx
@@ -32,6 +32,7 @@ import * as epics from '../epics/maptemplates';
  * @memberof plugins
  * @class
  * @name MapTemplates
+ * * @prop {object[]} cfg.templates: A list of objects with map template ids used to load templates when not in context
  */
 const mapTemplates = ({
     active,
--- a/web/client/plugins/MapTemplates.jsx
+++ b/web/client/plugins/MapTemplates.jsx
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useEffect } from 'react';
 import { get } from 'lodash';
 import { connect } from 'react-redux';
 import { Glyphicon } from 'react-bootstrap';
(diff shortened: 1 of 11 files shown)


AGMETEOR:
> This PR fixes the bug but I think that we should do it a little better.
> 
> I think that saving the templates in the context root was an error that breaks the idependency of plugins.
> 
> 
> 
> For this reason I should keep the context templates plugin selector only for retro-compatibility with saved contexts and I propose a better solution that works for both contexts and main viewer in the same way.
> 
> 
> 
> - On plugin mount an action stores the templates list from the config into the maptemplates state (reducer). (e.g. state.maptemplates.allowedTemplates)
> 
> - The selector check that piece of state, and if missing it tests also the context state (with soft-failing, only for retro-compatiblity)
> 
> - Context should save the mapTemplates in the configuration of the single plugin, instead of the root of the context.
> 
> 
> 
> What do you think, can you try to apply this alternative solution? 
> 
> 
> 
> 

Thsnks @offtherailz , let me work on the alternative solution you have suggested.

AGMETEOR:
> * On plugin mount an action stores the templates list from the config into the maptemplates state (reducer). (e.g. state.maptemplates.allowedTemplates)


@offtherailz for this point, I have noticed that localConfig MapTemplates configuration will not be loaded on mount when the plugin is opened from context.

AGMETEOR:
> This PR fixes the bug but I think that we should do it a little better.
> I think that saving the templates in the context root was an error that breaks the idependency of plugins.
> 
> For this reason I should keep the context templates plugin selector only for retro-compatibility with saved contexts and I propose a better solution that works for both contexts and main viewer in the same way.
> 
> * On plugin mount an action stores the templates list from the config into the maptemplates state (reducer). (e.g. state.maptemplates.allowedTemplates)
> * The selector check that piece of state, and if missing it tests also the context state (with soft-failing, only for retro-compatiblity)
> * Context should save the mapTemplates in the configuration of the single plugin, instead of the root of the context.
> 
> What do you think, can you try to apply this alternative solution?

@offtherailz I changed the implementation. I also think for the last point *Context should save the mapTemplates in the configuration of the single plugin, instead of the root of the context.* , we can separate it out into a new issue like you had suggested.

web/client/plugins/MapTemplates.jsx
```diff
@@ -32,16 +32,24 @@ import * as epics from '../epics/maptemplates';
  * @memberof plugins
  * @class
  * @name MapTemplates
+ * * @prop {object[]} cfg.allowedTemplates: A list of objects with map template ids used to load templates when not in context
```
offtherailz, line 35:
```suggestion
 * @prop {object[]} cfg.allowedTemplates: A list of objects with map template ids used to load templates when not in context
```

AGMETEOR: Fixed

offtherailz:
please @AGMETEOR backport the PR to 2020.02.01

Issue #5150, opened 2020-04-17
Configuration enhancement for Application Contexts


tdipisa:
## Description
Specific configurations for Desktop, Modile and Embedded should be removed from localConfig.js: MS must be improved to be aware of the context (Desktop, Modile or Embedded) and therefore automatically enable or disable plugins (or properly set plugin properties) that are not compatible with mobile or with the embedded.

tdipisa:
From @offtherailz 

Suggested implementation. Internal support to alternative configuration in mobile/embedded mode. 

Can be overridden by configuration.
```javascript
"cfg": { // valid for all
...
},
"modes": { // other modes
 "desktop": {} // specific for desktop
  "mobile": {} // specific for mobile
  "embedded": false // <-- Do not show in embedded mode.
}
```

Support compatibility with old setup. 
We should rewrite localConfig in this way, it can be more readable for the future, and this format can be used in the future for other pages.

PR #5691, opened 2020-07-23
Plugin config enhancement

vlt1:
## Description
First attempt. The idea is basically to generate an internal config of the following form:
```javascript
{
  [pageName]: {
    desktop: [pluginConf1, pluginConf2,...],
    mobile: [pluginConf1, pluginConf2,...],
    embedded: [pluginConf1, pluginConf2,...]
  }
}
```

This is to efficiently do the rendering since we need to quickly determine what props a plugin has in current mode in current page. I've gone through the old localConfig and mirrored it in the new format. The old localConfig is in the file localConfig-backup,json. The plugin config validation is contained in the configuration object alongside reducers and epics objects in plugin definition, e.g.:

```javascript
module.exports = {
    TOCPlugin: assign(TOCPlugin, {
        Toolbar: {
            name: 'toc',
            position: 7,
            exclusive: true,
            panel: true,
            help: <Message msgId="helptexts.layerSwitcher"/>,
            tooltip: "layers",
            wrap: true,
            title: 'layers',
            icon: <Glyphicon glyph="1-layer"/>,
            priority: 1
        },
        DrawerMenu: {
            name: 'toc',
            position: 1,
            glyph: "1-layer",
            icon: <img src={layersIcon}/>,
            buttonConfig: {
                buttonClassName: "square-button no-border",
                tooltip: "toc.layers"
            },
            priority: 2
        }
    }),
    configuration: {
        disabledModes: ['mobile'],
        main: (config) => ({
          ...config,
          someProp: 5
        }),
        desktop: (config) => ...,
        mobile: (config) => ...,
        embedded: (config) => ...
    },
    reducers: {
        queryform: require('../reducers/queryform'),
        query: require('../reducers/query')
    },
    epics: require("../epics/catalog").default(API)
};
```

Disabled modes array contains modes in which the plugin must be disabled. The functions main, desktop, mobile, embedded are fed a plugin configuration processed from localConfig for that particular mode. Before desktop, mobile and embedded get the config it first goes through main function. Main function can be added for convenience to do validation that is independent from the mode. All functions and disabledModes array are optional. I currently haven't added any validation, just mirrored the existing localConfig to the new one, since I don't really know how every plugin should be validated. If that is something that should be also added to the current PR, please provide information on that(maybe something should be moved from localConfig and enforced on plugin configuration level, etc.)

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#5150 

## Breaking change
 - [x] No


mbarto:
I try to give my big picture hints for this set of improvements.

First of all, our general targets for improvements should be, in my opinion:

 * simplify and streamline the core stuff: this means that every improvement in the core should first of all simplify something  (or at least avoid introducing new complexity) and streamline the existing mechanisms to make them more clear.
 * use core flexibility and configurability to implement custom behaviour (rules and constraints) at higher levels: this allows to use the same core stuff in different scenarios and applications

That said, I think a tipical MapStore application works on two or more orthogonal axes:

 * pages (home, mapviewer, embedded, dashboard, geostory, etc.)
 * user device
 * other...

We currently mix the two in the plugins configuration, so if we want to simplify and streamline, we need to make this axes visible in configuration.

My suggestion is to have a first level with pages, (removing the common section), each one with a list of plugins for that page:

```json
{
   "mapviewer": [{"name": "plugin1"}, {"name": "plugin2"}, ...],
   "home": [{"name": "plugin1"}, {"name": "plugin2"}, ...],
}
```

Please notice that this should be, in my opinion, the new format for localConfig.json plugins section, not just an internal format (we can think on how to allow seamless migration of old configurations to this new format, eventually)

For the second axis (user device), I would improve the current configuration, based on two simple categories (desktop, mobile) to have a broader set of choices (we can support tablets, or depend on device width and height directly, for example). The purpose here is to make the pages responsive, leveraging (also) plugins configuration.

To enable that, we can use (and eventually improve) the existing dynamic configuration functionalities. Every plugin configuration property can be either a constant value or an expression. For example, to disable a plugin based on the user device you can use the **disablePluginIf** property that every plugin has (no additional disabledModes is needed):

```json
{
  "mapviewer": [{
           "name": "Identify",
           "cfg": {
                "disablePluginIf": "{mode !== 'desktop'}"
           }
    }, ...],
}
```

or better:

```json
{
  "mapviewer": [{
           "name": "Identify",
           "cfg": {
                "disablePluginIf": "{screenWitdth < '100'}"
           }
    }, ...],
}
```

Another thing I would avoid, is to add more custom configuration directly in the plugin definition. My current thinking on this is to move containers cfg and other stuff from the plugins code, to configuration (remember: flexible core, configuration in the higher levels).

I  see some confusion in the original issue about embedded: currently it is a page, not a user device (we should make the embedded viewer responsive too). If we want embedded to use the same configuration of the mapviewer, we should consider it as part of the other axis:

```json
{
  "mapviewer": [{
           "name": "Identify",
           "cfg": {
                "disablePluginIf": "{screenWitdth < '100' || embedded === true}"
           }
    }, ...],
}
```

But in my opinion, the best model for embedded is still to be a page.
```

**Comment by offtherailz** [comment]
About the model proposed by @mbarto I'm agree on almost everything: 
 - the general idea of the structure of pages
- about the "orthogonal axes"

I'm not sure about the idea of _"move containers cfg and other stuff from the plugins code, to configuration"_ I'm afraid this could make the configuration quickly become very hard to maintain and to read. In fact, this is a reorganization of localConfig.json, that risks to complicate it, instead of simplify. 
Moreover I think that using dynamic configuration to implement responsiveness may generate confusion and deny the usage of dynamic configuration for other, more delicate purposes because it's hard to extend.
For instance `disablePluginIf` can easly become a big expression with a big set of options. 
For this reason i suggested to move all the default cfg overrides, depending by the mode, in the code,and make it overridable on needing. 
I see your point of "having one single way to do dynamic configuration", but I'm afraid that the configuration can  easely become an infinite list of :
```
disablePluginIf:  "(state('pdf') !== "enabled" || mapType !== 'cesium'  || displaySize >= {{variable_for_mobile}}" ....)
```
that may  become easely unmaintainable. What do you think about this?

About embedded @mbarto, this again correct for me. For now we have a mixed usage, for embedded and for API, we need to clarify. Anyway the "page" option looks also to me the best option.

mbarto:
@offtherailz it depends on what we decide about the SDK / product split.

If plugins belong to the product, then it's correct to have a default configuration that matches the product configuration.

If plugins belong to the SDK instead, they should be mroe flexible and delegate to the product configuration specific behaviours.

tdipisa:
This is out of date, closing it.

Issue #5566, opened 2020-06-30
Identify Mode not persisted when map saved


tdipisa:
## Description

The Identify mode selected in [Map Settings](https://mapstore.readthedocs.io/en/latest/user-guide/menu-bar/#burger-menu) (Click or Hover) should be persisted as soon the map is saved.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #5701, opened 2020-07-24
Identify Mode not persisted when map saved (#5566)

AGMETEOR:
## Description
This PR ensures that when a user selects identify mode from BurgerMenu > Settings as either Click or Hover, that mode is persisted with the data resource data.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Enhancement to existing feature

# Issue

#5566 

**What is the new behavior?**
Identify tool mode is persisted. For example if a user selects Hover, it's persisted with that particular map resource

## Breaking change
 - [x] No

## Other useful information
How to test thing feature
- Import a map with layers in MapStore
- Click Burger menu > Settings
- Click on `Trigger event for the display of the information sheet` and select either Hover or Click and close that modal
- Click on BurgerMenu > Save
- If you navigate away, the hover mode (if was selected) should still be activated

### DIFF
--- a/web/client/components/misc/FeatureInfoTriggerSelector.jsx
+++ b/web/client/components/misc/FeatureInfoTriggerSelector.jsx
@@ -15,12 +15,10 @@ const {FormControl, FormGroup, ControlLabel} = require('react-bootstrap');
 class FeatureInfoTriggerSelector extends React.Component {
     static propTypes = {
         trigger: PropTypes.string,
-        onTriggerChange: PropTypes.func,
         onSetMapTrigger: PropTypes.func
     }
 
     onChange = (event) => {
-        this.props.onTriggerChange(event);
         this.props.onSetMapTrigger(event.target.value);
     }
 
--- a/web/client/components/misc/FeatureInfoTriggerSelector.jsx
+++ b/web/client/components/misc/FeatureInfoTriggerSelector.jsx
@@ -15,8 +15,15 @@ const {FormControl, FormGroup, ControlLabel} = require('react-bootstrap');
 class FeatureInfoTriggerSelector extends React.Component {
     static propTypes = {
         trigger: PropTypes.string,
-        onTriggerChange: PropTypes.func
+        onTriggerChange: PropTypes.func,
+        onSetMapTrigger: PropTypes.func
     }
+
+    onChange = (event) => {
+        this.props.onTriggerChange(event);
+        this.props.onSetMapTrigger(event.target.value);
+    }
+
     render() {
         return (
             <FormGroup bsSize="small">
@@ -24,7 +31,7 @@ class FeatureInfoTriggerSelector extends React.Component {
                 <FormControl
                     value={this.props.trigger}
                     componentClass="select"
-                    onChange={this.props.onTriggerChange}>
+                    onChange={this.onChange}>
                     <option value="click" key="click">Click</option>
                     <option value="hover" key="hover">Hover</option>
                 </FormControl>
--- a/web/client/epics/map.js
+++ b/web/client/epics/map.js
@@ -29,6 +29,7 @@ const { mapPaddingSelector } = require('../selectors/maplayout');
 
 const {setControlProperty} = require('../actions/controls');
 const {MAP_CONFIG_LOADED, MAP_CONFIG_LOAD_ERROR} = require('../actions/config');
+const { registerEventListener, unRegisterEventListener } = require('../actions/map');
 const {isSupportedLayer} = require('../utils/LayersUtils');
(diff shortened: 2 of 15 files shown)


AGMETEOR:
> Good catch about the point used.
> Anyway mouseMoveListener is used essentially for performance reasons (having continuous mouse move action when not catched was causing a lot of unnecessary load).
> Moreover there are some other tools that can register to listen mouse move event, so you may have to make you sure that `identifyFloatingTool` is registered.
> 
> Anyway I see this is the only flag used in the current state to identify when the event is registered, and this is probably a little unclear or unoconsistent.
> For this reason I'd like to ask you to modify a little your PR in order to improve and make more clear the whole system.
> 
> There is an entry in the map saved called `mapInfoConfiguration` (and in the state inside `mapInfo.configuration`). I think you should save an a new entry here to save this information, it looks the correct place to me. (e,g, `trigger: 'hover'|'click'`, if missing click is understood). Saving the event listener is a little unclear and may also be difficoult to maintain in the future.
> 
> The action that changes the configuration, may also trigger the event listener register/unregister as a side effect (now is done by `onTriggerChange`) instead of being controlled by this (unclear) variable.
> On saved map load, this configuration must be read to setup the event listener correctly as well
> 
> Please during your develoment also check if:
> 
> * the floating identify tool works also in geostore after these changes.
> * switching between maps

Hello @offtherailz , I have changed the implementation according to this review.

web/client/epics/map.js
```diff
@@ -93,6 +95,15 @@ const handleCreationLayerError = (action$, store) =>
             ]) : Rx.Observable.empty();
         });
 
+// Epic to determine which trigger to use when map loads
```
offtherailz, line 98: This epic have to be moved go in mapInfo epics. 
I suggest to intercept **also** on the `setMapTrigger` event, so we have a single entry point  to register or unregister the identify tool, and you can remove the call the registerEventListener from the component.

web/client/plugins/Identify.jsx
```diff
@@ -233,6 +233,7 @@ const FeatureInfoFormatSelector = connect((state) => ({
 const FeatureInfoTriggerSelector = connect((state) => ({
     trigger: isMouseMoveIdentifyActiveSelector(state) ? 'hover' : 'click'
 }), {
+    onSetMapTrigger: setMapTrigger,
     onTriggerChange: (event) => event.target.value === 'hover' ? registerEventListener('mousemove', 'identifyFloatingTool') : unRegisterEventListener('mousemove', 'identifyFloatingTool')
```
offtherailz, line 237: Intercepting in the epic you can remove this

web/client/components/misc/FeatureInfoTriggerSelector.jsx
```diff
@@ -15,16 +15,23 @@ const {FormControl, FormGroup, ControlLabel} = require('react-bootstrap');
 class FeatureInfoTriggerSelector extends React.Component {
     static propTypes = {
         trigger: PropTypes.string,
-        onTriggerChange: PropTypes.func
+        onTriggerChange: PropTypes.func,
+        onSetMapTrigger: PropTypes.func
     }
+
+    onChange = (event) => {
+        this.props.onTriggerChange(event);
```
offtherailz, line 23: Intercepting the event trigger you can remove also this

AGMETEOR:
> Testing to save an existing map with hover set up, I don't see the data persisted in the JSON and in fact when I save, the map reloads and the mouse event is set to click again.
> 
> Please pay attention to the main focus of the issue to solve. You should add this to your unit tests.

I am testing the changes and it seems to work. I set up hovering and then save the map and the hovering is still working.

web/client/reducers/map.js
```diff
@@ -128,6 +130,9 @@ function mapConfig(state = {eventListeners: {}}, action) {
         }
         return data;
     }
+    case SET_MAP_TRIGGER: {
```
offtherailz, line 133: Why you added this? I see it is not read anywhere.

web/client/reducers/__tests__/map-test.js
```diff
@@ -258,4 +259,9 @@ describe('Test the map reducer', () => {
         expect(state.zoom).toEqual(zoom);
         expect(state.mapStateSource).toEqual(null);
     });
+    it('tests SET_MAP_TRIGGER to set trigger in mapInfoConfiguration', () => {
```
offtherailz, line 262: if state in map is not needed, you can also remove this test

offtherailz:
@AGMETEOR please backport this PR to 2020.02.xx, thank you :+1:

Issue #5067, opened 2020-03-25
Introduce a default color picker for all text content


allyoucanmap:
## Description

GeoStory settings provides theme customization only for the general background and text.
We should introduce also a global style that represent the default for all the text content overlay (Title and Immersive).
New theme picker should provide:
- background color
- text color
- shadow

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

## Other useful information
https://github.com/geosolutions-it/MapStore2/pull/5044#issuecomment-603200554

> General settings allows only to customize general background and text in paragraph. Title and immersive sections have to be customized one by one. I see when you select "Customize" in one of these sections it inherits the default background and text color", that is a little confusing.
> I think is more natural allow to set also the colors of text frames (title and immersive content) in general settings too (background, text color and shadow).

PR #5707, opened 2020-07-28
#5067. add color default color picker for geostory overlays

jstrR:
## Description

 - [x] Feature

## Issue

#5067 

**What is the new behavior?**

Text contents from Title and Immersive now has their own default color pickers
## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/geostory/builder/Settings.jsx
+++ b/web/client/components/geostory/builder/Settings.jsx
@@ -70,7 +70,6 @@ export default ({
             <CustomThemePicker
                 themeStyle={settings?.theme?.overlay}
                 placement="right"
-                disableBackgroundAlpha
                 onChange={(overlay) => onUpdateSettings("theme", { ...settings?.theme, overlay })}/>
         </FormGroup>
         <FormGroup>
--- a/web/client/components/geostory/builder/Settings.jsx
+++ b/web/client/components/geostory/builder/Settings.jsx
@@ -65,6 +65,16 @@ export default ({
         </FormGroup>
         <FormGroup>
             <div style={{ marginBottom: "10px"}}>
+                <ControlLabel><Message msgId="geostory.builder.settings.overlay"/></ControlLabel>
+            </div>
+            <CustomThemePicker
+                themeStyle={settings?.theme?.overlay}
+                placement="right"
+                disableBackgroundAlpha
+                onChange={(overlay) => onUpdateSettings("theme", { ...settings?.theme, overlay })}/>
+        </FormGroup>
+        <FormGroup>
+            <div style={{ marginBottom: "10px"}}>
                 <ControlLabel><Message msgId="geostory.builder.settings.logo"/></ControlLabel>
                 <SwitchButton
                     onChange={() => onToggleSettings("isLogoEnabled")}
--- a/web/client/components/geostory/contents/Contents.jsx
+++ b/web/client/components/geostory/contents/Contents.jsx
@@ -51,8 +51,7 @@ export default ({
     update = () => {},
     remove = () => {},
     bubblingTextEditing = () => {},
-    storyTheme,
-    overrideThemeTypeWithDefault
+    storyTheme
 }) =>
     (<div className={className}>
         {contents.reduce(( rendered = [], { id, ...props }) => {
@@ -73,7 +72,6 @@ export default ({
                     {...contentProps}
                     {...props}
                     storyTheme={storyTheme}
-                    defaultThemeForType={overrideThemeTypeWithDefault}
                     bubblingTextEditing={bubblingTextEditing}
                     tools={tools && tools[props.type]} />)];
             if (mode === Modes.EDIT && addButtons.length > 0) {
--- a/web/client/components/geostory/contents/Contents.jsx
+++ b/web/client/components/geostory/contents/Contents.jsx
@@ -51,7 +51,8 @@ export default ({
     update = () => {},
     remove = () => {},
     bubblingTextEditing = () => {},
-    storyTheme
+    storyTheme,
+    overrideThemeTypeWithDefault
 }) =>
     (<div className={className}>
         {contents.reduce(( rendered = [], { id, ...props }) => {
@@ -72,6 +73,7 @@ export default ({
                     {...contentProps}
                     {...props}
                     storyTheme={storyTheme}
+                    defaultThemeForType={overrideThemeTypeWithDefault}
                     bubblingTextEditing={bubblingTextEditing}
                     tools={tools && tools[props.type]} />)];
             if (mode === Modes.EDIT && addButtons.length > 0) {
--- a/web/client/components/geostory/layouts/sections/Immersive.jsx
+++ b/web/client/components/geostory/layouts/sections/Immersive.jsx
@@ -14,7 +14,7 @@ import Background from './Background';
 import AddBar from '../../common/AddBar';
 import { SectionTypes, ContentTypes, MediaTypes, Modes, SectionTemplates } from '../../../../utils/GeoStoryUtils';
 import pattern from './patterns/world.svg';
-import {get, isObject} from 'lodash';
+import {get} from 'lodash';
 
 /**
  * Immersive Section Type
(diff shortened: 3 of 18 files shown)


web/client/configs/newgeostory.json
```diff
@@ -7,6 +7,11 @@
         "color": "#333333",
         "backgroundColor": "#ffffff",
         "borderColor": "#e6e6e6"
+      },
+      "overlay": {
```
MV88, line 11: what happens if these values are not present? are these default present somewhere else?

opening an existing story has this behaviour
image

imo these defaults should not be set (only) here, @allyoucanmap do you agree?

web/client/components/geostory/layouts/sections/Title.jsx
```diff
@@ -56,6 +56,9 @@ export default compose(
     const hideContent = get(focusedContent, "target.id") === contentId;
     const visibility = hideContent ?  'hidden' : 'visible';
     const expandableBackgroundClassName = expandableMedia && background && background.type === 'map' ? ' ms-expandable-background' : '';
+    const overlayStoryTheme = isObject(storyTheme) && isObject(storyTheme.overlay) && storyTheme.overlay || {};
```
MV88, line 59: i think this is sligthly better :)
```suggestion
    const overlayStoryTheme = storyTheme?.overlay || {};
```

web/client/components/geostory/layouts/sections/Title.jsx
```diff
@@ -15,7 +15,7 @@ import AddBar from '../../common/AddBar';
 import { SectionTypes, ContentTypes, Modes, MediaTypes, SectionTemplates } from '../../../../utils/GeoStoryUtils';
 import titlePattern from './patterns/dots.png';
 import coverPattern from './patterns/grid.svg';
-import {get} from 'lodash';
+import {get, isObject} from 'lodash';
```
MV88, line 18:
```suggestion
import {get} from 'lodash';
```

web/client/components/geostory/layouts/sections/Immersive.jsx
```diff
@@ -50,6 +50,9 @@ const Immersive = ({
     const hideContent = focusedContent && focusedContent.hideContent && (get(focusedContent, "target.id") === contentId);
     const visibility = hideContent ? 'hidden' : 'visible';
     const expandableBackgroundClassName = expandableMedia && background && background.type === 'map' ? ' ms-expandable-background' : '';
+    const overlayStoryTheme = isObject(storyTheme) && isObject(storyTheme.overlay) && storyTheme.overlay || {};
```
MV88, line 53:
```suggestion
    const overlayStoryTheme = storyTheme?.overlay || {};
```

web/client/components/geostory/layouts/sections/Immersive.jsx
```diff
@@ -14,7 +14,7 @@ import Background from './Background';
 import AddBar from '../../common/AddBar';
 import { SectionTypes, ContentTypes, MediaTypes, Modes, SectionTemplates } from '../../../../utils/GeoStoryUtils';
 import pattern from './patterns/world.svg';
-import {get} from 'lodash';
+import {get, isObject} from 'lodash';
```
MV88, line 17:
```suggestion
import {get} from 'lodash';
```

jstrR:
@MV88 @tdipisa bug is fixed

web/client/plugins/GeoStory.jsx
```diff
@@ -43,7 +43,7 @@ const GeoStory = ({
 );
 
 const storyThemeSelector = (state) => {
-    return settingsSelector(state)?.theme?.general || {};
+    return settingsSelector(state)?.theme || {};
```
allyoucanmap, line 46: This change is fine but it has effect on the general theme picker background and text color.
eg; cascade component now gets a wrong style object
https://github.com/geosolutions-it/MapStore2/blob/c81244574a60bc5fcac312b95e75304847b6c9b3/web/client/components/geostory/layouts/Cascade.jsx#L120-L125

so in this case it needs to be changed in `storyTheme?.general` and the same should be done for all the storyTheme props passed to the nested component of the cascade component

web/client/components/geostory/layouts/sections/Immersive.jsx
```diff
@@ -50,6 +50,9 @@ const Immersive = ({
     const hideContent = focusedContent && focusedContent.hideContent && (get(focusedContent, "target.id") === contentId);
     const visibility = hideContent ? 'hidden' : 'visible';
     const expandableBackgroundClassName = expandableMedia && background && background.type === 'map' ? ' ms-expandable-background' : '';
+    const overlayStoryTheme = storyTheme?.overlay || {};
+    const DEFAULT = 'default';
```
allyoucanmap, line 54: With the new change on the theme selector now storyTheme contains `general` and `overlay` objects. The `general` is still needed by the background component. So we should pass `storyTheme?.overlay` to the SectionContents while `storyTheme?.general` to Background component. Please verify that both picker in the settings are working for background and overlay text.

allyoucanmap:
@tdipisa @MV88 I added a [review](https://github.com/geosolutions-it/MapStore2/pull/5707#pullrequestreview-458427299) because the previous color picker for background did not work with this changes.

> [ENHANCEMENT] - probably this is a desired behaviour, but i want to point it out.
> also the style in not applied if i'm not going to click on default again. i think this should be automatic is the content is using default style. @allyoucanmap @tdipisa what do you think ?

> Good catch @MV88, I think so. @allyoucanmap which is your opinion for this?

I agree that if it use the default style it should update directly on each changes of the overlay theme settings.
I checked the code and in this implementation the overlay style of settings was saved both  in the general settings but also in the style of the specific content. I think we shouldn't save the style in the specific content but read directly the general one.

web/client/components/geostory/contents/Contents.jsx
```diff
@@ -72,6 +73,7 @@ export default ({
                     {...contentProps}
                     {...props}
                     storyTheme={storyTheme}
+                    defaultThemeForType={overrideThemeTypeWithDefault}
```
allyoucanmap, line 76: Please remove `defaultThemeForType` and `overrideThemeTypeWithDefault` from here and from other components in this PR, they are not needed anymore.

web/client/configs/newgeostory.json
allyoucanmap: > opening an existing story has this behaviour

I don't see the issue in this behaviour because that represent correctly the configuration of the overlay theme. In previous version the default style toke the colors for default style from the general theme configuration so in existing story I would expect those color picker transpare or inactive but maybe I misunderstood your comment?

allyoucanmap:
@tdipisa 
1- with this enhancement we should select and define the color for default overlay in new stories, currently with default one is:

image


we could use something similar to the bright one:
image

this is the proposed configuration
```js
"overlay": {
  "backgroundColor": "rgba(255, 255, 255, 0.75)",
  "borderColor": "#dddddd",
  "boxShadow": "0 14px 28px rgba(0,0,0,0.25), 0 10px 10px rgba(0,0,0,0.22)",
  "color": "#333333"
}
```

and then we could use the `default` theme value instead of `bright` theme value for the first section of a new geostory and for the new added sections.

web/client/components/geostory/builder/Settings.jsx
```diff
@@ -63,6 +63,16 @@ export default ({
...
                 onChange={(general) => onUpdateSettings("theme", { ...settings?.theme, general })}/>
         </FormGroup>
+        <FormGroup>
+            <div style={{ marginBottom: "10px"}}>
+                <ControlLabel><Message msgId="geostory.builder.settings.overlay"/></ControlLabel>
+            </div>
+            <CustomThemePicker
+                themeStyle={settings?.theme?.overlay}
+                placement="right"
+                disableBackgroundAlpha
```
allyoucanmap, line 73: Just noticed we should enable the background alpha value for overlay so we should remove disableBackgroundAlpha.

web/client/utils/GeoStoryUtils.js
```diff
@@ -94,7 +94,10 @@ export const getClassNameFromProps = ({ theme = {}, align = 'center', size = 'fu
  * @prop {string} theme.value style key
  * @prop {string} theme[theme.value] a style object referred to the style key
  */
-export const getThemeStyleFromProps = ({ theme = {} }) => {
+export const getThemeStyleFromProps = ({ theme = {}, storyTheme }) => {
+    if (theme?.value === '') {
```
allyoucanmap, line 98: looking at the code the theme property can be represented by a string or an object so the default one could match one of these statements:

theme === ''
theme === undefined
theme?.value === ''
theme?.value === undefined

MV88:
@jstrR please backport this pr to 2020.02.xx

Issue #5657, opened 2020-07-16
Sync to map button state not retained when toggling editing on / off inside the featuregrid


giohappy:
## Description
On mapstore demo however I change the sync map button and activate editing it is always reverted to the default value (toggled) when I deactivate editing.

## How to reproduce
-  open the feature grid for a vector layer
- untoggle sync to map
- activate editing
- deactivate editing
- the button is reverted to default

*Expected Result*
The sync to map button state should be retained (and probably should also apply to the "editing mode", but as far as I can see it doesn't) when switching editing.

*Current Result*
The state od the sync button is reverted

![featuregird_editing](https://user-images.githubusercontent.com/571129/87653636-157f4a80-c756-11ea-917d-52ed9402e605.gif)

AGMETEOR:
The issue is caused by one of the epics in `web/client/epics/featuregrid.js`. In particular [this one](https://github.com/geosolutions-it/MapStore2/blob/6b08484549a06ace962cbd628f86fa8f3455b153/web/client/epics/featuregrid.js#L329). 

This epic listens to the `TOGGLE MODE` event that's triggered when editing is activated. Two scenarios arise from this 
-  One similar to that observed on this issue where if `sync with map filter` is `off`, the it'll be eventually toggled on at the end of this epic by this line `action$.ofType(TOGGLE_MODE, CLOSE_FEATURE_GRID, LOCATION_CHANGE).take(1).flatMap(() => Rx.Observable.of(toggleSyncWms()))`
- In the case where the `sync with map filter` is instead `on` this line will toggle `Rx.Observable.of(...(isSyncWmsActive(store.getState()) ? [toggleSyncWms()] : [])),` and then it i'll eventually be toggled on again by the last line.

## Solution 1
--------
- Remove this epic
## Solution 2
--------
- Remove listening to `TOGGLE_MODE`in this line `action$.ofType(TOGGLE_MODE, CLOSE_FEATURE_GRID, LOCATION_CHANGE).take(1).flatMap(() => Rx.Observable.of(toggleSyncWms()))`

PR #5157, opened 2020-04-17
Layer Editing - Editing tool

vlt1:
## Description
Adds geometry quick filter to Attribute Table as described [here](https://github.com/georchestra/mapstore2-georchestra/issues/45#issuecomment-614584059).

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [x] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue
[mapstore2-georchestra/#48](https://github.com/georchestra/mapstore2-georchestra/issues/48)

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #5732, opened 2020-08-04
Sync to map button state not retained when toggling editing on / off inside the featuregrid (#5657)

AGMETEOR:
## Description
This PR ensures that the sync to map button state in featuregrid is retained when toggling editing on / off

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#5657 

**What is the new behavior?**
Retain sync to map button state in featuregrid after edit toggle

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/featuregrid/toolbars/Toolbar.jsx
+++ b/web/client/components/data/featuregrid/toolbars/Toolbar.jsx
@@ -172,7 +172,7 @@ module.exports = ({
             tooltipId="featuregrid.toolbar.syncOnMap"
             disabled={disableToolbar}
             active={isSyncActive}
-            visible={mode === "VIEW" && showSyncOnMapButton}
+            visible={showSyncOnMapButton}
             onClick={events.sync}
             glyph="map-filter"
             renderPopover={syncPopover.showPopoverSync}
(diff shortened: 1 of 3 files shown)


offtherailz:
@vlt1 I noticed you introduced that epic in this PR.
https://github.com/geosolutions-it/MapStore2/pull/5157/commits/727786750562d248f68d28274c4c557ba284e58e#diff-c98dc3d4706d38df3c5cfb86e1491f73R307

Do you remember why? Maybe I remember to have talked about that, but I could be wrong I don't see any comment about it.
If you remember something, can you explain the reason, so we can document a little bit more the reason of that?

vlt1:
@offtherailz I don't remember the exact reason it was needed. I think it just needs to be tested without it and see if something breaks that worked before the deletion. Then proceed according to results of the testing and do the necessary changes.

offtherailz:
Yes, what I see is that the map filter can not be disabled while we are in edit mode, as well as the filters. @AGMETEOR what if the button of syncmap is present also in edit mode? I tried (`components/data/featuregrid/toolbars/Toolbar.jsx line 175, I tried a change like this and works well. Could you please add this change and test a little bit (and adjust unit tests, if needed). Everything looks working just fine.

```diff
-            visible={mode === "VIEW" && showSyncOnMapButton}
+            visible={showSyncOnMapButton}
```

offtherailz (approved):

Issue #5757, opened 2020-08-06
Story section URL problems


ElenaGallo:
### Issue n°1
The shared URL of a story section displays the URL of the **first** story section

**How to Reproduce**
- Open this story: https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/27159/7b688800-d58b-11ea-baa0-d1aae6c8107f
- Scroll down the story to the Banner section
- Click on Share 
- Enable the scrolling position 
- Copy the story link and open it in a new browser window

**Expected Result** 
The story should scroll to the Banner section at the URL https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/shared/27159/3d6ccdf1-2a63-446b-9372-297406a36d0d 

**Current Result**
The URL reported in the Share tool is correct: https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/shared/27159/3d6ccdf1-2a63-446b-9372-297406a36d0d 

But the link of the story becomes https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/shared/27159/7b688800-d58b-11ea-baa0-d1aae6c8107f without any scrolling to the Banner section

### Issue n°2
In the shared story, the UUIDs of each section are incorrect after clicking on NavBar to view a specific section

![share section1](https://user-images.githubusercontent.com/56537133/89559525-1451b280-d816-11ea-871d-7f8a0e1795f5.gif)

**How to Reproduce**
- On the Share story 
- Click for example on **Banner Section** on NavBar (same issue of issue n°6 below, the story does't scroll)
- Scroll the story

**Expected Result**  
The story URL changes based on the section you're viewing with a correct ID

**Current Result**
The story URL is not properly generated some time and the status of the navbar is not correct (items are not properly highlighted)

![2020_08_07_12_19_16_](https://user-images.githubusercontent.com/1280027/89636144-6e02bd00-d8a8-11ea-9d30-a8e7f36751ec.png)

### Issue n°3
Sometime the item in the navbar is not properly highlighted and a wrong section ID is reported in the URL **title_section_id1**

![2020_08_07_10_31_02_](https://user-images.githubusercontent.com/1280027/89626464-c8e0e800-d899-11ea-859c-42a23e04dd11.png)

**How to reproduce**

- Login as admin
- Open this story https://dev.mapstore2.geo-solutions.it/mapstore/#/geostory/11555/7b688800-d58b-11ea-baa0-d1aae6c8107f
[...]

PR #5767, opened 2020-08-11
#5757. add new url pattern, key navigation

jstrR:
## Description

 - [x] Bugfix

## Issue

#5757 
(resolves all cases except n°4, n°5)

**What is the new behavior?**

New url pattern which resolves the router problem . navigation bar rerender on each story
## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/geostory/navigation/Navigation.jsx
+++ b/web/client/components/geostory/navigation/Navigation.jsx
@@ -34,8 +34,7 @@ export default ({
     totalItems = 1,
     currentPosition = 0,
     router,
-    buttons = [],
-    geostoryId
+    buttons = []
 }) => {
     const theme = settings?.theme?.general;
     const {
@@ -106,7 +105,6 @@ export default ({
                             <ScrollMenu
                                 items={navigableItems}
                                 currentPage={currentPage}
-                                key={geostoryId}
                                 scrollTo={scrollTo}
                                 getItemStyle={(isSelected) => !isSelected
                                     ? {
--- a/web/client/components/geostory/navigation/Navigation.jsx
+++ b/web/client/components/geostory/navigation/Navigation.jsx
@@ -34,7 +34,8 @@ export default ({
     totalItems = 1,
     currentPosition = 0,
     router,
-    buttons = []
+    buttons = [],
+    geostoryId
 }) => {
     const theme = settings?.theme?.general;
     const {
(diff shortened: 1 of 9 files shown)


jstrR:
resolves all cases except n°4, n°5

web/client/components/geostory/navigation/Navigation.jsx
```diff
@@ -105,6 +106,7 @@ export default ({
                             <ScrollMenu
                                 items={navigableItems}
                                 currentPage={currentPage}
+                                key={geostoryId}
```
allyoucanmap, line 109: @jstrR Even if I suggested you to try to use the key to force the re-render based on story id, I think it's not enough.

This solve the point number three visually on the menu items but following our previous discussion, you pointed out that there is still this [updateCurrentPage](https://github.com/geosolutions-it/MapStore2/blob/f1b1326e41bde2ff29c7383aefa4c81fc58df76f/web/client/components/geostory/layouts/sections/enhancers/currentPageSectionManager.jsx#L43) that read the wrong id from the state.

So the issue is related the fact that the currentStory state is not cleared when a different story is selected in the same session. 

We should:

- [ ] remove this `key={geostoryId}` and the connection of `geostoryId` prop in the related plugin
- [ ] clear the currentStory state in the `loadGeostoryEpic`. @offtherailz suggested to add `.startWith(setCurrentStory({}))` at this [position](https://github.com/geosolutions-it/MapStore2/blob/dddf0be5c42051271cac6d54e179eb1969fe85e0/web/client/epics/geostory.js#L360) 

In this way we ensure that the new state is cleared with the setCurrentStory equal to empty object before the new story request.

Please try the suggested changes and let me know if this solve both: the error on the navbar and the ids used in updateCurrentPage.

web/client/plugins/GeoStoryNavigation.jsx
```diff
@@ -52,6 +54,7 @@ const GeoStoryNavigation = ({
             currentPage={currentPage}
             currentPosition={currentPosition}
             totalItems={totalItems}
+            geostoryId={geostoryId}
```
allyoucanmap, line 57: This should also be removed based on this https://github.com/geosolutions-it/MapStore2/pull/5767#discussion_r470463141

allyoucanmap:
please backport this fix to 2020.02.xx

Issue #5388, opened 2020-06-01
Search by bookmark


dsuren1:
## Description
It shall be possible to switch between different views (based on bbox function - https://mapstore.readthedocs.io/en/latest/user-guide/share/) directly in the map.
For this reason there shall be a drop down list left to the search bar. Whenever a drop down element is selected the map shall be opened at the specified bbox.

## Acceptance criteria

 a) The entries for the drop down list shall be configurable by the administrator only and there shall be a configuration button similar to the one for search services, but only visible to the administrator.
b) The administrator shall be able to set a name for the drop down list entry.
c) The administrator shall have two options to set a bbox: 1) manually by entering coordinates and 2) automatically get bbox parameters from the current view.
d) The administrator shall have a choice between the following settings
1) when a list entry is chosen the respective bbox shall be loaded and the visibility of the layers shall be reloaded as defined in the map configuration file and
2) when a list entry is chosen the respective bbox shall be loaded and the visibility of the layers shall be kept as currently (at the moment of the list entry selection) set by the user for each entry.
e) all settings shall be stored in the map configuration file

## Other useful information
Related to https://github.com/geosolutions-it/austrocontrol-C125/issues/152

ElenaGallo:
@dsuren1  If you are on Search by bookmark and log out, you can still search by bookmark.
![bookmark logout](https://user-images.githubusercontent.com/56537133/85567429-21815c00-b631-11ea-8972-13bcf1713832.gif)

**How to Reproduce**
- Open this [map(https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/26581)
- Click on Search by bookmark
- Click on Log out

PR #5784, opened 2020-08-14
#5388: Fix display of bookmark after logout

dsuren1:
## Description
Resets the active search tool when user logouts with search by bookmark selected

 - [x] Bugfix

## Issue

https://github.com/geosolutions-it/mapstore2/issues/5388#issuecomment-648836542

**What is the new behavior?**
With search by bookmark selected, when user logouts, the active search tool is reset to default (address search) 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/SearchByBookmark.jsx
+++ b/web/client/plugins/SearchByBookmark.jsx
@@ -11,7 +11,7 @@ import { connect } from 'react-redux';
 
 import { createSelector } from 'reselect';
 import {Glyphicon, Button, MenuItem} from 'react-bootstrap';
-import { get } from 'lodash';
+import { get, isNil } from 'lodash';
 
 import { createPlugin } from '../utils/PluginsUtils';
 import {getExtentFromViewport} from "../utils/CoordinatesUtils";
@@ -55,7 +55,7 @@ const SearchByBookmarkPanel = (props) => {
         bookmarkSearchConfig = {},
         editIdx,
         zoomOnSelect = true,
-        bookmarkEditing = "ALL",
+        bookmarkEditing = "EDIT",
         allowUser
     } = props;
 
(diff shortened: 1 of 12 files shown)


dsuren1:
> * [ ]  Bookmark should zoom when selected, without need to click on Zoom To

I have added a provision for optinally enabling this feature via plugin Cfg (zoomOnSelect). If not having a search icon will be meaningless and this gives the admin control on the behavior.  


> * [ ]  Search by bookmark should be visible as option if at least one bookmark has been set by the administrator. ( so disablePluginIf is not enough).
> * I think you should check if the user can edit the map (in case of new map or maps with editing permission) or, optionally, allow only users of role ADMIN , and show/hide settings for Bookmark tool in case the permission is denied. If the permission should be applied only to admin or to people with correct permission, should be indicated from a configuration of the plugin (e.g. bookmarkEditing: "ADMIN"|"EDIT"|"ALL")
>   Specific selectors for this flag are present for Save tool, for instance, or for admin role, in Security selectors.

I have now made the bookmark options available only when logged in and the user role is ADMIN. Since the requirement expects this feature to be available only for ADMIN, I'm not adding checks for other users. Also search bookmark config can be saved along with new map as well as on the existing map, I'm skipping the new map checks.

Kindly let me know your thoughts

web/client/plugins/SearchByBookmark.jsx
```diff
@@ -186,8 +192,11 @@ export default createPlugin('SearchByBookmark', {
     },
     containers: {
         Search: {
-            menuItem: searchMenuItem,
-            bookmarkConfig: searchByBookmarkConfig
+            // Display bookmark setting only when logged in user is an ADMIN
+            ...(isLoggedIn && isAdminUserSelector && {
```
offtherailz, line 196: I don't think this will affect the result of the visibility of this item... isLoggedIn and isAdminUserSelector are function, here you are testing their existience

web/client/plugins/SearchByBookmark.jsx
```diff
@@ -29,17 +30,22 @@ import searchbookmarkconfig from '../reducers/searchbookmarkconfig';
...
+ * rather than clicking search icon
  */
 const SearchByBookmarkPanel = (props) => {
     const { enabled, pages, page,
         onPropertyChange,
         bookmark,
         bookmarkSearchConfig = {},
-        editIdx
+        editIdx,
+        zoomOnSelect = false
```
offtherailz, line 42: I think this should be true by default.

offtherailz (changes requested):
Hi, probably there is a misunderstanding in this. I'll try to clarify the specification.

The **edit bookmark** is reserved to ADMIN role in the specification, but **bookmark functionality** (possibility to select and so zoom to the specific area ...) should be available for all users. 
Moreover, the ADMIN role specific is a customer requirement. Maps in MapStore can be edited by any logged user.
In order to fit both the customer specific requirement and the make this functionality coherent with the normal MapStore workflow, I suggested to add `bookmarkEditing` functionality with the allowed roles.
So the customer can configure this property to hide **bookmark settings button** (the gear) for no-admin users, but by default the bookmark editing should be visible by anyone. 
Last consideration. It is strange for the normal user to see the bookmarks entry in the menu, open and see no-bookmarks and not beeing able to edit them. For this reason the entry of the bookmarks should be visible under certain conditions. 

So, summarising: 
 - [ ] Bookmark entry in burger menu of search bar should be :
    - Always  visible to the users that can see the `gear` button.
    - Visible to anyone, if at least **one** bookmark is present.
- [ ] Bookmark gear button, for adding bookmarks, should be visible only to certain users. 
- [ ] `zoomOnSelect` should be true by default. 

Please check also my comments to the code and ask if you need some help or clarification. Thank you :+1: 

**Personal note:** 
I took a look at the code of existing bookmark functionality and coordinate search are too much hard coded inside the search bar.  My idea is to exernalize them (putting every component or custom thing of each tool in the relative plugin) to allow a better maintainance and simplify the code. 
I don't know if this can be done at this stage, but it should.

dsuren1:
Thanks @offtherailz . The requirement suggested is much clearer now.

offtherailz (changes requested):
The code looks good, but I have this bug so I can not proceed testing more and doesn't make sense to review the code now. Please check that the requirements are satified before submitting review requests:
- [ ] As admin user I'm not able to configure the bookmarks in new map or existing maps

offtherailz (approved):

Issue #5773, opened 2020-08-12
Issues with GFI popup in edit mode


offtherailz:
## Description
Several issues in popup of GFI (when set to burger menu --> settings --> Feature info trigger is set to `hover`).

See this map:
https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/27684

### Overlapping entries
In same cases the results come in column, instad of being paginated.
![popup1](https://user-images.githubusercontent.com/1279510/90042371-e3b4c180-dcca-11ea-9bef-bbc1a23607ed.jpg)

### Arrows are showing also with one single result
![image](https://user-images.githubusercontent.com/1279510/90042342-dbf51d00-dcca-11ea-8b8a-1afa5dc26d0d.png)

### A glitch
Sometimes the map pans even if not necessary (I think because of a temporary stacking of items in the popup). I was not able to replicate it, @tdipisa showed me in a different map.

All these are regressions not present in the latest release. So please check what recent changes introduced it and try to fix.

tdipisa:
Thank you @offtherailz for opening this issue. The reference map I used for my test is the following one:

https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10358

Below GIF thst shows the problem. It is a regression, I don't know if it has been introduced by the lates updates provided by @dsuren1 on the Identify tool. @dsuren1 please check it, thank you.

![info_popup_issue](https://user-images.githubusercontent.com/1280027/90046723-25e10180-dcd1-11ea-9a56-b3ce8bb01676.gif)

dsuren1:
Related to: https://github.com/geosolutions-it/austrocontrol-C125/issues/155

Issue #5465, opened 2020-06-11
GFI - Improvement


dsuren1:
## Description

- Order of GFI https://github.com/geosolutions-it/austrocontrol-C125/issues/155

- Selection of mulitple GFI in window https://github.com/geosolutions-it/austrocontrol-C125/issues/165

- Reduce space used for coordinates and buttons in GFI https://github.com/geosolutions-it/austrocontrol-C125/issues/166

ElenaGallo:
@dsuren1 here some observations after testing the Identify tool:

1- Sometimes when there are more then three features, the identify tool is displayed incorrectly, as follows:
![GFI hover](https://user-images.githubusercontent.com/56537133/91545502-6a9ba800-e921-11ea-8b05-9550a9ad77c0.gif)

2- Even though there is only one feature, the Identify tool displays arrows, as follows:
![Arrows GFI](https://user-images.githubusercontent.com/56537133/91545509-6cfe0200-e921-11ea-8215-fcdfaff1ab8d.png)

*How to reproduce*

- Open [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/27883) map
- Hover over the features on map

dsuren1:
@ElenaGallo The above issues have been addressed as part this PR https://github.com/geosolutions-it/MapStore2/pull/5787. Kindly retest when the PR is merged with master

PR #5787, opened 2020-08-14
#5773: Fix issues with GFI popup in edit mode

dsuren1:
## Description
Fixes the issues with GFI popup in edit mode

 - [X] Bugfix

## Issue

#5773 

**What is the new behavior?**
- Pagination instead of stacking of the viewer
- Restrict empty responses only to GFI panel on feature click 
- Display first response with data instead of empty viewer 

## Breaking change
 - [x] No

## Other useful information
Related to #5465


offtherailz (changes requested):
Hi,
- [ ] As far as I understood you fixed the problem by removing the empty results from the list, that cause some problem in rendering. Can you please explain it better the reason of the bug and the solution ? 
- [ ] IMO `identifyFloating`  property should not be propagated (at least with this name) to the default viewer. We can think to rename with `renderEmpty` flag (with the opposite meaning), anyway I want to make sure that this flag is really needed, and if there is a way to simplify it. 
Also because checking the code I noticed also this bug. https://github.com/geosolutions-it/MapStore2/issues/5811
What do you think about it? 

- [ ] Pointing the mouse in an area with only states layer, during the loading I see first the spinner, then the spinner disappears and I see the popup has this state, until the content is loaded (or the no-data is shown)
Peek 01-09-2020 10-53
image
Is it possible to keep the spinner until the some are visible, or the blue box as in the original popup 
Peek 01-09-2020 10-46
?

offtherailz (commented):
Why do you wait until all responses are received to show the results?
Some servers may take some time to respond, so we used to to start showing the results while downloading. Tipically, from the previous requirements, the first result that is received is shown.
Anyway I've seen it and it is not that bad. If possible, I should start showing something (maybe the loading spinner shown in a corner, until is finished).

@allyoucanmap , @tdipisa what do you think?

The rest of the PR works perfectly

dsuren1:
> Why do you wait until all responses are received to show the results?
> Some servers may take some time to respond, so we used to to start showing the results while downloading. Tipically, from the previous requirements, the first result that is received is shown.
> Anyway I've seen it and it is not that bad. If possible, I should start showing something (maybe the loading spinner shown in a corner, until is finished), because it can take seconds to get a response.
> 
> @allyoucanmap , @tdipisa what do you think?
> 
> The rest of the PR works perfectly

I have fixed this.  Instead of waiting for all responses to be received, from now, will start showing the results as soon as they are received

offtherailz (changes requested):
Great job. I only found a problem with Empty pages in GeoStory identify
image

offtherailz (approved):

PR #5782, opened 2020-08-13
Review of the story settings UI(#5739)

AGMETEOR:
## Description
This PR adds some more configuration UI components to the story settings UI and allows changing of title font size and changing of story fonts using both default fonts and fonts specified from config

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: Enhancement

## Issue

**What is the current behavior?**

#5739 

**What is the new behavior?**
Screenshot 2020-08-13 at 15 54 15

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

## Other useful information
Add this to localConfig.json
```json
          {
            "name": "GeoStoryEditor",
            "cfg": {
              "disablePluginIf": "{!!(state('browser') && state('browser').mobile)}",
              "containerPosition": "columns",
              "fontFamilies": [
                {
                  "family": "Lemonada",
                  "src": "https://fonts.googleapis.com/css2?family=Lemonada:wght@300&display=swap"
                },
                {
                  "family": "Red Rose",
                  "src": "https://fonts.googleapis.com/css2?family=Lemonada:wght@300&family=Red+Rose:wght@300&display=swap"
                }
              ]
            }
          },
```
- Need advice on where in state to store the fonts loaded from localConfig. The problem is that if we save them in `geostory.currentStory.settings` and the user saves, they're persisted. In case, the user removes fonts from localConfig, we have fonts in the story settings that're essentially useless.
Maybe we should instead save the loaded fonts in a more ephemeral way, such that in case the user removes them from localConfig, we don't have these fonts staying around in the story settings. Thoughts?

Issue #5727, opened 2020-08-03
More font families for story rich text editor


tdipisa:
## Description

Check the possibility to customize the configuration of the available font families in rich text editor for stories and improve the related [plugin documentation](https://mapstore.geo-solutions.it/mapstore/docs/api/plugins) explaining how to do it.

![image](https://user-images.githubusercontent.com/1280027/89648876-330c8380-d8c0-11ea-940e-b9bfdf703163.png)

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: configuration capabilities check and related documentation

AGMETEOR:
Approach to fix
----------
In `localConfig.json` we add configuration for GeoStory like so
```json
{
"name": "GeoStory",
"cfg": {
"fontFamilies": [
{
"family": "Lemonada",
"src": "https://fonts.googleapis.com/css2?family=Lemonada:wght@300&display=swap"
}
]
}
}
```
We can then read the config at runtime from GeoStory plugin and pass down the props to the editor to merge with the default families.

We also need a third party library to load in the font files from GeoStory plugin

Issue #5739, opened 2020-08-04
Review of the story settings UI


tdipisa:
## Description

The UI of the [story settings](https://mapstore.readthedocs.io/en/latest/user-guide/story-setting/) needs to be reviewed and small additional features included to improve the customization of the story header. The required updates are reported in the mockup below.

![story_settings_2](https://user-images.githubusercontent.com/1280027/89318011-81cadb00-d67e-11ea-8bc3-4fb8e9668f10.png)

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

AGMETEOR:
Approach to fix
-------
Changes to be made in `components/geostory/builder/Settings.jsx`

tdipisa:
@AGMETEOR take into account the following consideration. In my opinion the list of available font families should be shared between different components of the stories that need them.
First of all we have to establish that list of font families: by default it can be the same list available now in rich text editors. That list of font families should be included in a configuration property of the Story Settings component (plugin configuration) to be automatically used by all Rich Text Editors in the stories. In that way we can have a more coherent tool where the list of font families is defined in one single location.

PR #5797, opened 2020-08-18
Enable more font families for story rich text editor (#5727)

AGMETEOR:
## Description
This PR builds on work done in [this PR](https://github.com/geosolutions-it/MapStore2/pull/5782) to pass down fonts loaded from localConfig.json to the Editor level

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe: Enhancement

## Issue

#5727 

**What is the new behavior?**
In case story fonts are configured for the GeoStoryEditor plugin in the localConfig.json, they're loaded and put into state and they can then be passed down to Editor through GeoStory plugin.

## Breaking change
 - [x] No

## Other useful information
PR #5782 needs to merge first before this.


web/client/selectors/geostory.js
```diff
@@ -287,3 +287,9 @@ export const hasPendingChanges = (state = {}) => state?.geostory?.pendingChanges
  * @param {object} state application state
  */
 export const updateUrlOnScrollSelector = state => get(state, 'geostory.updateUrlOnScroll', false);
+
+/**
+ * Gets the currentStoryFonts that were loaded from config by useEffect hook in the GeoStoryEditor
+ * @param {object} state application state
+ */
+export const currentStoryFonts = state => get(state, "geostory.currentStory.settings.fontFamilies", []);
```
allyoucanmap, line 295: I tried the configuration proposed here https://github.com/geosolutions-it/MapStore2/issues/5727#issuecomment-671305781 inside the "GeoStoryEditor" but the list of fonts did not showed the new font.

tested configuration:

```js
{
    "name": "GeoStoryEditor",
    "cfg": {
        "disablePluginIf": "{!!(state('browser') && state('browser').mobile)}",
        "containerPosition": "columns",
        "fontFamilies": [
            {
                "family": "Lemonada",
                "src": "https://fonts.googleapis.com/css2?family=Lemonada:wght@300&display=swap"
            }
        ]
    }
}

```

I tried also the above fontFamilies configuration in GeoStory with the same result.

Note: in the screenshot below I changed the font in the editor settings but it's not available in the inline text editor

image

After a quick investigation on latest updates (https://github.com/geosolutions-it/MapStore2/pull/5782) I noticed following issues:

- [ ] It seems the `geostory.currentStory.settings.fontFamilies` property it's never updated so this means the currentStoryFonts will return always an empty array. We should decide where save the list of fonts:
  - should we provide them within the [newgeostory.json ](https://github.com/geosolutions-it/MapStore2/blob/b6412bd6dff00e96fc6a12fb66d9929b461aa235/web/client/configs/newgeostory.json#L4-L18)so the selected font family will be related to a specific story?
  - or should we have the possibility to change the list via localConfig.json?

In my opinion this information should be stored in the **newgeostory.json** similar to theme and be persisted in the saved story. In this way we will not have conflict if we decide to change the list of fonts in newgeostory.json because only new story will be affected while the older one will use their own list of fonts. If we want to use localConfig.json we should take into account that probably we should repeat two times the list of font family one for GeoStory and another for GeoStoryEditor or find a way to store this information temporary in the state to make it accessible to both plugins (I don't like this approach)

- [ ] the font loader is located in the [`GeoStoryEditor`](https://github.com/geosolutions-it/MapStore2/blob/b6412bd6dff00e96fc6a12fb66d9929b461aa235/web/client/plugins/GeoStoryEditor.jsx#L108-L123) plugin this means that if we remove this plugin from the localConfig the `GeoStory` plugin will not access to the loaded fonts.

- [ ] Same font is requested multiple times one for each times the GeoStoryEditor is rendered and I'm guessing: could the [fontFamilies](https://github.com/geosolutions-it/MapStore2/blob/b6412bd6dff00e96fc6a12fb66d9929b461aa235/web/client/plugins/GeoStoryEditor.jsx#L123) dependency cause this issue? Anyway we should ensure to not append the same css link multiple times also when mount or unmount a component.
 
image


### Configuration proposal

Move the fontFamilies list from localConfig to [newgeostory.json](https://github.com/geosolutions-it/MapStore2/blob/b6412bd6dff00e96fc6a12fb66d9929b461aa235/web/client/configs/newgeostory.json#L4-L18)

something like 

```js
{
    "type": "cascade",
    "resources": [],
    "settings": {
        "fontFamilies": [
            {
                "family": "Lemonada",
                "src": "https://fonts.googleapis.com/css2?family=Lemonada:wght@300&display=swap"
            }
        ],
        "theme": {
            "general": {
                "color": "#333333",
                "backgroundColor": "#ffffff",
                "borderColor": "#e6e6e6"
            },
            "overlay": {
                "backgroundColor": "rgba(255, 255, 255, 0.75)",
                "borderColor": "#dddddd",
                "boxShadow": "0 14px 28px rgba(0,0,0,0.25), 0 10px 10px rgba(0,0,0,0.22)",
                "color": "#333333"
            }
        }
    },
    ...
```

Pros

- easy to access from the state
- the list of font is related to a specific story

Cons

- the list is saved always in the story configuration
- update on newgeostory.json will affect only new created stories

in the future we could think to add a UI interaction to update the list of font families, to mitigate the cons of this approach 

please @tdipisa @offtherailz could you give me a feedback on this?

allyoucanmap (changes requested):

web/client/selectors/geostory.js
AGMETEOR: @allyoucanmap thanks for the review. PR #5782 needs to merge first.

allyoucanmap: @AGMETEOR I merged the master on this branch locally to test it.
Before proceed I'll add a comment with the expected configuration that I discussed with @offtherailz and @tdipisa this morning.

allyoucanmap: This should be the expected configuration:

- list of font families should be listed in the localConfig.json inside the **GeoStory** plugin. This list should **replace** the existing list, for this reason we should take into account also the representation of safe webfont

```js
{
    "name": "GeoStory",
    "cfg": {
        "fontFamilies": [
            {
                "family": "Georgia",
                "safe": true
            },
            {
                "family": "Lemonada",
                "src": "https://fonts.googleapis.com/css2?family=Lemonada:wght@300&display=swap"
            }
        ]
    }
}
```
the list in the example above should load the font only for the one with src

- when the GeoStory plugin mounts it should perform following actions
  - on GeoStory mount: merge the fontFamilies props provided by localConfig with the fontFamilies saved in the theme settings of a story (if they exists)
  - on GeoStory mount: from the merged list of fontFamilies we should load the ones with an src (not safe). Ensure that the tags link are not loaded multiple times if they are already loaded in the header.
  - on GeoStory mount: store the merged fontFamilies in the geostory redux state to make them accessible from the GeoStoryEditor plugin

- we need to store the list of merged font families in the geostory configuration when a user save the story. We should add this information inside the `settings.theme.fontFamilies` path.

- if both the localConfig.json and the settings.theme.fontFamilies don't provide a list of fonts we should use the default ones

please take into account also the issues in the previous comment https://github.com/geosolutions-it/MapStore2/pull/5797#discussion_r481940499

MV88: We discussed and the previous comment cover the needed use cases
@AGMETEOR  any questions?

AGMETEOR: None atm

web/client/plugins/GeoStory.jsx
```diff
@@ -28,19 +35,49 @@ import MediaContentToolbar from './geostory/MediaContentToolbar';
...
-    />
-</BorderLayout>
-);
+}) => {
+    const localize = useCallback((id) => getMessageById(messages, id), [messages]);
+    const addFunc = (path, position, element) => onAdd(path, position, element, localize);
+
+    useEffect(() => {
+        onUpdateSetting("fontFamilies", fontFamilies);
+    }, [ fontFamilies ]);
```
allyoucanmap, line 51: - [ ] If fontFamilies is not configured in localConfig this go in infinite loop because ad each render it will recieve a new empty array (see fontFamilies in prop list). Probably this could be solved using GeoStory.defaultProps instead. Steps:
  - remove fontFamilies from localConfig
  - use debug=true
  - login as admin
  - go to homepage
  - select 'Demo Observatories' story
  - infinite in console `GEOSTORY:UPDATE_SETTING` in console

web/client/plugins/GeoStory.jsx
```diff
@@ -28,19 +35,49 @@ import MediaContentToolbar from './geostory/MediaContentToolbar';
 const GeoStory = ({
```
allyoucanmap, line 35: We need to add documentation for this plugin for the new property as requested in the issue description. If possible It would be nice to have also some tests on this plugin in particular for the new introduced useEffect

allyoucanmap (changes requested):

allyoucanmap (changes requested):
- [ ] While reading the description of https://github.com/geosolutions-it/MapStore2/issues/5739 I noticed it was requested to apply the default style also to the story header (navigation bar)

image

but currently it's not applied

image

web/client/selectors/geostory.js
```diff
@@ -287,3 +287,9 @@ export const hasPendingChanges = (state = {}) => state?.geostory?.pendingChanges
  * @param {object} state application state
  */
 export const updateUrlOnScrollSelector = state => get(state, 'geostory.updateUrlOnScroll', false);
+
+/**
+ * Gets the currentStoryFonts that were loaded from config
+ * @param {object} state application state
+ */
+export const currentStoryFonts = state => get(state, "geostory.currentStory.settings.fontFamilies", []);
```
allyoucanmap, line 295: - [ ] we should move `geostory.currentStory.settings.fontFamilies` in `geostory.currentStory.settings.theme.fontFamilies`

web/client/plugins/GeoStory.jsx
```diff
@@ -28,19 +35,49 @@ import MediaContentToolbar from './geostory/MediaContentToolbar';
...
-        contentToolbar={MediaContentToolbar}
-    />
-</BorderLayout>
-);
+}) => {
+    const localize = useCallback((id) => getMessageById(messages, id), [messages]);
+    const addFunc = (path, position, element) => onAdd(path, position, element, localize);
+
+    useEffect(() => {
+        onUpdateSetting("fontFamilies", fontFamilies);
```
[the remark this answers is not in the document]
allyoucanmap: - [ ] we should move `geostory.currentStory.settings.fontFamilies` in `geostory.currentStory.settings.theme.fontFamilies`

allyoucanmap (changes requested):

allyoucanmap (approved):

Issue #5766, opened 2020-08-11
Lazy loading of story contents sometime generates issues


tdipisa:
## Description

Lazy loading of story contents sometime generate issues while automatically scrolling the story by clicking on an item in Navbar.
This issue comes from [this comment](https://github.com/geosolutions-it/MapStore2/issues/5757#issuecomment-671833274) in issue #5757 (problem n° 5 reported in the issue description).

## How to reproduce

-  See **issue n°5** in the #5757 issue description

*Expected Result*

See **issue n°5** in the #5757 issue description

*Current Result*

See **issue n°5** in the #5757 issue description

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #5800, opened 2020-08-19
Fix automatic scrolling issue due to lazy loading of story media(#5766)

AGMETEOR:
## Description
This PR solves the issue where automatic scrolling by pressing a button on the story navbar doesn't reach the target media section due to lazy loading of media.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#5766 

**What is the new behavior?**
The wrapper around the loader is giving a big enough height to allow the media section to come into view even when media is still loading.

## Breaking change
 - [x] No


web/client/components/geostory/media/index.js
```diff
@@ -23,7 +23,10 @@ export const typesMap = {
 };
 
 const ErrorComponent = () => <div className="ms-media-error"><Glyphicon glyph="exclamation-sign"/></div>;
-const LoaderComponent = () => <div className="ms-media-loader"><Loader size={52}/></div>;
+
+// The div wrapping <Loader /> is passed a style that includes a height e.g. 500px big enough such that we're able to scroll
+// to a section during lazy loading of media. See https://github.com/geosolutions-it/MapStore2/issues/5766
+const LoaderComponent = ({ style = {}, size = 52 }) => <div style={style} className="ms-media-loader"><Loader size={size}/></div>;
```
allyoucanmap, line 29: The issue is still visible even if we apply a height greater than the previous one because the height of media could change depending of the size property, the width of device or their own natural height.

You can see in this .gif the video is not correctly selected even with height equal to 500px
lazy-error

This issue is related to media contents rendered inside paragraph or column of immersive section. We should try to set the right height based on the type of media to ensure a consistent transition between loading and actual component and avoid issue with the scrollTo action. We could set the height of loader using CSS similar to the actual component. If you see the DOM nodes the ms-media-loader is contained inside a specific node with type, size and alignement

eg

image

- [ ] Video: it uses a [resolutions of 16/9](https://github.com/geosolutions-it/MapStore2/blob/0f5afda5d60a9962ec88954d77a6e292a5c3437e/web/client/components/geostory/media/Video.jsx#L208) so we could use [`padding-top: 56.25%;`](https://www.w3schools.com/howto/howto_css_aspect_ratio.asp) to the `ms-media-loader` class and `position: asbsolute` for the loader div child with top and left to 50% and transition to -50%, -50%. The loader should automatically adjust to 16/9 with this kind of css.

- [ ] Map: we should apply to loader component a similar CSS as for the actual component, see [paragraph](https://github.com/geosolutions-it/MapStore2/blob/0f5afda5d60a9962ec88954d77a6e292a5c3437e/web/client/themes/default/less/geostory.less#L621-L630) and [immersive](https://github.com/geosolutions-it/MapStore2/blob/0f5afda5d60a9962ec88954d77a6e292a5c3437e/web/client/themes/default/less/geostory.less#L735-L744) css of map contents

- [ ] Web page: we should apply to loader component a similar CSS as for the actual component, see [paragraph](https://github.com/geosolutions-it/MapStore2/blob/0f5afda5d60a9962ec88954d77a6e292a5c3437e/web/client/themes/default/less/geostory.less#L639-L645) and css [immersive](https://github.com/geosolutions-it/MapStore2/blob/0f5afda5d60a9962ec88954d77a6e292a5c3437e/web/client/themes/default/less/geostory.less#L754-L760) of web page contents. (I'm not sure if this needs changes to verify)

Exception

- [ ] Image: the height of images is computed by the browser based on the naturalHeight of the original source and we have not this information before loading. @offtherailz Should we need to store the naturalHeight and naturalWidth of an image in the story resources after adding or updating the src of an image in the media editor? If so we could calculate the height of the loader based on the resolution of original image. This approach needs inline style with the padding approach used by video using the resolution of the original image.


At the end we should not notice differences on the height before and after loading, eg: video media

image

image


Important: let's start from video content and see if the issue 5 described here is solved https://github.com/geosolutions-it/MapStore2/issues/5757 before to apply changes to every media contents

allyoucanmap (changes requested):

web/client/components/geostory/media/index.js
offtherailz: As discussed we should save the size information for image, and we can post-pone saving the resolution for video.

allyoucanmap (changes requested):
@AGMETEOR Loader for video, map and webpage (*) is using the right size now and the issue n 5 seems solved but there are some issue to fix with the images media, see my comments inline.

we verify that webpage did not need a style for loader because we don't apply lazy loading to this component.

allyoucanmap (approved):

Issue #5777, opened 2020-08-13
Jenkins has some issues with eslint, not intercepted locally or on travis


offtherailz:
## Description

- [This PR ](https://github.com/geosolutions-it/MapStore2/pull/5538 )had [a problem](https://github.com/geosolutions-it/MapStore2/pull/5538/files#diff-a345d109a2049b274bd46c8496bacaf2R672) in lint but passed the [travis build](https://travis-ci.org/github/geosolutions-it/MapStore2/builds/717309523?utm_source=github_status&utm_medium=notification)

**This error is not notified by vscode, neither npm run lint, but jenkins failed.**
```
> eslint web/client --ext .jsx,.js


/opt/jenkins_home/workspace/MapStore2_Dev_Build/web/client/utils/LayersUtils.js
  672:100  error  ["default"] is better written in dot notation  dot-notation

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

- [this other PR](https://github.com/geosolutions-it/MapStore2/pull/5466) had a [console statement](https://github.com/geosolutions-it/MapStore2/pull/5466/files#diff-284d539b152aa3826dd91ea967c14632R82) but passed [the travis build](https://travis-ci.org/github/geosolutions-it/MapStore2/builds/715104188?utm_source=github_status&utm_medium=notification)

jenkins build failed with this error:
```
✖ 2 problems (1 error, 1 warning)
  1 error and 0 warnings potentially fixable with the `--fix` option.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! mapstore2@0.1.0 lint: `eslint web/client --ext .jsx,.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the mapstore2@0.1.0 lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/tomcat/.npm/_logs/2020-08-13T11_33_00_892Z-debug.log
Process leaked file descriptors. See http://wiki.jenkins-ci.org/display/JENKINS/Spawning+processes+from+build for more information
Build step 'Execute shell' marked build as failure

```

Update (2020/9/7):
https://build.geo-solutions.it/jenkins/view/MapStore2/job/MapStore2_Dev_Build/1830/console
New Jenkins   failure: 
```
TypeError: Cannot read property 'range' of null
[...]

PR #5279, opened 2020-05-13
add eslint-plugin-import to improve eslint rule

MV88:
## Description

I found it useful to add this rule that save us to add wrong named import statement

it is really helpful to have also the [statusbar error](https://marketplace.visualstudio.com/items?itemName=JoeBerria.statusbarerror) plugin enabled in VS code

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: eslint improvement

## Issue

**What is the current behavior?**

you do not notice that some named import are wrong, you can lose time in the best scenario or introduce some runtime error in the worst

**What is the new behavior?**

an error appear if you use a wrong named import statement
import { FAKE } from "../actions/something"; 

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #5842, opened 2020-09-07
Update GeoStory.jsx t ofixx issues with Jenkins

offtherailz:
## Description
Note: I edited it on github, please @mbarto test this changes locally. 
It is related to #5777 It should fix temporary the build, but it doesn't solve the problem at all. 
I added the PR  #5279 as an attempt to fix issues with ESLint, based on comment by eslint dev team (or at least update to a not deprecated version of eslint).

### DIFF
--- a/web/client/plugins/GeoStory.jsx
+++ b/web/client/plugins/GeoStory.jsx
@@ -53,7 +53,7 @@ const GeoStory = ({
 
     useEffect(() => {
         if (storyFonts.length > 0) {
-            const storyFontsSrc = storyFonts.filter(({ src }) => src && !(document && document.head).querySelector(`link[href='${src}']`));
+            const storyFontsSrc = storyFonts.filter(({ src }) => src && !document.head.querySelector(`link[href='${src}']`));
             if (storyFontsSrc.length > 0) {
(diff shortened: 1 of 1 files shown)


web/client/plugins/GeoStory.jsx
```diff
@@ -53,7 +53,7 @@ const GeoStory = ({
 
     useEffect(() => {
         if (storyFonts.length > 0) {
-            const storyFontsSrc = storyFonts.filter(({ src }) => src && !document?.head.querySelector(`link[href='${src}']`));
+            const storyFontsSrc = storyFonts.filter(({ src }) => src && !(document && document.head).querySelector(`link[href='${src}']`));
```
mbarto, line 56: What is the scenario where document is undefined or null? And if it is, null.querySelector is going to crash anyway

offtherailz: mmm, yes I was trying to replicate the exact thing, but it doesn't make to much sense.
```suggestion
            const storyFontsSrc = storyFonts.filter(({ src }) => src && !document.head.querySelector(`link[href='${src}']`));
```

Issue #5386, opened 2020-06-01
Search Service - show GFI 


MV88:
## Description

When searching for features there shall be a possibility to directly show the GFI for this feature.
Preferably there shall be a button available for each search result that zooms to the feature and directly opens the GFI.
Clicking the whole bar shall still have the same effect as it has right now.

The latest MapStore already has a similar functionality described in the [User Guide](https://mapstore.readthedocs.io/en/latest/user-guide/menu-bar/#custom-wfs-searches) (see **Launch Info Panel** configuration options). 

![image](https://user-images.githubusercontent.com/1280027/77440773-f9435080-6de8-11ea-941f-04eaf04395cc.png)

There are only some additional features to include as agreed during the meeting in November:

* When _Search layer_ is enabled in **Launch info panel**, two new checkboxes should be added:
  - **Use button to open feature info**: to add a button to each result in the list to explicitly open GFI for it  

![image](https://user-images.githubusercontent.com/1280027/77440982-3a3b6500-6de9-11ea-8107-58bea2d6262a.png)

  - **Force search layer visibility**:  the layer (if present in the map) will be made visible automatically to always perform GFI operation on it too

MV88:
refers to [austrocontrol-C125/#169](https://github.com/geosolutions-it/austrocontrol-C125/issues/169)

ElenaGallo:
@vlt1 By enabling only the "Use to open feature information", no GFI buttons appear on the result in the list.

*How to reproduce*
- Open [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/25924) map
- Search  'Wall St' on the searching bar

*Current Result*
no GFI buttons appears on the result in the list

tdipisa:
@vlt1 this issue is still not working properly.

Please look at the steps described by @ElenaGallo in the above comment. The above description is not complete anyway, see also my comment below.

The GFI button should be showed when the option "Use button to open feature info" is enabled. In the map linked above this option is selected but the button doesn't appear: it appears only if both options are enabled. 

![image](https://user-images.githubusercontent.com/1280027/90253859-2351e980-de42-11ea-9e45-03c01e7ba081.png)

At the moment there is anyway confusion in the logic related to the two new checkboxes and associated functionalities. See the video below and check/compare again the requirements in the issue description. The implemented functinalities are not in line with the requirements and we have wrong behaviors.

[...]

PR #5872, opened 2020-09-14
Search Service - show GFI

vlt1:
## Description
Addressing [this](https://github.com/geosolutions-it/MapStore2/issues/5386#issuecomment-691067643). On my machine firefox works fine, but on Chromium if button is disabled the tooltips were not shown at all. This concerns any such button(for example location button in NavBar). The issue seems to be that if the button is disabled mouseover event handlers are not triggered for some reason. To resolve this I've found a solution in writing custom OverlayTrigger that will wrap the button in span that will handle showing the tooltip.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#5386 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/mapcontrols/search/SearchResultList.jsx
+++ b/web/client/components/mapcontrols/search/SearchResultList.jsx
@@ -14,6 +14,8 @@ import assign from 'object-assign';
 import SearchResult from './SearchResult';
 import I18N from '../../I18N/I18N';
 
+import OverlayTriggerCustom from '../../misc/OverlayTriggerCustom';
+
 import { showGFIForService, layerIsVisibleForGFI } from '../../../utils/SearchUtils';
 
 export default class SearchResultList extends React.Component {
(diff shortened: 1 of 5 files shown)


MV88 (changes requested):
@vlt1 Is it possible to change the current OverlayTrigger instead of creating a new one, by doing so we could fix the tooltip problem elsewhere. or am i wrong?

should not break anything doing that, right?

vlt1:
@MV88 OverlayTrigger is part of react-bootstrap, I can't change it. The other way to go about it is to upgrade react-bootstrap to 4 from 3, but I'm not sure if that isn't going to break something. Also the new OverlayTrigger can be later expanded to have all functionalities of the react-bootstrap one and substituted for the react-bootstrap one.

vlt1:
@MV88 OvelayTrigger in misc just wraps react-bootstrap OverlayTrigger with a hoc withContainer

offtherailz:
Yes @vlt1 , I'd like to see if we can solve the problem once for all. We have a lot of these problems, so adding also this fix of our custom version can be useful. 
Can you try and test the buttons?

MV88:
> Yes @vlt1 , I'd like to see if we can solve the problem once for all. We have a lot of these problems, so adding also this fix of our custom version can be useful.
> Can you try and test the buttons?

probably we can open  new issue to do this, @offtherailz what do you think ?

offtherailz:
To fix it only there now you're doing a sort of workaround. 
If you prefer to open a separate issue to speed-up the delivery of the functionality, ok, but generally speaking I'd prefer to fix the general issue rather then fixing it once for every special case. 
So, ok for opening an issue, but open a separate issue now and work on it very soon.

MV88 (approved):
this looks good to me, but as agreed, i will open a new issue for fixing the other buttons that have the same problems

Issue #5779, opened 2020-08-13
GeoStory Builder - Content Editor - Map Editor - Interactions


tdipisa:
This issue is referring to an improvement for the existing [Map Inline Editor](https://mapstore.readthedocs.io/en/latest/user-guide/configure-map/) currently used for stories. The aim is to include an enhancement to make the story editor able to define additional actions for story sections of type [map](https://mapstore.readthedocs.io/en/latest/user-guide/media-editor-window/#maps).

### Enable Locations
Possibility to define marker locations in map and associate some custom info content to visualize in a popup when clicking on them: this can be activated as an additional setting inside the [Map Inline Editor](https://mapstore.readthedocs.io/en/latest/user-guide/configure-map/), as reported in the mockup below.

![geostory_issue_31](https://user-images.githubusercontent.com/1280027/90135279-b9b6da00-dd72-11ea-9a0e-b74222a089d0.png)

As soon as the story editor activates **Enable Locations** it is possible to click on the plus button (+) to define a new location. Doing that, a new record is added in the locations list and the user can customize its name (by default it is "New Location"). At this point, with the new location record selected, the story editor can click on map to add a marker: as soon as he clicks on map, a marker is placed in the corresponding map coordinates and a popup opens so that it is possible to configure the popup info contents and the marker style (there will be two different tabs for this as reported in the mockup above).

The story editor can edit an existing location in two ways:
1. By clicking on the loation record in the list: the map will be centered in the marker position and the popup opened so that the user can modify contents

2. By clicking on the marker itself: the popup openes so that the user can modify contents. The corresponding location records in the list is also automatically selected.

In both cases it is possible to change the marker position by dragging it in the map.

**Note:** when a location record is selected in the list the identify tool (**if activated**) is temporarily disabled. As soon as the user clicks again on the location record in the list, the record is deselected, the marker's popup is closed and the identify tool re-enabled again.

**Popup Tab**
Use [the same rich text editor](https://mapstore.readthedocs.io/en/latest/user-guide/text-editor-toolbar/) (https://github.com/geosolutions-it/react-draft-wysiwyg) already in use for GeoStory: it is currently available as a npm dependency for MapStore. The Rich Text Editor component will include also the enhancement developed with #47. 

**Style Tab**
Use the same component already implemented for the new Annotation tool to make the user able to choose the desired style for each marker.

[...]

PR #5902, opened 2020-09-18
GeoStory Builder - Content Editor - Map Editor - Interactions(#5779)

AGMETEOR:
## Description
Adds ability to add map locations to geostory map media sections

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#5779 

**What is the new behavior?**
Adds ability to add map locations to geostory map media sections

## Breaking change
 - [x] No


MV88:
please solve conflict

web/client/components/geostory/common/enhancers/withLocationLayer.js
```diff
@@ -0,0 +1,271 @@
...
+                            features: [],
+                            featuresCrs: map.projection,
+                            title: "Locations",
+                            viewer: {}
+                        },
+                        queryParams: {
+                            lat: point.latlng.lat,
+                            lng: point.latlng.lng
+                        },
+                        response: `<body><p>${intersectingFeature[0].properties.html}</p></body>`
```
allyoucanmap, line 229: If intersectingFeature[0] is undefined the app crashes

web/client/components/geostory/common/map/LocationListItemEditable.jsx
```diff
@@ -0,0 +1,42 @@
...
+                active
+                    ? onChange("currentMapLocation", "")
+                    : onChange("currentMapLocation", locationId);
+            }}
+            className={`ms-geostory-map-locations-item ${active && "ms-geostory-map-locations-item-active"}`}>
+            <TitleEditable disabled={!active} title={locationName} onUpdate={(name) => {
+                const path = `layers[{"id": "locations"}].features[{"id": "locFeatureCollection"}].features[{"id": "${locationId}"}].properties.locationName`;
+                onChangeMap(path, name);
+            }} />
+            <GeneralButton onClick={() => {
```
allyoucanmap, line 29: We don't need a new GeneralButton, we could use the react-bootstrap one

web/client/components/geostory/common/map/LocationListItemEditable.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import TitleEditable from '../../builder/TitleEditable';
+import GeneralButton from '../../../misc/GeneralButton';
+const {Glyphicon} = require('react-bootstrap');
+const enhanceTooltip = require('../../../misc/enhancers/tooltip');
+const Glyph = enhanceTooltip(Glyphicon);
+
+const LocationItemEditable = (props) => {
```
allyoucanmap, line 8: If we need a list of card we could use components/misc/cardgrids/SideCard or components/misc/cardgrids/SideGrid instead of creating a new component

web/client/components/map/openlayers/DrawSupport.jsx
```diff
@@ -124,7 +125,7 @@ export default class DrawSupport extends React.Component {
         if (!newProps.drawStatus && this.selectInteraction) {
             this.selectInteraction.getFeatures().clear();
         }
-        if ( this.props.drawStatus !== newProps.drawStatus || this.props.drawMethod !== newProps.drawMethod || this.props.features !== newProps.features) {
+        if ( this.props.drawStatus !== newProps.drawStatus || this.props.drawMethod !== newProps.drawMethod || this.props.features !== newProps.features || this.props.drawStatus === "simpleDrag" || newProps.drawStatus === "simpleDrag") {
```
allyoucanmap, line 128: With this solution the addSimpleDragOperation is activated on every update of props (eg map move, add new location, ..) and each new translate interaction is added to ol and never removed

image

Following the approach of this component you should have a start and a end status (eg. startSmpleDrag, endSimpleDrag) and the end should remove all the interaction. The component that use the DragSupport should start and end it one per session without continuosly mount and unmount the component on each prop update.

@offtherailz please could you take a look at this including the part on the enhancers where the DrawSupport is used?

web/client/components/misc/GeneralButton.jsx
```diff
@@ -0,0 +1,21 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+const {Button} = require('react-bootstrap');
+
+const GeneralButton = props => {
```
allyoucanmap, line 12: We don't need this component use directly Button of react-bootstrap in the other component

allyoucanmap (commented):
I'll list here some issues I found while testing:

- [x] Missing style tab

image

- [ ] The text editor is activated by clicking on it and the available surface is small. Probably we should change the style a bit to make it fill the available space. Maybe we should keep always visible the toolbar for this edit case and keep it sticky to the.

image

- [ ] While working on the same point it seems that the popup are not cleared after select and unselect the same item. This need to be invastigated and see if it's causing issue with the usability

image

- [ ] Location/Identify intractions needs to be clarified. Now I does not seem possible to use both at the same time in the editor or in the story media map. Is it possible to use both at the same time instead to disable one or another while working on them?

current situation:
identify

I see that the issue descriptions suggests to temporary diasable the identify but imo we should avoid to disable if we can use them at the same time. @tdipisa what do you think?

Anyway if we want to temporary disable the filed we should use a different style from active.
to represent a temporary disabled field I would apply an opacity to all the form to make clear that the function is still activated, something like this:

image


- [ ] if the click does not intersect a loaction on a map an error show up. This happens only in the story media. Steps to reproduce:

  - edit a media map
  - activate and add a location
  - save the map settings
  - click on the media map (not on a location)

- [ ] position of location popup should be the location coordinates and not the click coordinates. We should also add the possibility to add custom inline style to the popover container to translate the y the same size of the height of the marker:

image

- [ ] It does not seem possible to change the title of the first element of the list
image

web/client/components/map/openlayers/DrawSupport.jsx
offtherailz: With the experience for support found in Swipe tool, maybe @AGMETEOR can find a solution to this. Please let me know if you still need my opinion here, when you start again working on it.

AGMETEOR: @offtherailz , should I create a new support maybe called "SimpleDrag"? If so, it should only be for OL implementation, right?

offtherailz: I Don't know, It depends how much time it takes and if this solves this problem. Anyway the crucial point is to solve the issue notified by @allyoucanmap, that knows how Swipe integration is done too, so I think you can solve this problem together. 
I only said to ask me if you need again my help.

tdipisa:
@allyoucanmap @AGMETEOR, see below for future reference.
 
I've updated the issue description to make more clear the interactin requirements between the edit of locations and the identify tool and other avalable setting options (see the first **Note** in the issue description).

Other pending issues here that still need to be solved are:

- Points reported by @allyoucanmap in [this comment](https://github.com/geosolutions-it/MapStore2/pull/5902#pullrequestreview-494499443) (some of them have been better clarified with the updated issue description as mentioned above)
- The current inability to open the DropDown for the location's type in the edit popup (**Style** tab)
- General review of Review of the L&F

tdipisa:
This is out of date. @offtherailz please create a new branch containing the current development state and refer it in the issue for future reference.

Issue #5539, opened 2020-06-23
Update Layer Title and Description


vlt1:
The title and description layers changes periodically, as we publish data every month.
MapStore does not automatically update this data, therefore a function shall be implemented to

a) update the title and description in the TOC from WMS and WMTS services for all used layers.
b) This function shall only be visible to the admin.
c) There shall be an option to update all layers at once
d) There shall be an option to update individually a single layers
e) There shall be an option to update a selection of layers (more than one, but not all)
f) The implemented function shall also give immediate feedback (warning), if the selected layers are not available in GeoServer anymore. (E.g. a layer has been implemented in MS, but the name in GeoServer has changed)

vlt1:
refers to [austrocontrol-C125/#168](https://github.com/geosolutions-it/austrocontrol-C125/issues/168)

offtherailz:
@tdipisa one question about this requirement:
```
At the end, the map state containing the updated information needs to be saved to make changes persistent

```
`needs to be saved` means that the user have to save it manually or you mean that this information will be automatically saved on the server when updated ?

tdipisa:
@offtherailz it means that the user have to save it manually

offtherailz:
I supposed.
The PR #5538 contains an automatic and sophisticated save, that updates only the layers saved. I think we should remove this part to avoid unnecessary complexity.

PR #5956, opened 2020-09-30
#5539 Update Layer Title and Description

vlt1:
## Description
Changes as per [this comment](https://github.com/geosolutions-it/austrocontrol-C125/issues/168#issuecomment-698444579).

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#5539 
[austrocontrol-C125/#168](https://github.com/geosolutions-it/austrocontrol-C125/issues/168)

#5539 

**What is the new behavior?**
Description

## Breaking change
 - [x] No

### DIFF
--- a/createProject.js
+++ b/createProject.js
@@ -84,6 +84,10 @@ function doWork(params) {
             return project.initGit(params.outFolder);
         })
         .then(() => {
+            process.stdout.write('git init\n');
+            return project.updateSubmoduleBranch(params.outFolder);
+        })
+        .then(() => {
             process.stdout.write('git repo OK!\n');
             process.exit();
         })
--- a/docs/developer-guide/configuration-files.md
+++ b/docs/developer-guide/configuration-files.md
@@ -13,7 +13,7 @@ This separation allows to:
 They are `.properties` files or `.xml` files, and they allow to configure the various parts of the back-end.
 They are located in `web/src/main/resources` and they will be copied in  `MapStore.war` under the directory `/WEB-INF/classes`.
 
-* `proxy.properties`: configuration for the internal proxy (for cross-origin requests)
+* `proxy.properties`: configuration for the internal proxy (for cross-origin requests). More information [here](https://github.com/geosolutions-it/http-proxy/wiki/Configuring-Http-Proxy>).
 * `geostore-datasource-ovr.properties`: provides settings for the database.
 * `log4j.properties`: configuration for back-end logging
 * `sample-categories.xml`: initial set of categories for back-end resources (MAP, DASHBOARD, GEOSTORY...)
--- a/docs/developer-guide/externalized-configuration.md
+++ b/docs/developer-guide/externalized-configuration.md
@@ -114,7 +114,7 @@ geostoreVendorAdapter.showSql=false
 
 ## Externalize front-end Configurations
 
-MapStore by default fetches configuration files directly via HTTP, without any needing of any special back-end service in the middle (this simplifies the usage in a no-back-end context, as a framework). The `app.jsx` of **custom projects** allows to customize these URLs for these static configuration files (usually relative paths) to use your custom versions.
+MapStore by default fetches configuration files directly via HTTP, without any needing of any special back-end service in the middle (this simplifies the usage in a no-back-end context, as a framework). The `app.jsx` of the **MapStore project** allows to customize these URLs for these static configuration files (usually relative paths) to use your custom versions.
 
 MapStore back-end provides a configuration service utility (at `/rest/config/load`) to provide the configuration files from the data directory.
 
--- a/docs/developer-guide/project-creation-script.md
+++ b/docs/developer-guide/project-creation-script.md
@@ -25,7 +25,7 @@ Finally, to create the project, use the following command:
 node ./createProject.js <projectType> <projectName> <projectVersion> <projectDescription> <gitRepositoryUrl> <outputFolder>
 ```
 
-All the arguments except gitRepositoryUrl are mandatory:
+Note that projectName and outputFolder are mandatory:
 
 * **projectName**: short project name that will be used as the repository name on github, webapp path and name in package.json
 * **projectType**: type of project to create, currently two types of projects are supported:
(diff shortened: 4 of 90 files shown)


offtherailz (changes requested):
- [ ] It works, but I think editing of the layer name should be a little more explicit
because the refresh button says nothing, and if you click on save on top, the name is not updated, it can be hard to understand for a user.

I think we should instead: 
- Start with disabled name editing, allow editing from a button (pencil icon)
image. tooltip for the pencil button is "Edit layer name"
- When click on the pencil the input enable and you can edit. The button on the right now is an "ok" icon ("v" like check icon). tooltip for ok icon is "Confirm layer name change"
- When click on confirm, we disable again the input and do the refresh, going back to initial state.

I think it should be better, in case of error, to leave the input enabled, but I think it is hard (the state) so I propose this solution.

- [ ] Unit tests

tdipisa:
> I think we should instead:
> 
> * Start with disabled name editing, allow editing from a button (pencil icon). tooltip for the pencil button is "Edit layer name"
>   image
> * When click on the pencil the input enable and you can edit. The button on the right now is an "ok" icon ("v" like check icon). tooltip for ok icon is "Confirm layer name change"
>   image
> * When click on confirm, we disable again the input and do the refresh, going back to initial state.

I agree with @offtherailz about the icons to use, it is also specified in the requirements:

>The layer name should be editable manually in Layer Settings (**a button will enable editing of the name and allow save**);

There is also the following part in the same requirement @offtherailz:

>once the settings are saved, the layer will be updated (name and description) and re-rendered. If name is wrong, a notification will show the error

We had the intention to persist changes only when the settings are saved. Let's talk about it one minute.

offtherailz:
it should already work this way

Beyond what you edit in the settings, if you close without saving it asks you for confirmation and eventually cancels the changes.

All changes are applied "live" to the map, but editing the layer name can be dangerous (for this reason is disabled by default) and while editing you should not update (for this reason we ask confirm).

vlt1:
@offtherailz @tdipisa changes persist only when pressing on save button

offtherailz:
@vlt1 I think there is too much logic in layers epic/actions/reducers for a functionality like this to maintain. Moreover, this is not a generic "layers" functionality, so It should be isolated. 
I think as much self contained as possible. For this reason I suggested to use internal state (useState), maybe a small dedicated component.

offtherailz (changes requested):
@vlt1 I think there is too much logic in layers epic/actions/reducers for a functionality like this to maintain. Moreover, this is not a generic "layers" functionality, so It should be isolated. 
I think as much self contained as possible. For this reason I suggested to use internal state (useState), maybe a small dedicated component.

Sorry if I didn't specified this implementation detail in the review, I though it was implicit in the evolution of your previous component (that had some internal state) to do it this way.

vlt1:
@offtherailz the external actions are needed for the epic to trigger them correctly on layer load event

offtherailz:
@vlt1 the first implementation contained it [here](https://github.com/geosolutions-it/MapStore2/pull/5956/commits/613577cf422c1a568f79e590374747eacc46d224#diff-4b3be91e5c684f0262d1806a1a14f4c9R159).  I think it was enough

vlt1:
@offtherailz current implementation offers better user experience

offtherailz:
@vlt1 yes, I like very much the user interface you did. 

But having the self contained solution works also in other use case of the TOC (e.g. maps in dashboards or in stories). 
For this reason a good compromise can be having an acceptable user experience with the functionality reusable also in the other use cases. 
Otherwise the we have to find a solution for each TOC use case (and maintain them). I think is easier for us.

vlt1:
@offtherailz I've just realized layer loading and error information is kept and updated within the layer object(in General component it is the element prop I believe), so I can use that information, which means I can delete all actions and move edit state to component state and that should allow me to be able to keep the current UX. So new actions, state, reducers introduced can be removed.

offtherailz:
Great. Thanks, 
Please check also if your changes work in dashboards map widget TOC and geostory map editor inline or mediaeditor, so we are sure that the solution is reusable in a lot of use case. In case of error handling problem in other contexts, do not implement complex error handling, but try to keep it simple as much as possible.

vlt1:
@offtherailz Does name editing need to work also in geostory and dashboard? If so that it will need its own implementation anyway, because that MapView is separate from the main view and it's epics. Currently implementation is for map viewer only.

offtherailz:
@vlt1 the old editor you made here (613577cf422c1a568f79e590374747eacc46d224) was working in all the contexts, except for the notification of the error, and I think it was acceptable. For this reason I asked to continue on that way in a self contained component, and only to change the internal behavior, limiting as much as possible the external interactions.

Of course, the TOCSettings should continue to work and be consistent in every use context inside the application.

vlt1:
@offtherailz In other contexts there is no mechanism for notifying the state of the layer (is it loading, has been loaded, with what error) it seems to me. That is a problem I think, because the user can save a layer with a wrong layer name that doesn't work, since to check whether layer load or not we need to update it to force it to load. I'm not sure though need to investigate it more, since I've never worked with that code before.

offtherailz:
@vlt1 for the moment you can avoid to notify the error in other contexts then the main map viewer, because only the map can support error catching due to tile loading errors. In the future we may allow to catch the errors also from other maps then the main viewer, or doing a request directly in the component that do this check.

For the moment, take only care of protecting the editing of the layer name the way I asked, in a self contained component.

vlt1:
@offtherailz The functionality is contained within General component, I've added a prop that enables more feedback for the user if it is possible within the context the General component is used.

offtherailz (changes requested):
- You forgot the tooltips for the edit/confirm buttons ;) . Just this to merge this PR
- maybe all the logic of edit confirm etc... stays well in a separated component "LayerNameEditor", instead of mixing all in General.jsx, anyway, this is a plus and there is no to much time.

offtherailz (approved):

PR #5955, opened 2020-09-30
SpyGlassSupport for Openlayers(#5899)

AGMETEOR:
## Description
This PR adds the SpyGlass mode of Swipe tool

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: Enhancement

## Issue

**What is the current behavior?**

#5899 

**What is the new behavior?**
Screen-Recording-2020-09-30-at-19 47 57

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

Issue #5898, opened 2020-09-18
Swipe Tool Configuration / Switcher


offtherailz:
This tool is available when the swipe tool is active. It allows to:
- Change swipe or spyglass tool
- Allows to select the radius for the spyglass.
- (future change could allow to change horizontal or vertical cut for swipe)

AGMETEOR:
@tdipisa is it possible to get some mockups for this issue?

tdipisa:
@AGMETEOR yes of course, we will provide it asap. in the meantime you can continue with the implementation of the core functionality by developing a floating config panel, we will probably opt for a button somewhere to open this once the tool is active.

tdipisa:
@AGMETEOR 
below a simple mockup of what the tool activation button  should looks like to switch between the two modes (swipe/spyglass)

![5893_mockup](https://user-images.githubusercontent.com/1280027/94675916-cab7ac80-031a-11eb-84f6-b4f9eef5801e.png)

If the user clicks on **Configure**, the panel illustrated below opens to set the spyglass size (in px i guess) if the spyglass mode is active. 

![mockup_spyglass](https://user-images.githubusercontent.com/1280027/94677081-9c3ad100-031c-11eb-925c-104b918ac97b.png)

In case the of Swipe active instead, the same panel opens but with the possibility to set the swipe mode (vertical/horizontal) through two radio buttons or a dropdown (maybe it is better)

AGMETEOR:
@tdipisa there's also some form of configuration to do for the Swipe tool i.e. switching between Horizontal and vertical swipe

tdipisa:
> @tdipisa there's also some form of configuration to do for the Swipe tool i.e. switching between Horizontal and vertical swipe

@AGMETEOR it is already considered in my comment above.

PR #5969, opened 2020-10-02
#5898: Swipe Tool Configuration / Switcher

AGMETEOR:
## Description
This PR adds UI to allow switching to different modes of MapSwipe support and configure the different modes

- x ] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#5898 

**What is the new behavior?**
Adds UI for MapSwipe tool configuration
Screenshot 2020-10-05 at 17 53 20
Screenshot 2020-10-05 at 17 54 57
Screenshot 2020-10-05 at 17 55 15

## Breaking change
 - [x] No

## Other useful information
Needs #5955 to merge first


web/client/components/TOC/Toolbar.jsx
```diff
@@ -330,14 +341,14 @@ class Toolbar extends React.Component {
...
                     {this.props.activateTool.activateSwipeOnLayer && (status === 'LAYER') &&
-                        <OverlayTrigger
-                            key="layerSwipe"
-                            placement="top"
-                            overlay={<Tooltip id="layer-tooltip-swipe"><Message msgId="toc.compareTool" /></Tooltip>}>
-                            <Button key="layer-swipe" bsStyle={this.props?.swipeSettings?.active ? "success" : "primary"} className="square-button-md" onClick={() => this.showSwipeSettings(status)}>
-                                <Glyphicon glyph="transfer" />
-                            </Button>
-                        </OverlayTrigger>}
+                        <DropdownButtonT
```
offtherailz, line 344: Can you try to include this with the diff I provided to you for toolbar ? I'd like to start to separate and reduce dependencies between plugin as much as possible, so we can make also the other buttons pluggable, in the near future?

web/client/plugins/SwipeSettings.jsx
```diff
@@ -8,11 +8,100 @@
...
+        toolMode,
+        swipeModeSettings,
+        spyModeSettings,
+        onSetConfigurationActive = () => {},
+        onSetSpyToolRadius = () => {},
+        onSetSwipeToolDirection = () => {}
+    }) => {
+    return (<div className="mapstore-swipe-settings">
+        <DockabePanel
+            title={toolMode === "spy"
```
offtherailz, line 71: Please use 2 different labels for this. msgParams are ok for numbers or fixed strings, but in some languages also "splyglass" or "swipe" strings may be localized.

web/client/components/TOC/Toolbar.jsx
```diff
@@ -330,14 +341,14 @@ class Toolbar extends React.Component {
...
+                            bsStyle={this.props?.swipeSettings?.active ? "success" : "primary"}
+                            noCaret {...swipeToolButtonConfig}>
+                            <MenuItem active={this.props.swipeSettings.mode === "swipe"} onClick={() => this.showSwipeTools(status, "swipe")}>Swipe</MenuItem>
```
offtherailz, line 347: Please localized tool titles. Use "Swipe" and "Spy glass" for all the languages for the moment

offtherailz (changes requested):
- [ ] Without a separation between the button and the dropdown, I can not turn off the swipe anymore. This is a little missleading. 
image
please use `SplitButton` from react-bootstrap. When the user clicks, the swipe vertical is active by default. The dropdown menu should highlight the current tool line. Maybe also some icons may help. There is a problem with dropdown menu style that should be solved. What do you think @allyoucanmap and @tdipisa ? 
image
- [ ] I can't click on "Configure" when no tool is not active, and there has no effect.
- [ ] Dragging the configure window has a slowed down effect, probably due to a CSS effect. Please remove this effect. 
- [ ] The config window is a little to oversized for the controls that it contains. reduce a little the height, and the width
image
image

- [ ] The radius extremes go from 50px to 80px. I think we should expand a little this interval. It should be 
    - min 20, 
    - max: 200
    - step: 10
Peek 05-10-2020 10-19

allyoucanmap:
> There is a problem with dropdown menu style that should be solved. What do you think @allyoucanmap and @tdipisa ?

All the buttons in the toolbar are using the `@square-btn-medium-size` .less variable to set the width and the height so I suggest to use `@square-btn-medium-size` to set the height of the dropdown button and make it aligned with others buttons. We could also narrow the width by shrinking a bit the horizontal padding.

web/client/components/TOC/Toolbar.jsx
```diff
@@ -330,14 +341,26 @@ class Toolbar extends React.Component {
...
+                            <MenuItem
+                                active={this.props.swipeSettings.mode === "spy"}
+                                onClick={() => this.props.onToolsActions.onSetSwipeMode("spy")}>
+                                <Glyphicon glyph="search" /><Message msgId="toc.spyGlass" />
+                            </MenuItem>
+                            <MenuItem
+                                onClick={() => this.showConfiguration(status)}>
+                                <Glyphicon glyph="cog" /><Message msgId="toc.configureTool" />
+                            </MenuItem>
+                        </SplitButtonT>
```
offtherailz, line 362: Can you wrap all the logic here into a single "SwipeButton" component? this is useful for future extraction that will come soon.
showSwipeTools, showConfiguration, splitToolButtonConfig etc... should go there.

web/client/themes/default/less/swipe.less
```diff
@@ -13,11 +13,42 @@
 }
 
 .mapstore-swipe-settings {
+
+    .mapstore-swipe-settings-modal {
+        min-height: 22% !important;
+        height: auto !important;
+        width: 25% !important;
+        overflow: visible !important;
```
offtherailz, line 21: you should avoid `!important` as much as possible. They should be used only to override some inline styles. Otherwise please use css rules priority system to make these rules win over the others

offtherailz (changes requested):
- [ ] See my comments
- [ ] About the dialog I think adding this in modal.less can fix the issue for all modals in  client/themes/default/less/modal.less . 
```
.modal.fade.in .modal-dialog.modal-dialog-draggable{
    .transition(none);
}
```
- [ ] Here the drop-down menu looks another button. 
image
We should reduce a the `dropdown-toggle` width for that toolbar (maybe is a good idea to make it general for that toolbar size) and, maybe, make it appear only when the tool is active. I tried this and it looks good to me
See this suggestion:
https://github.com/geosolutions-it/MapStore2/pull/5969#discussion_r500242360
image

Another small thing. When the button is not active, if I select swipe or spyglass, the tool should activate, as well as when I click on the button.

allyoucanmap:
@offtherailz 

> About the dialog I think adding this in modal.less can fix the issue for all modals. Confirm @allyoucanmap

What's the issue with the modals?

> but I didn't investigated if and where it should be placed in general theme, @allyoucanmap?, any suggestion ?

it's a custom style for the dropdown button of swipe plugin so it could be placed in the swipe.less.

> Maybe we should show the dropdown only when the button is active?

I would avoid to change size or structure of this button if possible

web/client/themes/default/less/toc.less
```diff
@@ -76,6 +75,15 @@
...
             margin: (@square-btn-size - @square-btn-medium-size)/2 0 0 0;
+
+            .btn-group {
+                margin-top: 0px;
+
+                .dropdown-toggle {
+                    width: @square-btn-medium-size;
+                    height: @square-btn-medium-size;
+                }
+            }
```
offtherailz, line 86: As discussed with @allyoucanmap  Remove this 
```suggestion
```

in favor of this to put in `square-button.less`
```
.dropdown.btn-group > .square-button-md+button.dropdown-toggle {
    height: @square-btn-medium-size;
    padding: 2px;
}
```

offtherailz (approved):

Issue #5952, opened 2020-09-30
Align/Review of some dependencies versions


tdipisa:
We have to review the alignment for next MS releases. Versions to use are reported below (those selected are versions yet to be created):

**master**
- [x] print-lib `2.0-SNAPSHOT`
- [x] http-proxy `1.2-SNAPSHOT`
- [x] geostore `1.6-SNAPSHOT`
- [x] mapstore-backend `1.0-SNAPSHOT`

**2020.02.02** and **2020.02.xx**
- [x] print-lib `geosolutions-2.0`
- [x] http-proxy `1.1.0`
- [x] geostore `1.5.0`
- [x] mapstore-backend `1.0`

**2020.03.00** and **2020.03.xx** (as soon as the branch is created, it doesn't exists yet)
- [ ] print-lib `geosolutions-2.0`
- [ ] http-proxy `1.1.0`
- [ ] geostore `1.6.0`
- [ ] mapstore-backend `1.0`

edit from @offtherailz : 
- [ ] For mapstore back-end, we should also schedule deploy on our maven report from jenkins

offtherailz:
can not be tested by hand

PR #5990, opened 2020-10-09
#5952: set mapstore-backend and print-lib to use a tagged version, added machinery to deploy mapstore-backend to maven with mvn deploy

mbarto:
## Description

 - [x] Set fixed version for Java packages to 1.0-2020.02
 - [x] Set fixed version for print-lib dependency to 2.0-2020.02
 - [x] Added configuration to deploy java artifacts for mapstore-backend to GeoSolutions maven

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: release versioning

## Issue
#5952

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -4,7 +4,7 @@
   <groupId>it.geosolutions.mapstore</groupId>
   <artifactId>mapstore-backend</artifactId>
   <packaging>jar</packaging>
-  <version>1.0-SNAPSHOT</version>
+  <version>1.0-2020.02</version>
   <name>MapStore 2 - Backend</name>
   <url>http://www.geo-solutions.it</url>
 
@@ -97,7 +97,23 @@
                 </configuration>
             </plugin>
     </plugins>
+    <extensions>
+        <!--.............................................-->
+        <!--       GeoSolutions (using wagon ftp)       -->
+        <!--.............................................-->
+      <extension>
+       <groupId>org.apache.maven.wagon</groupId>
+       <artifactId>wagon-ftp</artifactId>
+       <version>1.0-beta-2</version>
+      </extension>
+    </extensions>
   </build>
+  <distributionManagement>
+   <repository>
+    <id>geosolutions</id>
+    <url>ftp://maven.geo-solutions.it/</url>
+   </repository>
+  </distributionManagement>
   <profiles>
   </profiles>
     <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <groupId>it.geosolutions.mapstore</groupId>
   <artifactId>mapstore-root</artifactId>
   <packaging>pom</packaging>
-  <version>1.0-SNAPSHOT</version>
+  <version>1.0-2020.02</version>
   <name>MapStore 2</name>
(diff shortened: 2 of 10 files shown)


backend/pom.xml
```diff
@@ -4,7 +4,7 @@
   <groupId>it.geosolutions.mapstore</groupId>
   <artifactId>mapstore-backend</artifactId>
   <packaging>jar</packaging>
-  <version>1.0-SNAPSHOT</version>
+  <version>1.0-2020.02</version>
```
offtherailz, line 7: Seems good. 
One question: Is not clear to me why 1.0-2020.02.
What is the naming convension intended of this?
I think one of the following is more rational:
 - 1.0 and for 2020.02 
 - 1.1 or 2020.03
 and so on...

Or:
- 1.0-2020.02
- 1.0-2020.03
and so on.

I'm sure there is a reason behind it but I can't get it.

offtherailz (changes requested):
We have to: 
- [x] Update the release procedure with the necessary version to increase and naming convention to use (when clarified)
- [ ] Schedule the deploy for all these versions (stable and snapshot) --> this is not part of this PR but for sure of the task. We should indicate what builds should deploy in order to get the correct pacakges deployed and available.

backend/pom.xml
mbarto: The idea is to make the relationship between the MapStore version and the backend version clear, so I think it is worth having 2020.02 as part of the version.
Removing 1.0 (so having only 2020.02) would work the same, feel free to use the versioning schema you prefer.

offtherailz: Yes, I supposed. I said because I've seen your edit on the issue, and it was not clear: 

> **2020.02.02** and **2020.02.xx**
> - [ ] print-lib v2.0-2020.02
> - [x] http-proxy v1.1.0
> - [x] geostore v1.5.0
> - [ ] mapstore-backend v1.0-2020.02
> 
> **2020.03.00** and **2020.03.xx** (as soon as the branch is created, it doesn't exists yet)
> - [ ] print-lib v2.0.0
> - [ ] http-proxy v1.1.0
> - [ ] geostore v1.6.0
> - [ ] mapstore-backend v1.0.0
> 
@tdipisa what do you think? Having this makes sense:
**2020.02.02** and **2020.02.xx**
- [ ] print-lib v2.0-2020.02
- [x] http-proxy v1.1.0
- [x] geostore v1.5.0
- [ ] mapstore-backend **v2020.02**

**2020.03.00** and **2020.03.xx** (as soon as the branch is created, it doesn't exists yet)
- [ ] print-lib v2.0.0
- [ ] http-proxy v1.1.0
- [ ] geostore v1.6.0
- [ ] mapstore-backend **v2020.03**

tdipisa: @offtherailz @mbarto in my opinion it would be better to have something as @mbarto suggested so that we are able to better provide minor versions of the backend if needed. Therefore, having something like @mbarto provided in this PR is fine in my opinion.

offtherailz: good point.
so I suggest 2020.02-1.0 etc...

offtherailz (approved):

PR #6000, opened 2020-10-12
Ables Tabular view to display GeoJSON data from a vector layer

FNI18300:
## Description
Current MapStore release could display data in tabular view only for a WMS/WFS layer.
The current proposal is designed to display data in tabular view for a vector layer containing GeoJSON data.
The original data are stored in a temporary variable and filtering (from the header of the tabular view) is done locally from original data.
The behaviour is not modified for WMS/WFS layer.

Two files are modified to be able to check the layer type "Vector" and do the same job that the original block does for WMS/WFS.

- [ x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

when we add a vector layer (from GeoJSON datas), opening the the tabular view display nothing.

#<issue>
when we add a vector layer, opening the tabular view display the features is the layer. All existing tools in the tabular view still working and filtering from the header too.

## Breaking change
 - [x] No

## Other useful information
N/A

### DIFF
--- a/web/client/epics/wfsquery.js
+++ b/web/client/epics/wfsquery.js
@@ -6,6 +6,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+const {head} = require('lodash');
+
 const Rx = require('rxjs');
 const axios = require('../libs/ajax');
 
@@ -16,6 +18,9 @@ const {paginationInfo, isDescribeLoaded, layerDescribeSelector} = require('../se
 const {mapSelector} = require('../selectors/map');
 const {authkeyParamNameSelector} = require('../selectors/catalog');
 const {getSelectedLayer} = require('../selectors/layers');
+const {selectedLayerSelector} = require('../selectors/featuregrid');
+const {layerLoad} = require('../actions/layers');
+
 
 const CoordinatesUtils = require('../utils/CoordinatesUtils');
 const { addTimeParameter } = require('../utils/WFSTimeUtils');
@@ -86,6 +91,39 @@ const featureTypeSelectedEpic = (action$, store) =>
                 const geometry = info.geometry[0] && info.geometry[0].attribute ? info.geometry[0].attribute : 'the_geom';
                 return Rx.Observable.of(changeSpatialAttribute(geometry));
             }
+
+            const setedLayer = selectedLayerSelector(state);
+            if (setedLayer.type ==='vector') {
+                return Rx.Observable.defer( () =>axios.get(action.url))
+                    .map((response) => {
+
+                        var originalData={
+                            featureTypes: [response.data.featureTypes[0]]
+                        }
+                        var info = {
+                            geometry: originalData.featureTypes[0].properties
+                                .filter((attribute) => attribute.name.indexOf('geometry') === 0)
+                                .map((attribute) => attribute
+                                ),
+                            original: originalData,
+                            attributes: describeFeatureTypeToAttributes(originalData)
+                        }
+
+                        const geometry = info.geometry[0] && info.geometry[0].attribute ? info.geometry[0].attribute : 'the_geom';
+                        var changeSpatialAttributePromise = new Promise((resolve) => {
+                            resolve(changeSpatialAttribute(geometry))
+                        });
+                        var promiseFeatureType = changeSpatialAttributePromise.then(() => {
+                            return new Promise((resolve) => {
+                                resolve(featureTypeLoaded(action.typeName, info))
+                            });
+                        })
+                        return Rx.Observable.defer( () => promiseFeatureType);
+                    })
+                    .mergeAll()
+
+            }
+
             return Rx.Observable.defer( () => axios.get(ConfigUtils.filterUrlParams(action.url, authkeyParamNameSelector(store.getState())) + '?service=WFS&version=1.1.0&request=DescribeFeatureType&typeName=' + action.typeName + '&outputFormat=application/json'))
                 .map((response) => {
                     if (typeof response.data === 'object' && response.data.featureTypes && response.data.featureTypes[0]) {
@@ -122,11 +160,14 @@ const featureTypeSelectedEpic = (action$, store) =>
 const wfsQueryEpic = (action$, store) =>
     action$.ofType(QUERY)
         .switchMap(action => {
+            const state = store.getState();
             const sortOptions = getDefaultSortOptions(getFirstAttribute(store.getState()));
             const totalFeatures = paginationInfo.totalFeatures(store.getState());
             const searchUrl = ConfigUtils.filterUrlParams(action.searchUrl, authkeyParamNameSelector(store.getState()));
             // getSelected Layer and merge layerFilter and cql_filter in params  with action filter
             const layer = getSelectedLayer(store.getState()) || {};
+            const setedLayer = selectedLayerSelector(state);
+
             const {layerFilter, params} = layer;
             const cqlFilter = find(Object.keys(params || {}), (k = "") => k.toLowerCase() === "cql_filter");
             const cqlFilterRules = cqlFilter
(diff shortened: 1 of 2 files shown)


tdipisa:
@offtherailz we need to check which milestone must be assigned to this PR considering the release plan we have [here](https://github.com/georchestra/mapstore2-georchestra).

offtherailz (changes requested):
This is going to be a good feature. :+1: 

This doesn't break anything and is a partial implementation of tabular view for geojson, so it can be merged when accepted. It is better to make it available in 2020.03. if possible (where georchestra will be aligned after the official release, now is on 2020.02.xx).

**Here my requests:**
- [ ] I did this PR for adjust and improve the functionalities you introduced https://github.com/FNI18300/MapStore2/pull/1. 
    -  fixing problems with state handling and making it analogue to the filtering workflow in WMS.
    - commenting the number filtering, that doesn't work. If you need that feature, you have to fix it

- [ ] Because it is disabled by default, in order to maintain it, I suggest to add some unit tests (to avoid regressions and guarantee future maintainance).


Off topic:
If I understood correctly, you open the feature grid on your own in your project, isn't it?
To enable it in MapStore it is sufficent to add something it like this and add the `search` part in the layer :
```
diff --git a/web/client/components/TOC/Toolbar.jsx b/web/client/components/TOC/Toolbar.jsx
index 09b744d44..34636fab2 100644
--- a/web/client/components/TOC/Toolbar.jsx
+++ b/web/client/components/TOC/Toolbar.jsx
@@ -269,7 +269,7 @@ class Toolbar extends React.Component {
                             </Button>
                         </OverlayTrigger>
                         : null}
-                    {this.props.activateTool.activateQueryTool && status === 'LAYER' && this.props.selectedLayers[0].search && !this.props.settings.expanded && !this.props.layerMetadata.expanded && !this.props.wfsdownload.expanded ?
+                    {this.props.activateTool.activateQueryTool && status === 'LAYER' && (this.props.selectedLayers[0].search || this.props.selectedLayers[0].type === 'vector') && !this.props.settings.expanded && !this.props.layerMetadata.expanded && !this.props.wfsdownload.expanded ?
                         <OverlayTrigger
                             key="featuresGrid"
                             placement="top"
@@ -358,7 +358,7 @@ class Toolbar extends React.Component {
 
     browseData = () => {
         this.props.onToolsActions.onBrowseData({
-            url: this.props.selectedLayers[0].search.url || this.props.selectedLayers[0].url,
+            url: this.props.selectedLayers[0].search?.url || this.props.selectedLayers[0].url,
             name: this.props.selectedLayers[0].name,
             id: this.props.selectedLayers[0].id
         });
```
For the moment it is still a partial implementation (no advanced filter, number and boolean still to be impemented, maybe need to disable cross origin filters... ) and so can not be enabled by default. But when completedI think it is possible.

offtherailz (changes requested):
Thank you very much. :+1: 
In order to make this PR mergiable on MapStore:
- [ ] conflicts have to be solved (recent changes may generate them)
- [ ] Unit tests should pass. (see the [travis](https://travis-ci.org/github/geosolutions-it/MapStore2/builds/734949086?utm_source=github_status&utm_medium=notification) build) I noticed a couple of test are failing, but I didn't checked them yet. Please, if you can, fix them and, if you want your changes to be more robust, add your own. 

**hint**: in order to speed up  tests fix 
- edit the file `test.webpack.js` 
```javascript
var context = require.context('../web/client/...PATH OF FILES TO TEST', true, /(-test\.jsx?)|(-test-chrome\.jsx?)$/);
```
- use `npm run continuoustest` to re-run the tests as you save the files.

This partial implementation is very useful for a future full implementation of a full support of attribute table for local vector layer. In order to make this feature available in the official MapStore (from the table button of the feature grid) we will need to make all the UI consistent (e.g. advanced filters and any kind of filtering working, or disable the one not enabled). This feature will stay not accessible in MapStore because of the button "open attribute table is not shown",  until we do not finalize these parts, but, as far as I understood, these changes are sufficient for your use case.

offtherailz:
Hi @FNI18300 , I see 410 files that differs from the current master and 3 files in conflict. Could you please check if everything is good? 
I'm afraid something get wrong during merge.

FNI18300:
During the 15 past days, many changes have been done on files mainly to replace this kind of import 
"import LocaleUtils from '../../utils/LocaleUtils';" by  this one "import {getMessageById} from '../../utils/LocaleUtils';" 

I check all changes and it ok for me
I check the 3 conflicting files and the merge is ok because the change made in import describre at the beginning of my message is already integrated in my code.
The rename of file FilterUtils.jsx into FilterUtil.js is also integrated.

offtherailz:
Thank you. 
I Don't know why the diff is so long. It should show only your changes, not the ones from the master. Maybe you did some rebase?
Anyway, if you do a merge with master branch, solving the current conflicts, we should be able to see only differences between master and your branch. I tried on my fork to check and it seems to work, but I'm not sure of my changes. 
https://github.com/geosolutions-it/MapStore2/compare/master...offtherailz:tabular-view-for-geojson-data

web/client/plugins/Map.jsx
```diff
@@ -27,6 +27,7 @@ import mapTypeReducer from "../reducers/maptype";
 import additionalLayersReducer from "../reducers/additionallayers";
 import mapEpics from "../epics/map";
 import pluginsCreator from "./map/index";
+import createFeatureFilter from '../utils/FilterUtils';
```
offtherailz, line 30: This causes an error when trying to load a vector layer.
```suggestion
import {createFeatureFilter} from '../utils/FilterUtils';
```

web/client/epics/wfsquery.js
```diff
@@ -153,11 +192,11 @@ export const wfsQueryEpic = (action$, store) =>
                 ...queryOptions
             };
             return Rx.Observable.merge(
-                getLayerJSONFeature({...layer, search: {...layer.search, url}}, ogcFilter, options)
+                (typeof filterObj === 'object' && getJSONFeatureWA(searchUrl, action.filterObj, options, selectedLayer) || getLayerJSONFeature({...layer, search: {...layer.search, url}}, ogcFilter, options))
                     .map(data => querySearchResponse(data, action.searchUrl, action.filterObj, action.queryOptions, action.reason))
                     .catch(error => Rx.Observable.of(queryError(error)))
                     .startWith(featureLoading(true))
-                    .concat(Rx.Observable.of(featureLoading(false)))
+                    .concat(Rx.Observable.of(featureLoading(false)), Rx.Observable.of(layerLoad(layer.id)))
```
offtherailz, line 199: I don't understand the reason of these changes.  
I think you can avoid almost for sure the first row. Not sure about the `layerLoad` event at the end of the second changed row.

offtherailz (commented):
@FNI18300 can you explain the rows I highlighted ? 
The rest seems not to influence MapStore  in general, so it's ok and a good starting point for supporting attribute browsing :+1:

web/client/epics/wfsquery.js
FNI18300: the line 195 is changed because the getJSONFeatureWA signature from observables/wfs.js is changed => we add a parameter layer (set to selectedLayer in this call) which is used into getJSONFeature methods.
The original code was build this way with a call to getJSONFeatureWA and getLayerJSONFeature. We just add the new parameter

the line 199 is changed to make the layer loaded when it was a vector layer. This help to take into accound filter and so on.

offtherailz: You should use instread the new method, to keep the things in sync. Can you use the new code?

offtherailz (changes requested):
Hi, the code looks good, I'd like to test it, but I noticed you introduced a `FilterUtils.jsx` (with the final 'x') aside from FilterUtils.js. Can you please check and make your changes converge to FilterUtils.js?

offtherailz (approved):

tdipisa:
@ElenaGallo please do a test tour of the attribute table in DEV as soon as merged to identify possiblel regressions due to this update.

ElenaGallo:
@tdipisa we have no regressions with this PR

tdipisa:
Thank you @ElenaGallo

Issue #5136, opened 2020-04-09
Improvements to TMS catalog and layers


offtherailz:
## Description
Here a list of things we could improve in current TMS catalog support 

- [ ] Support for variants in custom TMS services (to allow a list)
- [ ] Update and/or make extrernalizable the list of well known services
- [ ] Add thumbnails, clear titles and descriptions for well known services (OSM, etc...) and their variants.

Related to #5088 

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

PR #6020, opened 2020-10-14
#5136: Fix issue in TMS printing

offtherailz:
## Description
If the map has custom resolutions the TMS do not print correctly. 
This because the print spec is generated on map resolutions, instead of "x,y,z EPSG:3857 standard resolutions".
Moreover, it adds a filter to exclude resolution higher than the `maxNativeZoom` so the server can get the tiles at the best resolution ( e.g. NASAGIBS for higher level of zoom)

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#5136

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/PrintUtils.js
+++ b/web/client/utils/PrintUtils.js
@@ -473,7 +473,7 @@ const PrintUtils = {
         },
         tileprovider: {
             map: (layer) => {
-
+                // details here: http://www.mapfish.org/doc/print/protocol.html#xyz
                 const [providerURL, layerConfig] = getLayerConfig(layer.provider, layer);
                 if (!isEmpty(layerConfig)) {
                     let validURL = extractValidBaseURL({ ...layerConfig, url: layerConfig?.url ?? providerURL });
@@ -488,6 +488,7 @@ const PrintUtils = {
                         .replace("{x}", "${x}")
                         .replace("{y}", "${y}")
                         .replace("{z}", "${z}");
+                    // TODO: support bounds
                     return {
                         baseURL,
                         path_format: pathFormat,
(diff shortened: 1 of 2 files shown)


web/client/utils/PrintUtils.js
```diff
@@ -496,7 +497,33 @@ const PrintUtils = {
                         "opacity": getOpacity(layer),
                         "tileSize": [256, 256],
                         "maxExtent": [-20037508.3392, -20037508.3392, 20037508.3392, 20037508.3392],
-                        "resolutions": MapUtils.getResolutions()
+                        "resolutions": [
```
mbarto, line 500: What is the difference between this hardcoded list and the list previously returned by MapUtils.getResolutions ?

mbarto: Isn't getGoogleMercatorResolutions in MapUtils returing the same?

offtherailz: For getResolutions it is not the same, and this is the cause of the issue. If the map has different resolutions from the server, we have the bug.

Yes I think it is the same for `getGoogleMercatorResolutions`, I see it now that it should.
I've seen the same hardcoded resolutions in osm and other entries of print utils (where I copied it), so I supposed we needed to use that and didn't noticed if we could use it. 
Maybe an improvement is to get them with max as `maxNativeZoom`, but I'm not sure this works in every tile provider in the list.

offtherailz: I noticed that certain layers have problem with their maxNative in tests, 
So I should keep this explicitly hardcoded, as osm or mapquest, until we find out a better way to handle it.

Issue #6057, opened 2020-10-19
Allow OSM background selection for map with custom CRS


dsuren1:
## Description
Currently Mapstore allows only specific CRS and layers of type WMS, compatible WMTS and empty to be selectable/clickable from the background selector. 
For projects with custom CRS, we need to at least allow OSM regardless of the CRS, as the Openlayers can reproject on the fly based on the projection defs initialized from local config for the custom CRS

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

ElenaGallo:
@dsuren1 Opening a New map the background is empty and it is not possible to select any background from the Background Selector
![no_background](https://user-images.githubusercontent.com/56537133/97160504-bf6e6a00-177c-11eb-973d-90d45272e20e.png)

**How to Reproduce**
- Open MapStore
- Open New Map
- Click on the Background Selector

dsuren1:
@ElenaGallo 
I believe the new openlayer map is not loaded in your case and that's the reason for background selector thumbnails are disabled.
This change just allow OSM for custom CRS, and hasn't modified the existing flow. 
Also I tried to reproduce and unfortunately I couldn't.
![new_map_background_selector](https://user-images.githubusercontent.com/26929983/97254595-b8a32e00-1834-11eb-9014-f6735aad88f9.gif)

dsuren1:
@ElenaGallo In order to test with custom CRS, you need to modify the configs of Mapstore or you can test the same in the decsira application https://decsira-dev.geo-solutions.it/decsiraweb2/map.html#/map which uses custom CRS EPSG:32632

ElenaGallo:
@dsuren1 thank you for the note

PR #6060, opened 2020-10-20
#6057: Allow OSM background selection for custom CRS

dsuren1:
## Description
This PR adds a feature to allow the selection of OSM background layer from the background selector regardless of the CRS. As the Openlayer will be able to project based on the projection defs given in the local config for the custom CRS used.

 - [x] Enhancement

## Issue

#6057 

**What is the new behavior?**
User can select the background layer of type OSM for the custom CRS

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/background/PreviewIcon.jsx
+++ b/web/client/components/background/PreviewIcon.jsx
@@ -46,7 +46,7 @@ class PreviewIcon extends React.Component {
         const compatibleWmts = this.props.layer.type === "wmts" && has(this.props.layer.allowedSRS, this.props.projection);
         const containerClass = this.props.vertical ? 'background-preview-icon-container-vertical' : 'background-preview-icon-container-horizontal';
         const type = this.props.layer.visibility ? ' bg-primary' : ' bg-body';
-        const valid = ((validCrs || compatibleWmts || this.props.layer.type === "wms" || this.props.layer.type === "empty") && !this.props.layer.invalid );
+        const valid = ((validCrs || compatibleWmts || ["wms", "empty", "osm"].includes(this.props.layer.type)) && !this.props.layer.invalid );
 
         const click = !valid ? () => {} : () => {
             this.props.onToggle();
--- a/web/client/components/background/PreviewIcon.jsx
+++ b/web/client/components/background/PreviewIcon.jsx
@@ -8,7 +8,7 @@
 
 const React = require('react');
 const PropTypes = require('prop-types');
-const {indexOf, has} = require('lodash');
+const {indexOf, has, includes} = require('lodash');
 require('./css/previewicon.css');
 
 class PreviewIcon extends React.Component {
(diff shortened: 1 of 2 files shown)


web/client/components/background/PreviewIcon.jsx
```diff
@@ -46,7 +46,7 @@ class PreviewIcon extends React.Component {
         const compatibleWmts = this.props.layer.type === "wmts" && has(this.props.layer.allowedSRS, this.props.projection);
         const containerClass = this.props.vertical ? 'background-preview-icon-container-vertical' : 'background-preview-icon-container-horizontal';
         const type = this.props.layer.visibility ? ' bg-primary' : ' bg-body';
-        const valid = ((validCrs || compatibleWmts || this.props.layer.type === "wms" || this.props.layer.type === "empty") && !this.props.layer.invalid );
+        const valid = ((validCrs || compatibleWmts || ["wms", "empty", "osm"].includes(this.props.layer.type)) && !this.props.layer.invalid );
```
mbarto, line 49: includes is not supported by every browser, can you use a lodash function?

web/client/components/background/__tests__/PreviewIcon-test.jsx
```diff
@@ -36,29 +36,43 @@ describe("test the PreviewIcon", () => {
...
-        expect(node).toExist();
-        const image = node.querySelector('.background-preview-icon-frame').children[0];
-        ReactTestUtils.Simulate.mouseOver(image);
-        expect(spyLayerChange).toHaveBeenCalled();
-        ReactTestUtils.Simulate.click(image);
-        expect(spyPropertiesChange).toHaveBeenCalled();
-        expect(spyToggle).toHaveBeenCalled();
-        expect(spyPropertiesChange).toHaveBeenCalledWith("layer", {visibility: true});
-        expect(spySetCurrentBackgroundLayer).toHaveBeenCalledWith("layer");
+        layers.map((layer, i)=>{
```
mbarto, line 60: If you are doing side effects and not collecting the map function result, I prefer to use forEach, it makes the intent more clear, in my opinion

mbarto (changes requested):

tdipisa:
Hi, I would like to have also the @offtherailz review before merging this.

web/client/components/background/PreviewIcon.jsx
offtherailz: osm is valid in every CRS ?

offtherailz:
I know it works for sure in EPSG4326. Not sure of the final effect in other CRS. If it doesn't cause particular problem or errors, it can be merged.
For instance you could try EPSG:3003 or a couple of other Projections. I guess the project has some custom projections like this, so I'm quite confident that you already tested with them

dsuren1:
> I know it works for sure in EPSG4326. Not sure of the final effect in other CRS. If it doesn't cause particular problem or errors, it can be merged.
> For instance you could try EPSG:3003 or a couple of other Projections. I guess the project has some custom projections like this, so I'm quite confident that you already tested with them

Also, I have tried with EPSG:32632 and it works fine as well.

mbarto:
@offtherailz @dsuren1 openlayers can reproject OSM on the fly, so it should work on every supported CRS

web/client/components/background/PreviewIcon.jsx
dsuren1: Since we would have already specified the projection defs for the custom CRS, openlayer should be able to reproject, if I'm being right. Also this would help us select and deselect the OSM background, as of now we can enable but not re-select from other background.

mbarto (approved):

offtherailz (approved):

Issue #6077, opened 2020-10-22
Multiple selection on Map


tdipisa:
## Description

The filtering on map will be extended with the following:
    - Ctrl + click: add / remove features from filter list

At the moment using the "filter on map" tool the user can select only one feature by clicking on map and the attribute table is filtered accordingly (this happens both in vew and edit mode of the attribute table). There should be the possibility to perform a selection of multiple features with multiple clicks on map (as explained above), the attribute table will be filtered in the same way (in both vew and edit mode as it is now).

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: improvement on an existing feature

## Other useful information
The user guide related to the attribute table component is available [here](https://mapstore.readthedocs.io/en/latest/user-guide/attributes-table/).

ElenaGallo:
@AGMETEOR when more than one feature has been selected on the map and the SyncMap button is enabled, only the last selected feature is displayed on the map
![filter_map](https://user-images.githubusercontent.com/56537133/98276347-1c94c780-1f96-11eb-99c1-fe1eab3bc62b.gif)


**How to Reproduce**
- Open [this map]() 
- Open a layer's attribute table
- Click on Filter on map
- Select two features on map
- Click on SYNCMap

*Current Result*
It is visbile on map only the last feature selected

AGMETEOR:
Hi @ElenaGallo , seems to be working well. Could you please try again?

tdipisa:
@AGMETEOR I have the same result in DEV 

![image](https://user-images.githubusercontent.com/1280027/98371517-94b5c880-203c-11eb-8621-8224b70e1e34.png)

tdipisa:
[...]

PR #6121, opened 2020-10-28
#6077: Multiple selection on Map

AGMETEOR:
## Description

Allow Multiple selection and deselection when using Geometry filter

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: Enhancement

## Issue

#6077

**What is the new behavior?**

Screen-Recording-2020-10-29-at-14 20 34

## Breaking change
 - [x] No

## Other useful information
On Mac OS, use the command key instead of Ctrl


web/client/epics/featuregrid.js
```diff
@@ -290,10 +291,21 @@ module.exports = {
             .map(updateFilterFunc(store))
     ),
     /**
-     * Enables the Geometry filter when entering edit mode in feature grid.
-     * No filter value should have been set otherwise nothing is enabled.
-     * @memberof epics.featuregrid
+     * Resets multiselect and ctrl, metaKey modifiers in case they were true during map clicks
+     * This happens when filter is disabled
+     * @memberof epics.featuregird
```
MV88, line 296:
```suggestion
     * @memberof epics.featuregrid
```

web/client/reducers/mapInfo.js
```diff
@@ -288,6 +289,12 @@ function mapInfo(state = initState, action) {
             filterNameList: action.filterNameList || null
         });
     }
+    case UPDATE_MAPINFO_STATE_WITH_CLICK_POINT: {
```
MV88, line 292: could you not reuse FEATURE_INFO_CLICK ?

MV88 (changes requested):
Hi, seems good, but the only problem i've found is that the numbers are not updated with the current rows.

it is always shown **1 item of 1**

web/client/reducers/mapInfo.js
offtherailz: More precisely I think you should intercept the CLICK_ON_MAP action as well as the actual filter does. Click on map action should contain all the info, included the modifiers, so you don't need anything else, i think.

offtherailz: Map info is dedicated to feature info tool. In this case the click on map is handled by the other tool, so the mapinfo actions/apics should not be included at all

AGMETEOR:
@MV88 about this [comment](https://github.com/geosolutions-it/MapStore2/pull/6121#pullrequestreview-519826589) I can't seem to see the numbers info anymore. Maybe something was changed in the FeatureEditor code?

web/client/reducers/mapInfo.js
MV88: > @MV88 about this [comment](https://github.com/geosolutions-it/MapStore2/pull/6121#pullrequestreview-519826589) I can't seem to see the numbers info anymore. Maybe something was changed in the FeatureEditor code?

i mean the text at the bottom left when you have multiple rows selected
image

AGMETEOR: Alright. I had a small screen size with console open so I wasn't seeing it. I have fixed the count issue.

tdipisa:
@AGMETEOR is it ready for a new review?

MV88 (changes requested):
I've found three points that needs to be addressed:

1) improve tooltip for the selection because the user cannot be aware of using the CTRL for multiple selection

2) when you have selected a feature, then you enter query panel and exit, without searching, you 
 2.1) lose the highlight style on the selected features see gif below
 2.2) only get one row instead of the previous selected ones  (3 in the gif)

multiple selection

AGMETEOR:
@MV88 requested a new review on this. Fixed above issues in your latest comment

MV88 (changes requested):
Other points i have found:

1) when you have selected something with the clicks and you open the query panel they should be hidden (temporarily)

then 

1.1)  when clicking on the left arrow, going back they should be shown again

1.2) when instead i apply a search they should be cleared and the filter coming from query panel be applied.

Another minor thing, the message passes from **1 items of 1** to 2 items, 3 items. 
can't you have **1 item** as initial text?

web/client/translations/data.it-IT.json
```diff
@@ -1541,7 +1541,7 @@
             "deselectall": "Deseleziona tutti",
             "backtosearch": "Ritorna alla ricerca",
             "resultInfo": "{total, plural, =0 {Nessun elemento} =1 {{total} di {total}} other {{start}-{end} di {total}}}",
-            "resultInfoVirtual": "{total, plural, =0 {Nessun elemento} =1 {{total} di {total}} other {{total} Elementi}}",
+            "resultInfoVirtual": "{total, plural, =0 {Nessun elemento} =1 {{total} Elementi} other {{total} Elementi}}",
```
MV88, line 1544:
```suggestion
            "resultInfoVirtual": "{total, plural, =0 {Nessun elemento} =1 {{total} Elemento} other {{total} Elementi}}",
```

MV88 (changes requested):
can you check it works properly_? i see that it loses the number of selected features when closing the query panel

MV88 (approved):
LGTM

Issue #6101, opened 2020-10-23
Ability to disable Identify for a layer


tdipisa:
## Description

It is required to allow users to disable the Identify functionality per layer.
A new option must be available in Layer Settings (Feature Info Tab) so that the user can disable the identify request at layer level

![image.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/c66d6e32-6481-4f9d-9c90-59cceec8cd00)

A new item in the list above (**DISABLE IDENTIFY**) should be added on top (in capital case like the other ones) for this purpose.

Item title: DISABLE IDENTIFY
Item description: Disable the feature info for this layer

For that item it is not needed to expand a panel to show a preview when the user clicks on it since the option is needed to disable the Identify at all for the layer.

Texts above need to be translated in all supported languages in MS and it must be possilble to persist that setting when the user save the map. 
As a result of this setting, Identify requests are not performed for the layer if **DISABLE IDENTIFY** has been selected. This means that the layer will not be present at all in the Identify side panel dropdown list.

In MapStore there are to possible use cases for Identify operations:

1. **One or more layers are selected in TOC**: in that case MapStore performs Identify requests only for the selected layers (that are available in the dropdown of the Identify side panel). In the context of this issue, if the identify mode of one of the selected layers is set to **DISABLE IDENTIFY**,  the layer will not appear in the list (same behavior of a hidden layer or not queryable layer)

2. **No specific layer is selected in TOC**: in that case MapStore performs Identify requests for all TOC layers in the clicked position. The behavior for layers with **DISABLE IDENTIFY** option enabled is the same reported in the above point: the layer will not appear in the list (same behavior of a hidden layer or not queryable layer)

The **DISABLE IDENTIFY** option **must not be available** in general settings options:

![image](https://user-images.githubusercontent.com/1280027/98359144-08010f80-2028-11eb-94ac-2c02c156c839.png)

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

ElenaGallo:
[...]

PR #6128, opened 2020-10-29
Ability to disable Identify for a layer #6101

myrdstom:
## Description

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#6101

**What is the new behavior?**

## Breaking change
 - [x] No

## Other useful information


offtherailz (changes requested):
image
I don't think this is the desired behavior. 
I think that the request should not be performed at all, and nothing should be shown, filtering out the layer from the feature info list, as well as the one that not support it (e.g. TMS ...).

Correct me if I'm wrong @tdipisa 

Moreover, here I have a doubt about the "Feature info only on selected layer" what the expected behavior if the selected layer is disabled ? 
I think the usual no info layer modal should be shown. Can you confirm @tdipisa

myrdstom:
Yes, it isn't, I will be adjusting as specified

tdipisa:
@offtherailz yes, I've already discussed your comment with @myrdstom. He is working to complete the implementation.

And yes, in the following case:

>Moreover, here I have a doubt about the "Feature info only on selected layer" what the expected behavior if the selected layer is disabled ?

It must be as you have reported:

>I think the usual no info layer modal should be shown. Can you confirm @tdipisa

web/client/epics/identify.js
```diff
@@ -162,6 +162,13 @@ export const getFeatureInfoOnFeatureInfoClick = (action$, { getState = () => { }
                         const itemId = itemIdSelector(getState());
                         const reqId = uuid.v1();
                         const param = { ...appParams, ...requestParams };
+                        const infoFormat = getState()?.layers?.flat[6]?.featureInfo?.format;
```
offtherailz, line 165: This is uncorrect, here you are getting for every layer the state of the 6th layer. Moreover, you should not create a request for this layer, anyway. You can filter them on queraible layer selector (line 125, 126), and there is already a flag `queryable` for this purpose. So I think you don't have to touch this file, but simply, from the UI, set the queryable flag of the layer, when the format is selected.

web/client/utils/FeatureInfoUtils.js
```diff
@@ -1,5 +1,6 @@
 
 const INFO_FORMATS = {
+    "HIDDEN": "text/html",
```
offtherailz, line 3: I don't think this is needed

web/client/utils/MapInfoUtils.js
```diff
@@ -17,7 +17,7 @@ const MapInfoUtils = {
      * specifies which info formats are currently supported
      */
     //           default format ↴
-    AVAILABLE_FORMAT: ['TEXT', 'PROPERTIES', 'HTML', 'TEMPLATE'],
+    AVAILABLE_FORMAT: ['HIDDEN', 'TEXT', 'PROPERTIES', 'HTML', 'TEMPLATE'],
```
offtherailz, line 20: I don't think this is not needed

offtherailz (changes requested):
The current solution is not correct (you hardcoded to get the only the 7th layer). I think you should move your check in a different place.
Please see comments inline.
Moreover:
- [ ] unit tests should be provided
- [ ] The icon should be different
image
- [ ] I think the info format is not the proper value to store in the layer. Internally (see MapInfoUtils line 213) there is already a flag, `queryable` that provides this functionality. You should instead set that flag when the user select "HIDDEN" to disable info.

web/client/epics/identify.js
myrdstom: I must have pushed this by mistake. I have readjusted to: queryableLayers && queryableLayers[0] && queryableLayers[0].featureInfo && queryableLayers[0].featureInfo.format;

web/client/utils/MapInfoUtils.js
myrdstom: Screenshot 2020-11-04 at 13 14 31

DefaultSettings calls the function `MapInfoUtils.getAvailableInfoFormat()` which determines what settings are available in TOC, is there another way to go about it?

web/client/epics/identify.js
myrdstom: With regards the `DISABLE IDENTITY` icon being different, I had a brief chat with @tdipisa and he mentioned he would discuss it... I can't recall whom exactly but I think you or Stefano. I can raise it with him again

web/client/utils/MapInfoUtils.js
offtherailz: `HIDDEN` is not a real featureInfo  format, so I think you should handle the accordion to set/unset properly the `queryable` flag, and keeping the feature info format unchanged, as well as the utilities dedicated to it.
You can append it here, as a fake info format, for the accordion. Then when click, check it and set/unset queryable accordingly, or set the current info format.

web/client/epics/identify.js
tdipisa: @offtherailz I will involve Stefano to provide a better icon for this as soon as the PR is merged.

web/client/utils/MapInfoUtils.js
myrdstom: So what I have done is append `HIDDEN` to the result of `getAvailableInfoFormat()`. I also Append it to the MapInfoUtils.INFO_FORMATS as that is used throughout the file

MV88:
@myrdstom  a small suggestion: avoid to use so much force push on a created pr, it will be hard to identify the changes between commits.

do that locally before creating a pr, but once ti is created is better to avoid that

myrdstom:
> @myrdstom a small suggestion: please, avoid to use so much force push on a created pr, it will be hard to identify the changes between commits.
> 
> maybe you can do that locally before creating a pr, but once it is created is better to avoid that operation
> commits description should describe what the commit does in a few words

Sorry for the new commit mess... Early morning. This won't happen moving forward

offtherailz (changes requested):
I think you committed the first solution. (both behavior and files are very similar)
Are you still working on it?

myrdstom:
> I think you committed the first solution. (both behavior and files are very similar)
> Are you still working on it?

No, It's ready for review

web/client/epics/identify.js
```diff
@@ -162,6 +162,13 @@ export const getFeatureInfoOnFeatureInfoClick = (action$, { getState = () => { }
                         const itemId = itemIdSelector(getState());
                         const reqId = uuid.v1();
                         const param = { ...appParams, ...requestParams };
+                        const infoFormat = queryableLayers[0]?.featureInfo?.format;
+                        if (infoFormat === 'HIDDEN') {
+                            return [
+                                newMapInfoRequest(reqId, param),
+                                loadFeatureInfo(reqId, [{}], requestParams, { ...lMetaData, features: 0, featuresCrs: `EPSG:4326` }, layer)
+                            ];
+                        }
```
offtherailz, line 171: filter the layers at lines 125-126

web/client/components/data/identify/viewers/HiddenViewer.jsx
```diff
@@ -0,0 +1,19 @@
...
+            <div id="container">
+                <Alert bsStyle={"danger"}>
+                    <h4><HTML msgId="noFeatureInfo"/></h4>
+                </Alert>
+            </div>
+        );
+    }
+}
+
+module.exports = HiddenViewer;
```
offtherailz, line 19: hidden viewer and it's test are not needed anymore

Issue #6126, opened 2020-10-29
Annotations & circles can raise an error when a circle is deleted


MV88:
## Description

Interactions on editing in annotation can go inconsistent state due to an error, then selecting the circle in the ui it recovers.

Expand details below to se the error
<details>
<img src="https://user-images.githubusercontent.com/11991428/97598512-5b130b00-1a07-11eb-928f-e27878ccff0e.png"/>
</details>

## How to reproduce

- open new map
- start adding three circles
- delete second one (two circles are deleted instead of only one)
- click on map 

*Expected Result*

no error occurs, and only that sepcific circles is deleted
*Current Result*

see the above error

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

for more details see this gif
<img src="https://user-images.githubusercontent.com/11991428/97599171-fad09900-1a07-11eb-8606-3d15d06a371f.gif"/>

it is also strange that the highlight  remains on the second last feature drawn

PR #6129, opened 2020-10-30
#6126 - Annotations highlight and circle draw fix

dsuren1:
## Description
This PR adds fix to previous annotations remaining highlighted and issue with adding Circle annotations.

 - [x] Bugfix

## Issue

#6126 

**What is the new behavior?**
- The previous annotations will not stay highlighted unless hovered upon
- Circle annotations can be added without any draw issues

## Breaking change
 - [x] No


MV88 (changes requested):
i noticed that when you delete something , you lose the edit interaction and you can recover that editing the field lat lon or by clicking again on the selected one.

honestly i think that the highlight style should be different from the edit style, but this is my opinion

dsuren1:
> i noticed that when you delete something , you lose the edit interaction and you can recover that editing the field lat lon or by clicking again on the selected one.

I think we can just disable the delete buttons on the geometries until the user is out of editing, it makes the scenario more clear and less complex? This will also prevent any cosmetic glitches during these combined actions. Your thoughts @MV88 @tdipisa 

> honestly i think that the highlight style should be different from the edit style, but this is my opinion
 
Yes, it makes sense @MV88 . @tdipisa your thoughts on this. If so, any color preference ?

tdipisa:
@dsuren1 @MV88 I think what we have to do is to manage selections properly

MV88:
> @dsuren1 @MV88 I think what we have to do here is to manage selections properly. I think that introducing now additional behavior could be counterproductive if it isn't strictly necessary.

@dsuren1 
The thing that the highlight is disabling the current edit interaction. this is probably the cause of the thing i noticed. 

please solve this behaviour and test that the edit interactions remain active while you are in that scenario i.e. while having the coordinate editor visible and one feature is selected

MV88 (approved):

Issue #6133, opened 2020-11-02
Migration to PlotlyJS


tdipisa:
## Description

Apart for some graphical decision to take, accordingly to the possibilities provided from the library and the usage inside the widgets, here there is a list of the tasks to do:

- [x] Make the data transformation in the correct place
- [ ] Adapt to small widgets sizes
   - [x] Provide the correct margins
   - [x] remove/hide ModeTools (tooblar) for small widgets, because it stays on top on the chart making the interaction with it hard
legend
- [ ] Colors
- [ ] Adjust options: 
   - [x] Oblique labels -> tickangle option
   - [x] showticklabels, showgrid, showline options of xaxis and yaxis in layout options to support hide/show grid, hide/show y-values and so on.
   - [ ] I’ve not found a way to customize them. For long strings it may be problematic. And adapt the relative UI.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: Charts lib update

## Other useful information

MapStore has some advanced features to maintain in the migration. They are indicated in the widget creation panel:

![image](https://user-images.githubusercontent.com/1280027/97859383-cf4bf800-1d00-11eb-8e6e-618be6c474c1.png)

- Colour: possibility to select a colour for the pie/line chart or a color scale in case of pie chart.
- Display legend: show/hide the legend for the chart. The legend is only 1 element for line or bar. Contains all the elements for pie. 
- Advanced (only for line/bar)
   - Hide Grid: hide/show the grid behind the charts
   - Hide Y Axis: do not show the Y axis at all
   - X Axis label rotation angle
   - Legend label: instead of using sum(ATTRIBUTE)

Issue #6134, opened 2020-11-02
Linear/Series Charts


tdipisa:
## Description

Using PlotyJS, this can be implemented adding to the UI a tool that allows to select this option to the chart. It is supported by PlotyJS.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

offtherailz:
Here how the Advanced Options change to support this:

![image](https://user-images.githubusercontent.com/1279510/98361432-ceca9e80-202b-11eb-82d1-65416e4207d8.png)

The entries are better organized, and some new options are added: 

New options are :
 - Axis type (see below) 
- X-Axis Hide  / Never skip labels: can be useful to adjust the look and feel of small charts with long labels. 

The Axis types are these:
![image](https://user-images.githubusercontent.com/1279510/98361499-ee61c700-202b-11eb-83c8-b23645da4512.png)
- **Auto** (automatic recognition, the default)
- **Category**: Each x value is a column
- **Linear**: the axis is linear, so with values for x 1 and 3, we have the 2 in the middle
- **Log**: log scale
- **Date**: the axis value is parsed as a date.

These options are valid for both X and Y and they open the possibilitis of a big set of charts, expecially when not grouping.

PR #6163, opened 2020-11-05
Plotly charts

offtherailz:
## Description
This is the new integration with PlotlyJS
Here a sample
image

Different presentation of charts
image

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#6133

**What is the new behavior?**

## Breaking change
 - [x] No


web/client/components/charts/__tests__/sample_data.js
```diff
@@ -0,0 +1,27 @@
+
+export const DATASET_1 = {
```
MV88, line 2: in general do we have this in the resources folder, or not?

web/client/components/charts/SimpleChart.jsx
```diff
@@ -1,71 +1,6 @@
...
-        {...{legend, tooltip}}
-    >
-        {tooltip !== false ? type === "pie" ? <Tooltip content={CustomTooltip}/> : <Tooltip {...tooltip}/> : null}
-        {legend !== false ? <Legend {...legend} wrapperStyle={{bottom: 0}} /> : null}
-    </Component>
-    );
-};
-
-module.exports = SimpleChart;
+export default SimpleChart;
```
MV88, line 6: not sure about the need of this component

web/client/components/widgets/builder/wizard/ChartWizard.jsx
```diff
@@ -118,13 +122,14 @@ module.exports = enhanceWizard(({ onChange = () => { }, onFinish = () => { }, se
...
             key="widget-options"
             data={data}
             onChange={onChange}
             layer={data.layer || layer}
-            sampleChart={renderPreview({ data, layer: data.layer || layer, dependencies, setValid: v => setValid(v && isChartOptionsValid(data.options)), shortenChartLabelThreshold })}
+            sampleChart={renderPreview({ data, layer: data.layer || layer, dependencies, setValid: v => setValid(v && isChartOptionsValid(data.options)) })}
         />
-    </Wizard>));
+    </Wizard>);
+export default enhanceWizard(ChartWizard);
```
MV88, line 135: this is missing the display name, maybe is better to export a function called ChartWizard which contains the enhancer...
so it is easier to debug

web/client/components/widgets/view/WidgetsView.jsx
```diff
@@ -17,7 +17,7 @@ const ResponsiveReactGridLayout =
         widthProvider
     )(Responsive);
 const withGroupColor = require('../enhancers/withGroupColor');
-const DefaultWidget = withGroupColor(require('../widget/DefaultWidget'));
+const DefaultWidget = withGroupColor(require('../widget/DefaultWidget').default);
```
MV88, line 20: no conversion of all require here ;) ?

web/client/components/widgets/widget/ChartWidget.jsx
```diff
@@ -22,7 +22,7 @@ const renderHeaderLeftTopItem = ({ showTable, toggleTableView = () => {}} = {})
 };
 
 
-module.exports = ({
+export default ({
```
MV88, line 25: i would avoid in general this, better to export a named function

web/client/components/widgets/widget/DefaultWidget.jsx
```diff
@@ -5,79 +5,20 @@
...
+    TableWidget,
+    TextWidget,
+    LegendWidget
+} from './enhancedWidgets';
 
 /**
  * Renders proper widget by widgetType, binding props and methods
  */
-module.exports = ({
+export default ({
```
MV88, line 21: same here

web/client/components/widgets/widget/enhancedWidgets.js
```diff
@@ -0,0 +1,121 @@
...
+import dependenciesToExtent from '../enhancers/dependenciesToExtent';
+import dependenciesToLayers from '../enhancers/dependenciesToLayers';
+import dependenciesToMapProp from '../enhancers/dependenciesToMapProp';
+
+
+import BaseChartWidget from './ChartWidget';
+import BaseTextWidget from './TextWidget';
+import BaseMapWidget from './MapWidget';
+import BaseTableWidget from './TableWidget';
+import BaseCounterWidget from "./CounterWidget";
```
MV88, line 30: align quotes, sometimes they have " some other '

maybe we can think of an eslint rule it it exists

web/client/plugins/DashboardEditor.jsx
```diff
@@ -145,10 +147,10 @@ const Plugin = connect(
         onUnmount: () => setEditorAvailable(false)
     }
 )(DashboardEditorComponent);
-module.exports = {
+export default {
```
MV88, line 150: can't you convert it also to create plugin?

MV88 (changes requested):
First thanks for migrating some of the components/plugins etc.

I noticed that new files are missing the copyrights comment at the start

I left you some comments here and there
In general the pr seems well done, and the chart are working well

I found some minor things here
image

web/client/components/charts/__tests__/sample_data.js
offtherailz: I didn't want to put it so far from the specific usage, or we loose the reason why some data is in global test-resources folder. They was originally inline, I put them out only to separate data sets, that can grow, from effective tests.
I should put them out also if are common things (like generic responses of well know services). Do you think I should move them anyway?

web/client/components/charts/SimpleChart.jsx
offtherailz: Yes, It became an alias so I can remove it, and use WidgetChart anywhere

web/client/components/widgets/view/WidgetsView.jsx
offtherailz: I didn't want to convert all the files at once :smile: or the PR can become easely not reviewable. but if you prefer I can convert also all the Widgets plugin (actually I converted only the Dashboard part)

offtherailz:
@MV88 thank you for the review.
Here my comments about the image you provided in the review.
- [ ] **colors for te tooltip** --> we can add it to a list of cosmethic things to keep in consideration in a separate PR?
- [ ] **no separation lines** --> we can add it to a list of cosmethic things to keep in consideration in a separate PR?
- [ ] **legend overlaps** --> on my current branch I'm working to improve general chart margins. I'll look if I found a general solution for this, or anyway this will be solved by my next PR #6134
- [ ] **null value,** I'll investigate
- [ ] histogram weird max.** : plotly automatically recognize numeric values as a linear axis. In #6134 I'll add the possibility to select if we want the x axis to be automatic, linear, category (as the default behaviour of recharts). So I'll change this in #6134 
- [x] **Toggle spike lines** if you put the mouse on the effective point/bar top, it shows dashed lines for that value, in oder to see the value compared to others.
- [x] You have to click on the home button or rescale to back to original visibility.

tdipisa:
@offtherailz 
>colors for te tooltip --> we can add it to a list of cosmethic things to keep in consideration in a separate PR?

I think we have do it now and not later, it is part of the basic migration.

>no separation lines --> we can add it to a list of cosmethic things to keep in consideration in a separate PR?

I can see them in the pie chart instead? What this means? Which separation lines?

Ok for the other points.

MV88:
> > no separation lines --> we can add it to a list of cosmethic things to keep in consideration in a separate PR?

> I can see them in the pie chart instead. What this means? Which separation lines?

the white line at the top between the two values, between the two pieces

tdipisa:
Ok @MV88 now I understood, I didn't notice it. @offtherailz we should at lease investigate also this: comparing it with other pie charts in the screenshot above, it seems to me a particular case.

web/client/components/charts/__tests__/sample_data.js
MV88: no it is ok, just wondering when to use test-resources and when not

web/client/components/charts/SimpleChart.jsx
MV88: yes i would remove this since it adds nothing beside a renaming which can be done in the import statement

web/client/components/widgets/view/WidgetsView.jsx
MV88: we can do it later on :)

offtherailz:
Yes @tdipisa is a 'bug' of plotly. I think we can review the default initial color for pie chart to be the automatic default of plotly, (and add it as an entry of the menu) that is better than our 'random' (that btw is a 'hue', not properly 'random') and it is also a better default for a pie chart.
Something like this: 
image

web/client/components/widgets/widget/ChartWidget.jsx
offtherailz: I used const but I think is the same

MV88 (approved):

MV88:
going to merge as the travis build finished

Issue #6137, opened 2020-11-02
Time Series


tdipisa:
## Description

Time series are only a mix of the following improvements:
- Recognizing the time x-axis condition
- Linear x-axis mode (instead of series/category)
- Formatting the X axis accordingly

Plotly should recognizes automatically the timestamp as X value so no work is needed.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes 
- [X] Other... Please describe: Charts improvements

## Other useful information
See also #6055.

offtherailz:
#6169 solves also this. So as well as it is merged, we can start testing also this

offtherailz:
reopen because of https://github.com/geosolutions-it/MapStore2/pull/6855

PR #6169, opened 2020-11-09
#6134: Add types and other options for X/Y Axis

offtherailz:
## Description
This PR adds the requirements for #6134 . Moreover, it adds more options to allow the user to adjust the chart layout and improve the appearance. 

In order to disable controls that does not make sense in the XAxis UI on certain configuration, I added also the `disabled` flag to the SwitchButton, and to the Slider, with the needed CSS.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Options for axis  are missing
**What is the new behavior?**
#6134 (see also comments)
#6137

## Breaking change
 - [x] No


MV88 (changes requested):
HI, Here is some thing I noticed

- [ ] when data are a lot and you select never skip labels, there soem performance issue, 
use the [demo](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10358) map "tests chart" for an example using 
image



- [ ] there are some index problems with the type
image

- [ ] **X Axis:** Linear Log and Date type does not show anything
image
for testing you can use the tests charts mentioned before.

For me the Date seems not clear enough and prone to error or misunderstandings. but is just my opinion

offtherailz:
> - [ ]  when data are a lot and you select never skip labels, there soem performance issue, use the demo map "tests chart" for an example using
I preferred to make this option more explicit because in the old implementation it was implicitly activated by oblique labels > 0. Effectively ploltly seems less performant that recharts for this particular task. Recharts takes around 1 second to do it, while plotly takes 5 seconds. I decided to add it because I remember it was a requirement (c040 if I remember well). But these new charts have also zoom and other features that may make this feature unuseful.
@tdipisa what do you think? I can remove this feature and avoid these performance problems at all
> - [ ] here are some index problems with the type 
Good catch, I fixed it in dashboard editor but not in the map. Thank you, going to fix
>  - [ ] X Axis: Linear Log and Date type does not show anything
**X Axis** : Auto is ok in most of the case. You may need to force it depending on your needings to category, linear, Log, Date etc... 
Examples: 
 - numbers that must be interpreted as categories 
 - You need numbers on a log scale or a linear scale
 - You need to force a number to be interpreted as date.
...


This depends mostly from the data on the back-end. With different data sources (e.g. WFS) this options become even more clear, allowing to do vertical charts, unloking some combinations, for instance vertical charts and so on.

offtherailz:
@MV88 I fixed the performance issue finding another property to limit the number of labels.

tdipisa:
@offtherailz

>@MV88 I fixed the performance issue finding another property to limit the number of labels.

As we agree, please put something like a tooltip message to warn the user about this

MV88 (approved):

Issue #6157, opened 2020-11-03
WPS Export Data UI components


tdipisa:
## Description

The current Export Data window can be reused for the time being, It will be launched by:

- the already available “export data” button from the FeatureGrid, with export data button replaced with glyphicon-download

![image](https://user-images.githubusercontent.com/571129/97861079-7cc00b00-1d03-11eb-90af-46533af7f528.png)

- a new button inside the layer toolbar (as already done for AustroControl)

![image](https://user-images.githubusercontent.com/571129/97865575-b8aa9e80-1d0a-11eb-89e1-b99000bed4e2.png)

### Important Note
_Currently the Export Data tool is using WFS: we need to include a preventive check to verify if the WPS process is available or not and enable the WPS UI part accordingly, otherwise the current one is provided to the user using WFS as it is now._

Two options will be visible to the user:

- “**crop dataset to current viewport**”: this will set the ROI and cropToROI execution parameters, that will clip the vector / crop the raster to the visible map viewport
- “**export filtered dataset” (ONLY FOR VECTOR LAYERS)** in case a global filter has been set and/or a filter has been set inside the FeatureGrid
“**advanced options“ (ONLY FOR RASTER LAYERS -for the moment)**:  collapsible area showing the following advanced parameters:
  - “Compression”: select box
    - Deflate
    - LZW
    - JPEG
    - ZLib
    - PackBits
    - CCITT RLE
  - “Tile height”: (numeric box)
  - “Tile width”: (numeric box)
  - "quality" (missing in the image below) which is a double in the range [0 : 1] where 0 is for worst quality/higher compression and 1 is for best quality/lower compression

![image](https://user-images.githubusercontent.com/571129/97863436-41bfd680-1d07-11eb-844f-21a48b31e829.png)

![image](https://user-images.githubusercontent.com/571129/97860374-4df56500-1d02-11eb-8fec-f08ff805645f.png)

![image](https://user-images.githubusercontent.com/571129/97863469-52704c80-1d07-11eb-8111-c12dcfee12e6.png)

### Execution status and Results management
Being an asynchronous execution we must keep track of the status and implement a panel to let the user inspect the export procedure status and download the final files.

Ideally the status should be bound to the user's session**, but this requires the support from a backend.
Session bound status will be developed at a future stage**. **For the moment the status will be only serialized to a cookie**, so that at least the user won't loose ongoing exports if he refreshes or closes / reopens the page.

[...]

Issue #6156, opened 2020-11-03
WPS Download client


tdipisa:
## Description

A critical enhancement to the current use cases in MapStore is the implementation of an asynchronous interface. This means that the following key points must be designed and implemented:

- download estimator request to validate the download request (gs:DownloadEstimator) according to the configured limits for the WPS process
- request async execution (gs:Download)
- track execution status
- manage execution results, which will be requested as “reference” results to be obtained as link to downloadable files

The execution request will use the following parameters:

- layerName 
- outputFormat 
- targetCRS 
- filter (only for vector data)
- ROI
- RoiCRS
- cropToROI 
- writing parameters 

First three parameters are those already adopted by the current “Export Data” window. 
The filter parameter will restrict the export to a subset of data. The filter will be defined with the already existing MapStore filter panel and/or (in case of vector layers) through the FeatureGrid filters.

The ROI+RoiCRS+cropToROI parameters will be used to support cropping the dataset to the map viewport.

The [writing parameters](https://docs.geoserver.org/stable/en/user/community/wps-download/rawDownload.html#writing-params) are especially useful for raster data to, for example, request a specific compression format or the tile sizes of the downloaded file.

### Important Note:
_The implementation of the WPS async interface client will be as generic as possible, so that it will be possible to adopt it for different processes and contexts with minimal effort. A specific business logic layer will be implemented on top of it for the specific purposes of the “data export” process._

## Other useful information
- Geoserver WPS module: https://docs.geoserver.org/stable/en/user/community/wps-download/index.html
- OGC WPS Standard: https://www.ogc.org/standards/wps

offtherailz:
This can be unblocked as well as the PR #6156 is merged

Issue #6144, opened 2020-11-02
WPS based Data Export


giohappy:
## Description

We want to extend the “Export Data” tool, that currently is only available for vector layers from the FeatureGrid component, with the following enhancements:

- overcome the limitation of WFS/WCS requests (heavy load on Geoserver, sync execution, etc.) using the [Geoserver WPS Download](https://docs.geoserver.org/stable/en/user/community/wps-download/index.html) community module
- export both vector and raster layers
- expose the export tool as a layer tool
- pave the path to more advanced options (ROI, GeoTiff writer options, etc.)

We are going to rely on the [Raw data download process for GeoServer](https://docs.geoserver.org/stable/en/user/community/wps-download/rawDownload.html).

### WPS Download client
MapStore already adopts WPS for some components (example data “aggregate” functionalities). We might therefore investigate if something from the existing tools can be reused or factored out to obtain a common base interface for any functionality that requires to interact with WPS processes.

The option to reuse something from previous implementations (MapStore1 with OL2) has been discarded since the technologies have changed too much meanwhile, and OL3/4 don’t support WPS natively anymore.

A critical enhancement to the current use cases in MapStore is the implementation of an asynchronous interface. This means that the following key points must be designed and implemented:

- download estimator request to validate the download request (gs:DownloadEstimator) according to the configured limits for the WPS process
- request async execution (gs:Download)
- track execution status
- manage execution results, which will be requested as “reference” results to be obtained as link to downloadable files

The execution request will use the following parameters (with more to come in the future):

- layerName 
- outputFormat 
- targetCRS 
- filter (only for vector data)
- ROI
- RoiCRS
- cropToROI 
- writing parameters 

First three parameters are those already adopted by the current “Export Data” window. 
The filter parameter will restrict the export to a subset of data. The filter will be defined with the already existing MapStore filter panel and/or (in case of vector layers) through the FeatureGrid filters.

The ROI+RoiCRS+cropToROI parameters will be used to support cropping the dataset to the map viewport.

The [writing parameters](https://docs.geoserver.org/stable/en/user/community/wps-download/rawDownload.html#writing-params) are especially useful for raster data to, for example, request a specific compression format or the tile sizes of the downloaded file.

### Important Note:
_The implementation of the WPS async interface client will be as generic as possible, so that it will be possible to adopt it for different processes and contexts with minimal effort. A specific business logic layer will be implemented on top of it for the specific purposes of the “data export” process._

### Export Data UI components
[...]

PR #6204, opened 2020-11-17
WPS based Data Export

vlt1:
## Description
WPS based data export implementation. Contains WPS download client.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#6144 
#6156 
#6157 

## Breaking change
 - [x] No


vlt1:
I will now proceed with adding docs and tests first for #6156 then for #6157

offtherailz:
@vlt1 can you solve the conflicts ? (we migrated syntax from require to import) they should be very easy to solve.

web/client/translations/data.it-IT.json
```diff
@@ -1610,16 +1610,43 @@
...
             "format": "Formato file",
             "srs": "Sistema di riferimento",
             "export": "Esporta",
             "downloadonlycurrentpage": "Scarica solo la pagina corrente",
+            "cropDataSet": "Ritaglia il set di dati nella finestra corrente",
+            "downloadFilteredDataSet": "Scarica set di dati filtrati",
+            "advancedOptions": "Opzioni avanzate",
+            "compression": "Compressione",
+            "tileWidth": "Larghezza piastrella",
+            "tileHeight": "Altezza piastrella",
```
offtherailz, line 1624:
```suggestion
            "tileWidth": "Larghezza tile",
            "tileHeight": "Altezza tile",
```

web/client/product/plugins.js
```diff
@@ -118,8 +120,7 @@ export default {
         UserManagerPlugin: require('../plugins/manager/UserManager').default,
         UserExtensionsPlugin: require('../plugins/UserExtensions').default,
         UserSessionPlugin: require('../plugins/UserSession').default,
-        VersionPlugin: require('../plugins/Version').default,
-        WFSDownloadPlugin: require('../plugins/WFSDownload').default,
+        VersionPlugin: require('../plugins/Version'),
```
offtherailz, line 123:
```suggestion
        VersionPlugin: require('../plugins/Version').default,
```

offtherailz (changes requested):
- [ ] see my inline comments
- [ ] when WPS is missing, so using ordinary WFS service the download doesn't work anymore (the `typeName` is sent as `undefined` to the server)
```xml
<wfs:GetFeature service="WFS" version="2.0" xmlns:wfs="http://www.opengis.net/wfs/2.0" xmlns:fes="http://www.opengis.net/fes/2.0" xmlns:gml="http://www.opengis.net/gml/3.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd"><wfs:Query typeNames="undefined" ></wfs:Query></wfs:GetFeature>
```
- [ ] Actually if I have WPS I have this list of formats, that is a big limitation over the normal WFS download list. 
image 

the normal list is this. 

image.  

Moreover, proposing these 4 formats for every layer maybe confusing, because only a sub-set of them is available, depending on the feature type. @giohappy is this the desired behaviour ? 


Minor things:
- [ ] Instead of display "no results found" if the download list is empty, we should show something more clear, like. "Export list is empty"
- [ ] Maybe the two plugins should be merged in one single plugin (the ExportResultsList can not live without download and vice-versa) to simplify the management. 
- [ ] Maybe the download status button should not be always present, but only when I have some downloads in the list.
image

vlt1:
> Actually if I have WPS I have this list of formats, that is a big limitation over the normal WFS download list.

The formats in WPS list are the ones that actually work with WPS.

> Moreover, proposing these 4 formats for every layer maybe confusing, because only a sub-set of them is available, depending on the feature type.

How do I determine if a particular format is available for a specific layer?

offtherailz:
Thanks @vlt1 for clarification. @tdipisa is this acceptable or should we somehow allow to switch to "WFS only mode" or disable WPS option with a flag? 
One more question for @vlt1 : where the list is stored, to recover it in a second time after browser reload?

tdipisa:
@vlt1 few more comments below.

**1.** 
>Actually if I have WPS I have this list of formats, that is a big limitation over the normal WFS download list.

At the moment there is no direct way in GeoServer to retrieve the list of mime types supported by the WPS Download process (**this could be an interesting improvement to do in GeoServer in my opinion**). The supported mime types list can be also related to eventual additional extensions in GeoServer (like DXF). For this reason, now we need a default configuration list of mime types defined for the download plugin in MapStore. This configuration must be different for vector and raster layers. Below the list we can use for the default configuration in MapStore:

Vector layers (**a proper label need to be associated to all of them in the MS UI**):

```
- text/xml; subtype=wfs-collection/1.0
- text/xml; subtype=wfs-collection/1.1
- application/wfs-collection-1.0
- application/wfs-collection-1.1
- application/zip
- text/csv
- application/json
```

Raster layers (**a proper label need to be associated to all of them in the MS UI**):

```
- image/tiff 
- application/arcgrid
- image/png
- image/jpeg
```

**2.**
When activating the MS tool for the Download process, we have to check if the layer is a raster layer or a vector layer and then provide only the corresponding list of formats (one of the above lists depending on the layer).

At the moment there is no clean way to understand if a layer is a vector or a raster. Available options are the following ones:

**Option 1**
a. While activating the MS tool, perform a WCS GetCapabilities request to see if the layer is reported there: if yes, the layer is a raster layer
b. If the layer is not reported by GeoServer in the WCS GetCapabilities output, perform a second WFS GetCapabilities request to see if the layer is reported there: if yes, the layer is a vector layer

_**The limitation here are:**_
- Possible performance issues 
- The WCS and WFS can be disabled in GeoServer (in this case the MS tool should inform the user that the Download is not possible) 

**Option 2**
a. While adding a layer from the MS Catalog tool (WMS source), check if the keyword WPS is present and put this information in the layer state so that MS is able to reconize if the layer is a raster or not and so provide the corresponding mime types list accordingly.

_**The limitation here are:**_
- This is a weak check because usually the keyword WPS is automatically set for raster layers by GeoServer, but the keyword can be removed/changed from the GeoServer layer configuration panel

Providing a better way to check if a layer is vector or raster could be a really interesting improvement for GeoServer.
**For the moment, in my opinion, we should opt for Option 2** (as I already told you @vlt1).

**3.**
The download tool must not be available in some cases (e.g. for local vector layers like annotations). Also for this we have to do a preventive check to enable the tool only for WFS/WMS layers (this is an information we have in MS).

web/client/plugins/LayerDownload.jsx
```diff
@@ -0,0 +1,129 @@
...
+import { createPlugin } from '../utils/PluginsUtils';
+
+/**
+ * Provides advanced export functionalities using WFS and WPS.
+ * @memberof plugins
+ * @name LayerDownload
+ * @class
+ * @prop {object[]} formats An array of name-label objects for the allowed formats available.
+ * @prop {object[]} srsList An array of name-label objects for the allowed srs available. Use name:'native' to omit srsName param in wfs filter
+ * @prop {string} defaultSrs Deafult selected srs
```
offtherailz, line 57:
```suggestion
 * @prop {string} defaultSrs Default selected srs
```

offtherailz (approved):
I think is very good. :+1:

Issue #6226, opened 2020-11-20
Support combination of advanced filters and Quick filter for feature grid


offtherailz:
## Description
An current limitation of the filtering system is that the `spatialField` is an array of spatial filter that can have only one operator. 
With the introduction of multiple spatial quick filters (from feature grid) we now need to manage the combination of AdvancedFilterSpatialFilter AND (quickSpatialFilter1 OR quickSpatialFilter2). 

For the moment the quick filter wins over the advanced spatial filter, if present.

## How to reproduce
-  Open the feature grid
- Open advanced filter and select a spatial filter, click on search
- select the quick spatial filter and draw on map one (or more, with ctrl pressed) filters

*Expected Result*
The resulting filter is the cobination of `Advanced * (QF1 + QF2)

*Current Result*
The advanced is skipped (Because AND + OR is not supported yet)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

offtherailz:
## RFC

@mbarto , @allyoucanmap , @tdipisa , @MV88 

Here some notes about the current filtering functionalities, to understand the complexity of the current system and the things to change in order to have a system that can adapt to future requirements. 

### Filter structure

The internal fliter object structure is made on the model of the query panel and not well separated from the WFS: 
```javascript
{
    featureTypeName: 'gs:us_states', // ← Used for select a different layer, see #6430
    filterType: 'OGC',
    ogcVersion: '1.1.0',
    pagination: {
      startIndex: 0,
      maxFeatures: 100000 // ← this is used to deactivate virtual scrolling when multiple selection is active
    },
    spatialField: [ // this is the original spatial field of the query panel (1 element) extended to array because now in feature grid we need multiple geometric filters
      {
        attribute: 'the_geom', 
geometry: {center: [ // includes circle geometry info, to repopulate the query panel in a second time
            -12997963.78583765,
            4693845.032936105
          ],
          coordinates:  [...]
          extent: [
            -13017531.665078655,
            4674277.1536951,
            -12978395.906596646,
            4713412.912177109
          ],
          projection: 'EPSG:900913',
          radius: 19567.879241004586,
          type: 'Polygon'
        },
[...]

Issue #6224, opened 2020-11-20
Feature grid bugs 


MV88:
## Description

There are some bugs not resolved from [this ](https://github.com/geosolutions-it/MapStore2/pull/6220)PR
three bugs:

- [ ] feature selection is not active both via click or box if you open and close advanced query form without applying a filter

- [ ] if then you disable the feature selection it keeps the selection, no clear is applied

- [ ] in edit mode the box selection selects all the features

## How to reproduce

![bugs feature grid feature selection](https://user-images.githubusercontent.com/11991428/99790923-1e799180-2b25-11eb-8d64-0d62c299b2fd.gif)

*Expected Result*

- The box selections should be active when closing the feature grid,
- disabling the feature selection should reset the list of selected features
- the box selection in edit should select only the features inside box

*Current Result*

see gif

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #6227, opened 2020-11-20
#6224: fixed issues with advanced filter

offtherailz:
## Description
In order to avoid the problems notified in #6224 i did the following changes:
- Because the advanced filter can not go in AND with the sequence of OR of the geometric quick filter, I make temporary the quick filter win. See #6226. The issue of All selection was caused by the empty filter advanced in OR with the spatial quick filter (where the empty wins so every feature was selected).
- When the user open the advanced filter panel, the current quick filter on the feaure grid is removed. This avoid UI inconsistencies. This because the current logic reset the interaction with the spatial filters on close, and the panel is closed when advanced filter is open. Refactoring the logic is complicated and can not be done at this stage. Instead resetting it when the advanced filter is open prevents problems).
- one more fix - when the first quick spatial filter is applied I remove all selections from the checkboxes (that can not be reused in the filtered context), so the selected feature are all in map. 

**Previous** (press control when click on map)
Peek 20-11-2020 17-46
**new**
Peek 20-11-2020 17-48

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#<issue>

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/__tests__/featuregrid-test.js
+++ b/web/client/epics/__tests__/featuregrid-test.js
@@ -1050,17 +1050,12 @@ describe('featuregrid Epics', () => {
     });
 
     it('test askChangesConfirmOnFeatureGridClose', (done) => {
-        testEpic(askChangesConfirmOnFeatureGridClose, 2, closeFeatureGridConfirm(), actions => {
-            expect(actions.length).toBe(2);
+        testEpic(askChangesConfirmOnFeatureGridClose, 1, closeFeatureGridConfirm(), actions => {
+            expect(actions.length).toBe(1);
             actions.map((action) => {
                 switch (action.type) {
                 case CLOSE_FEATURE_GRID:
                     expect(action.type).toBe(CLOSE_FEATURE_GRID);
-                    break;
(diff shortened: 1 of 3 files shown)


MV88 (changes requested):
i think i have found a flow that breaks some of the logic.
box selection and quick filter and advanced


Also i do not understand why we should remove the advanced filter when we create the box selection, the moment i re-enter the advanced filter the spatial one is lost.. and re applying is not working anymore. 

in general advanced filters should win over the rest

offtherailz:
> Because the advanced filter can not go in AND with the sequence of OR of the geometric quick filter, I make temporary the quick filter win. See #6226. The issue of All selection was caused by the empty filter advanced in OR with the spatial quick filter (where the empty wins so every feature was selected).

I had to do this workaround as well as #6226 is fixed. I'll check about the problem that the advanced filter is not applied the second time, it looks strange, probably a little bug, good catch.

MV88 (approved):

Issue #6196, opened 2020-11-17
There is an error with click on map in Cesium


MV88:
## Description

Here is the error i get when clicking on a 3d map.

![image](https://user-images.githubusercontent.com/11991428/99368091-0dc5e300-28bb-11eb-97f4-b1db2e58e2f5.png)

## How to reproduce

- open demo map 
- go to 3d and open dev tools
- click on map and move the mouse

*Expected Result*

To see the identify info panel with correct results

*Current Result*

the errorr in console

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

A crash happen in the release

tdipisa:
@MV88 I also see a lot of console logs while loading the 3D map

![image](https://user-images.githubusercontent.com/1280027/99376385-2804be80-28c5-11eb-8067-a7ac3e4aac1e.png)

dsuren1:
**Question**
1. Looking at the error log (screenshot), the error was generated from leaflet (specifically from hook registered for _GET_PIXEL_FROM_COORDINATES_HOOK_)
2. Was the map click supported earlier in the cesium map implementation we have? Because I don't see any support for this. Kindly correct me if I'm wrong

tdipisa:
@dsuren1 

>1. Looking at the error log (screenshot), the error was generated from leaflet (specifically from hook registered for GET_PIXEL_FROM_COORDINATES_HOOK)

Which is the question?

>2. Was the map click supported earlier from the cesium map implementation we have? Because I don't see any support for this. Kindly correct me if I'm wrong

It should be, @offtherailz @MV88 ?

MV88:
> * Looking at the error log (screenshot), the error was generated from leaflet (specifically from hook registered for _GET_PIXEL_FROM_COORDINATES_HOOK_)

i don't think leaflet is involved here. i mean, i went to 3D from openlayers.

> * Was the map click supported earlier from the cesium map implementation we have? Because I don't see any support for this. Kindly correct me if I'm wrong

i do not remember if the feature info was supported before. honestly i think it was not. anyway is there a way to prevent that error?

offtherailz:
[...]

PR #6230, opened 2020-11-23
#6196: Cesium map on click error fix

dsuren1:
## Description
This PR adds a fix for error when on click cesium map

 - [x] Bugfix

## Issue

#6196 

**What is the new behavior?**
- Unregister the hooks causing issue when in cesium map
- GFI panel open with data of the clicked point on map

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/map/cesium/Map.jsx
+++ b/web/client/components/map/cesium/Map.jsx
@@ -13,7 +13,12 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import ConfigUtils from '../../../utils/ConfigUtils';
 import ClickUtils from '../../../utils/cesium/ClickUtils';
-import { ZOOM_TO_EXTENT_HOOK, registerHook } from '../../../utils/MapUtils';
+import {
+    ZOOM_TO_EXTENT_HOOK,
+    registerHook,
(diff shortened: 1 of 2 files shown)


web/client/components/map/cesium/Map.jsx
```diff
@@ -335,6 +340,12 @@ class CesiumMap extends React.Component {
         this.pauserStream$ = pauserStream$;
     };
     registerHooks = () => {
+        // Unregister hooks as coming from a leaflet or openlayer map retains hooks
+        // causing issue in feature info click
+        this.props.hookRegister.registerHook(GET_PIXEL_FROM_COORDINATES_HOOK);
```
MV88, line 345: in the comment you say unregister, but then you use **registerHook**

can you clarify?

MV88 (changes requested):

web/client/components/map/cesium/Map.jsx
dsuren1: Yes. Here I'm registering the hook GET_PIXEL_FROM_COORDINATES_HOOK with an undefined value which should return undefined when performing a getHook('GET_PIXEL_FROM_COORDINATES_HOOK'). So in a sense we are unregistering the hook.

MV88 (approved):

Issue #3916, opened 2019-07-05
CQL_FILTER could exceed URL length limit 


kappu72:
### Description
The CQL_FILTER builded by composing all different filters (layerFilter, FG advanced filter and columns filter and cql_filter from layer configuration) could easily exceed  [server URL max length limit](https://stackoverflow.com/questions/1289585/what-is-apaches-maximum-url-length)
### In case of Bug 

*Steps to reproduce*
1) Open filterLayer and set up a spatial filter type circle and save it.
2) Open FG and then advanced filter panel.
3) Create a new spatial filter type circle and apply it and activate map-sync
4) Check in developer console the network requests

![CQL_ISSUE](https://user-images.githubusercontent.com/9317650/60752381-2cc51d00-9fc5-11e9-91d6-5c1f972bb825.gif)

*Expected Result*
GetMap requests don't have to fail.

*Current Result*
GetMap requests fail

PR #6260, opened 2020-12-01
3916 cql exceed

MV88:
## Description

This pr changes the way WMS requests are sent (POST) when the url of the image is longer than 2000 chars
It is done only for when using ImageWMS in openalyers

this was problematic in some custom project like c039

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#3916

**What is the new behavior?**

wms requests longer than 2k chars are sent using xhr POST

## Breaking change
 - [x] No i don't think is a breaking change


web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -154,7 +154,50 @@ const createLayer = (options, map) => {
                 crossOrigin: options.crossOrigin,
                 attributions: toOLAttributions(options.credits),
                 params: queryParameters,
-                ratio: options.ratio || 1
+                ratio: options.ratio || 1,
+                imageLoadFunction: function(image, src) {
```
mbarto, line 158: What about TileWMS? We should handle also that one I think

web/client/components/map/openlayers/__tests__/Layer-test.jsx
```diff
@@ -285,6 +285,30 @@ describe('Openlayers layer', () => {
...
+        };
+        // create layers
+        let layer = ReactDOM.render(
+            <OpenlayersLayer type="wms"
+                options={options} map={map}/>, document.getElementById("container"));
+
+
+        expect(layer).toExist();
+        // count layers
+        expect(map.getLayers().getLength()).toBe(1);
```
mbarto, line 310: I don't see any expect that checks what we are changing with this PR

mbarto: I think the only meaningful test is to check we are sending a request with POST method

mbarto (changes requested):

web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -154,7 +154,50 @@ const createLayer = (options, map) => {
...
+                            if (i === 0) {
+                                url = dataEntries[i];
+                            } else {
+                                params = params + "&" + dataEntries[i];
+                            }
+                        }
+                        xhr.open('POST', url, true);
+
+                        xhr.responseType = 'arraybuffer';
+                        xhr.onload = function() {
```
offtherailz, line 178: Here we are using XHR object directly, this could fall into same origin policy violations, that normal images do not have.
@mbarto what do you think we should do here?
We could use axios, but we should check how it works with authentication (e.g. authkey, it is applied as additional parameter)

offtherailz (commented):
I only leave the inline comment, other than the review from @mbarto .

web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -154,7 +154,50 @@ const createLayer = (options, map) => {
                 crossOrigin: options.crossOrigin,
                 attributions: toOLAttributions(options.credits),
                 params: queryParameters,
-                ratio: options.ratio || 1
+                ratio: options.ratio || 1,
+                imageLoadFunction: function(image, src) {
+                    // fixes #3916, see https://gis.stackexchange.com/questions/175057/openlayers-3-wms-styling-using-sld-body-and-post-request
+                    var img = image.getImage();
+
+                    if (typeof window.btoa === 'function' && src.length >= 2000) {
```
MV88, line 162: use a default value very big like INFINITY  and pick it from options if present

web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -170,7 +206,38 @@ const createLayer = (options, map) => {
             resolutions: options.resolutions || MapUtils.getResolutions(),
             tileSize: options.tileSize ? options.tileSize : 256,
             origin: options.origin ? options.origin : [extent[0], extent[1]]
-        })
+        }),
+        tileLoadFunction: (tile, src) => {
+            if (typeof window.btoa === 'function' && src.length >= (options.maxLengthUrl || Infinity)) {
```
mbarto, line 211: What are the differences in imageLoadFunction for single tile and tileLoadFunction for tiles? Is there any chance we can have a simpler function for both?

mbarto (commented):
Look if you can make the loading function simpler and handle both single tile and tiled cases, if not, go ahead and merge, LGTM

web/client/components/map/openlayers/plugins/WMSLayer.js
MV88: i will see if we can unify them in a single load function
i'm sure we care simplify more, any idea apart from abstracting it?

mbarto (approved):

Issue #6266, opened 2020-12-01
Allow override of default user locale via localConfig.json


allyoucanmap:
## Description

This configuration could be used to declare the default user locale of the app without rely on the browser language or the first available supported locale
```js
{
    "defaultUserLocale": "it",

    "defaultState": {
        "locales": {
            "supportedLocales": {
                "en": {
                    "code": "en-EN",
                    "description": "English"
                },
                "it": {
                    "code": "it-IT",
                    "description": "Italiano"
                }
            }
        }
    }
}
```

or inside custom application that get the locale from other sources rather than browser language
```js
import { getConfigProp } from '@mapstore/framework/utils/ConfigUtils';
import cookies from 'js-cookie';

const defaltLocale = cookies.get('default_app_locale');
if (defaltLocale) {
  setConfigProp('defaultUserLocale', defaltLocale);
}
```

We should keep the current behaviour if the defaultUserLocale is not available.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

PR #6268, opened 2020-12-02
#6266: Allow override of default user locale via localConfig.json 

allyoucanmap:
## Description

This PR introduces the `locale` property inside of localConfig to allow override of the initial user locale.
Some examples:

- force default initial language
```
{
    "locale": "it-IT",

    "defaultState": {
        "locales": {
            "supportedLocales": {
                "en": {
                    "code": "en-EN",
                    "description": "English"
                },
                "it": {
                    "code": "it-IT",
                    "description": "Italiano"
                }
            }
        }
    }
}
```
- custom app
```js
import { getConfigProp } from '@mapstore/framework/utils/ConfigUtils';
import cookies from 'js-cookie';
const locale = cookies.get('locale');
if (locale) {
  setConfigProp('locale', locale);
}
```

This PR removes also `getUserLocale()` inside StandardApp because it's a duplication. See [loadLocale](https://github.com/geosolutions-it/MapStore2/blob/v2020.02.02/web/client/actions/locale.js#L68) thunk where the fallback locale is `getUserLocale()` if the language is undefined.

 - [x] Feature

## Issue

#6266

**What is the new behavior?**

It's possible to set `locale` property in localConfig.json to override the initial user locale of StandardApp. This enhancement should not change current behaviour of MapStore product so it should display the language of the browser as before.

## Breaking change
 - [x] No

### DIFF
--- a/docs/developer-guide/internationalization.md
+++ b/docs/developer-guide/internationalization.md
@@ -48,6 +48,43 @@ Setting locales in localConfig.json file is doable only for supported locales pr
 The default behavior is to use those already configured in "supportedLocales" object.
 You can customize the messages by editing the data.code-CODE.json files.
 
+You can also add the `defaultUserLocale` property to localConfig.json to override the initial user locale:
+```javascript
+{
+    "defaultUserLocale": "it",
+
+    "defaultState": {
+        "locales": {
+            "supportedLocales": {
+                "en": {
+                    "code": "en-EN",
+                    "description": "English"
+                },
+                "it": {
+                    "code": "it-IT",
+                    "description": "Italiano"
+                }
+            }
+        }
+    }
+}
+```
+MapStore will select the initial user locale based on the browser language and the supported locales if the `defaultUserLocale` is not defined.
+The property `defaultUserLocale` could be useful inside custom application where the locale is stored in other sources rather than using the browser language:
+
+```javascript
+// example
+import { getConfigProp } from '@mapstore/framework/utils/ConfigUtils';
+import cookies from 'js-cookie';
+
+// ...
+const defaultUserLocale = cookies.get('default_app_locale');
+if (defaultUserLocale) {
+  setConfigProp('defaultUserLocale', defaultUserLocale);
+}
+// ...
+```
+
 # How to add a new language
 
 Let's say we want to add the russian language.
--- a/docs/developer-guide/internationalization.md
+++ b/docs/developer-guide/internationalization.md
@@ -51,7 +51,7 @@ You can customize the messages by editing the data.code-CODE.json files.
 You can also add the `defaultUserLocale` property to localConfig.json to override the initial user locale:
 ```javascript
 {
-    "defaultUserLocale": "it",
+    "defaultUserLocale": "it-IT",
 
     "defaultState": {
         "locales": {
--- a/docs/developer-guide/internationalization.md
+++ b/docs/developer-guide/internationalization.md
@@ -32,7 +32,7 @@ You can configure MapStore to provide to the user only a restricted list of sele
     "locales": {
         "supportedLocales": {
             "en": {
-                "code": "en-EN",
+                "code": "en-US",
                 "description": "English"
             },
             "it": {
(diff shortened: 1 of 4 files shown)


docs/developer-guide/internationalization.md
```diff
@@ -48,6 +48,43 @@ Setting locales in localConfig.json file is doable only for supported locales pr
 The default behavior is to use those already configured in "supportedLocales" object.
 You can customize the messages by editing the data.code-CODE.json files.
 
+You can also add the `defaultUserLocale` property to localConfig.json to override the initial user locale:
+```javascript
+{
+    "defaultUserLocale": "it",
```
MV88: maybe i would call it defaultLocale only

docs/developer-guide/internationalization.md
```diff
@@ -48,6 +48,43 @@ Setting locales in localConfig.json file is doable only for supported locales pr
...
+MapStore will select the initial user locale based on the browser language and the supported locales if the `defaultUserLocale` is not defined.
+The property `defaultUserLocale` could be useful inside custom application where the locale is stored in other sources rather than using the browser language:
+
+```javascript
+// example
+import { getConfigProp } from '@mapstore/framework/utils/ConfigUtils';
+import cookies from 'js-cookie';
+
+// ...
+const defaultUserLocale = cookies.get('default_app_locale');
```
MV88, line 81:
```suggestion
const defaultUserLocale = cookies.get('default_app_locale'); // it-IT for example
```

docs/developer-guide/internationalization.md
allyoucanmap, line 54:
```suggestion
    "defaultUserLocale": "it-IT",
```

MV88:
```suggestion
    "defaultUserLocale": "it-IT",
```

docs/developer-guide/internationalization.md
```diff
@@ -48,6 +48,43 @@ Setting locales in localConfig.json file is doable only for supported locales pr
 The default behavior is to use those already configured in "supportedLocales" object.
 You can customize the messages by editing the data.code-CODE.json files.
 
+You can also add the `defaultUserLocale` property to localConfig.json to override the initial user locale:
```
MV88, line 51:
```suggestion
You can also add the `defaultUserLocale`, the locale **code**,  property to localConfig.json to override the initial user locale:
```

MV88 (changes requested):
all good, maybe we can think to call it defaultLocale instead of defaultUserLocale.

anyway if you compile wrongly the above-mentioned property the application does not load.
i think we should add a fallback mechanism to switch to english in this scenario.
possibly in a new issue @offtherailz @allyoucanmap @mbarto  @tdipisa what do you think?

docs/developer-guide/internationalization.md
offtherailz: If I understood correctly, it forces the initial locale. if so, maybe is not properly `defaultUserLocale` but `forceUserLocale` , `initialLocale` or something similar, what do you thing @allyoucanmap

allyoucanmap: @MV88 @offtherailz I called `defaultUserLocale` because we use the function `getUserLocale()` so I did not think much about this. In my opinion `initialLocale` or `defaultLocale` could work. Let me know which one you prefer.

allyoucanmap:
@offtherailz @MV88 I updated the PR with 'locale' instead of 'defaultUserLocale'

allyoucanmap:
> anyway if you compile wrongly the above-mentioned property the application does not load.
> i think we should add a fallback mechanism to switch to english in this scenario.
> possibly in a new issue @offtherailz @allyoucanmap @mbarto @tdipisa what do you think?

@offtherailz @allyoucanmap @mbarto @tdipisa

This issue is related to the failing request of loadLocale and it could happen for two main reasons:

1 - the locale property in localConfig is not correct (introdiced in this PR)
2 - the translations folder or a translation file is missing in the compiled app directory

So the fallback to english could work in the first case but not in the second one. In my opinion both cases should not provide a fallback to english because they depend on a static configuration and an error message could help the user to correctly seup the application.

MV88 (approved):

mbarto:
> So the fallback to english could work in the first case but not in the second one. In my opinion both cases should not provide a fallback to english because they depend on a static configuration and an error message could help the user to correctly setup the application.

I agree, it's not a good thing to ignore configuration errors.

offtherailz (approved):

Issue #6179, opened 2020-11-10
MapSwipte shouldn't be deactivated when the layer is deactivated


giohappy:
## Description
To me this is an unexpected behaviour: when the MapSwipe tool is activated for a layer it must be activated again if we select/unselect the layer itself. In my opinion the MapSwipe tool should remain selected when toggling the layer.

![ezgif-2-aa9bb3fccdc0](https://user-images.githubusercontent.com/571129/98704592-1da26c00-237d-11eb-97f6-ac218a848f08.gif)

ElenaGallo:
@AGMETEOR Now if I have activated the MapSwipe for a layer, it remains active even when opening another map

**How to reproduce**
Follows the steps shown in the gif below
![swipe](https://user-images.githubusercontent.com/56537133/100632839-9ae25080-332d-11eb-9c10-a5dd8750d9ee.gif)


*Expected Result**
The MapSwipe tool is disabled when a new map is opened

AGMETEOR:
> @AGMETEOR Now if I have activated the MapSwipe for a layer, it remains active even when opening another map
> 
> **How to reproduce**
> Follows the steps shown in the gif below
> ![swipe](https://user-images.githubusercontent.com/56537133/100632839-9ae25080-332d-11eb-9c10-a5dd8750d9ee.gif)
> 
> _Expected Result_*
> The MapSwipe tool is disabled when a new map is opened

Thanks @ElenaGallo . Pushed a fix here https://github.com/geosolutions-it/MapStore2/pull/6277

PR #6277, opened 2020-12-04
#6179: Deactivate swipe tool on location change

AGMETEOR:
## Description

This PR addresses this [comment](https://github.com/geosolutions-it/MapStore2/issues/6179#issuecomment-735877274)

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#6179 

**What is the new behavior?**

On location change deactivate swipe tool if it was active

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/__tests__/swipe-test.js
+++ b/web/client/epics/__tests__/swipe-test.js
@@ -7,8 +7,9 @@
  */
 import expect from 'expect';
 
-import { resetLayerSwipeSettingsEpic } from '../swipe';
+import { resetLayerSwipeSettingsEpic, deactivateSwipeToolOnSwitchMaps } from '../swipe';
 import { selectNode } from '../../actions/layers';
+import { onLocationChanged } from 'connected-react-router';
 import { testEpic } from './epicTestUtils';
 import { SET_ACTIVE } from '../../actions/swipe';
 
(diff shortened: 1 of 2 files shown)


MV88 (changes requested):
the PR is good per se, but we are using too much epics even when they are not needed.

in this simple case this could have been solved with a new case in the swipe reducer.
we should try to minimize the use of epics since they are handful to manage side effects (which are considered "bad" things")

may I ask if you can move the current implementation to the reducer?

also i want ot ask to my colleagues what do they think
@mbarto @allyoucanmap @offtherailz

offtherailz:
@MV88 I agree, but location change is not properly plugin logic, but app logic. For this reason I personally like also this solution, but what you say is correct, we can do it in a simpler way by resetting the reducer.

mbarto:
@offtherailz @MV88 looking here: https://github.com/geosolutions-it/MapStore2/blob/ff233b7bb63fdc2f0f7571773a7e70cd88806ed0/web/client/plugins/Swipe.jsx#L16

To me it looks like both the reducer and the epic are included at the plugin level, not at the application level

MV88:
> but location change is not properly plugin logic, but app logic

@offtherailz

is your point but to seems more logical to place the logic inside the plugin,
try to change the perspective.
"Me swipe plugin, what i need to do when it happen a location change? i need to reset my flags."

in this way you delegate to the plugin the behaviour of clean up procedures and so on

AGMETEOR:
@MV88 what's final verdict from the discussion above?

offtherailz:
@mbarto @MV88 Yes, the result is the same. I see reducers more "core", so seeing routing stuff inside them makes me wander "what if I don't have react-redux".
I can easly remove the epic from the plugin, It is less easy to remove the logic from the reducer. For this reason only I prefer epic solution, and in general I prefer to see less actions as possible in the reducer coming from other plugins/stuff...
 But is ok and works in both ways.  No strong opinion, proceed as you want

Issue #6282, opened 2020-12-08
Annotation layer in TOC and annotation panel visibility sync


dsuren1:
## Description
Annotation layer in TOC and annotation panel visibility sync

## Acceptance criteria
- As soon as I hide all annotations in the annotation panel, also the layer in TOC changes his state.
- If then I show one annotation in the annotation panel, also the layer in TOC changes his state accordingly
- As soon as I hide the annotation layer in TOC all the annotations are hidden accordingly

## Other useful information
https://github.com/geosolutions-it/austrocontrol-C125/issues/246#issuecomment-731289024

ElenaGallo:
@dsuren1 As soon as I hide all the annotations in the annotation panel, the layer in TOC changes its state and also on the map the annotations are hidden but once the annotation panel is closed the annotations are visible again.
![annotation](https://user-images.githubusercontent.com/56537133/101930342-4f0a9380-3bd8-11eb-9853-95b101228264.gif)


**How to Reproduce**
- Open a Demo Map on MS
- From burger Menu click on Annotations
- Hide all the annotations
- Click on X from the annotations panel

*Current Result*
The annotations become visible again

*Expected Result*
The annotations remain hidden

dsuren1:
Note: 
After this PR, the above issue should be fixed and also the annotation's visibility will always be in sync. i.e When adding/editing one annotation the other annotations will not be hidden. The visibility of the annotation will be retained unless changed from TOC or annotation panel by clicking on the visibility icon as per the requirement requested by the client 

Reference:
> Yes, it's no requirement. Still it would be useful to give the user the possibility to see other annotations in case they are drawing based in other annotations. Probably we will come back to this.

https://github.com/geosolutions-it/austrocontrol-C125/issues/246#issuecomment-732969958

PR #6284, opened 2020-12-08
#6282: Annotation layer visibility sync

dsuren1:
## Description
This PR adds commits to sync the visibility of annotation layer in TOC with that of annotations panel 

 - [x] Feature

## Issue

#6282

**What is the new behavior?**
- As soon as I hide all annotations in the annotation panel, also the layer in TOC changes his state.
- If then I show one annotation in the annotation panel, also the layer in TOC changes his state accordingly
- As soon as I hide the annotation layer in TOC all the annotations are hidden accordingly

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/annotations.js
+++ b/web/client/epics/annotations.js
@@ -7,7 +7,7 @@
  */
 
 import Rx from 'rxjs';
-import { head, findIndex, castArray, isArray, find, values } from 'lodash';
+import { head, findIndex, castArray, isArray, find, values, isEmpty } from 'lodash';
 import assign from 'object-assign';
 import axios from 'axios';
 import uuidv1 from 'uuid/v1';
@@ -72,7 +72,7 @@ import {
     STYLE_POINT_MARKER,
     STYLE_POINT_SYMBOL,
     DEFAULT_SHAPE,
-    DEFAULT_PATH } from '../utils/AnnotationsUtils';
+    DEFAULT_PATH, ANNOTATIONS } from '../utils/AnnotationsUtils';
 import { MEASURE_TYPE } from '../utils/MeasurementUtils';
 import { createSvgUrl } from '../utils/VectorStyleUtils';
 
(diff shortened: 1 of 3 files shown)


web/client/epics/annotations.js
```diff
@@ -436,15 +438,16 @@ export default (viewer) => ({
...
-                    features: annotationsLayerSelector(store.getState()).features.map(f => f.properties.id === action.id ? assign({}, f, {
-                        properties: {...f.properties, visibility: !isUndefined(f.properties.visibility) ? !f.properties.visibility : false}
-                    }) : f)
-                })
-            );
+            const feature = (f, visibility = false) => assign({}, f, {
+                properties: {...f.properties, visibility}
+            });
+            let isLayerPropertyChange = action.layer === "annotations";
+            const features = annotationsLayerSelector(store.getState()).features.map(f => isLayerPropertyChange ? feature(f, action?.newProperties?.visibility)
```
offtherailz, line 447: This epic reacts on every `CHANGE_LAYER_PROPERTIES`. If annotation layer is not present, there is an error. Please prevent this.

web/client/epics/annotations.js
```diff
@@ -202,23 +202,25 @@ export default (viewer) => ({
...
+                            }
+                            if (styleType === "LineString") {
+                                extraStyles.concat(getStartEndPointsForLinestring());
+                            }
+                            return {...ft,
+                                style: isArray(ft.style) ? ft.style.map(ftStyle => {
+                                    const {symbolUrlCustomized, ...filteredStyle} = ftStyle;
+                                    return filteredStyle;
+                                }) : [{...ftColl.style[styleType], id: ftColl.style[styleType].id || uuidv1(), symbolUrlCustomized: undefined}].concat(extraStyles)};
+                        })};
```
offtherailz, line 223: Can you describe a little bit with comments what happens here. Is not clear reading the code, expecially the "why". It could be a good imrovement.

offtherailz (changes requested):
Please fix the issue and comment a little bit more, on both methods, the reason of certain operations

offtherailz (commented):
If I change the visibility of one layer, the annotations on annotation layer, if present, is always updated with the previous status. 
We should avoid any action when the layer you are updating is not the your's one, to prevent side effects. Can we do it?

offtherailz (approved):

PR #6297, opened 2020-12-11
SaveLayer config with extended fields

mgogh:
## Description
Add the possibility to save extended fields for layers (added froms extension or plugins for example)

 - [ ] Bugfix
 - [ x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

If we create an extension which will add some particular fields to layers (ex: keywords, etc). We will not save these extended fields.
#<issue>

**What is the new behavior?**
Extended fields will be saved and reused as other standard fields.

## Breaking change
 - [x] No


mgogh:
Will check to make all test succeed

mgogh:
Tests "saveMapConfiguration" don't succeed cause there are more fields than expected (which is the purpose of this modification).

Issue #6363, opened 2021-01-08
Edit features problem when two layers are selected on map


ElenaGallo:
## Description

Opening the edit feature from the GFI panel, the attribute table was not showing the selected feature
![riga7](https://user-images.githubusercontent.com/56537133/104033346-adbf2f00-51cf-11eb-9f56-137c1bae99ea.gif)

## How to reproduce

- Open [this map](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/32214) or [this](http://localhost:8081/?debug=true#/viewer/openlayers/32524) one
- Click a point on map with state and observatories feature
- Click on edit 

*Expected Result*

The clicked observatories feature is visible on the attribute table

*Current Result*

The clicked observatories feature is not visible on the attribute table

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MV88:
Currently there are various issues:
- the pencil button in the GFI does not appear for [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/32214) map for states (click only on a state), it appears only for observatories (click only on an observatory)
- when clicking on both the pencil is activated for both, but the wfs request is wrong: because it contains <wfs:Query typeName="undefined", it works only if the layer you have in the GFI panel is selected in the toc, but it should work also without


- if you open instead [this](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10358) map the pencil does not appear for the layer states, while it should

Issue #6365, opened 2021-01-08
Highlight feature from GFI panel problem


ElenaGallo:
## Description

Only the features of the first layer of the TOC (observatories in this case) are highlighted on the map 

![riga8](https://user-images.githubusercontent.com/56537133/104034316-04793880-51d1-11eb-9586-ec686f96081f.gif)

## How to reproduce

- Open  [this map](https://dev.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/32214)
- Click a states feature on map
- Enable the highlight fetures

*Expected Result*

the states feature is highlighted on the map

*Current Result*

The states feature is not highlighted on the map

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #6412, opened 2021-01-15
#6363 Fix editing of features from GFI

MV88:
## Description

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#6363

**What is the new behavior?**

the pencil in gfi now appears correctly, i was using an index over a filtered array
and the wfs request that was failing is now using the correct feature type name.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/wfsquery.js
+++ b/web/client/epics/wfsquery.js
@@ -154,7 +154,7 @@ export const wfsQueryEpic = (action$, store) =>
                 ...queryOptions
             };
             return Rx.Observable.merge(
-                getLayerJSONFeature({...layer, search: {...layer.search, url}}, ogcFilter, options)
+                getLayerJSONFeature({...layer, name: layer.name || action.filterObj.featureTypeName, search: {...layer.search, url}}, ogcFilter, options)
                     .map(data => querySearchResponse(data, action.searchUrl, action.filterObj, action.queryOptions, action.reason))
                     .catch(error => Rx.Observable.of(queryError(error)))
                     .startWith(featureLoading(true))
--- a/web/client/epics/wfsquery.js
+++ b/web/client/epics/wfsquery.js
@@ -153,8 +153,9 @@ export const wfsQueryEpic = (action$, store) =>
                 sortOptions,
                 ...queryOptions
             };
+            // TODO refactor, the layer that should be used should be the used when the feature grid is opened from the toc, see #6430
             return Rx.Observable.merge(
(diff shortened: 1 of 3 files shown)


web/client/epics/wfsquery.js
```diff
@@ -154,7 +154,7 @@ export const wfsQueryEpic = (action$, store) =>
                 ...queryOptions
             };
             return Rx.Observable.merge(
-                getLayerJSONFeature({...layer, search: {...layer.search, url}}, ogcFilter, options)
+                getLayerJSONFeature({...layer, name: layer.name || action.filterObj.featureTypeName, search: {...layer.search, url}}, ogcFilter, options)
```
offtherailz, line 157: Here I don't understand why the layer set by `setLayer` here doesn't work..., maybe you need to set layer before triggering click point Selector action? Give it a try. 
https://github.com/geosolutions-it/MapStore2/blob/e34489ca12b06e958f57052d068976bc7080059d/web/client/epics/identify.js#L355

Because I'd like to keep as much simpler as possible. If it doesn't work, please Can you add a unit test for this, specifying why you need this additional check.

offtherailz (changes requested):
Please check my comment, I think the can be solved inverting the actions in EDIT_LAYER_FEATURES

web/client/epics/wfsquery.js
offtherailz: mmm... i tried #6363 with #6365 on this map.
http://localhost:8081/#/viewer/openlayers/10358

Tried to zoom to new york 
click here:
image
Change to state layer in the selector and tried to edit. It doesn't work. I think you should try the suggested solution.

MV88: i have tested the both pr together and the fix is to revert the change [here](https://github.com/geosolutions-it/MapStore2/pull/6412/files#diff-e8488af38db4e66cbfd306ca62e231eff32ea2138ca547086e739f19c91490beL74)

since i'm using only valid responses there

offtherailz: I opened https://github.com/geosolutions-it/MapStore2/issues/6430 for fixing this structural issue.

offtherailz (approved):

tdipisa:
@MV88 please backport this to 2021.01.xx for the release.

ElenaGallo:
@MV88 the PR passed the test on dev, please backport this to the stable branch.

Issue #6293, opened 2020-12-10
Embed Dashboard into existing web page


tdipisa:
## Description

Improve the Share tool of Dashboards by including the possibility to embed a [Dashboard](https://mapstore.readthedocs.io/en/latest/user-guide/exploring-dashboards/) in a web page, just as with maps. Below is the current share tool for Dashboards.

![image](https://user-images.githubusercontent.com/1280027/101796561-5151ec80-3b09-11eb-9f47-2e88a7f0616c.png)

## Acceptance criteria
The activities needed for this improvement are (similar to GeoStory embedded, see issue #):

- Add a new application entry with related files and configuration for dashboard embedded
- Add the embedded tab inside the share panel plugin for dashboard
- Update the project with the new files and app entry for dashboard embedded created in the point a) and b)
- An embedded dashboard is only in view mode: it is not possible to add widgets as it is now if you open a dashboard in a mobile device

For all of these three points we need to take into account the current responsiveness of the dashboard layout: we did a quick test with the current implementation removing temporary the top bar and displaying the results on different devices. The Generic Laptop size (~1280px) and the mobile size (~400px)  could be fine but the **medium** size (~768px) **has not a good visual fit**. 

![image](https://user-images.githubusercontent.com/19175505/98950620-8282d100-24f9-11eb-960d-0e25ad09bcd4.png)

## a) Add a new application entry with related files and configuration for dashboard embedded

We should add a new application entry that import only the minumum components and configuration to make the dashboard work in view mode to keep the bundle as light as possible.

We will need following files:

- dashboardEmbedded.jsx app entry (this should be added to webpack)
- appConfigDashboardEmbedded.js for routing and other configurations
- pluginsDashboardEmbedded.js the list of plugins to make the dashboard work only in view mode)
- dashboard-embedded.html the html template that contain the new embedded dashboard
- configuration for this new page in the localConfig that list all the needed plugins
- (? to be verified while developing) probably we will need to create a new page that request the dashboard config based on the id on mount
- (? to be verified while developing) add all needed reducers and epics to the app level

The name of the new files in the list above are only a reference they can/should be changed if needed while developing.

If some file is missing in the list above use the embedded.jsx of map as reference.

While working on this new app entry we should take particular attention to the following aspects:

- the dashboard should resize and fit the given container (see comment above)

## b) Add the embedded tab inside the share panel plugin for dashboard

[...]

PR #6443, opened 2021-01-21
#6293: Embed Dashboard into existing web page

allyoucanmap:
## Description

This PR introduces the embed page for dashboard and it's possible to get a snippet from the Share plugin under the embed tab.

image

The option "Show connection" will create a link with the connections parameter set to true. The connections parameter will display widgets with the colored border, same color means connected widgets.

- `dashboard-embedded.html#/:id`

image

- `dashboard-embedded.html?connections=true#/:id`

image

The new parameter `minLayoutWidth` in the Dashboard cfg allow to change the minimum width of the layout if the width is less or equal to the configured value the widgets are listed in a single column:

- dashboard in embedded mode has a value of 768px for  minLayoutWidth

image

 - [x] Feature

## Issue

#6293

**What is the new behavior?**

see description

## Breaking change
 - [x] No


web/client/localConfig.json
```diff
@@ -656,8 +656,14 @@
...
           "cfg": {
-            "embedPanel": false,
-            "advancedSettings": false
+            "showAPI": false,
+            "advancedSettings": false,
+            "shareUrlRegex": "(h[^#]*)#\\/dashboard\\/([A-Za-z0-9]*)",
+            "shareUrlReplaceString": "$1dashboard-embedded.html#/$2",
+            "embedOptions": {
+              "showTOCToggle": false,
+              "showConnectionsParamToggle": true
```
MV88, line 665: @tdipisa should this be true by default?

web/client/plugins/Dashboard.jsx
```diff
@@ -120,27 +123,42 @@ const WidgetsView = compose(
...
 
 export default {
-    DashboardPlugin,
+    DashboardPlugin: withResizeDetector(DashboardPlugin),
     reducers: {
+        dashboard: dashboardReducers,
         widgets: widgetsReducers
+    },
+    epics: {
+        ...dashboardEpics,
```
MV88, line 161: do we need all of them here?

web/client/product/appConfigDashboardEmbedded.js
```diff
@@ -0,0 +1,44 @@
...
+        defaultState: {},
+        mobile: {}
+    },
+    baseReducers: {
+        version,
+        dashboard,
+        widgets
+    },
+    baseEpics: {
+        ...dashboardEpics,
```
MV88, line 36: if these are included by the plugin, this are not needed here, right?

MV88 (changes requested):
Please, see my comments here and there :)

i think we need to add embed also here
can embed be present for dashboard
image

web/client/localConfig.json
```diff
@@ -656,8 +656,14 @@
         }, "Login", "Language", "NavMenu", "DashboardSave", "DashboardSaveAs", "Attribution", "Home", {
           "name": "Share",
           "cfg": {
-            "embedPanel": false,
-            "advancedSettings": false
+            "showAPI": false,
+            "advancedSettings": false,
+            "shareUrlRegex": "(h[^#]*)#\\/dashboard\\/([A-Za-z0-9]*)",
```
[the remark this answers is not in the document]
allyoucanmap: Share plugin is used by three different applications: Map, Dashboard and GeoStory. The default values are used by embedded map:

https://github.com/geosolutions-it/MapStore2/blob/75c8631cae65ddf5660d211e6953f1e0bbecfcde/web/client/components/share/SharePanel.jsx#L97-L98

fo this reason I had to add specific rules for dashboard.

web/client/localConfig.json
allyoucanmap: In my opinion is better to keep this value false by default because is needed only for dashboard. If we put this value true by default we should instead add a rule to share plugin of map with showConnectionsParamToggle set to false.

web/client/plugins/Dashboard.jsx
allyoucanmap: If you try to add only a Dashboard plugin in a application most of the interactions were missing for example loading of a dashboard (epics/dashboard) or filter of maps synch with table (epics/widgets).

allyoucanmap: > do we need all of them here?
@MV88 Do you think this is causing some issue?

web/client/product/appConfigDashboardEmbedded.js
allyoucanmap: I could remove them from the plugin and keep them only here.

web/client/plugins/Dashboard.jsx
allyoucanmap: I'll remove them from here

MV88: i thought it was only redundant, maybe is better to have them only in the plugin?

web/client/plugins/Dashboards.jsx
```diff
@@ -71,7 +73,17 @@ class Dashboards extends React.Component {
             md: 2,
             className: 'ms-map-card-col'
         },
-        maps: []
+        maps: [],
+        shareOptions: {
```
MV88, line 77: this is a clone of the one in featured maps, maybe we can use a constant pulled from the shareUtils
so in the future we will edit only in one place

MV88 (commented):
seems good, but please have a look at my answers and one new comment i'm adding now.

MV88 (approved):

tdipisa:
@ElenaGallo please let us know as soon as tests on DEV are ok so that we will proceed with the needed backport to 2021.01.xx.

ElenaGallo:
Test done in Dev. @allyoucanmap please, backport to 2021.01.xx.

Issue #6495, opened 2021-01-29
Incorrect placement of widgets


tdipisa:
## Description

Regression from 2021.01.00 that affect DEV and QA instances. In latest mapstore there is a small bug in the widgets placement: with the current placement grid it is not possible to properly align widgets on the right.
![image.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/026a3c9e-243e-40f1-a024-78955e327dfc)

The correct position should be the following:

![image.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/58a8c623-74c2-4f37-9559-42163e221463)

**Steps to reproduce**

![widgets_issue](https://user-images.githubusercontent.com/1280027/106454327-da8efb00-648a-11eb-87b1-5d05e2d067a3.gif)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MV88:
so far i have noticed that the update of the version of react-grid -layout is causing this issue
working on a fix

PR #6500, opened 2021-02-01
#6495 fix for widgets width when adding them on the map viewer

MV88:
## Description

the width is passed correctly when it is present solving the issue,
this was caused by incorrect calculation of width on first render

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#6495

**What is the new behavior?**

the initial width of a widget card is 213px which is the requested one
## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/widgets/enhancers/emptyTextState.js
+++ b/web/client/components/widgets/enhancers/emptyTextState.js
@@ -4,8 +4,12 @@ import emptyState from '../../misc/enhancers/emptyState';
 
 export default emptyState(
     ({text} = {}) => !text,
-    ({iconFit} = {}) => ({
-        iconFit,
-        tooltip: <Message msgId="widgets.errors.notext" />
+    () => ({
+        iconFit: false,
+        glyph: false,
+        style: {
+            margin: "auto"
+        },
+        title: <Message msgId="widgets.errors.notext" />
     })
 );
--- a/web/client/components/widgets/view/__tests__/WidgetsView-test.jsx
+++ b/web/client/components/widgets/view/__tests__/WidgetsView-test.jsx
@@ -115,49 +115,6 @@ describe('WidgetsView component', () => {
         const el = container.querySelector('.react-grid-item');
         expect(el).toExist();
     });
-    it('tests correct width of widgets', () => {
-        ReactDOM.render(<WidgetsView
-            useDefaultWidthProvider
-            isResizable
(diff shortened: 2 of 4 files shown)


allyoucanmap (changes requested):
I tested this locally and it does not to seem to solve the issue. It seems the widget on map is using two different widthProvider switching them with the `useDefaultWidthProvider` prop. When the WidgetView is in normal state (not maximize) the `useDefaultWidthProvider` is set to true so we are using the [WidthProvider](https://github.com/STRML/react-grid-layout/blob/1.2.0/lib/components/WidthProvider.jsx) of the react-data-grid library that seems to listen only on the mount event and the window resize event. Lately we updated the [resize listener](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/layout/enhancers/gridLayout.jsx#L27) we are using to check for the container width with a different library to avoid the issue to compute also the space of the scrollbar but this new resize detector seems to return undefined width on the first renders. This behaviour is generating a wrong CSS [here](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/Widgets.jsx#L87-L96) and probably when empty the react-grid-layout WidthProvider does not compute the right width on mount while after tirggering the maximize  or resize event we reset the WidthProvider and get the right size of the container.

I think we could approach this width a direct control on the width also for the normal view of the widgets as we are doing with the maximize props https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/Widgets.jsx#L60-L100

What do you think @MV88 @offtherailz ?

MV88:
@tdipisa this pr also fixes this problem by showing only text when an empty text widget has been added

https://user-images.githubusercontent.com/11991428/107750541-7850b880-6d1c-11eb-9896-00b35ff6ff3f.mp4

allyoucanmap (approved):

allyoucanmap:
> @tdipisa this pr also fixes this problem by showing only text when an empty text widget has been added
> 
>  empty.text.widget.problem.mp4

@tdipisa  I approved the PR, let me know if I can merge it or not

tdipisa:
@ElenaGallo please test this in DEV as soon as deployed and let us know if we can backport to 2021.01.xx

ElenaGallo:
@MV88 please backport this Pr to the stable branch

Issue #6560, opened 2021-02-22
Show informative messages inside widgets rather then on hover


giohappy:
## Description
As an example of what I'm proposing, here below an image of a chart widget when map synced and the unlderying data is not inside the map viewport. Since no data are available in this are a not much informative "i" icon is shown.

![image](https://user-images.githubusercontent.com/571129/108696047-50720980-7501-11eb-8963-ef614ecd1d79.png)

The user has not hints of what is happening. In a case it lead the user to think it was a bug.
He will discover what's happening only if he hover the mouse on the icon

![image](https://user-images.githubusercontent.com/571129/108696153-7697a980-7501-11eb-88c1-798c85a2c0ab.png)

It shouldn't be a big deal if we show the message directly inside the widget, maybe above (or in place) of the current icon. It would be a much more effective and informative wai to make the user aware of what's happening.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

MuhweziDeo:
hi @tdipisa check the design below
<img width="418" alt="Screenshot 2021-03-01 at 14 43 54" src="https://user-images.githubusercontent.com/39124174/109492841-a106db00-7a9c-11eb-8fe5-ab3792d33abc.png">

giohappy:
@MuhweziDeo the message is half true. No data is available **for this area**. This is important for the user...

tdipisa:
I agree with @giohappy @MuhweziDeo, the message is already there and visible on mouse hover: we have to make it more visible.

MuhweziDeo:
okay let me see how to generate the message

PR #6596, opened 2021-03-01
#6560: Show messages in widget other than hover

MuhweziDeo:
## Description

 - [x] Feature

## Issue

108696153-7697a980-7501-11eb-88c1-798c85a2c0ab

#6560

**What is the new behavior?**

Screenshot 2021-03-01 at 16 08 05

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/featuregrid/EmptyRowsView.jsx
+++ b/web/client/components/data/featuregrid/EmptyRowsView.jsx
@@ -6,9 +6,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 import React from 'react';
-import {Glyphicon} from 'react-bootstrap';
 import PropTypes from 'prop-types';
 import Message from '../../I18N/Message';
+import WidgetEmptyMessage from '../../widgets/widget/WidgetEmptyMessage';
+
 const nodataStyle = {
     width: "100%",
     height: "100%",
@@ -16,24 +17,13 @@ const nodataStyle = {
     verticalAlign: "center"
 };
 
-const emptyRowsStyle = {
-    position: 'absolute',
-    width: '100%',
-    height: '100%',
-    display: 'flex',
-    alignItems: 'center',
-    justifyContent: 'center'
-};
 class EmptyRowsView extends React.PureComponent {
     static propTypes = {
         loading: PropTypes.bool
     }
     render() {
         return this.props.loading
-            ? (<div style={nodataStyle}><Message msgId="loading" /></div>) :
-            <div style={emptyRowsStyle}>
-                <Glyphicon glyph="list-alt"/>&nbsp;<Message msgId="featuregrid.noFeaturesAvailable" />
-            </div>;
+            ? (<div style={nodataStyle}><Message msgId="loading" /></div>) : (<WidgetEmptyMessage messageId="featuregrid.noFeaturesAvailable" glyph="features-grid"/>);
     }
 }
 
--- a/web/client/components/data/featuregrid/EmptyRowsView.jsx
+++ b/web/client/components/data/featuregrid/EmptyRowsView.jsx
@@ -8,7 +8,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Message from '../../I18N/Message';
-import WidgetEmptyMessage from '../../widgets/widget/WidgetEmptyMessage';
 
 const nodataStyle = {
     width: "100%",
@@ -23,7 +22,7 @@ class EmptyRowsView extends React.PureComponent {
     }
     render() {
         return this.props.loading
-            ? (<div style={nodataStyle}><Message msgId="loading" /></div>) : (<WidgetEmptyMessage messageId="featuregrid.noFeaturesAvailable" glyph="features-grid"/>);
+            ? (<div style={nodataStyle}><Message msgId="loading" /></div>) : (this.props.children ? this.props.children : <div style={nodataStyle}><Message msgId="featuregrid.noFeaturesAvailable" /></div>);
     }
 }
 
--- a/web/client/components/data/featuregrid/EmptyRowsView.jsx
+++ b/web/client/components/data/featuregrid/EmptyRowsView.jsx
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 import React from 'react';
-
+import {Glyphicon} from 'react-bootstrap';
 import PropTypes from 'prop-types';
 import Message from '../../I18N/Message';
 const nodataStyle = {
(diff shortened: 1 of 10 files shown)


web/client/components/widgets/enhancers/emptyChartState.js
```diff
@@ -1,11 +1,13 @@
...
+import EmptyView from '../../misc/EmptyView';
 
 export default emptyState(
     ({data = []}) => !data || data.length === 0,
     ({mapSync, iconFit} = {}) => ({
         iconFit,
         tooltip: mapSync ? <Message msgId="widgets.errors.nodatainviewport" /> : <Message msgId="widgets.errors.nodata" />
-    })
+    }),
+    () => <EmptyView content={<Message msgId="widgets.errors.nodatainviewport" />}/>
```
allyoucanmap, line 12: in this case we are adding a custom component for the empty state and in this way we are overriding the ternary that select a different message based on the mapSync props.
I think we should keep the two different messages with something like this:
```js
import React from 'react';
import { Glyphicon } from 'react-bootstrap';
import Message from '../../I18N/Message';
import emptyState from '../../misc/enhancers/emptyState';

function ChartEmptyState({ message }) {
    return (
        <div
            style={{
                position: 'absolute',
                width: '100%',
                height: '100%',
                display: 'flex',
                alignItems: 'center',
                justifyContent: 'center'
            }}>
            <Glyphicon glyph="stats"/>&nbsp;{message}
        </div>
    );
}

export default emptyState(
    ({data = []}) => !data || data.length === 0,
    ({mapSync} = {}) => ({
        message: mapSync ? <Message msgId="widgets.errors.nodatainviewport" /> : <Message msgId="widgets.errors.nodata" />
    }),
    ChartEmptyState
);
```

Note: the above code is only an example to discuss and on how we could implement this.

allyoucanmap (changes requested):
This fix replaces the previous empty state message only for the chart so this is the current situation:

image

I think we should align all empty messages

image

or with an icon to reference the widget type (I'm not sure if it's currently possible with the table component)

image

@giohappy could you please give us a feedback on this?

giohappy:
I agree @allyoucanmap. Visual cohesiveness is important for these kind of things.

MuhweziDeo:
@giohappy  should we go with the one with text Aligned center or use the one with Icons. Such I can implement it for all the 4 items

giohappy:
I think the icon will help to identify which widget is what, but the text is already sufficently clear to me.
If using the icon is not problematic ok, otherwise it shouldn't be a big deal.

web/client/components/data/featuregrid/EmptyRowsView.jsx
```diff
@@ -15,14 +15,25 @@ const nodataStyle = {
...
 class EmptyRowsView extends React.PureComponent {
     static propTypes = {
         loading: PropTypes.bool
     }
     render() {
         return this.props.loading
             ? (<div style={nodataStyle}><Message msgId="loading" /></div>) :
-            <div style={nodataStyle}><Message msgId="featuregrid.noFeaturesAvailable" /></div>;
+            <div style={emptyRowsStyle}>
+                <Glyphicon glyph="list-alt"/>&nbsp;<Message msgId="featuregrid.noFeaturesAvailable" />
```
allyoucanmap, line 35: Please use the `features-grid` glyph

allyoucanmap (changes requested):

MuhweziDeo:
thanks, @allyoucanmap I will get to implementing these changes right away

web/client/components/data/featuregrid/EmptyRowsView.jsx
```diff
@@ -6,23 +6,24 @@
...
+
 class EmptyRowsView extends React.PureComponent {
     static propTypes = {
         loading: PropTypes.bool
     }
     render() {
         return this.props.loading
-            ? (<div style={nodataStyle}><Message msgId="loading" /></div>) :
-            <div style={nodataStyle}><Message msgId="featuregrid.noFeaturesAvailable" /></div>;
+            ? (<div style={nodataStyle}><Message msgId="loading" /></div>) : (<WidgetEmptyMessage messageId="featuregrid.noFeaturesAvailable" glyph="features-grid"/>);
```
allyoucanmap, line 26: Instead of modify this component used inside different grid we could give the possibility to customize the node message :

```js
class EmptyRowsView extends React.PureComponent {
    static propTypes = {
        loading: PropTypes.bool
    }
    render() {
        return this.props.loading
            ? (<div style={nodataStyle}><Message msgId="loading" /></div>) :
            this.props.children
                ? this.props.children
                : <div style={nodataStyle}><Message msgId="featuregrid.noFeaturesAvailable" /></div>;
    }
}
```

then we could pass the WidgetEmptyMessage only for the TableWidgets

https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/widgets/widget/TableWidget.jsx#L70

eg:

```js
// ...
emptyRowsView={() =>
    <EmptyRowsView loading={loading}>
        <WidgetEmptyMessage messageId="featuregrid.noFeaturesAvailable" glyph="features-grid"/>
    </EmptyRowsView>}
// ...
```

in this way the WidgetEmptyMessage is related only to the Widgets components.
If you think the this.props.children is not enough explicit you could use a props to pass the custom message node.

web/client/themes/default/less/widget.less
```diff
@@ -316,3 +316,20 @@
...
+.ms-widget-empty-message {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.ms-widget-empty-text-message {
```
allyoucanmap, line 328: I would like to avoid duplicate this class only for the position property so I propose this approach:

```css
.ms-widget-empty-message {
    position: absolute;
    /* we ensure the empty table message is centered and cover the header with top at 0 and filled background */
    top: 0; 
    background-color: @ms2-color-background;
    width: 100%;
    height: 100%;
    display: flex;
    align-items: center;
    justify-content: center;
}
/* needed to compute the correct position of empty messages */
.mapstore-widget-card .ms2-border-layout-body {
    position: relative;
}
```

and remove the .ms-widget-empty-text-message class

web/client/components/widgets/enhancers/emptyTextState.js
```diff
@@ -1,15 +1,12 @@
...
-        iconFit: false,
-        glyph: false,
-        style: {
-            margin: "auto"
-        },
-        title: <Message msgId="widgets.errors.notext" />
-    })
+        glyph: 'sheet',
+        messageId: 'widgets.errors.notext',
+        className: 'ms-widget-empty-text-message'
```
allyoucanmap, line 9: remove the ms-widget-empty-text-message following previous proposal

allyoucanmap (changes requested):

web/client/components/data/featuregrid/EmptyRowsView.jsx
MuhweziDeo: this makes it a better implementation. Using props.children is enough

allyoucanmap (approved):

ElenaGallo:
@MuhweziDeo test done, please backport this PR to stable branch, thank you.

Issue #6539, opened 2021-02-16
Not proper validation for HTML featureinfo response


offtherailz:
## Description
If the GFI of a WMS replies with HTML that starts with `<?xml`, instead of displaying data, MapStore reply with an error. 
HTML may start with `<?xml` without any problem. The feature info parsing try to catch this string to identify XML exception, but this is not valid, expecially with html or xml (GML) responses.

https://github.com/geosolutions-it/MapStore2/blob/64175ad83801593a135d2f83026cbe8a2ce1c047/web/client/utils/FeatureInfoUtils.js#L73

Note: the xml check is not a good check anyway. A check like this is more precise. Anyway it should be converted in a Promise version.

https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/ObservableUtils.js#L33

## How to reproduce

-  Open a layer that replies to GFI in html with the `<?xml ` prefix. (configure it, or use [this](http://ovc.catastro.meh.es/cartografia/INSPIRE/spadgcwms.aspx) catalog source and add some layers)
- Select HTML as feature info format and zoom to the zone with data (in the passed wms you can zoom on **Madrid**).
- Click to get the FeatureInfo. 

*Expected Result*
- If the service replies with a valid HTML (not always, it depends on the point clicked, for the example service) MapStore shows the HTML

*Current Result*
- MapStore tells "No feature found for any result.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MuhweziDeo:
@offtherailz please help me share  a layer that replies to GFI in HTML

MuhweziDeo:
@offtherailz is it necessary to convert the `interceptOGCError` to return a Promise version.

offtherailz:
@MuhweziDeo The linked interceptOGCError is only an example, but it is not buggy. 
I think for the moment you can proceed improving the check in featureInfoUtils to do a real check (strings starting with `<?xml` are still valid html, an error should intercept the `<ExceptionReport` tag. To do a test about the error response you can try to do the same http request with an invalid format.

Please let me know if it is more clear now?

offtherailz:
[...]

PR #6598, opened 2021-03-02
 6539: Not proper validation for HTML featureinfo response

MuhweziDeo:
## Description

 - [x] Bugfix

#6539 

MapStore tells "No feature found for any result.

#<issue>

**What is the new behavior?**
If the service replies with a valid HTML MapStore shows the HTML

## Breaking change
 - [x] No

## Other useful information
I added a check response includes <xml> since the xml reponse can contain html. So with this the second check for `regexpBody` will verify if reponse contains html body

### DIFF
--- a/web/client/utils/FeatureInfoUtils.js
+++ b/web/client/utils/FeatureInfoUtils.js
@@ -32,7 +32,7 @@ const regexpBody = /^[\s\S]*<body[^>]*>([\s\S]*)<\/body>[\s\S]*$/i;
 const regexpStyle = /(<style[\s\=\w\/\"]*>[^<]*<\/style>)/i;
 
 export function parseHTMLResponse(res) {
-    if ( typeof res.response === "string" && res.response.indexOf("<?xml") !== 0 ) {
+    if ( typeof res.response === "string" && res.response.indexOf("<?xml") !== -1 ) {
         let match = res.response.match(regexpBody);
         if ( res.layerMetadata && res.layerMetadata.regex ) {
             return match && match[1] && match[1].match(res.layerMetadata.regex);
--- a/web/client/utils/FeatureInfoUtils.js
+++ b/web/client/utils/FeatureInfoUtils.js
@@ -32,7 +32,7 @@ const regexpBody = /^[\s\S]*<body[^>]*>([\s\S]*)<\/body>[\s\S]*$/i;
 const regexpStyle = /(<style[\s\=\w\/\"]*>[^<]*<\/style>)/i;
 
 export function parseHTMLResponse(res) {
-    if ( typeof res.response === "string" && res.response.indexOf("<?xml") !== -1 ) {
+    if ( typeof res.response === "string" && (res.response.indexOf("<?xml") !== -1 || res.response.indexOf("<?xml") !== 0) ) {
         let match = res.response.match(regexpBody);
         if ( res.layerMetadata && res.layerMetadata.regex ) {
             return match && match[1] && match[1].match(res.layerMetadata.regex);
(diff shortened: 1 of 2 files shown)


web/client/utils/FeatureInfoUtils.js
```diff
@@ -32,7 +32,7 @@ const regexpBody = /^[\s\S]*<body[^>]*>([\s\S]*)<\/body>[\s\S]*$/i;
 const regexpStyle = /(<style[\s\=\w\/\"]*>[^<]*<\/style>)/i;
 
 export function parseHTMLResponse(res) {
-    if ( typeof res.response === "string" && res.response.indexOf("<?xml") !== 0 ) {
+    if ( typeof res.response === "string" && (res.response.indexOf("<?xml") !== -1 || res.response.indexOf("<?xml") !== 0) ) {
```
allyoucanmap, line 35: With this new statement the part `(res.response.indexOf("<?xml") !== -1 || res.response.indexOf("<?xml") !== 0)` will return true for each index position so it has not a real effect.

from @offtherailz 
> @MuhweziDeo The linked interceptOGCError is only an example, but it is not buggy.
> I think for the moment you can proceed improving the check in featureInfoUtils to do a real check (strings starting with <?xml are still valid html, an error should intercept the <ExceptionReport tag. To do a test about the error response you can try to do the same http request with an invalid format.

Before we used to check the initial xml definition to ensure we were parsing an error or the actual valid response but this is not valid for all the servers.
We know that OGC error uses the 200 status and it should contain the Exception tag, we have an example from the https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/ObservableUtils.js#L35. (from the issue description).

I tried to generate an error in their instance by changing the `layers` field with an invalid layer and it seems their service return a ServiceException tag
image

We could apply this statement "is a string, does not contain <ExceptionReport and does not contain <ServiceExceptionReport tags"
- `typeof res.response === "string" && res.response.indexOf("<ExceptionReport") === -1 && res.response.indexOf("<ServiceExceptionReport") === -1`

What do you think @offtherailz ?

allyoucanmap (changes requested):

web/client/utils/FeatureInfoUtils.js
MuhweziDeo: @allyoucanmap this makes sense but what are the chances that there isnt other exceptions that can be returned

allyoucanmap: @MuhweziDeo, I talked with @offtherailz and the xml check is an old control so we should remove it.
So we should keep this one:
`if ( typeof res.response === "string") {`

We need 2 unit tests one with an error response and one with a valid html.

Here an error example:
```xml
<?xml version='1.0' encoding="ISO-8859-1"  standalone="no" ?>
<ServiceExceptionReport version="1.1.1">
  <ServiceException code="InvalidFormat">
    <![CDATA[
    Parámetros erroneos:
    formato = image/png
    XMin = -412208.172942018
    YMin =  4928258.28942967
    XMax = -411725.664200968
    YMax =  4928740.79817072
    AnchoPixels =  101
    AltoPixels =  101
    Transparente = No
    Descripción error:
    layers (AD.ADDRESSA) No soportada.]]>
   </ServiceException>
</ServiceExceptionReport>
```

MuhweziDeo: thanks @allyoucanmap

allyoucanmap (approved):

tdipisa:
@ElenaGallo can you please check this in DEV and let us know if we can proceed with the backport to 2021.01.xx?

ElenaGallo:
Test done on DEV,  @MuhweziDeo please backport this to the stable branch, thx.

Issue #6592, opened 2021-02-25
Compare tool problem


magnusnil:
## Description
The vertical swipe compare tool does not match the map content.

![CompareTool](https://user-images.githubusercontent.com/43540361/109211974-44ab6f00-77af-11eb-9f73-fabc118776b7.JPG)

## How to reproduce
- Add two layers and pick the swipe compare tool 
- The slider map content and the swiper are not synced

Tested on Chrome 88.0.4324.182 and Edge 88.0.705.74 using MapStore 2021.01.00.

wdcsimon:
We are using MapStore 2021.01.00 as well with the same issue.
The slider and swipe are in the correct place if it is hard against the left hand side, but as you move the slider to the right the gap between slider and swipe gets increasingly wider.
Same issue with the slider and swipe set to horizontal, with the gap increasing moving from top to bottom.

<img width="1918" alt="Slider2" src="https://user-images.githubusercontent.com/62117895/109214705-0c3f7b80-7817-11eb-92b2-49b7b7a6e822.png">

offtherailz:
Looking @wdcsimon screenshot I noticed the big resolution ( or he changed the zoom level from chrome in order to reduce the size of the control). So I tried to use the zoom tools and I as able to replicate a similar issue, reducing page zoom from chrome at 75%.

![image](https://user-images.githubusercontent.com/1279510/109280871-dbae1080-781b-11eb-8da0-93205afbf6d1.png)

Maybe also the original issue by @magnusnil is due to a similar thing ? 
To well identify the issue, can you (both) confirm you altered the zoom of chrome (or tell if you have a big screen resolution instead, for instance ) ?

magnusnil:
I use 1920 x 1080 and zoom is set to 100% in Chrome.

I noticed that the Windows setting for scaling was set to 125%. When I set this to 100%, the compare tool works as expected.

wdcsimon:
Myself and the GIS team here have screens set at 3840x2160 with 150% scaling.  We all have the issue. 
Other staff in our organization mainly have screens set at 1920 x1080 with 100% scaling - they do not have the issue.
So it seems to be related to screen resolution and/or scaling.

MuhweziDeo:
This issue is can strange I have tried to scale my chrome zoom level but the outcome is still the same even when it's not zoomed. However, I do notice that the swiper width increase the more you scroll either left or right. Not sure what exact issue is currently to fix @offtherailz

[...]

PR #6612, opened 2021-03-05
6592: Compare tool Problem

MuhweziDeo:
## Description

 - [x] Bugfix

#6592 

The default vertical and horizontal heights are not in inline with the slider and swipe height 
The slider map content and the swiper are not synced
#6592 

**What is the new behavior?**
The default vertical and horizontal heights are in inline with the slider and swipe height 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/map/openlayers/swipe/SliderSwipeSupport.jsx
+++ b/web/client/components/map/openlayers/swipe/SliderSwipeSupport.jsx
@@ -36,7 +36,13 @@ const VSlider = ({ type, map, widthRef }) => {
     }, [ type ]);
 
     useEffect(() => {
-        widthRef.current = map.getProperties().size[0] / 2;
+        const callback = (event) => {
+            let ctx = event.context;
+            widthRef.current = ctx.canvas.width / 2;
+            map.un('precompose', callback);
+        };
+        map.on('precompose', callback);
+        return () => map.un('precompose', callback);
     }, [ type ]);
 
     return (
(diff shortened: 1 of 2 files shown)


web/client/components/map/openlayers/swipe/SliderSwipeSupport.jsx
```diff
@@ -36,7 +36,13 @@ const VSlider = ({ type, map, widthRef }) => {
     }, [ type ]);
 
     useEffect(() => {
-        widthRef.current = map.getProperties().size[0] / 2;
+        const callback = (event) => {
```
allyoucanmap, line 39: After an investigation on the compare behaviour the issue seems related to the devicePixelRatio.

These are the values printed after changing the display settings of the pc to 125%:

image

The map canvas still uses the original screen width 1920px because it computes the size width multiply by the device pixel ratio so probably the solution is to multiply the getWidth() inside the EffectSupport by the pixel ratio.

https://github.com/geosolutions-it/MapStore2/blob/f7698720254b4028afbf78cffde6b7a4002cf2bd/web/client/components/map/openlayers/swipe/EffectSupport.jsx#L29
```js
const width = getWidth() * map.pixelRatio_;
```
At this point we should also fix the getHeight();

So please remove this callback and restore the previous code in this file.

allyoucanmap (changes requested):

web/client/components/map/openlayers/swipe/SliderSwipeSupport.jsx
```diff
@@ -78,11 +78,11 @@ const HSlider = ({ type, map, heightRef }) => {
 
     const onWindowResize = () => {
         setPos({x: 0, y: 0});
-        heightRef.current = map.getProperties().size[1] / 2;
+        heightRef.current = map.getProperties().size[1];
```
allyoucanmap, line 81: I don't think we need to change this file because the `map.pixelRatio_` is applied in the EffectSupport.jsx file

allyoucanmap (changes requested):

web/client/components/map/openlayers/swipe/SliderSwipeSupport.jsx
MuhweziDeo: Okay let me apply the change

allyoucanmap (approved):

Issue #6127, opened 2020-10-29
If the services are in HTTP, then GetCapabilities requests must be sent in HTTP


ElenaGallo:
## Description

If  a map with HTTPs URL has the services in HTTP the service requests must remain in HTTP and not in HTTPS as it happens now.
## How to reproduce

-  Open [this map](https://dev.mapstore.geo-solutions.it/mapstore/#/viewer/openlayers/5849)
-  On console open a layer request

*Expected Result*

Request is in  HTTP, as follows:

> http://encws.geo-solutions.it/geoserver/ows?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=EMSA%3AS52%20Standard&STYLES=&SRS=EPSG%3A4326&CRS=EPSG%3A4326&TILED=false&WIDTH=1085&HEIGHT=879&BBOX=35.71878433227539%2C14.203777313232422%2C36.020565032958984%2C14.576282501220703

*Current Result*

Request is in HTTPs, as follows:

> https://encws.geo-solutions.it/geoserver/ows?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=EMSA%3AS52%20Standard&STYLES=&SRS=EPSG%3A4326&CRS=EPSG%3A4326&TILED=false&WIDTH=1085&HEIGHT=879&BBOX=35.71878433227539%2C14.203777313232422%2C36.020565032958984%2C14.576282501220703

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

tdipisa:
@ElenaGallo I don't think it is possible for security reasons of mixed content.

simboss:
I understand what you mean but the MapStore behaviour is not correct.

It should prevent me from adding to the catalog something which is in HTTP while instead right now I can.
This means it sends a GetCapabilities and allows me to add layers from it which then fails.

This will confuse users even very smart ones like me ;)

tdipisa:
@simboss ok I understand your point.

tdipisa:
[...]

PR #6626, opened 2021-03-10
6127 : Ensure Http urls added in Catalog conform to Mapstore Http protocol

MuhweziDeo:
## Description

 - [x] Bug

#6127 

If a map with HTTPS URL has the services in HTTP the service requests must remain in HTTP and not in HTTPS as it happens now.
#<issue>

**What is the new behavior?**
URLs will be allowed to be input if they match the protocol of mapstore for example. if mapstore is on HTTP then only URLs with `http://` will be allowed
Screenshot 2021-03-10 at 15 58 12

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/catalog/editor/MainForm.jsx
+++ b/web/client/components/catalog/editor/MainForm.jsx
@@ -5,7 +5,7 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import React from 'react';
+import React, { useState } from 'react';
 import {get, find} from 'lodash';
 import Message from '../../I18N/Message';
 import HTML from '../../I18N/HTML';
@@ -16,7 +16,7 @@ import InfoPopover from '../../widgets/widget/InfoPopover';
 import { FormControl as FC, Form, Col, FormGroup, ControlLabel, Alert } from "react-bootstrap";
 
 import localizedProps from '../../misc/enhancers/localizedProps';
-import {defaultPlaceholder, checkIfUrlMatchesProtocol} from "./MainFormUtils";
+import {defaultPlaceholder, isHttps} from "./MainFormUtils";
 
 const FormControl = localizedProps('placeholder')(FC);
 
@@ -123,9 +123,17 @@ export default ({
     onChangeUrl,
     onChangeServiceProperty,
     onChangeType,
-    invalidProtocol = false,
-    setInvalidProtocol = () => {}
+    setValid = () => {}
 }) => {
+    const [invalidProtocol, setInvalidProtocol] = useState(false);
+    function handleProtocolValidity(url) {
+        onChangeUrl(url);
+        if (url) {
+            const isInvalidProtocol = !isHttps(url);
+            setInvalidProtocol(isInvalidProtocol);
+            setValid(!isInvalidProtocol);
+        }
+    }
     const URLEditor = service.type === "tms" ? TmsURLEditor : DefaultURLEditor;
     return (
         <Form horizontal >
@@ -151,13 +159,10 @@ export default ({
                         onChange={(e) => onChangeTitle(e.target.value)} />
                 </Col>
             </FormGroup>
-            <URLEditor key="url-row" serviceTypes={serviceTypes} service={service} onChangeUrl={(url) => {
-                onChangeUrl(url);
-                url && setInvalidProtocol(!checkIfUrlMatchesProtocol(url));
-            }} onChangeTitle={onChangeTitle} onChangeServiceProperty={onChangeServiceProperty} />
+            <URLEditor key="url-row" serviceTypes={serviceTypes} service={service} onChangeUrl={handleProtocolValidity} onChangeTitle={onChangeTitle} onChangeServiceProperty={onChangeServiceProperty} />
 
             {invalidProtocol ? <Alert bsStyle="danger">
-                <Message msgId="catalog.invalidUrlHttpProtocol" msgParams={{protocol: window.location.href.split("://")[0].toUpperCase()}}/>
+                <Message msgId="catalog.invalidUrlHttpProtocol" />
             </Alert> : null}
 
         </Form>);
--- a/web/client/components/catalog/editor/MainForm.jsx
+++ b/web/client/components/catalog/editor/MainForm.jsx
@@ -13,10 +13,10 @@ import HTML from '../../I18N/HTML';
 import {getConfigProp} from '../../../utils/ConfigUtils';
 
 import InfoPopover from '../../widgets/widget/InfoPopover';
-import { FormControl as FC, Form, Col, FormGroup, ControlLabel } from "react-bootstrap";
+import { FormControl as FC, Form, Col, FormGroup, ControlLabel, Alert } from "react-bootstrap";
 
 import localizedProps from '../../misc/enhancers/localizedProps';
-import {defaultPlaceholder} from "./MainFormUtils";
+import {defaultPlaceholder, checkIfUrlMatchesProtocol} from "./MainFormUtils";
 
 const FormControl = localizedProps('placeholder')(FC);
 
@@ -122,7 +122,9 @@ export default ({
     onChangeTitle,
     onChangeUrl,
     onChangeServiceProperty,
-    onChangeType
+    onChangeType,
+    invalidProtocol = false,
+    setInvalidProtocol = () => {}
 }) => {
     const URLEditor = service.type === "tms" ? TmsURLEditor : DefaultURLEditor;
     return (
@@ -149,6 +151,14 @@ export default ({
                         onChange={(e) => onChangeTitle(e.target.value)} />
                 </Col>
             </FormGroup>
-            <URLEditor key="url-row" serviceTypes={serviceTypes} service={service} onChangeUrl={onChangeUrl} onChangeTitle={onChangeTitle} onChangeServiceProperty={onChangeServiceProperty} />
+            <URLEditor key="url-row" serviceTypes={serviceTypes} service={service} onChangeUrl={(url) => {
+                onChangeUrl(url);
+                url && setInvalidProtocol(!checkIfUrlMatchesProtocol(url));
+            }} onChangeTitle={onChangeTitle} onChangeServiceProperty={onChangeServiceProperty} />
+
+            {invalidProtocol ? <Alert bsStyle="danger">
+                <Message msgId="catalog.invalidUrlHttpProtocol" msgParams={{protocol: window.location.href.split("://")[0].toUpperCase()}}/>
+            </Alert> : null}
+
         </Form>);
 };
--- a/web/client/components/catalog/editor/MainFormUtils.js
+++ b/web/client/components/catalog/editor/MainFormUtils.js
@@ -1,3 +1,5 @@
+import url from 'url';
+
 export const defaultPlaceholder = (service) => {
     let urlPlaceholder = {
         wfs: "e.g. https://mydomain.com/geoserver/wfs",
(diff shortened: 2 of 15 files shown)


allyoucanmap (commented):
I double checked the reason of the conversion from http to https and they seem due to the value of the content header `Referrer Policy`  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy.
Chrome 89.0.4389.82 sends by default the value `strict-origin-when-cross-origin` while Firefox 86.0 `no-referrer-when-downgrade` and they have different behaviour with the same [catalog](http://encws.geo-solutions.it/geoserver/ows) on dev.mapstore:
- Chrome request use `https`
image

- Firefox request use `http`
image

These are the default behaviours of the browser so we could not do much about these.

@tdipisa before to make a review on the code I quickly tested this PR locally and the implementation introduces an error message and prevent to add the catalog if the protocol does not match the protocol of mapstore. So now I'm not allowed to add an https catalog if my mapstore is under http so we will not be able to add an https catalog in the local enviroment with localhost with this fix.

Mixed content happens only when a HTTPS page include an HTTP content and not viceversa

https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content

> An HTTPS page that includes content fetched using cleartext HTTP is called a mixed content page.

Should we allow `http` page to include catalog with `https` protocol?

web/client/translations/data.en-US.json
```diff
@@ -1413,6 +1413,7 @@
             "showPreview": "Show preview",
             "advancedSettings": "Advanced settings",
             "templateMetadataAvailable": "Metadata available from Dublin Core format: abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "The URL doesn't match the MapStore HTTP protocol {protocol}",
```
allyoucanmap, line 1416:
```suggestion
            "invalidUrlHttpProtocol": "Thìs catalog cannot be added to the available ones because it uses an http protocol. Please provide a catalog url that uses https protocol",
```

web/client/translations/data.es-ES.json
```diff
@@ -1413,6 +1413,7 @@
             "showPreview": "Mostrar vista previa",
             "advancedSettings": "Ajustes avanzados",
             "templateMetadataAvailable": "Metadatos disponibles en formato Dublin Core: abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "La URL no coincide con el protocolo HTTP de MapStore {protocol}",
```
allyoucanmap, line 1416: Please translate this accordingly to the proposed suggestions

web/client/translations/data.de-DE.json
```diff
@@ -1413,6 +1413,7 @@
             "showPreview": "Vorschau zeigen",
             "advancedSettings": "Erweiterte Einstellungen",
             "templateMetadataAvailable": "Metadaten im Dublin Core-Format verfügbar: abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "Die URL stimmt nicht mit dem MapStore-HTTP-Protokoll überein {protocol}",
```
allyoucanmap, line 1416: Please translate this accordingly to the proposed suggestions

web/client/translations/data.it-IT.json
```diff
@@ -1413,6 +1413,7 @@
             "showPreview": "Mostra la preview",
             "advancedSettings": "Impostazioni avanzate",
             "templateMetadataAvailable": "Metadati disponibili del formato Dublin Core: abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "L'URL non corrisponde al protocollo HTTP di MapStore {protocol}",
```
allyoucanmap, line 1416:
```suggestion
            "invalidUrlHttpProtocol": "Questo catalogo non può essere aggiunto a quelli disponibili perché utilizza un protocollo http. Fornisci uno url di catalogo che utilizzi il protocollo https.",
```

web/client/translations/data.fr-FR.json
```diff
@@ -1414,6 +1414,7 @@
             "showPreview": "Afficher l'aperçu",
             "advancedSettings": "Réglages avancés",
             "templateMetadataAvailable": "Métadonnées disponibles pour le format Dublin Core : abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "L'URL ne correspond pas au protocole HTTP MapStore {protocol}",
```
allyoucanmap, line 1417: Please translate this accordingly to the proposed suggestions

web/client/translations/data.fi-FI.json
```diff
@@ -1195,6 +1195,7 @@
       "showPreview": "Näytä esikatselu",
       "advancedSettings": "Lisäasetukset",
       "templateMetadataAvailable": "Metatiedot saatavissa Dublin Core -muodossa: abstrakti, borderingBox, kontribuuttorit, luojat, kuvaus, formaatti, tunniste, viitteet, oikeudet, lähde, aihe, aika, otsikko, tyyppi, uri",
+      "invalidUrlHttpProtocol": "URL-osoite ei vastaa MapStore HTTP -protokollaa {protocol}",
```
allyoucanmap, line 1198: We usually translate only the default languages supported by mapstore DE, US, ES, FR and IT so this could be removed

web/client/translations/data.hr-HR.json
```diff
@@ -1153,6 +1153,7 @@
             "showPreview": "Prikaži pretpregled",
             "advancedSettings": "Napredne mogućnosti",
             "templateMetadataAvailable": "Metapodaci distupni iz Dublin Core formata: abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "URL se ne podudara s protokolom MapStore HTTP {protocol}",
```
allyoucanmap, line 1156: We usually translate only the default languages supported by mapstore DE, US, ES, FR and IT so this could be removed

web/client/translations/data.nl-NL.json
```diff
@@ -1413,6 +1413,7 @@
             "showPreview": "Voorbeeld weergeven",
             "advancedSettings": "Geavanceerde instellingen",
             "templateMetadataAvailable": "Metadata beschikbaar in Dublin Core-formaat: abstract, boundingBox, bijdrager, maker, beschrijving, formaat, identifier, referenties, rechten, bron, onderwerp, tijdelijk, titel, type, uri",
+            "invalidUrlHttpProtocol": "De URL komt niet overeen met het MapStore HTTP-protocol {protocol}",
```
allyoucanmap, line 1416: We usually translate only the default languages supported by mapstore DE, US, ES, FR and IT so this could be removed

web/client/translations/data.pt-PT.json
```diff
@@ -1130,6 +1130,7 @@
             "showPreview": "Show preview",
             "advancedSettings": "Advanced Settings",
             "templateMetadataAvailable": "Metadata available from Dublin Core format: abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "O URL não corresponde ao protocolo MapStore HTTP {protocol}",
```
allyoucanmap, line 1133: We usually translate only the default languages supported by mapstore DE, US, ES, FR and IT so this could be removed

web/client/translations/data.vi-VN.json
```diff
@@ -144,6 +144,7 @@
             "noRecordsMatched": "Không có bản ghi phù hợp",
             "noResultsText": "Không có kết quả",
             "notAvailable": "Không có sẵn",
+            "invalidUrlHttpProtocol": "URL không khớp với giao thức HTTP của MapStore {protocol}",
```
allyoucanmap, line 147: We usually translate only the default languages supported by mapstore DE, US, ES, FR and IT so this could be removed

web/client/translations/data.zh-ZH.json
```diff
@@ -1107,6 +1107,7 @@
             "showPreview": "Show preview",
             "advancedSettings": "Advanced Settings",
             "templateMetadataAvailable": "Metadata available from Dublin Core format: abstract, boundingBox, contributor, creator, description, format, identifier, references, rights, source, subject, temporal, title, type, uri",
+            "invalidUrlHttpProtocol": "The URL doesn't match the MapStore HTTP protocol {protocol}",
```
allyoucanmap, line 1110: We usually translate only the default languages supported by mapstore DE, US, ES, FR and IT so this could be removed

web/client/components/catalog/editor/MainFormUtils.js
```diff
@@ -13,3 +13,11 @@ export const defaultPlaceholder = (service) => {
     }
     return true;
 };
+
+export const checkIfUrlMatchesProtocol = (url = '') => {
```
allyoucanmap, line 17: This check needs to match only for the mapstore instances with https.

The expected behaviours is:
- mapstore with `http:` protocol should allow catalog with `http:` or `https:`
- mapstore with `https:` protocol should allow only catalog with `https:`

I would suggets to use an url parser instead of indexOf, usually we use the 'url' library so it could be implemented like this:

```js
import url from 'url';
// ...

export const checkIfUrlMatchesProtocol = (catalogUrl = '') => {
    const { protocol: mapStoreProtocol } = url.parse(window.location.href);
    const { protocol: catalogProtocol } = url.parse(catalogUrl);
    if (mapStoreProtocol === 'https:') {
        return mapStoreProtocol === catalogProtocol;
    }
    return true;
};
```
Please if you are using this code check if it's working as expected because I did not test it properly (maybe also the name of the function needs to change due to the new behaviour)

web/client/components/catalog/editor/MainForm.jsx
```diff
@@ -149,6 +151,14 @@ export default ({
...
                 </Col>
             </FormGroup>
-            <URLEditor key="url-row" serviceTypes={serviceTypes} service={service} onChangeUrl={onChangeUrl} onChangeTitle={onChangeTitle} onChangeServiceProperty={onChangeServiceProperty} />
+            <URLEditor key="url-row" serviceTypes={serviceTypes} service={service} onChangeUrl={(url) => {
+                onChangeUrl(url);
+                url && setInvalidProtocol(!checkIfUrlMatchesProtocol(url));
+            }} onChangeTitle={onChangeTitle} onChangeServiceProperty={onChangeServiceProperty} />
+
+            {invalidProtocol ? <Alert bsStyle="danger">
+                <Message msgId="catalog.invalidUrlHttpProtocol" msgParams={{protocol: window.location.href.split("://")[0].toUpperCase()}}/>
```
allyoucanmap, line 160: I don't think we need the msgParams anymore

allyoucanmap (changes requested):

MuhweziDeo:
Thanks, @allyoucanmap for the review. I will quickly start implementing the feedback

web/client/components/catalog/editor/MainFormUtils.js
MuhweziDeo: this Works @allyoucanmap

allyoucanmap (approved):

tdipisa:
@ElenaGallo please test this in DEV and let us know if we can proceed with the backport to 2021.01.xx. Thank you.

ElenaGallo:
@MuhweziDeo test done, please bakport this PR to 2021.01.xx. Thank you.

Issue #5138, opened 2020-04-10
Exported measurements issues with import/styiling


offtherailz:
## Description
If you export measures and re-import, the styler is shown but the styles are not applied.

## How to reproduce

- Open the measurement tool and create a measure
- Export as GeoJSON
- Close measure tool 
- Open Import tool
- Import the just exported json
- Apply the style

*Expected Result*
The style is applied (or the styler is not shown, if it was exported as annotation
*Current Result*
The style is not applied.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

offtherailz:
Just as reference (missing the stage of this solution) was implemented in #5399

MuhweziDeo:
Hi @MV88 I just want to confirm if to check if one of the `features.style` has a value. Incase of an array the length should be greater than 1 and incase of an object it should have atleast one item. Does the the `feature.type` matter in any way.

MuhweziDeo:
Hi, @MV88 @tdipisa @offtherailz. does the design below work. If not do have any recommendations.

![Screenshot 2021-03-17 at 16.10.26.png](https://images.zenhubusercontent.com/60350ba854742f543a9fa5ff/05c12226-3db1-40da-8a12-301c7ffcc360)

MV88:
> Hi @MV88 I just want to confirm if to check if one of the `feature.style` has a value. Incase of` feature.style` of an array the length should be greater than 1 and in the case of an object, it should have at least one item. Does the `feature. type` matter in any way?

by has a value it means  
- feature.style.length > 0 if is an array
- !lodash.isEmpty(feature.style) if is not an array

for the message please change it to "This layer has already a style defined and is not possible to customize it"

PR #6663, opened 2021-03-17
5138: Exported measurements issues with import/styling #5138

MuhweziDeo:
## Description

 - [x] Others Describe. Enhancement

## Issue

The style is not applied.
#5138 

**What is the new behavior?**
If the imported map has inline styles defined. Check and Disable Default Style checkbox and notify the user with a message.
Screenshot 2021-03-17 at 16 10 26

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/import/style/StylePanel.jsx
+++ b/web/client/components/import/style/StylePanel.jsx
@@ -67,11 +67,13 @@ class StylePanel extends React.Component {
         useDefaultStyle: false,
         zoomOnShapefiles: true,
         overrideAnnotation: false,
-        initialLayers: []
+        initialLayers: [],
+        disableStyleCustomization: false
     };
 
     componentDidMount() {
         this.setState({initialLayers: [...this.props.layers]});
+        this.checkAndDisableStyleCustomization();
     }
 
     getGeometryType = (geometry) => {
@@ -138,6 +140,17 @@ class StylePanel extends React.Component {
         </Row>);
     };
 
+    renderDisableStyleCustomization = () => {
+        if (this.state.disableStyleCustomization) {
+            return (
+                <div className="alert alert-info mb-2 style-customisation-disabled-container">
+                    <Message msgId="shapefile.styleCustomizationDisabled"/>
+                </div>
+            );
+        }
+        return null;
+    }
+
     render() {
         return (
             <Grid role="body" style={{width: "400px"}} fluid>
@@ -151,13 +164,14 @@ class StylePanel extends React.Component {
                 <Row key="styler" style={{marginBottom: 10}}>
                     {this.state.useDefaultStyle ? null : this.props.stylers[this.getGeomType(this.props.selected)]}
                 </Row>
+                {this.renderDisableStyleCustomization()}
                 <Row key="options">
                     {isAnnotation(this.props.selected) ?
                         this.annotationOptions()
                         :
                         <>
                             <Col xs={2}>
-                                <input aria-label="..." type="checkbox" defaultChecked={this.state.useDefaultStyle} onChange={(e) => { this.setState({ useDefaultStyle: e.target.checked }); }} />
+                                <input aria-label="..." type="checkbox" disabled={this.state.disableStyleCustomization} checked={this.state.useDefaultStyle} defaultChecked={this.state.useDefaultStyle} onChange={(e) => { this.setState({ useDefaultStyle: e.target.checked }); }} />
                             </Col>
                             <Col style={{ paddingLeft: 0, paddingTop: 1 }} xs={10}>
                                 <label><Message msgId="shapefile.defaultStyle" /></label>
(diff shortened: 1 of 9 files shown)


MV88 (changes requested):
@MuhweziDeo hi, there is a problem when you try to import two different layers and the first does not have the style defined and the second does.

you are always checking the first layer, instead you should change the check and evaluate the selected one, my suggestion is to see if the selected prop is what you need 
image

web/client/components/import/style/StylePanel.jsx
```diff
@@ -67,11 +67,19 @@ class StylePanel extends React.Component {
...
         zoomOnShapefiles: true,
         overrideAnnotation: false,
-        initialLayers: []
+        initialLayers: [],
+        disableStyleCustomization: false
     };
 
     componentDidMount() {
         this.setState({initialLayers: [...this.props.layers]});
+        this.setState({currentActiveLayer: this.props.selected});
```
MV88, line 76: you are not using this
```suggestion
```

web/client/components/import/style/StylePanel.jsx
```diff
@@ -239,6 +258,31 @@ class StylePanel extends React.Component {
             </Col>
         </>
     );
+
+    checkAndDisableStyleCustomization = () => {
+        this.setState({disableStyleCustomization: false, useDefaultStyle: false});
+        const layer =  this.props.selected;
+        if (layer) {
+            if (layer.features.length) {
```
MV88, line 266: you can remove one if in this case
```suggestion
        if (layer && layer.features.length) {
```

web/client/components/import/style/StylePanel.jsx
```diff
@@ -239,6 +258,31 @@ class StylePanel extends React.Component {
...
+                        } else {
+                            if (Object.keys(feature.style).length) {
+                                this.setState({disableStyleCustomization: true, useDefaultStyle: true});
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
MV88, line 284:
```suggestion
```

web/client/components/import/style/StylePanel.jsx
```diff
@@ -67,11 +67,19 @@ class StylePanel extends React.Component {
...
+        initialLayers: [],
+        disableStyleCustomization: false
     };
 
     componentDidMount() {
         this.setState({initialLayers: [...this.props.layers]});
+        this.setState({currentActiveLayer: this.props.selected});
+        this.checkAndDisableStyleCustomization();
+    }
+    componentDidUpdate(prevProps) {
```
MV88, line 79: i think there is no need for this

web/client/components/import/style/StylePanel.jsx
```diff
@@ -138,6 +146,17 @@ class StylePanel extends React.Component {
...
+    renderDisableStyleCustomization = () => {
+        if (this.state.disableStyleCustomization) {
+            return (
+                <div className="alert alert-info mb-2 style-customisation-disabled-container">
+                    <Message msgId="shapefile.styleCustomizationDisabled"/>
+                </div>
+            );
+        }
+        return null;
+    }
```
MV88, line 158:
```suggestion
    renderDisableStyleCustomization = (hasCustomStyle) => {
            return hasCustomStyle ? (
                <div className="alert alert-info mb-2 style-customisation-disabled-container">
                    <Message msgId="shapefile.styleCustomizationDisabled"/>
                </div>
            );
        } : null;
    }
```

MV88 (changes requested):
i think that this internal state is superflous
you can run checkAndDisableStyleCustomization in the render method
like

`const hasCustomStyle = this.checkAndDisableStyleCustomization(this.props.selected);`
and then use it where it needs to be used

in this way you can avoid using componentDidUpdate 
can you please refactor this and introduce this new flag

i have added some changes, but check for other when doing the refactor
thanks

web/client/components/import/style/StylePanel.jsx
MuhweziDeo: @MV88 I need this trigger the `this.checkAndDisableStyleCustomization()` . How would you recommend I do it because without it. I cant get the check to run again

web/client/components/import/style/StylePanel.jsx
MuhweziDeo: Sorry left this by mistake

web/client/components/import/style/StylePanel.jsx
MuhweziDeo: Ohh I get it now. Just need to call `renderDisableStyleCustomization()` and it will always get the latest props.selected value. Thanks @MV88

web/client/components/import/style/StylePanel.jsx
MuhweziDeo: Hey @MV88  How do I get the has `hasCustomStyle ` value.

MV88 (approved):

Issue #6723, opened 2021-04-01
Error creating circle annotation


tdipisa:
## Description

MapStore crashes when entering coordinates for the creation of a circle (decimal and aeronautical), if no radius is set yet. It should be possible to enter coordinates before entering a radius without throwing an error.

## How to reproduce

Tested on https://qa.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10358

-  Enter LAT first and then LON (or vice versa)

![annotation_circle_issue.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/701b1a7c-c3b7-46fe-a8c3-dc76b349081e)

*Expected Result*

 It should be possible to enter coordinates before entering a radius without throwing an error.

*Current Result*

MapStore craches before having the chance to enter a readius.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #6737, opened 2021-04-02
6723: Error creating circle annotation

MuhweziDeo:
## Description

 - [x] Bugfix

## 6723

MapStore crashes before having the chance to enter a radius while entering Lat and Lng

#6723

**What is the new behavior?**
It should be possible to enter coordinates before entering a radius without throwing an error.

#### how the issue was resolved
I found that the package https://www.npmjs.com/package/@turf/turf causes the app to crash at this line https://github.com/geosolutions-it/MapStore2/blob/master/web/client/reducers/annotations.js#L158. It's my guess that `circle` needs the radius. Without radius it causes the app to crash. So wrapping it a `try-catch solves the problem

## Breaking change
 - [x] No

### DIFF
--- a/web/client/reducers/annotations.js
+++ b/web/client/reducers/annotations.js
@@ -155,11 +155,16 @@ function annotations(state = {validationErrors: {}}, action) {
             if (validateCoordsArray(selected.properties.center)) {
                 center = selected.properties.center;
                 // turf/circle by default use km unit hence we divide by 1000 the radius(in meters)
-                c = circle(
-                    center,
-                    action.crs === "EPSG:4326" ? action.radius : action.radius / 1000,
-                    { steps: 100, units: action.crs === "EPSG:4326" ? "degrees" : "kilometers" }
-                ).geometry;
+                // this try catch prevents the app from crashing when there is no radius maybe we can use a default value for radius incase action.radius === undefined
+                try {
+                    c = circle(
+                        center,
+                        action.crs === "EPSG:4326" ? action.radius : action.radius / 1000,
+                        { steps: 100, units: action.crs === "EPSG:4326" ? "degrees" : "kilometers" }
+                    ).geometry;
+                } catch (e) {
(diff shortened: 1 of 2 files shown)


MV88 (changes requested):
Hi @MuhweziDeo 

is there a way to avoid the problem instead doing this try catch just to prevent the app from crashing. also in the catch you are not doing anything and the console is not helping the user.

can we use a very little value for radius if not defined so even it it is drawn then it is not visible
please make a few tries and let me know

i would

MuhweziDeo:
> Hi @MuhweziDeo
> 
> is there a way to avoid the problem instead doing this try catch just to prevent the app from crashing. also in the catch you are not doing anything and the console is not helping the user.
> 
> can we use a very little value for radius if not defined so even it it is drawn then it is not visible
> please make a few tries and let me know
> 
> i would

Okay let try that thanks

web/client/reducers/annotations.js
```diff
@@ -155,9 +155,11 @@ function annotations(state = {validationErrors: {}}, action) {
             if (validateCoordsArray(selected.properties.center)) {
                 center = selected.properties.center;
                 // turf/circle by default use km unit hence we divide by 1000 the radius(in meters)
+                // this try catch prevents the app from crashing when there is no radius maybe we can use a default value for radius incase action.radius === undefined
```
MV88, line 158:
```suggestion

```

MV88 (changes requested):
can you add a unit test?

web/client/reducers/annotations.js
MuhweziDeo: okay

MV88 (approved):

MV88:
@ElenaGallo can you please test this in DEV?

ElenaGallo:
@MuhweziDeo test done, please backport this PR to the stable branch. Thank you

Issue #6727, opened 2021-04-01
Infobox issue DB side?


tdipisa:
## Description

We have the following user feedback:

>Every change, not only in the [infobox](https://mapstore.readthedocs.io/en/latest/user-guide/resources-properties/#details) itself, but also when simply opening the settings window and closing it by pressing “save” or storing a change in the map, will store a new separate entry for the info box in the database, while old entries are not deleted. 
This includes at least changes in tables gs_stored_data, gs_category, gs_resource, gs_security. Therefore a lot of data junk is created in the database.
We would expect only one database entry for the info box per map and that changes are made by updating this one entry.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

[This map](https://qa.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10358) already contains Details configured. However, It would be good probably to reproduce the issue, is it really exists, in a local clean DB.

dsuren1:
When checked locally with clean DB, on update, an additional entry to existing resource is created in gs_stored_data, gs_resource, gs_security every time when user tries to save map with Infobox/Details present.

![2021-04-02 20_22_34-MapStore HomePage](https://user-images.githubusercontent.com/26929983/113426488-2c9fbe80-93f1-11eb-99f6-9da303bbdae6.jpg)

![2021-04-02 20_22_15-pgAdmin 4](https://user-images.githubusercontent.com/26929983/113426493-2e698200-93f1-11eb-99da-a58e0b24551e.jpg)

![2021-04-02 20_24_17-pgAdmin 4](https://user-images.githubusercontent.com/26929983/113426613-56f17c00-93f1-11eb-97a6-5054b74f9212.jpg)


We need to restrict creation of resource for Detail/Infobox when resource is updated.

PR #6742, opened 2021-04-06
#6727: Details update fix

dsuren1:
## Description
This PR adds commit to fix, the duplication resource when updating the existing resource.

 - [x] Bugfix

## Issue

#6727 

**What is the new behavior?**
When updating the details resource, the specific resource will only get updated and no duplicate creation of the same resource.
**i.e** Only one entry specific to the resource will be available in the DB (tables - gs_stored_data, gs_resource, gs_security) 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/observables/__tests__/geostore-test.js
+++ b/web/client/observables/__tests__/geostore-test.js
@@ -211,5 +211,33 @@ describe('geostore observables for resources management', () => {
                 e => done(e)
             );
         });
+        it('updateResource linked resource is updated', done => {
+            const testResource = {
+                id: 10,
+                data: {},
+                category: "TEST",
+                metadata: {
+                    name: "RES2"
+                },
+                linkedResources: {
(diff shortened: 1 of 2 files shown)


MV88 (approved):
LGTM 

nice catch

tdipisa:
@ElenaGallo please test (maybe locally? @your though @offtherailz) and let us know if we can backport.

offtherailz:
Looking at the code, it looks that the update flow was stopped because of wrong method call (in case detail was existing). If an error occours during update process, MapStore deduces that the linked resource was previously deleted (or permissions somehow not syncronized correctly) and so creates a new one. This should happen only as a b-plan in case of some inconsistence with previous data, but because of this error in code, it was happening every time a detail card was updated.

In order to test, you need to see the URL of the original resource and check that is updated after the second save.
image

ElenaGallo:
@dsuren1 test done, please backport this Pr to the stable branch. Thank you

Issue #6724, opened 2021-04-01
The first GFI result visualized in Identify panel is not the expected one


tdipisa:
## Description

We agreed on the solution, that the first element that is received will be selected initially. This is the case so far. Nevertheless as soon as all elements are loaded, the first item in the drop down list is automatically selected, leading to a sudden change in the info panel, in case the first element in list wasn’t the first element to be loaded. We would expect that the first loaded feature stays selected, until the user changes it.

## How to reproduce

Tested on https://qa.mapstore2.geo-solutions.it/mapstore/#/viewer/openlayers/10358

![gfi_order_issue](https://user-images.githubusercontent.com/1280027/113276086-c6774680-92df-11eb-884e-bf022e31e4a5.gif)

*Expected Result*

We would expect that the first loaded feature stays selected, until the user changes it.

*Current Result*

The first item in the drop down list is automatically selected even if it is not related to the first GFI respose loaded.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MuhweziDeo:
Hey @tdipisa. Just want to confirm that the `first loaded feature` is determined the time it takes to complete the request as seen in the network tab

tdipisa:
> Hey @tdipisa. Just want to confirm that the `first loaded feature` is determined by the time it takes to complete the request as seen in the network tab from the demo link shared above

@MuhweziDeo, yes. when the user performs an identify request clickinf on the map, MS performs N requests one for each layer visible in map: the first response MS receives for one of these layers should determines the selection of the corresponding item in DD, but currently always the first one in the list is slected at the end.

PR #6744, opened 2021-04-07
6724: Fix Select First GFI response in Dropdown

MuhweziDeo:
## Description

 - [x] Bugfix

## Issue

The first item in the drop-down list is automatically selected even if it is not related to the first GFI response loaded.

#6724 

**What is the new behavior?**
The First loaded feature is selected

## Breaking change
 - [x] No

### DIFF
--- a/web/client/reducers/mapInfo.js
+++ b/web/client/reducers/mapInfo.js
@@ -59,9 +59,9 @@ function receiveResponse(state, action, type) {
         }
 
         // Handle data and vector responses
-        const {requests} = state;
+        const {configuration: config, requests} = state;
         let responses = state.responses || [];
-
+        const isHover = (config?.trigger === "hover"); // Display info trigger
         if (!isVector) {
             const updateResponse = {
                 response: action[type],
@@ -69,14 +69,20 @@ function receiveResponse(state, action, type) {
                 layerMetadata: action.layerMetadata,
                 layer: action.layer
             };
-            responses = [...responses, updateResponse];
+            if (isHover) {
+                // Add response upon it is received
+                responses = [...responses, updateResponse];
+            } else {
+                // Add response in same order it was requested
+                responses[requestIndex] = updateResponse;
+            }
         }
         let indexObj = {loaded: true, index: 0};
         // Set responses and index as first response is received
         return assign({}, state, {
             ...(isVector && {requests}),
             ...(!isUndefined(indexObj) && indexObj),
-            responses: [...responses.filter((res) => res)]}
+            responses: [...responses]}
         );
     }
     return state;
--- a/web/client/reducers/mapInfo.js
+++ b/web/client/reducers/mapInfo.js
@@ -85,6 +85,7 @@ function receiveResponse(state, action, type) {
         let firstLayerId = state.firstLayerId;
         let indexObj = {loaded: true, index: 0};
         if (!isHover) {
+            // pull out the recievedResponses
             const recievedResponses = responses.filter((res) => res);
             if (recievedResponses.length && !firstLayerId) {
                 const format = state.requests[requestIndex]?.request?.info_format;
--- a/web/client/reducers/mapInfo.js
+++ b/web/client/reducers/mapInfo.js
@@ -104,6 +104,7 @@ function receiveResponse(state, action, type) {
                     const correctIndex = findIndex(filteredResponses, (filteredResp) => {
                         return filteredResp?.layerMetadata?.title === firstLayerId;
                     });
+                    console.log({format, filteredResponses, correctIndex, responses}, "==================>");
                     indexObj = {...indexObj, index: correctIndex !== -1 ? correctIndex : indexObj.index};
                 }
             }
--- a/web/client/reducers/mapInfo.js
+++ b/web/client/reducers/mapInfo.js
@@ -83,7 +83,7 @@ function receiveResponse(state, action, type) {
             }
         }
         let firstLayerId = state.firstLayerId;
-        let indexObj;
+        let indexObj = {loaded: true, index: 0};
         const allResponsesLoaded = responses.filter(r => !isUndefined(r)).length === requests.length;
         if (!isHover) {
             // pull out the recievedResponses
--- a/web/client/reducers/mapInfo.js
+++ b/web/client/reducers/mapInfo.js
@@ -37,28 +37,11 @@ import {
 import { MAP_CONFIG_LOADED } from '../actions/config';
 import { RESET_CONTROLS } from '../actions/controls';
 import assign from 'object-assign';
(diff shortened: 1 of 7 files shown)


tdipisa:
@dsuren1 Since you have developed the Identify tool updates for c125, I would like to have also your feedback on this issue.

dsuren1:
I did see only one issue. The loader/spinner is not displayed in the GFI panel between user requests now
spinner_missing
Try with this [map](https://dev.mapstore2.geo-solutions.it/mapstore/?debug=true#/viewer/openlayers/10358) And query only **states** as shown above

dsuren1 (changes requested):
The loader/spinner is not displayed in the GFI panel between user requests now
spinner_missing
Try with this [map](https://dev.mapstore2.geo-solutions.it/mapstore/?debug=true#/viewer/openlayers/10358) And query only **states** as shown above

web/client/reducers/mapInfo.js
```diff
@@ -72,21 +69,14 @@ function receiveResponse(state, action, type) {
...
                 layer: action.layer
             };
-            if (isHover) {
-                // Add response upon it is received
-                responses = [...responses, updateResponse];
-            } else {
-                // Add response in same order it was requested
-                responses[requestIndex] = updateResponse;
-            }
+            responses = [...responses, updateResponse];
```
MV88, line 72: this is where the responses loses the order in the TOC

MV88 (changes requested):
the first result selected appear to be ordered based on the first response received.
but all the responses are not matching the TOC order. @tdipisa is this somehow accepted or maybe a further refinement to the PR is needed? Was there a particular reason to have the responses in the order of the requests were sent?
image





@dsuren1 the problem you noticed is already present in current dev and can be a separate issue. @tdipisa do you agree?
i noticed that the loading spinner start appearing after a layer is zoomed to its extent

tdipisa:
@dsuren1 
>The loader/spinner is not displayed in the GFI panel between user requests now
I think this is a regression but a different issue that is affecting the current DEV and QA (so not directly related to this PR). Please open a separate issue for this and assign it to 2021.01.02 milestone.

@MV88 the order of items in the DD must respect the order of layers in TOC. This is the original requirement for the Identify tool  and how it is working now in DEV (this must be maintained).

MV88 (changes requested):
the check of length here is always true, did you considered this? i believe that this needs to be done when all all responses have arrived hence:
`if (responses.filter(v => v).length === requests.length) {`

image


also can you fix this graphical problem?
image


i have user an [url throttler](https://chrome.google.com/webstore/detail/url-throttler/kpkeghonflnkockcnaegmphgdldfnden/related?hl=en) to delay requests coming from poi and ny_roads layer so that States is the first valid response that comes
there is also an update of the viewer when other responses comes,
this is not a big problem, but can you see why it re-renders like this

https://user-images.githubusercontent.com/11991428/114890348-3838af80-9e0b-11eb-9abc-68b83beb47a9.mp4

MuhweziDeo:
> the check of length here is always true, did you considered this? i believe that this needs to be done when all all responses have arrived hence:
> `if (responses.filter(v => v).length === requests.length) {`
> 
> image
> 
> also can you fix this graphical problem?
> image
> 
> i have user an [url throttler](https://chrome.google.com/webstore/detail/url-throttler/kpkeghonflnkockcnaegmphgdldfnden/related?hl=en) to delay requests coming from poi and ny_roads layer so that States is the first valid response that comes
> there is also an update of the viewer when other responses comes,
> this is not a big problem, but can you see why it re-renders like this
> 
>  6724.first.gfi.mp4

Thanks the graphical problem is being fixed here https://github.com/geosolutions-it/MapStore2/pull/6775

dsuren1:
@MuhweziDeo 
The menu auto switches the selection, as first received response is trying to catch up it's index position when other responses are received.  This also happens when there is vector responses (Measurement & Annotations). 
**Layers**
auto_menu_switching_4
**Vector**
auto_menu_switching

dsuren1 (changes requested):
@MuhweziDeo 
The menu auto switches the selection, as first received response is trying to catch up it's index position when other responses are received.  This also happens when there is vector responses (Measurement & Annotations). 
**Layers**
auto_menu_switching_4
**Vector**
auto_menu_switching

web/client/reducers/mapInfo.js
```diff
@@ -80,10 +82,38 @@ function receiveResponse(state, action, type) {
...
+            if (recievedResponses.length && !firstLayerId) {
+                const format = state.requests[requestIndex]?.request?.info_format;
+                const layer = responses[requestIndex];
+                const validator = getValidator(format);
+                const valid = validator?.getValidResponses(recievedResponses);
+                if (valid.length) {
+                    firstLayerId = layer?.layerMetadata?.title || "";
+                }
+            }
+            if (allResponsesLoaded) {
```
dsuren1, line 100: As per this implementation, it waits for all response to complete before showing the result in GFI panel. I believe the requirement was to show the first response as soon as it is received, because some servers might take time to respond. 
Your thoughts @tdipisa @MV88

dsuren1 (changes requested):
It fixes the above issues. However it contradicts with the pervious requirement of GFI

MuhweziDeo:
> @MuhweziDeo
> It fixes the above issues. However it contradicts with the pervious requirement of GFI if I'm not wrong. Let's wait for the feedback

Okay thats the drawback I so since need allResponses to properly get the right index.

web/client/reducers/mapInfo.js
tdipisa: @dsuren1 yes, the first element for which we receive a response must be selected. This was the original requirement @dsuren1 you should know about that. It is also reported in the description of issue connected to this PR.
Is there a blocker with this? @dsuren1 @MuhweziDeo

MuhweziDeo: In  a way it is a blocker because currently unable to prevent `the menu from auto switches the selection`. as reported by @dsuren1 because I have to wait for all the responses. The first response received will be selected by the loader will not stop and the element won't be selected until all responses are received.

dsuren1:
@MuhweziDeo 
As per the discussion,

- Kindly use these as reference and maintain the [original requirement](https://github.com/geosolutions-it/MapStore2/issues/5465) with fixes (highlight features, etc..) that came later on with respect to GFI panel.  Use this [branch ](https://github.com/dsuren1/MapStore2/tree/6275_gfi_layer) as reference for the behavior (For layer selector and the viewer panel). **_Note:_** This branch contains a bug for highlight which was later fixed as part of https://github.com/geosolutions-it/MapStore2/issues/6365

As first step, you could try fixing the glitching of the layer selector and viewer while maintaining other part of the GFI features intact.

Issue #6728, opened 2021-04-01
Infobox issue with image content left/right aligned


tdipisa:
## Description

We have the following user feedback:

>When using left/right-align for an image it is not possible to position a text below the image, although the WYSIWYG editor shows it. 
It will always be positioned next to it. We would expect the possibility to position text below an image, independent of the images position.

## How to reproduce

![infobox_issue.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/c50bd143-90fc-4ca4-a8cf-823cfa12a57e)

*Expected Result*

We would expect the possibility to position text below an image, independent of the images position.

*Current Result*

 It is not possible to position a text below the image, although the WYSIWYG editor shows it.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

dsuren1:
_Note_: 
The issue happens because of empty tags created when user manually pushes the content by hitting '**Enter**'. Fix is required on handling the same.

PR #6745, opened 2021-04-07
#6728: Infobox viewer alignment fix

dsuren1:
## Description
This PR add commit to fix the detail viewer in handling the empty tag created from the editor which affects alignment (when manually pushing content by hitting "Enter") 

 - [x] Bugfix

## Issue

#6728 

**What is the new behavior?**
The empty tags will have enough spacing such that it doesn't break the alignment and viewer will match the representation in the editor tool

## Breaking change
 - [x] No

## Other useful information
Map ref: https://dev.mapstore2.geo-solutions.it/mapstore/?debug=true#/viewer/openlayers/10358

### DIFF
--- a/web/client/themes/default/less/details.less
+++ b/web/client/themes/default/less/details.less
@@ -23,6 +23,9 @@
 
 }
 
+// Workaround: This emulates the br tags inside p or h tags in the editor, when pressing enter,
+// that are not included in the output of draftToHtml(DraftJS converter)
+// see https://github.com/geosolutions-it/MapStore2/issues/6728
 .ms-details-preview {
     p{
         margin: 1em;
--- a/web/client/themes/default/less/details.less
+++ b/web/client/themes/default/less/details.less
@@ -23,6 +23,18 @@
 
 }
 
+.ms-details-preview {
+    p{
+        margin: 1em;
+    }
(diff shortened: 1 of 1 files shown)


web/client/themes/default/less/details.less
```diff
@@ -23,6 +23,18 @@
 
 }
 
+.ms-details-preview {
```
offtherailz, line 26:
```suggestion
// this emulates the br tags inside p or h tags in the editor, when pressing enter, that are not included in the final html by DraftJS converter, see #6728
.ms-details-preview {
```

offtherailz (commented):
I see the problem is that while the editor inserts <br /> tags during editing, this convertion do not happen in the final convertion.

I think is a bug of https://github.com/jpuri/draftjs-to-html, isn't it?
If so, I added a comment to explain the workaround

dsuren1:
> I see the problem is that while the editor inserts
> tags during editing, this convertion do not happen in the final convertion.
> 
> I think is a bug of https://github.com/jpuri/draftjs-to-html, isn't it?
> If so, I added a comment to explain the workaround

Yes, you are absolutely right. 
Let me add that comment to as suggested above and update the PR accordingly then

offtherailz (approved):

PR #6746, opened 2021-04-07
fix SEARCH_CANCEL_ITEM error when search state is null

MV88:
## Description

porting in master branch a fix made for https://github.com/geosolutions-it/webmapper-halliburton/issues/1643
in https://github.com/geosolutions-it/MapStore2/tree/c039_identify_error


mbarto (changes requested):
I guess adding a small unit test should not be difficult

mbarto (approved):

Issue #6730, opened 2021-04-01
Share by bounding box shows wrong result ?


tdipisa:
## Description

We have the following user feedback.

>Share by bounding box shows wrong result. Attached you find a video and images showing the issue. In this case sharing the link: https://austrocontrol.geo-solutions.it/austrocontrol-ms2/#/viewer/openlayers/91?bbox=-1.2365477937083256,33.82441087264053,29.18639649513955,45.6371696422341
which seems to use the correct bbox for the shown map, results in opening a map with different extent, although using the same browser window size (switching to a different browser has the same result). 
When creating a new sharing link from this map, the following link is created: https://austrocontrol.geo-solutions.it/austrocontrol-ms2/#/viewer/openlayers/91?bbox=6.369188278503656,36.69457516832796,21.580660422927593,42.63895027595234
which again opens a map at a **higher zoomlevel**.

>The effect on our map is even higher. Sharing this link: https://maps.austrocontrol.at/mapstore/#/viewer/openlayers/121?bbox=7.087595302292528,45.56075722169766,19.256773017831676,49.72440265750633
which should show whole Austria results in a map only showing parts of the district of Salisbury.

**BBOX Shared**
![bbox_shared.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/22d2f4ec-5eba-43ae-80e1-6b73fad81b83)

**BBOX Result in shared map**
![bbox_result.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/961312de-a18d-4746-879b-18a9bcb4537e)

## How to reproduce

https://images.zenhubusercontent.com/39382076/1d061152-bcf4-4fe7-bff4-782a6a32e45e/bounding_box.mp4

*Expected Result*

The zoom level should be the same in the shared map.

*Current Result*

It seems a different zoom level is choosen by MS/OL for the provided BBOX

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

[...]

PR #6759, opened 2021-04-09
6730: Share by bounding box shows wrong result

MuhweziDeo:
## Description
Share by bounding box shows the wrong result which seems to use the correct box for the shown map, results in opening a map to a different extent, although using the same browser window size (switching to a different browser has the same result).
When creating a new sharing link from this map, the link is created opens a map at a higher zoom level.

- [ ] Tests for the changes have been added (for bug fixes/features)

 - [x] Bugfix

## Issue

It seems a different zoom level is chosen by MS/OL for the provided BBOX

#6730 

**What is the new behavior?**
The zoom level is almost similar but always less by 1 than the original

## Breaking change
 - [x] No

### DIFF
--- a/web/client/actions/map.js
+++ b/web/client/actions/map.js
@@ -126,13 +126,15 @@ export function panTo(center) {
  * @param {number[]} extent in the form of [minx, miny, maxx, maxy]
  * @param {string} crs related the extent
  * @param {number} maxZoom the max zoom limit
-*/
-export function zoomToExtent(extent, crs, maxZoom) {
+ * @param {object} options additional options {nearest: true}
+ */
+export function zoomToExtent(extent, crs, maxZoom, options) {
     return {
         type: ZOOM_TO_EXTENT,
         extent,
         crs,
-        maxZoom
+        maxZoom,
+        options
     };
 }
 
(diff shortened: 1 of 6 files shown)


web/client/actions/map.js
```diff
@@ -126,13 +126,15 @@ export function panTo(center) {
  * @param {number[]} extent in the form of [minx, miny, maxx, maxy]
  * @param {string} crs related the extent
  * @param {number} maxZoom the max zoom limit
-*/
-export function zoomToExtent(extent, crs, maxZoom) {
+ * @param {object} options additional options {nearest: true}
```
offtherailz, line 129:
```suggestion
 * @param {object} options additional options `{nearest: true}` is the only supported (See {@link https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#fit| Openlayers View, fit method} )
```

offtherailz (approved):

tdipisa:
@ElenaGallo can you please check this in DEV and let us know if we can backport to 2021.01.xx?

ElenaGallo:
@MuhweziDeo test done, please bckport this PR to the stable branch. Thank you

Issue #6813, opened 2021-04-28
Add filter to CSW Catalog to request only layers


allyoucanmap:
## Description

The CSW requests include all the resources available so the Catalog cards will show an error when the record item is not a layer. An example is the geonode mapstore client that using the CSW catalog of GeoNode shows all other contents such as maps, documents, ... with an error

To avoid listing these additional resources inside the MapStore client catalog a filter must be always included:

```xml
<csw:Constraint version="1.1.0">
  <Filter xmlns="http://www.opengis.net/ogc">
	 <Or>
	 <PropertyIsEqualTo>
		 <PropertyName>dc:type</PropertyName>
		 <Literal>dataset</Literal>
	 </PropertyIsEqualTo>
	 <PropertyIsEqualTo>
		 <PropertyName>dc:type</PropertyName>
		 <Literal>http://purl.org/dc/dcmitype/Dataset</Literal>
	 </PropertyIsEqualTo>
	</Or>
  </Filter>
</csw:Constraint>
```

Below screenshot of the current Catalog showing resources different from layers.

![image](https://user-images.githubusercontent.com/19175505/116238474-aa8d8600-a761-11eb-8428-42e545171394.png)

The CSW catalog must continue working on: 
- GeoNode (https://master.demo.geonode.org/catalogue/csw?service=CSW&version=2.0.2&request=)
- Geoserver (https://gs-stable.geo-solutions.it/geoserver/csw?service=CSW&version=2.0.2&request=)
- GeoNetwork (http://gn3-rndt.geo-solutions.it/geonetwork/srv/eng/csw)

Showing only datasets, not services.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

offtherailz:
One note: http://gn3-rndt.geo-solutions.it/geonetwork/srv/eng/csw doesn't seems to be compatible, so you can skip this test.

allyoucanmap:
please @MuhweziDeo backport this fix to https://github.com/geosolutions-it/MapStore2/tree/geonode-mapstore-client-2.1.x

offtherailz:
Can be related to this https://github.com/georchestra/mapstore2-georchestra/issues/257

offtherailz:
can be related to https://github.com/georchestra/mapstore2-georchestra/issues/257

tdipisa:
@offtherailz is it possible to configure this filter?

offtherailz:
[...]

PR #6818, opened 2021-04-28
Fix 6813 filter request layers only

MuhweziDeo:
## Description

 - [x] Enhancement

## Issue

Record Items that are not Layers are shown
#6813 

**What is the new behavior?**
 Record Items that are not Layers are hidden

## Breaking change
 - [x] No

### DIFF
--- a/web/client/api/CSW.js
+++ b/web/client/api/CSW.js
@@ -39,12 +39,18 @@ export const constructXMLBody = (startPosition, maxRecords, searchText) => {
         <csw:Query typeNames="csw:Record">
             <csw:ElementSetName>full</csw:ElementSetName>
             <csw:Constraint version="1.1.0">
-                <ogc:Filter>
-                    <ogc:PropertyIsEqualTo>
-                        <ogc:PropertyName>dc:type</ogc:PropertyName>
-                    <ogc:Literal>dataset</ogc:Literal>
-                    </ogc:PropertyIsEqualTo>
-                </ogc:Filter>
+        <ogc:Filter>
+            <ogc:Or>
+            <ogc:PropertyIsEqualTo>
+                <ogc:PropertyName>dc:type</ogc:PropertyName>
+                <ogc:Literal>dataset</ogc:Literal>
+            </ogc:PropertyIsEqualTo>
+            <ogc:PropertyIsEqualTo>
+                <ogc:PropertyName>dc:type</ogc:PropertyName>
+                <ogc:Literal>http://purl.org/dc/dcmitype/Dataset</ogc:Literal>
+            </ogc:PropertyIsEqualTo>
+           </ogc:Or>
+        </ogc:Filter>
             </csw:Constraint>
         </csw:Query>
     </csw:GetRecords>`;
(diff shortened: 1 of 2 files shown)


offtherailz (changes requested):
This filter makes `https://gs-stable.geo-solutions.it/geoserver/csw` not working anymore. (now empty response).
So the filter should be optional. 
(if text filter and configured filter(s) are present, add the contrain tag, in with the filter, in `And` if both are present.)

I'm wandering if we should make this additional filter configurable (via UI and/or json configuration). @giohappy ?

MuhweziDeo:
> This filter makes `https://gs-stable.geo-solutions.it/geoserver/csw` not working anymore. (now empty response).
> So the filter should be optional.
> (if text filter and configured filter(s) are present, add the contrain tag, in with the filter, in `And` if both are present.)
> 
> I'm wandering if we should make this additional filter configurable (via UI and/or json configuration). @giohappy ?

There is a pending fix on the backend to take this in consideration already reported this issue

offtherailz:
mmm... Anyway we may need to support csw without this "fix". I think this filter have to be optional **in any case**. So I suggest to make this filter optional, configured where needed, and continue. @giohappy do you agree?
Can you link the ticket for back-end fixing?

giohappy:
@offtherailz I'm fineto have it as optional. The filter is what should be used to filter layers only according to the CSW specs. FYI @etj is checking why Geoserver returns that unexpected value instead of "dataset". By the way I'm fine to have it configured only inside the GeoNode MapStore client.

giohappy:
@offtherailz I'm fine, but if you set the "dc:dataset" filter by default now Geoserver won't return anything, because as I was saying Geoserver CSW is using a different label (apparently not standard).

giohappy:
@tdipisa this is the original issue https://github.com/GeoNode/geonode-mapstore-client/issues/103
It's not connected to a specific project because it spans multiple clients (ADBPO and Almaviva at elast)

offtherailz:
So, as discussed with @etj  and @giohappy adding modifying the filter with a more complex one, that includes all the possible `dataset` filters should be the solution for all the catalogs available.
@giohappy can you send to @MuhweziDeo the final filter so we can update this PR?

giohappy:
The following filter was tested succesfully on:

 - GeoNode (https://master.demo.geonode.org/catalogue/csw?service=CSW&version=2.0.2&request=)
 - Geoserver (https://gs-stable.geo-solutions.it/geoserver/csw?service=CSW&version=2.0.2&request=)
 - GeoNetwork (http://gn3-rndt.geo-solutions.it/geonetwork/srv/eng/csw)

```xml
<ogc:Filter>
	 <ogc:Or>
	 <ogc:PropertyIsEqualTo>
		 <ogc:PropertyName>dc:type</ogc:PropertyName>
		 <ogc:Literal>dataset</ogc:Literal>
	 </ogc:PropertyIsEqualTo>
	 <ogc:PropertyIsEqualTo>
		 <ogc:PropertyName>dc:type</ogc:PropertyName>
		 <ogc:Literal>http://purl.org/dc/dcmitype/Dataset</ogc:Literal>
	 </ogc:PropertyIsEqualTo>
	</ogc:Or>
 </ogc:Filter>
```

@MuhweziDeo you can proceed with this

offtherailz:
I removed the "blocked" label

MuhweziDeo:
> I removed the "blocked" label



> I removed the "blocked" label

Do you mind updating the issue description for the issue as well to have better alignment with the expectations and also keep track of the requirements @offtherailz

offtherailz:
Ok. Sumarizing, nothing changed except the filter that should be something like in the comment https://github.com/geosolutions-it/MapStore2/pull/6818#issuecomment-839646401 
use the services in the same comment as test.

offtherailz (approved):

Issue #6817, opened 2021-04-28
Remove 'allowFullScreen' Iframe attribute from dashboard and geostory share


tdipisa:
## Description

With #6761 the **allowFullScreen** attribute has been added also on the embed snippet of dashboard and geostory but it is not needed there so it should be removed for them.

## How to reproduce

-  Open the share->embed tool of a dashboard or a geostory

![2021_04_28_17_24_55_MapStore_HomePage.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/372738ca-19aa-4b19-ac8a-2b6c071874d7)

*Expected Result*

The **allowFullScreen** should not be present in the embedded snippet of dashboards and geostories.

*Current Result*

The **allowFullScreen** is present in all cases.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

ElenaGallo:
@MuhweziDeo The **allowFullScreen** attribute is still present from the dashboard and geostory cards on the home page

**How to Reproduce**
Follow the steps shown in the gif below:
![share_dash_story](https://user-images.githubusercontent.com/56537133/117954515-62e92b80-b317-11eb-841b-79f05a6fa4f2.gif)

PR #6861, opened 2021-05-13
6817: Hide full screen when sharing links

MuhweziDeo:
## Description
See issue [comment](https://github.com/geosolutions-it/MapStore2/issues/6817#event-4681857812) 

Hide allowFullScreen when sharing on the home screen.

 - [x] Bugfix

## Issue

See issue [comment](https://github.com/geosolutions-it/MapStore2/issues/6817#event-4681857812) 
#6817 

**What is the new behavior?**

allowFullScreen is not present 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/share/ShareEmbed.jsx
+++ b/web/client/components/share/ShareEmbed.jsx
@@ -141,7 +141,8 @@ class ShareEmbed extends React.Component {
 
     };
     hideAllowFullScreen = () => {
-        return window.location.hash?.includes("geostory") || window.location.hash?.includes("dashboard");
+        return window.location.hash?.includes("geostory") || window.location.hash?.includes("dashboard") ||
+            this.props.shareUrl.includes("dashboard") || this.props.shareUrl.includes("geostory");
     }
 }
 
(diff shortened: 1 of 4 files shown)


offtherailz (commented):
This solution is becoming very tricky.
Can I propose to change it ?
Instead of guessing from URL or ShareURL, I think is more useful and solid to use `embedOptions` parameter. 
The new option `allowFullscreen`: `false` for dashboards and stories, as default, `true` for maps.  

I see ShareUtils DASHBOARD_DEFAULT_SHARE_OPTIONS  and GEOSTORY_DEFAULT_SHARE_OPTIONS that can may be helpful. 

@allyoucanmap you worked with this more then me. The idea is to show `allowFullScreen **only** for maps and disable it for every other one. Can you indicate a way to do it clean?

offtherailz:
@MuhweziDeo can you try that way instead? Due to the usage in several environments of share tool, I think this solution should be used to prevent unwanted bahaviours.

offtherailz (commented):
I'd prefer to have allowFullscreen true in maps only, instead of as default. it's looks more natural to me. Anyway also this solution may work. 

@allyoucanmap any suggestion

offtherailz (approved):

PR #6880, opened 2021-05-18
#6870 The Identify panel opens when it shouldn't in one use case

QSafariallahkheili:
## Description

This PR fixes the the status of identify panel by clearing mapinfo results when the trigger method is changed

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/M1apStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**

#6870 

**What is the new behavior?**

it fixes the problems related to trigger event change in Settings Panel

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

Issue #6870, opened 2021-05-14
The Identify panel opens when it shouldn't in one use case


tdipisa:
## Description

Switching from **Hover** to **Click** mode the Identify panel opens but it shouldn't.

## How to reproduce

- Open [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/10358)
- Perform an identify request by clicking on a layer in map
- Switch to **Hover** mode in Option->Settings
- Perform an identify request by hovering the mouse over a layer in the map
- Switch again to **Click** mode in Option->Settings

*Expected Result*

The Identify panel should not be opened in this case because an identify request has not been performed.

*Current Result*

The Identify panel opens.

![gfi_bug_panel.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/098ac280-da4c-4083-bde7-0f15dc24ff37)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #6929, opened 2021-05-28
backport_2021.01.xx - #6870 The Identify panel opens when it shouldn't in one use case (#6880)

QSafariallahkheili:
## Description

backport_2021.01.xx - #6870 The Identify panel opens when it shouldn't in one use case (#6880)


MV88 (approved):

MV88:
@ElenaGallo please test in QA

Issue #6892, opened 2021-05-21
WFS layers of geometry type `Geometry` crash the layer style panel


dhlambert:
## Description
WFS layers of geometry type `Geometry` crashes the layer style panel.

## How to reproduce

1. Navigate to the maps page
2. Connect to a wfs catalog
3. Add a wfs layer that returns a geometry type `gml:Geometry` from a `DescribeFeatureType` request
4. Select the added layer and  click the wrench icon labeled `Selected Layer Settings`
5. Click on the color picker for style.
6. Observe the application crashing like the following screenshot

![image](https://user-images.githubusercontent.com/38519933/119073484-8c096a80-ba41-11eb-8084-e12675685694.png)

*Expected Result*
The application should not crash and/or an appropriate style panel should appear.

*Current Result*
The application crashes when opening the style panel with wfs layers of geometry type `Geometry`.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information
It appears that the issue lies with the geometry type of the layer which is used to select the `styler` in the `SingleVectorStyleEditor` see [here](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/tocitemssettings/SimpleVectorStyleEditor.jsx#L147).

PR #6966, opened 2021-06-09
Backport 2021.01.xx - #6892 Fix geometry type bug in wfs layer style panel (#6893)

dhlambert:
## Description
Backport 2021.01.xx - #6892 Fix geometry type bug in wfs layer style panel (#6893)


offtherailz (approved):

offtherailz:
Thank you very much for contributing :+1:

Issue #6843, opened 2021-05-10
Improve strings for french translation


MaelREBOUX:
## Description
For the georchestra project we need to change some strings that are leading users to confusions between some concepts.

See https://github.com/georchestra/mapstore2-georchestra/issues/367 for details.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Code style update (formatting, local variables)

MaelREBOUX:
TODO

- [ ] Afficher la légende et les outils -> Légende et visibilité

MaelREBOUX:
I have not yet worked on the 4 string listed above. PR to come

MaelREBOUX:
Hello @ElenaGallo 

I cannot find string to change the name of the measurement layer added when doing :

![image](https://user-images.githubusercontent.com/6370443/121690628-5f87c200-cac6-11eb-9a98-97370a668f5c.png)

We want it named in the user langage. Need a specific issu maybe ?

tdipisa:
@MaelREBOUX this is a closed issue, so yes please open a new one. Thank you so much for contributing.

PR #6972, opened 2021-06-11
Improve localization strings

MaelREBOUX:
## Description

This PR to propose some enhancement on some strings visible in the UI

See https://github.com/geosolutions-it/MapStore2/issues/6843#issuecomment-848632106

May be discussed.

### DIFF
--- a/web/client/translations/data.de-DE.json
+++ b/web/client/translations/data.de-DE.json
@@ -514,7 +514,7 @@
         "toc": {
             "toggleLayerVisibility": "Ebenensichtbarkeit umschalten",
             "toggleGroupVisibility": "Gruppensichtbarkeit umschalten",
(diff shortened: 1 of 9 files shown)


offtherailz (changes requested):
Hi, one thing to change

- [ ] We should leave "Search by location Name"  for MapStore. (for all the languages):
 It is consistent with the menu. (search by --> locatioon name, coordinate, bookmark)
image

Of course you can put this customization in your project, if you prefer (take into account this menu, anyway)

Of course if you think that the translation in french should be improved, you can do it. 


thank you very much for your contribution :+1:

MaelREBOUX:
I revert the commit.

MaelREBOUX (commented):
I revert this commit

MaelREBOUX (commented):
I revert this commit

offtherailz (approved):

PR #7008, opened 2021-06-22
#6853 Fix crash when using an unescaped template

MV88:
## Description

The problem was related to the name of the feature attributes, especially if they contained a - like `${ properties.data-name }`
the crash happened inside template of lodash here
https://github.com/geosolutions-it/MapStore2/blob/b7b366d7d35491f659e4946dec800faff7cca887/web/client/components/data/identify/viewers/TemplateViewer.jsx#L19
So what i did is to prevent the crash with a try catch and to use square bracket notation if it finds a minus inside templates segments

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**

#6853

**What is the new behavior?**

no crash
the minus properties are managed correctly even if already escaped

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

Issue #6853, opened 2021-05-11
MapStore crashes when the JSON response from a GetFeatureInfo contains a property name with a dash


giohappy:
## Description
MapStore crashes if it receives a property name conainting a dash ("-") inside the JSON response to a GetFeatureInfo request.

## How to reproduce
[This layer](https://master.demo.geonode.org/layers/geonode_master_data:geonode:alberi) on GeoNode master demo contains an attribute named "data-acquisizione"
- execute [this GFI request](https://master.demo.geonode.org/geoserver/ows?service=WMS&version=1.1.1&request=GetFeatureInfo&exceptions=application%2Fjson&id=geonode:alberi__4&layers=geonode:alberi&query_layers=geonode:alberi&styles=&x=51&y=51&height=101&width=101&srs=EPSG:3857&bbox=1389752.3702168849,5148380.807455325,1389872.9974021476,5148501.434640587&feature_count=10&info_format=application%2Fjson&ENV=&access_token=fGVTE4n5db10M5xJPY6mgj3tMSKLNc)  
- a JSON response (reported below) is generated correctly and is a valid JSON

JSON response
```json
[...]

PR #7029, opened 2021-06-25
backport 2021.01.xx - #6853 Fix crash when using an unescaped template (#7008)

MV88:
## Description

backport 2021.01.xx - #6853 Fix crash when using an unescaped template (#7008)


offtherailz (approved):
I notice right now `templateFormatInfoAlertExample` and `templateFormatInfoAlertExample2` that should have been merged to make it simpler, but it is acceptable anyway.

Issue #7074, opened 2021-07-07
Visual Style Editor crashes the app with raster layers


allyoucanmap:
## Description

![123404977-afc45100-d5a9-11eb-85ab-bb0d0f08b1d2](https://user-images.githubusercontent.com/19175505/124720995-01090480-df09-11eb-94d5-05603fcc7fe8.gif)

## How to reproduce

-  open a map in MapStore and add a raster layer (eg arcGridSample from WMS Catalog)
- open the TOC and select the layer
- open the settings and select style tab
- select a style and then click on edit style
- switch to visual style editor

*Expected Result*

- visual style editor is visible

*Current Result*

- application crashes

![image](https://user-images.githubusercontent.com/19175505/124723183-0e26f300-df0b-11eb-86a0-c1c90ba27902.png)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

The fix should be backported to the geonode-mapstore-client-2.1.x branch

PR #7075, opened 2021-07-07
#7074: Visual Style Editor crashes the app with raster layers 

allyoucanmap:
## Description

Sometimes attributes inside the visual style editor could return `false` so valid for the `?.` operator but not valid for the .filter function

 - [x] Bugfix

## Issue

#7074

**What is the new behavior?**

App does not crash after switching to visual style editor with raster layers

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/styleeditor/RulesEditor.jsx
+++ b/web/client/components/styleeditor/RulesEditor.jsx
@@ -9,6 +9,7 @@
 import React, { useRef, forwardRef } from 'react';
 import PropTypes from 'prop-types';
 import find from 'lodash/find';
+import isArray from 'lodash/isArray';
 import { Glyphicon, FormControl as FormControlRB, FormGroup } from 'react-bootstrap';
 import Fields from './Fields';
 import uuidv1 from 'uuid/v1';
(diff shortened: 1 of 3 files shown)


web/client/components/styleeditor/RulesEditor.jsx
```diff
@@ -253,7 +253,9 @@ const RulesEditor = forwardRef(({
                         hideScaleDenominator,
                         classificationType
                     } = ruleBlock[ruleKind] || {};
-                    const isCustomNumber =  attributes?.filter(({label}) => label === rule?.attribute)?.[0]?.type === 'number';
+
+                    // attributes could be false so it is not possible to use ?. operator
+                    const isCustomNumber =  attributes && attributes.filter(({label}) => label === rule?.attribute)?.[0]?.type === 'number';
```
offtherailz, line 258: Does it makes sense the `attributes` to become false?
if attributes is **only** false or array, that this makes sense. otherwise, if it is true, it may fail again. 
I suggest to use `isArray` instead, to ensure the method filter is present, instead.

offtherailz (commented):
Given that `false` is a valid value for `arrays`, I should use `isArray` that is more correct and less implicit.It  is exactly what you are looking for.

If false is not a valid value, maybe you should identify the reason.

web/client/components/styleeditor/RulesEditor.jsx
offtherailz: Anyway this is strange. it is always true, isn't it?

allyoucanmap: The PR is updated with following changes:
- return an empty array of attributes if false
- use isArray before to look inside the attributes list

This control is needed to change the disable state of the attributes options in the classification symbolizer and this state changes based on the current selected attribute type

offtherailz (approved):

Issue #7134, opened 2021-07-22
Improve configurability of map catalog in dashboard editor 


allyoucanmap:
## Description

We should allow to configure the map catalog of dashboard by:

- disable the empty map entry in the list
- use the persistence API to get the map resource

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

PR #7135, opened 2021-07-22
#7134: Improve configurability of map catalog in dashboard editor

allyoucanmap:
## Description

This PR adds:

- a new property disableEmptyMap to remove the empty map entry from map catalog
- use persistence api after selecting a map in the map catalog in this way we can override the request in custom project

 - [x] Improvement

## Issue

#7134

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/maps/enhancers/enhancers.js
+++ b/web/client/components/maps/enhancers/enhancers.js
@@ -60,7 +60,7 @@ const scrollSpyOptions = { querySelector: ".ms2-border-layout-body .ms2-border-l
  * @param {function} fn the original loadPage function
  */
 const emptyMap = fn => (opts, page) => {
-    if (page === 0 && opts && !opts.text) {
+    if (!opts.disableEmptyMap && page === 0 && opts && !opts.text) {
         return fn(opts, page).map(({ items, total, ...props}) => ({
             ...props,
             total,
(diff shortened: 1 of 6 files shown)


web/client/components/maps/enhancers/enhancers.js
```diff
@@ -91,7 +91,7 @@ export const withEmptyMapVirtualScrollProperties = ({ loadPage: lp, scrollSpyOpt
     loadPage: emptyMap(lp),
     hasMore: ({total, items}) => {
         if (items && items.length >= 1 && items[0].id === "EMPTY_MAP") {
-            return total > (items.length + 1);
+            return total > (items.length - 1);
```
offtherailz, line 94: Didn't get this change

allyoucanmap: if the total of resources from the server is 11 and the items per page are 10 the first request will return 10 items + 1 added locally as empty map. So I think we should remove 1 from the item length that represent the local empty map.

We could add 1 but I should to the total instead, either ways should work.

offtherailz: Ok, so it's a bug fix. Now I get it and it seems correct
- total: 11
- items: 11 (10 + 1)
- 11 > 11 - 1

offtherailz (approved):

Issue #7185, opened 2021-08-05
Geolocation orientation in mobile


dsuren1:
### Description
In mobile move put the arrow relating to the orientation of the geolocation. At the moment in MS2 there is the circle symbol to represent only the position. It is also required to have the map move when the subject moves.

### Please indicate if this issue is related to a bug or a new feature request
- [ ] Bug
- [X] New Feature

### Other useful information:

- **Heading** is a property of coordinateshat we can use for this: when this property is set (the subject is moving), we will have to change the point to an arrow or a triangle, orienting it in the direction (heading is expressed in deg, clockwise, start from north). If the property is not valorized (the subject is stationary) the point symbol will be used (as it is now). Implementation for leaflets and openlayers is required.

- Tracking is already supported by MS and it will be maintained (by default the MS map follows the user until the map is manually moved, then he can click again on the geolocation tool to reactivate the tracking). Nothing to to for this since it is already supported.

**Refer**: https://github.com/geosolutions-it/MapStore2-C040/issues/458

PR #7197, opened 2021-08-09
#7185: Geolocation orientation in mobile for OL and Leaflet map

dsuren1:
## Description
This PR adds commits to add new feature to OL and Leaflet maps to show Geolocation orientation in devices that supports heading prop of Geolocation API

 - [x] Feature

## Issue

#7185 

**What is the new behavior?**
When geolocation is activated and in supported devices (Mobile, tablet, etc.) with `enableHighAccuracy: true` , the heading props is used to display navigation arrow marker with rotation angle determined by the heading radian or degree value based on OL and leaflet map respectively
2021-08-09 20_10_16-MapStore HomePage

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/openlayers/OlLocate.js
+++ b/web/client/utils/openlayers/OlLocate.js
@@ -22,6 +22,9 @@ import GeometryCollection from 'ol/geom/GeometryCollection';
 import {Style, Fill, Stroke} from 'ol/style';
 import CircleStyle from 'ol/style/Circle';
 import throttle from 'lodash/throttle';
+import isNil from 'lodash/isNil';
+import {getMarkerStyleLegacy} from "../../components/map/openlayers/VectorStyle";
+import {createSvgUrl} from "../VectorStyleUtils";
 
 
 const popUp = olPopUp();
@@ -30,6 +33,7 @@ const popUp = olPopUp();
 const OlLocate = function(map, optOptions) {
     BaseObject.call(this, {state: "DISABLED"});
     this.map = map;
+    const { heading, circleAccuracy, locate } = this._getDefaultStyles() || {};
     let defOptions = {
         drawCircle: true, // draw accuracy circle
         follow: true, // follow with zoom and pan the user's location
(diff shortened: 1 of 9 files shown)


offtherailz (commented):
# Both
- + The arrow is not well centered.
Maybe if the speed is after a treshold, a point should be used. 


### Openlayers
The arrow blinks on each update. Moreover I think that 0 heading or speed is not contemplated.

Screen_Recording_20210825-124428_Chrome_1

### Leaflet
Can you try with a deploy. Just to confirm that leaflet rotatedmarker works when compiled

web/client/utils/openlayers/OlLocate.js
```diff
@@ -210,15 +237,37 @@ OlLocate.prototype.mapClick = function(evt) {
...
+            iconAnchor: [0.5, 0.5],
+            anchorXUnits: 'fraction',
+            anchorYUnits: 'fraction',
+            color,
+            fillColor: color,
+            opacity: 1,
+            size: 41,
+            fillOpacity: 1,
+            shape,
+            symbolUrl: `product/assets/symbols/${shape}.svg`
```
offtherailz, line 268: Use directly the raw xml. it is small and can be included in the bundle without requiring URL.

offtherailz (changes requested):
# Both
- [ ] Include the SVG of the arrow in the bundle, not as URL. It is very small and can be included without need of symbols system (raw import of svg + data-url can do the job.
- [ ] The arrow is not well centered.

I tested in leaflet but the same problem looks to be present in OL, as you can see the back corners go outside the circle (they seems also **asymmetric** in leaflet) while the main arrow remain inside. I try to guess, maybe the svg is not perfectly symmetrical, you can modify it to make it perfectly aligned. 
- [ ] We could  introduce a threshold to show the arrow (very small speed can be an error)

### Openlayers
- [ ] The arrow blinks on each update. Moreover 
- [ ] I think that 0 heading or speed is not contemplated.

Screen_Recording_20210825-124428_Chrome_1

### Leaflet
Can you try with a deploy. Just to confirm that leaflet rotatedmarker works when compiled

MV88:
closing in favor of https://github.com/geosolutions-it/MapStore2/pull/7264

Issue #7231, opened 2021-08-25
Always activate synch map button when feature grid opens if configured so


MV88:
## Description

If enableMapFilterSync in FeatureEditor plugin is set to true always activate synch map button when feature grid opens

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #7232, opened 2021-08-25
#7231 make synch to map always active if configured so

MV88:
## Description

@offtherailz the did mount of feature editor is run only once, even if i render null in case props.open is false
so in this way we always activate it when feature grid opens

let me know if this solution is acceptable

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7231

**What is the new behavior?**

when feature grid opens if enableMapFilterSync is true it ill set to true also the sync map flag

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/FeatureEditor.jsx
+++ b/web/client/plugins/FeatureEditor.jsx
@@ -285,6 +285,9 @@ const EditorPlugin = compose(
             if (!isEqual(newOptions, oldOptions) ) {
                 this.props.onMount(newOptions);
             }
+            if (this.props.enableMapFilterSync) {
+                this.props.setSyncTool(true);
+            }
         }
     }),
     connect(selector,


offtherailz (changes requested):
- [ ] There is no way to set it to `false`. Only to set it to `true`. Maybe we should use the value of the flag, and always apply. 
I think is better if we remove `setSyncTool` at all and add this flag to the `setUp` flags.  What do you think?

offtherailz (approved):
It works. Only a doubt. when the user deactivate sync, close and reopen the feature grid, what is the expected behavior @tdipisa

MV88:
@tdipisa 
i am merging this and backporting to 2021.01.xx and  to 2021.02.xx

Issue #7276, opened 2021-09-03
Improve feedback when empty results are returned from GetFeatureInfo Requests


MV88:
## Description

Actually an empty feature info viewer is displayed when you click on this map that has 2  
layers that points to geocollect geoserver instance
Due to a proxy property missing (see [here](https://github.com/geosolutions-it/MapStore2-C040/issues/445#issuecomment-904527487)) they return an empty body

what we can do is to show some message for this use case

so open [this](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/35736) map and click on a feature

expected: 
to see the message notifying empty results has been returned from the server

current 
empty white page

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

MV88:
The requirements for this feature have changed:

Basically we should avoid to manage and show dedicated messages for empty responses
instead we will treat them as failing responses  @offtherailz  please confirm this

so if we do that we should obtain:
![image](https://user-images.githubusercontent.com/11991428/132889372-c7e0714d-8860-4c86-a646-da64bd8e620f.png)

instead of current behavior

![image](https://user-images.githubusercontent.com/11991428/132889400-d3555b54-de05-4ceb-849d-f39e058b0199.png)

MV88:
@tdipisa we can postpone this issue and since it depends on #7296

PR #7292, opened 2021-09-10
#7276 Improve feedback about feature info

MV88:
## Description

This PR provides changes to how messages and feedback are presented i will list before and after for the different use cases that we can have.

[in this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/35736) i have configured 4 layers:
- segnalazione muri (it will return empty response always)
- sopralluoghi (it will return empty response always)
- linea costa (a valid layer with correct responses)
- regioni italiane (a valid layer with correct responses)

by clicking on different points and turning on and off the layers you can test different combinations

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7276

**What is the new behavior?**

this has been removed
image
and a different way to identify invalid responses have been included

- when an invalid response with an empty body is returned we show this error message
image

- when an invalid response with a valid body is returned we show this info message
image

## Breaking change
 - [x] No

## Other useful information

Here is a list of all combination tested

image

image

image

image

image

image

### DIFF
--- a/web/client/components/data/identify/DefaultViewer.jsx
+++ b/web/client/components/data/identify/DefaultViewer.jsx
@@ -95,6 +95,22 @@ class DefaultViewer extends React.Component {
         return validator.getValidResponses([response]);
     }
 
+    getInvalidTitles = () => {
+        const {invalidResponses} = this.getResponseProperties();
+        const titlesWithResponse = invalidResponses
+            .filter(({response}) => response !== "")
+            .map((res) => {
+                const {layerMetadata} = res;
+                return layerMetadata.title;
+            });
+        const titlesEmptyResponse = invalidResponses
+            .filter(({response}) => response === "")
+            .map((res) => {
+                const {layerMetadata} = res;
+                return layerMetadata.title;
+            });
+        return {titlesWithResponse, titlesEmptyResponse};
+    };
     renderEmptyLayers = () => {
         const {invalidResponses, emptyResponses} = this.getResponseProperties();
         if (this.props.missingResponses === 0 && emptyResponses) {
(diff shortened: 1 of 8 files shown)


web/client/components/data/identify/DefaultViewer.jsx
```diff
@@ -192,7 +218,7 @@ class DefaultViewer extends React.Component {
         componentOrder = this.props.isMobile ? componentOrder : reverse(componentOrder);
         return (
             <div className="mapstore-identify-viewer">
-                {!emptyResponses ? componentOrder.map((c)=> c) : this.renderEmptyPages()}
+                {!emptyResponses ? componentOrder.map((c)=> c) : this.renderEmptyPages(this.props)}
```
offtherailz, line 221: I don't think this is needed, renderEmptyPages takes no arguments
```suggestion
                {!emptyResponses ? componentOrder.map((c)=> c) : this.renderEmptyPages()}
```

web/client/components/data/identify/DefaultViewer.jsx
```diff
@@ -95,6 +95,22 @@ class DefaultViewer extends React.Component {
...
+                return layerMetadata.title;
+            });
+        const titlesEmptyResponse = invalidResponses
+            .filter(({response}) => response === "")
+            .map((res) => {
+                const {layerMetadata} = res;
+                return layerMetadata.title;
+            });
+        return {titlesWithResponse, titlesEmptyResponse};
+    };
```
offtherailz, line 113: I suggest to create a `getAlerts` that returns  an array of 
```j
{
  type: "info" // or "danger"
  msgId
  msgParams // layers
}
```

I think you can use HTML in any case.
It is easier to parse and handle errors of any kind this way. 

```javascript
<Alert bsStyle={type}>
  <HTML msgId={msgId} msgParams={msgParams}/>
</Alert>
```

offtherailz (changes requested):
Apart for the solution, that is ok, 

@tdipisa I don't like the idea to handle the empty response of our **specific** proxy in the code of GFI code to manage a very specific use case than may disappear in the future (Java8 will be used in future MapStore and the error seems to be related only to Java7) and that anyway is not limited to GFI (any request can have this SSL handshake problem, not only GFI)

I'd prefer to fix http-proxy, returning a real http error in these case, and handle it as a normal error in GFI, makes more sense to me.
Also not merging this PR (or at least the part that handle empty string) makes more sense to me.

Improvements to the error handling of GFI are anyway welcome.

web/client/components/data/identify/DefaultViewer.jsx
MV88: good catch

offtherailz (changes requested):
Please remove all the response = "" handling only the list of not-responding layer.

MV88:
Closed because requirements have changed

Issue #7367, opened 2021-09-27
Edit a *Map widget* couse an error


ElenaGallo:
## Description

Error while trying to edit a *Map widget*

## How to reproduce

-  Open [this dashorad](https://qa-mapstore.geosolutionsgroup.com/mapstore/#/dashboard/21694)
- Click on Edit from the *Map widget*

*Expected Result*

The *Map widget* is editable.
*Current Result*

A gray error page opens. Error on console are:
> TypeError: e.forEach is not a function
    at j (mapstore2.js?8beef3696d5cd156e798:2)
    at mapstore2.js?8beef3696d5cd156e798:2
    at Array.reduce (<anonymous>)
    at mapstore2.js?8beef3696d5cd156e798:2
    at Array.reduce (<anonymous>)
    at Object.$ [as getLayersByGroup] (mapstore2.js?8beef3696d5cd156e798:2)
    at re (mapstore2.js?8beef3696d5cd156e798:2)
    at mapstore2.js?8beef3696d5cd156e798:2
    at mapstore2.js?8beef3696d5cd156e798:2
    at mapstore2.js?8beef3696d5cd156e798:2
    at _ (mapstore2.js?8beef3696d5cd156e798:2)
    at mapstore2.js?8beef3696d5cd156e798:2
    at mapstore2.js?8beef3696d5cd156e798:2
    at na (mapstore2.js?8beef3696d5cd156e798:2)
    at Fs (mapstore2.js?8beef3696d5cd156e798:2)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #7390, opened 2021-09-30
Issue #7367; fix error when editing map widget

tonym3nsa:
## Description

 - [X] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7367 

**What is the new behavior?**

The edit map widget works without breaking
## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/LayersUtils.js
+++ b/web/client/utils/LayersUtils.js
@@ -158,12 +158,14 @@ export const getNestedGroupTitle = (id, groups = []) => {
  */
 export const flattenArrayOfObjects = (array) => {
     let result = [];
-    array?.forEach((a) => {
-        result.push(a);
-        if (Array.isArray(a.nodes)) {
-            result = result.concat(flattenArrayOfObjects(a.nodes));
-        }
-    });
+    if (array && Array.isArray(array)) {
+        array?.forEach((a) => {
+            result.push(a);
+            if (Array.isArray(a.nodes)) {
+                result = result.concat(flattenArrayOfObjects(a.nodes));
+            }
+        });
+    }
     return result;
 };
 
@@ -175,10 +177,13 @@ export const flattenArrayOfObjects = (array) => {
  */
 
 export const displayTitle = (id, groups) => {
-    for (let group of groups) {
-        if (group?.id === id) {
-            return group.title;
+    if (groups && Array.isArray(groups)) {
+        for (let group of groups) {
+            if (group?.id === id) {
+                return group.title;
+            }
         }
+        return 'Default';
     }
     return 'Default';
 };


web/client/utils/LayersUtils.js
```diff
@@ -153,17 +153,19 @@ export const getNestedGroupTitle = (id, groups = []) => {
 
 /**
  * Flatten nested array to a one-level array
- * @param {object[]} array of objects
- * @returns {object[]} flattened array
+ * @param {Object[]} array of objects
```
MV88, line 156: @tonym3nsa where did you see that Object type needs to be in capital case?
I think we can keep it lowercase since the whole project is using this convention
probably we did this because docma lib was having issues with it
@offtherailz right?

web/client/utils/LayersUtils.js
```diff
@@ -153,17 +153,19 @@ export const getNestedGroupTitle = (id, groups = []) => {
...
+ * @returns {Object[]} flattened array
  */
-export const flattenArrayOfObjects = (array) => {
+export const flattenArrayOfObjects = (arrayOfObj) => {
```
MV88, line 159: what if we use this https://lodash.com/docs/4.17.15#flattenDeep ?

MV88 (changes requested):
@tonym3nsa 
Exactly why was crashing?

tonym3nsa:
> @tonym3nsa Exactly why was crashing?

The function was expecting an array of object but what's passed in this case is a single object not in an array

web/client/utils/LayersUtils.js
tonym3nsa: @MV88 Well noted. I have however noted that there are various instances in the project where `@param {object} `starts with an uppercase O. Is it something we need to take care of?

web/client/utils/LayersUtils.js
tonym3nsa: `_.flattenDeep` seems to be expecting an array but in some cases what is passed is just a single object

tdipisa:
> > @tonym3nsa Exactly why was crashing?
> 
> The function was expecting an array of object but what's passed in this case is a single object not in an array

Thank you @tonym3nsa. Please, for each regression issue belonginh to 2021.02.00 I need a comment in the issue that explain why there has been that regression (if possible indicating the commit). Thank you.

web/client/utils/LayersUtils.js
offtherailz: I asked to make you try because it was originally causing problems when generating documentation. I did some tests and now it seems not to be a problem anymore (and the more correct syntax may be "Object").
So for me it's ok.

MV88:
@ElenaGallo please test

Issue #7422, opened 2021-10-07
StyleEditor plugin must not rely on the workspace prefix inside the style names returned by GetCapabilities


allyoucanmap:
## Description

> Geoserver 2.19.x fixed a wrong behaviour where style names isnide workspace-specific GetCapabilities reponse were returned with the workspace prefix.
> MapStores Styled editor relied on this prefix for its functionalities, which now are broken with Geoserver 2.19.x.
> ...

see https://github.com/GeoNode/geonode-mapstore-client/issues/519

## How to reproduce

- login as admin (eg #/viewer/openlayers/10358)
- open a map and add a wms layer from catalog with style editing permissions (eg States of US)
- select the wms layer and open settings
- select the style tab

*Expected Result*

- all the styles should be available for edit or delete action

*Current Result*

- style included in a specific workspace cannot be edited or deleted because mapstore is not able to build the correct url due to latest fix

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #7423, opened 2021-10-07
#7422: StyleEditor plugin must not rely on the workspace prefix inside the style names returned by GetCapabilities 

allyoucanmap:
## Description

This PR replaces the request GetCapabilities of the style editor with a rest/layers request to get the correct styles information. If the rest/layer request fails it fallback to GetCapabilities request to enable the styles selector

The Style Editor will work only with the styles included in the same workspace.

 - [x] Bugfix

## Issue

#7422

**What is the new behavior?**

Styles can be edited and deleted if the users have permission

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/__tests__/styleeditor-test.js
+++ b/web/client/epics/__tests__/styleeditor-test.js
@@ -1407,8 +1407,8 @@ describe('Test styleeditor epics, with mock axios', () => {
                         name: 'layerWorkspace:style_02',
                         workspace: 'layerWorkspace'
                     }, {
-                        name: 'notLayerWorkspace:style_03',
-                        workspace: 'notLayerWorkspace'
+                        name: 'style_03',
+                        workspace: ''
                     }]
                 }
             }}];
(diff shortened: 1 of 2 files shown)


dsuren1:
Hi @allyoucanmap, I have few questions and apart from that everything looks fine to me.

Map used: _Demo map -_ **10358**
**Questions**
Since we are showing only styles of current workspace of the layer (for authenticated user)
   - In some instance, it is not clear which style is selected  (Ex: with ny_roads, actual style selected is _tiger_roads_ but I can see _Base SLD_ style is selected (since only one is shown)
   - And unable to make a new selection from the listed style
   
Example
   - So can we not update styles of other workspace (since we cannot edit without authentication)
   
Style selected:  **_tiger_roads_** (layer - ny_roads)

giohappy:
let's recap @allyoucanmap. From the discussion with @aaime global styles or styles from other workspaces souldn't leask into the GetCapabilities under a workspace. So, if the style editor only relies on the workspaced GC the user shouldn never see styles from outside the workspace itself.

**BUT** currently Geoserver [has a bug](https://github.com/geosolutions-it/geoserver/issues/315#issuecomment-938536200)  and external styles are also listed, although without any means to tell where those styles are coming from. 
So, I guess this is still a problem in case you cannot rely on the REST API, since you will never now if that style is local to the workspace ore coming from elsewhere. Right @allyoucanmap?

aaime:
Global style can legitimately leak into a workspace specific caps document, unless they end up being covered by a same-named local style (in that case, the latter wins). Styles from other workspaces should not be there instead.

allyoucanmap:
@tdipisa @giohappy 

removed the restriction of editing in workspace here https://github.com/geosolutions-it/MapStore2/pull/7423/commits/e2a402dc0f2e97ebc1a18ec11ce90cba0ec3ea27

now the editing works also for global styles

Issue #7287, opened 2021-09-08
Debounce time too low and validation missing for attribute filter in dashboard


tdipisa:
## Description

Typing filter values produces in certain use cases an inconsistent state of the widget. In general we should ensure the following.

- A proper debounce time before applying the filter: now it seems to be applied immediately but we should consider at least 1-2 seconds in order to avoid possible unuseful requests sent by the client 
- A proper validation check to verify that the entered filter is correct before applying it: a kind of validation is seems to be already present but it fails in certain cases, such as:
   - "<0.02,>": this is failing
   - ">0.02,>": this is not failing and regularly notified as wrong filter by the validator
- In case of a wrong filter entered and applied, it should be anyway possible to change it: currently the widget goes in a wrong and inconsistent state 

## How to reproduce

-  To reproduce what reported above [this dashboard](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/dashboard/21694) has been used
- Follow the steps depicted in the GIF below to reproduce

![dash_table.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/41864178-4a70-4a39-a2fe-2459a7c51b53)

*Expected Result*

The table widget should not be in a inconsistent state in case of a wrong filter entered and the filter validation is correctly reported

*Current Result*

The current state is reported in the above description

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #7522, opened 2021-11-09
#7287: Debounce attribute filter evaluation and validation

dsuren1:
## Description
This PR adds debounce to attribute filter and validation causing inconsistent widget state

 - [x] Bugfix

## Issue

#7287 

**What is the new behavior?**
- Entering incorrect expression in Number filter will not cause the widget to go into invalid state
- Incorrect expression will be evaluated and allow filtering only upon correct expression
- Debounce the filter expression input by 1 sec for consistency

## Breaking change
 - [x] No

## Other useful information
Ref: https://github.com/geosolutions-it/MapStore2-C040/issues/493

### DIFF
--- a/web/client/components/data/featuregrid/filterRenderers/__tests__/NumberFilter-test.jsx
+++ b/web/client/components/data/featuregrid/filterRenderers/__tests__/NumberFilter-test.jsx
@@ -9,7 +9,7 @@
 import expect from 'expect';
 import React from 'react';
 import ReactDOM from 'react-dom';
-import ReactTestUtils, {act} from 'react-dom/test-utils';
+import ReactTestUtils from 'react-dom/test-utils';
 
 import NumberFilter from '../NumberFilter';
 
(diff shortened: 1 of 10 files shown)


web/client/components/data/featuregrid/filterRenderers/__tests__/NumberFilter-test.jsx
```diff
@@ -67,37 +62,60 @@ describe('Test for NumberFilter component', () => {
         const input = document.getElementsByTagName("input")[0];
         expect(input.value).toBe("1");
     });
-    it('Test NumberFilter onChange', () => {
```
offtherailz, line 70: For this and all the tests, onChange handler call have to be tested.

offtherailz (changes requested):
Seems to be good. 2 Observations. 

- [x] Applying the debounce to the AttributeFilter make the feature grid of the main map (that uses the same filter) very slow, because it is already debounced on itself. Applying this debounce on an upper level in the dashboard could avoid to influence the main feature grid. I personally think that can be avoided at all. to avoid complicating things. @dsuren1 please check if it is possible to debounce the onChange handler of the widget, instead of the component's one.
- [x] debounce time in any case can be reduced to 500 ms, to speed up a little bit the things. 
- [x] `onChange` handlers have to be tested anyway. They are the main interface provided by the component (see comments)

tdipisa:
@dsuren1 please proceed with this accordingly asap.

offtherailz (approved):

offtherailz:
Please @ElenaGallo test on dev instance

ElenaGallo:
Test passed, @offtherailz please backport to 2021.02.xx

Issue #7416, opened 2021-10-06
Filter not applied correctly in attribute table


tdipisa:
## Description

When defining a layer filter from TOC using the[ Filter Layer](https://mapstore.readthedocs.io/en/latest/user-guide/filtering-layers/#layer-filter) tool, that filter is applied to the attribute table but it isn't applied to the [WFS export tool](https://mapstore.readthedocs.io/en/latest/user-guide/attributes-table/#download-the-grid-data).

## How to reproduce

- [This](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/10358) test map can be used 
- Click on the states layer
- Enable the layer filter on that layer 
- Open the attribute table

*Expected Result*

The filter defined with the Layer filter tool is applied also to the WFS download tool as expected.

*Current Result*

- The table entries are correctly filtered according to the defined filter
Try to download the table's datas
- All entries are downloaded, without filter

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #7529, opened 2021-11-10
#7416: Filter not applied correctly in attribute table

DavidQuartz:
## Description

Layer filters are now applied in WFS export tool.
Previously `getWFSFeature` did not consider any filters when querying WFS for download. This PR introduces a new functionality to properly query WFS with layer filters.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7416 

**What is the new behavior?**

The filter defined with the Layer filter tool is applied also to the WFS download tool as expected.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/layerdownload.js
+++ b/web/client/epics/layerdownload.js
@@ -110,7 +110,10 @@ const getWFSFeature = ({ url, filterObj = {}, layerFilter, downloadOptions = {},
 
     const data = mergeFiltersToOGC({ ogcVersion: '1.0.0', addXmlnsToRoot: true, xmlnsToAdd: ['xmlns:ogc="http://www.opengis.net/ogc"', 'xmlns:gml="http://www.opengis.net/gml"'] }, layerFilter, filterObj);
 
-    return getXMLFeature(url, getFilterFeature(query(filterObj.featureTypeName, [...(sortOptions ? [sortBy(sortOptions.sortBy, sortOptions.sortOrder)] : []), ...(pn ? [propertyName(pn)] : []), ...(data ? castArray(data) : [])]), options), options, downloadOptions.selectedFormat);
+    return getXMLFeature(url, getFilterFeature(query(
+        filterObj.featureTypeName, [...(sortOptions ? [sortBy(sortOptions.sortBy, sortOptions.sortOrder)] : []), ...(pn ? [propertyName(pn)] : []), ...(data ? castArray(data) : [])],
+        { srsName: downloadOptions.selectedSrs })
+    ), options, downloadOptions.selectedFormat);
 
 };
 
--- a/web/client/epics/layerdownload.js
+++ b/web/client/epics/layerdownload.js
@@ -247,7 +247,7 @@ export const startFeatureExportDownload = (action$, store) =>
             }
         })
             .do(({ data, headers }) => {
-                if (headers && headers["content-type"] === "application/xml") { // TODO add expected mimetypes in the case you want application/dxf
+                if (headers["content-type"] === "application/xml") { // TODO add expected mimetypes in the case you want application/dxf
                     let xml = String.fromCharCode.apply(null, new Uint8Array(data));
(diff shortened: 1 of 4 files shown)


web/client/epics/layerdownload.js
```diff
@@ -103,15 +105,15 @@ const hasOutputFormat = (data) => {
...
-const getWFSFeature = ({url, filterObj = {}, downloadOptions = {}} = {}) => {
-    const data = toOGCFilter(filterObj.featureTypeName, filterObj, filterObj.ogcVersion, filterObj.sortOptions, false, null, null, downloadOptions.selectedSrs);
-    return Rx.Observable.defer( () =>
-        axios.post(cleanDuplicatedQuestionMarks(url + `?service=WFS&outputFormat=${downloadOptions.selectedFormat}`), data, {
-            timeout: 60000,
-            responseType: 'arraybuffer',
-            headers: {'Content-Type': 'application/xml'}
-        }));
+const getWFSFeature = ({ url, filterObj = {}, layerFilter, downloadOptions = {}, options } = {}) => {
+    const { sortOptions, propertyName: pn } = options;
```
allyoucanmap, line 109: it seems that the `downloadOptions.selectedSrs` is not used anymore, I think we should not remove this option.
Could you explain where is located in the current implementation?

web/client/observables/wfs.js
```diff
@@ -149,6 +148,50 @@ export const getJSONFeature = (searchUrl, filterObj, options = {}) => {
...
+ * @param {string} searchUrl URL of WFS service
+ * @param {object} filterObj FilterObject
+ * @param {*} downloadOption selected format for download
+ * @returns {Observable} a stream that emits the XML
+ */
+export const getXMLFeature = (searchUrl, filterObj, options = {}, downloadOption) => {
+
+    if (options.layer && options.layer.type === 'vector') {
+        return getFeatureUtilities(searchUrl, filterObj, options, downloadOption);
+    }
```
allyoucanmap, line 168: Are we sure we need to repeat this control?
It's a bit strange that the getFeatureUtilities can return an object or a observable

web/client/observables/wfs.js
```diff
@@ -149,6 +148,50 @@ export const getJSONFeature = (searchUrl, filterObj, options = {}) => {
...
+ * Get Features in json format. Intercepts request with 200 errors and workarounds GEOS-7233 if `totalFeatures` is passed
+ * @param {string} searchUrl URL of WFS service
+ * @param {object} filterObj FilterObject
+ * @param {number} totalFeatures optional number to use in case of a previews request, needed to workaround GEOS-7233.
+ * @return {Observable} a stream that emits the GeoJSON or an error.
+ */
+export const getJSONFeature = (searchUrl, filterObj, options = {}) => {
+
+    if (options.layer && options.layer.type === 'vector') {
+        return getFeatureUtilities(searchUrl, filterObj, options);
```
allyoucanmap, line 190: same as previous comment

allyoucanmap (changes requested):

web/client/observables/wfs.js
DavidQuartz: Both `getJSONFeature` and `getXMLFeature` return observables. However, they both have similar logic and may return the same observable if `options.layer.type === 'vector'`. To ensure optimisation, they share a `getFeatureUtilities` utility function to provide them with data they need to build their respective observables. However, in the case where. In the case where `options.layer.type === 'vector'`, `getFeatureUtilities` will return the Observable needed so that both functions can send it back to their outer epics.

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo  moving this to test, please let us now when we can backport this fix

ElenaGallo:
Test passed, @DavidQuartz please backport to 2021.02.xx. Thanks

Issue #7508, opened 2021-11-05
Annotation plugin crash because of invalid geometry


alexander-fedorenko:
## Description
There is a set of use-cases that leads to app getting crashed while editing annotation.
All these cases are related to the invalid geometry (empty coordinates)

### Case 1: Attempt to edit annotation that was saved when it had invalid geometry.
1. Create new annotation or edit existing one. Specify title.
2. Create new geometry and add one point with empty coordinates, line or polygon.
3. Deselect geometry and save annotation.
4. Open annotation.

*Expected Result*
App opens toolbar to edit annotation

*Current Result*
App crashes

https://user-images.githubusercontent.com/4226620/140523658-6a6a47a5-dc41-4705-9b7f-e926665cfcb4.mp4

### Case 2: Attempt to create new geometry while geometry with empty coordinate exists in the list.
1. Create new annotation or edit existing one.
2. Create new geometry and add one point with empty coordinates, line or polygon.
3. Deselect geometry.
4. Try to create new geometry of any type.

*Expected Result*
App opens toolbar to edit geometry

*Current Result*
App crashes

https://user-images.githubusercontent.com/4226620/140523688-e0a5803b-5a4a-47f8-8f12-c815150573d7.mp4

### Case 3 Geometry is marked as incorrect while it's correct ([https://github.com/geosolutions-it/MapStore2/issues/7395#issuecomment-945618853](link)):

1. Go to annotation plugin
2. Click on Plus icon button
3. Enter title value in textbox
4. Click on line from geometries section at bottom.
5. Add 2 Coordinates for the line then Deselect that.
6. Add One more line geometry.
7. Deselect that.

Issue:
    Even if 2 Coordinates are given in the first geometry while selecting any other geometry it shows error sign.

https://user-images.githubusercontent.com/92375187/137712440-d6f1e578-e866-4b87-a2d7-f1526dc0d733.mp4

### Case 4 "Point" geometry is created unexpectedly on attempt to modify "Text" geometry coordinates
1. Go to annotation plugin
2. Create new or edit existing annotation.
3. Add "Text" geometry.
4. Click on the map to get coordinates pre-filled.
5. Remove value from one coordinate to make geometry invalid.
6. Click on the map somewhere.

*Expected Result*
Text geometry coordinates updated to the point clicked.

*Current Result*
New "Point" geometry is created and gets selected for editing. This geometry can't be removed and app will crash on attempt to edit point coordinated before text validation errors are resolved.
[...]

PR #7531, opened 2021-11-10
#7508 Annotation plugin crash because of invalid geometry

alexander-fedorenko:
## Description
There is a set of conditions that lead to app crash when geometries attached to the annotation are edited.
All these cases related to invalid geometry because one of its coordinates is null.

This PR provides following changes:
1. User will have to fix all validation errors of the selected geometry before change to another geometry or new geometry creation.
2. Geometries that are not selected will be inactive (like if they were disabled elements) in the list if currently selected geometry has validation errors. 
3. User's click on inactive geometries won't trigger change of selected geometry.
4. User's click on active geometry won't deselect it if there are validation errors.
5. Validation mark is removed from geometries list and kept inside geometry edit form. No sense to keep it in two places, only one geometry at a time might be invalid.
6. It's not possible to save invalid geometry. It's not possible to save annotation with invalid geometries.

This concept will make impossible to save geometries with invalid coordinates.

Other changes:
- Changed behavior when user click on the map while invalid "Text" or "Marker" geometry is inactive.
- Invalidate Circle geometry properly if radius is not set.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7508 

**What is the new behavior?**
- It's not possible to change selected geometry when there are validation errors.
- Click on the map while "Text" and "Marker" geometry is invalid won't trigger new geometry creation.
- Circle geometry was not invalidated inside of editing.features state branch, this led to the issue when geometry marked as invalid in edit form, but remains valid on the level of geometries list. This behavior is fixed.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/actions/annotations.js
+++ b/web/client/actions/annotations.js
@@ -63,7 +63,7 @@ export const HIDE_MEASURE_WARNING = 'ANNOTATIONS:HIDE_MEASURE_WARNING';
 export const TOGGLE_SHOW_AGAIN = 'ANNOTATIONS:TOGGLE_SHOW_AGAIN';
 export const GEOMETRY_HIGHLIGHT = 'ANNOTATIONS:GEOMETRY_HIGHLIGHT';
 export const UNSELECT_FEATURE = 'ANNOTATIONS:UNSELECT_FEATURE';
-export const SET_IS_VALID_FEATURE = 'ANNOTATIONS:SET_IS_VALID_FEATURE';
+export const VALIDATE_FEATURE = 'ANNOTATIONS:VALIDATE_FEATURE';
 
 export const initPlugin = () => ({
     type: INIT_PLUGIN
@@ -423,7 +423,7 @@ export const toggleShowAgain = () => ({
     type: TOGGLE_SHOW_AGAIN
 });
 
-export const setIsValidFeature = () => ({
-    type: SET_IS_VALID_FEATURE
+export const featureValidationCheck = () => ({
+    type: VALIDATE_FEATURE
 });
 
--- a/web/client/actions/annotations.js
+++ b/web/client/actions/annotations.js
@@ -423,7 +423,7 @@ export const toggleShowAgain = () => ({
     type: TOGGLE_SHOW_AGAIN
 });
 
-export const featureValidationCheck = () => ({
+export const validateFeature = () => ({
     type: VALIDATE_FEATURE
 });
 
--- a/web/client/actions/annotations.js
+++ b/web/client/actions/annotations.js
@@ -63,6 +63,7 @@ export const HIDE_MEASURE_WARNING = 'ANNOTATIONS:HIDE_MEASURE_WARNING';
 export const TOGGLE_SHOW_AGAIN = 'ANNOTATIONS:TOGGLE_SHOW_AGAIN';
 export const GEOMETRY_HIGHLIGHT = 'ANNOTATIONS:GEOMETRY_HIGHLIGHT';
 export const UNSELECT_FEATURE = 'ANNOTATIONS:UNSELECT_FEATURE';
+export const SET_IS_VALID_FEATURE = 'ANNOTATIONS:SET_IS_VALID_FEATURE';
 
 export const initPlugin = () => ({
     type: INIT_PLUGIN
@@ -421,3 +422,8 @@ export const hideMeasureWarning = () => ({
 export const toggleShowAgain = () => ({
     type: TOGGLE_SHOW_AGAIN
 });
+
+export const setIsValidFeature = () => ({
+    type: SET_IS_VALID_FEATURE
+});
+
--- a/web/client/components/mapcontrols/annotations/AnnotationsEditor.jsx
+++ b/web/client/components/mapcontrols/annotations/AnnotationsEditor.jsx
@@ -244,7 +244,7 @@ class AnnotationsEditor extends React.Component {
         onInitPlugin: PropTypes.func,
         onGeometryHighlight: PropTypes.func,
         onUnSelectFeature: PropTypes.func,
-        onFeatureValidationCheck: PropTypes.func
+        onValidateFeature: PropTypes.func
     };
 
     static defaultProps = {
@@ -351,7 +351,7 @@ class AnnotationsEditor extends React.Component {
     };
 
     renderEditingCoordButtons = () => {
-        const allGeometryIsValid = this.allGeometryIsValid();
+        const areAllFeaturesValid = this.validateFeatures();
         return (<Grid className="mapstore-annotations-info-viewer-buttons" fluid>
             <Row className="text-center noTopMargin">
                 <Col xs={12}>
@@ -361,6 +361,7 @@ class AnnotationsEditor extends React.Component {
                             {
                                 glyph: 'arrow-left',
                                 tooltipId: "annotations.back",
+                                tooltipPosition: 'bottom',
                                 visible: true,
                                 disabled: this.props?.selected?.properties
                                     && !this.props?.selected?.properties?.isValidFeature || false,
@@ -386,6 +387,7 @@ class AnnotationsEditor extends React.Component {
                             }, {
                                 glyph: 'trash',
                                 tooltipId: "annotations.remove",
+                                tooltipPosition: 'bottom',
                                 disabled: !this.props.annotations.length,
                                 visible: !this.props.selected,
                                 onClick: () => {
@@ -393,13 +395,15 @@ class AnnotationsEditor extends React.Component {
                                 }
                             }, {
                                 glyph: 'floppy-disk',
-                                tooltipId: !allGeometryIsValid ? "annotations.geometryError" : !isEmpty(this.props.selected) ? "annotations.saveGeometry" : "annotations.save",
-                                disabled: (this.props.selected && this.props.selected.properties && !this.props.selected.properties.isValidFeature) || !allGeometryIsValid,
+                                tooltipPosition: 'bottom',
+                                tooltipId: !areAllFeaturesValid ? "annotations.annotationSaveGeometryError" : !isEmpty(this.props.selected) ? "annotations.saveGeometry" : "annotations.save",
+                                disabled: (this.props.selected && this.props.selected.properties && !this.props.selected.properties.isValidFeature) || !areAllFeaturesValid,
                                 onClick: () => this.save()
                             },
                             {
                                 glyph: 'download',
                                 tooltip: <Message msgId="annotations.downloadcurrenttooltip" />,
+                                tooltipPosition: 'bottom',
                                 disabled: Object.keys(this.validate()).length !== 0 || this.props.unsavedChanges,
                                 visible: !this.props.selected,
                                 onClick: () => {
@@ -500,8 +504,8 @@ class AnnotationsEditor extends React.Component {
                     geodesic={this.props.geodesic}
                     defaultPointType={this.getConfig().defaultPointType}
                     defaultStyles={this.props.defaultStyles}
-                    onFeatureValidationCheck={this.props.onFeatureValidationCheck}
-                    allGeometryIsValid={this.allGeometryIsValid}
+                    onValidateFeature={this.props.onValidateFeature}
+                    validateFeatures={this.validateFeatures}
                 />
                 }
             </div>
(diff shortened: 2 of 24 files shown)


alexander-fedorenko:
Documentation update is pending to be done after review.
New tests to check that annotation and/or geometry cannot be saved when selected geometry is invalid are TBD.

web/client/components/mapcontrols/annotations/AnnotationsEditor.jsx
```diff
@@ -764,6 +769,16 @@ class AnnotationsEditor extends React.Component {
...
 
+    allGeometryIsValid = () => {
+        let allGeometryIsValid = true;
+        if (this.props?.editing?.features && this.props.editing?.features?.length > 0) {
+            this.props.editing.features.map((p) => {
+                if (!p?.properties?.isValidFeature) { allGeometryIsValid = false; }
+            });
+        }
+        return allGeometryIsValid;
+    };
```
MV88, line 780: personally i don't like this too much because the check "this.props.editing?.features?.length > 0" is superfluous and in this case it would have been better to use a .forEach instead of a .map

but i think this one it's shorter and easier to read
```suggestion
const areAllGeometriesValid = this.props?.editing?.features?.reduce((valid, feature) => {
    return valid && p?.properties?.isValidFeature;
}, true);
return areAllGeometriesValid;
```

or better with every makes it even clearer

```suggestion
let areAllGeometriesValid = true;
if (this.props?.editing?.features) {
    areAllGeometriesValid = this.props?.editing?.features?.every(feature => feature?.properties?.isValidFeature)
}
return areAllGeometriesValid ;
```

web/client/components/mapcontrols/annotations/FeaturesList.jsx
```diff
@@ -36,10 +37,13 @@ const FeaturesList = (props) => {
...
         geodesic,
         defaultStyles,
-        defaultPointType
+        defaultPointType,
+        onFeatureValidationCheck,
+        allGeometryIsValid
     } = props;
     const {features = []} = editing || {};
     const isValidFeature = get(props, "selected.properties.isValidFeature", true);
+    const allFeaturesAreValid = allGeometryIsValid();
```
MV88, line 46: in general conditional flags should be named starting with a verb
or at least using correct plural form

```suggestion
    const areAllFeaturesValid = validateGeometries();
```

or 

```suggestion
    const areFeaturesValid = validateFeatures();
```
maybe we can call it validateFeatures

web/client/epics/annotations.js
```diff
@@ -619,14 +623,14 @@ export default {
             // TODO update selected feature in editing features
 
             let selectedIndex = findIndex(feature.features, (f) => f.properties.id === selected.properties.id);
-            if (selected.properties.isValidFeature || selected.geometry.type === "LineString" || selected.geometry.type === "MultiPoint" || selected.geometry.type === "Polygon") {
+            if (selected.properties.isValidFeature || includes(featureTypes, selected.geometry.type)) {
```
MV88, line 626: in this way you are changing the number of types that the previous condition was setup
for example here it will be included Point where before it wasn't
is it intended?

web/client/reducers/annotations.js
```diff
@@ -197,6 +198,8 @@ function annotations(state = {validationErrors: {}}, action) {
     }
     case FEATURES_SELECTED: {
         let newState = state;
+
```
MV88, line 201: not a big deal, but before creating the PR check for things like this

empty lines are not needed and increase a little bit entropy in the code review
unless they are intentionally added to increase readability

web/client/reducers/annotations.js
```diff
@@ -743,6 +754,21 @@ function annotations(state = {validationErrors: {}}, action) {
     }
     case START_DRAWING:
         return {...state, config: {...state.config, geodesic: get(action.options, 'geodesic', false)}};
+    case VALIDATE_FEATURE:
+
+        let updatedFeatures = state.editing.features;
```
MV88, line 759: doing a map and not assigning to anything is not correct and mutating things like this are not wrong per se, but we try to use a different way
```suggestion
        let updatedFeatures = state.editing.features.map((f) => {
             const isValidFeature = validateFeature({
                properties: f.properties,
                components: getComponents(f.geometry),
                type: f.geometry.type
            });
        return {...f, properties: {...f.properties, isValidFeature }}
        });
```
OR
```suggestion
        let updatedFeatures = state.editing.features.map((f) => {
             const isValidFeature = validateFeature({
                properties: f.properties,
                components: getComponents(f.geometry),
                type: f.geometry.type
            });
            return set(`properties.isValidFeature`, isValidFeature, f);
        });
```
with set we do not mutate the original object but we return a new object mutated

web/client/translations/data.it-IT.json
```diff
@@ -1153,7 +1153,9 @@
                 "notValidText": "Inserisci un testo ed una coordinata valida (+|- 90° lat, +|-180° lon)",
                 "notValidCircle": "Inserisci un raggio ed una coordinata valida (+|- 90° lat, +|-180° lon)",
                 "newFeature": "Nuovo"
-            }
+            },
+            "geometryError": "La configurazione della geometria non è valida",
+            "resolveAllErrors": "Si prega di correggere eventuali errori di convalida prima di modificare altre geometrie"
```
MV88, line 1158:
```suggestion
            "resolveAllErrors": "Si prega di correggere gli errori prima di modificare altre geometrie"
```

web/client/translations/data.nl-NL.json
```diff
@@ -1172,7 +1172,7 @@
                 "notValidPolyline": "Alle coördinaten moeten geldig zijn (+ | - 90 ° lat, + | -180 ° lon)",
                 "notValidText": "Voeg een tekstwaarde en een geldige coördinaat in (+ | - 90 ° lat, + | -180 ° lon)",
                 "notValidCircle": "Voeg een straalwaarde en een geldige coördinaat in (+ | - 90 ° lat, + | -180 ° lon)",
-                "newFeature": "Nieuw"
+                "newFeature": "Nieuw"
```
MV88, line 1175: not needed

web/client/translations/data.sv-SE.json
```diff
@@ -1153,7 +1153,7 @@
                 "notValidPolyline": "All koordinat måste vara giltig ( +|- 90 ° lat, +| -180 ° lon)",
                 "notValidText": "Infoga ett textvärde och en giltig koordinat ( +|- 90 ° lat, +| -180 ° lon)",
                 "notValidCircle": "Infoga ett radievärde och en giltig koordinat ( +|- 90 ° lat, +| -180 ° lon)",
-                "newFeature": "Ny"
+                "newFeature": "Ny"
```
MV88, line 1156: not needed

web/client/utils/AnnotationsUtils.js
```diff
@@ -622,8 +621,11 @@ export const isAMissingSymbol = (style) => {
  * @return {boolean} true if it is a valid polygon, false otherwise
 */
 export const isCompletePolygon = (coords = [[[]]]) => {
-    const validCoords = coords[0].filter(AnnotationsUtils.validateCoordsArray);
-    return validCoords.length > 3 && head(validCoords)[0] === last(validCoords)[0] && head(validCoords)[1] === last(validCoords)[1];
+    if (coords && coords[0]) {
```
MV88, line 624: please add or amend a unit test

MV88 (changes requested):
@alexander-fedorenko overall good job

the new workflow seems working good, but i've found something wrong plus there are a few comments here and there

- [ ] improvement
probably we can do different here when we have an invalid geometry, because the users can think that also the buttons delete an zoom on "disabled features" are not active
image
@allyoucanmap what if we keep the toolbar on the same background as they are normally

- [ ] BUG 1
i'm able to deselect circles if radius is invalid

- [ ] BUG 2
circles are no longer editable from map when they pass trough an invalid state

for bugs refer to this video
https://user-images.githubusercontent.com/11991428/141277118-66490708-08d5-4db5-a1d3-d1c32199c064.mp4

- [ ] enhancement
please add missing unit tests
you can test them one by one by using npm run continoustest and editing tests.webpack.js
see [here](https://mapstore.readthedocs.io/en/latest/developer-guide/developing/#frontend-unit-tests) for more details

web/client/epics/annotations.js
alexander-fedorenko: Yes, this is intended. This specific change resolves the issue when new "Point" geometry is created when user clicks on the map while currently selected Text or Marker is invalid. Both Marker end text are "Point" feature with custom properties, and old condition was forcing creation of a new feature if selected is invalid. With new condition invalid feature is not being filtered out and eventually coordinates of the point are being updated after click.

web/client/utils/AnnotationsUtils.js
alexander-fedorenko: Updated unit tests for cases when polygon has 0 and 1 coordinate added.

web/client/translations/data.sv-SE.json
alexander-fedorenko: Removed

web/client/translations/data.nl-NL.json
alexander-fedorenko: Removed

web/client/translations/data.it-IT.json
alexander-fedorenko: Thanks, corrected

web/client/reducers/annotations.js
alexander-fedorenko: Thanks, corrected!

web/client/reducers/annotations.js
alexander-fedorenko: Thanks, got rid of unneeded empty lines.

web/client/components/mapcontrols/annotations/FeaturesList.jsx
alexander-fedorenko: Thanks, renamed into areFeaturesValid & validateFeatures

web/client/components/mapcontrols/annotations/AnnotationsEditor.jsx
alexander-fedorenko: Thanks, used second version of suggested changes but modified condition to check if this.props?.editing?.features is array and removed optional chaining before "every" since we already check array existence in condition.

web/client/components/mapcontrols/annotations/AnnotationsEditor.jsx
```diff
@@ -350,6 +351,7 @@ class AnnotationsEditor extends React.Component {
     };
 
     renderEditingCoordButtons = () => {
+        const allGeometryIsValid = this.allGeometryIsValid();
```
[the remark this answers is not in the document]
alexander-fedorenko: Thanks, renamed as in second suggestion.

web/client/actions/annotations.js
```diff
@@ -421,3 +422,8 @@ export const hideMeasureWarning = () => ({
 export const toggleShowAgain = () => ({
     type: TOGGLE_SHOW_AGAIN
 });
+
+export const featureValidationCheck = () => ({
```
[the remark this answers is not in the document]
alexander-fedorenko: Added 2 unit tests to check how validation works when all geometry types are valid, invalid. Also added 4 specific cases for text and circle when entity become invalid if either of two inputs (coordinates/text, center/radius) is invalid.

allyoucanmap:
> @allyoucanmap what if we keep the toolbar on the same background as they are normally

@MV88 your proposal seems consistent with the state of the button so it's ok for me

web/client/actions/annotations.js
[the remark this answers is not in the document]
MV88: ok nice, we need one for action creator too

MV88 (changes requested):
ok seems to work fine, 

- [ ] please add a unit test for the new action
- [ ] remove graying out of toolbar background in the list, since they should be always with default background

MV88 (approved):
LGTM

MV88:
@ElenaGallo please test this in DEV and keep note of amending the user guide

ElenaGallo:
@alexander-fedorenko reported here a small issue related to this PR:

*  The gray error page is displayed when there is invalid geometry and the user attempts to delete another geometry.

**How to Reproduce**
See the steps reported in the gif below:

1_anno

Issue #7516, opened 2021-11-09
Implement bearer authentication support for WMS GetMap requests


mbarto:
## Description

We want to add support for authentication through a bearer Authorization header in WMS GetMap requests (as needed by https://github.com/geosolutions-it/webmapper-halliburton/issues/2080).

Both ol and cesium WMS layer implementation will support this.

## Acceptance criteria

- [ ] WMS layers implementation for ol and Cesium will send the user access_token as a bearer Authorization header if configured through the localConfig authenticationRules with the bearer method

offtherailz:
See this comment. @mbarto Probably the PR causes cross-origin requests where before were not needed. Please investigate.
https://github.com/geosolutions-it/MapStore2/pull/7518#issuecomment-966438590

mbarto:
@offtherailz I should have fixed the issue with https://github.com/geosolutions-it/MapStore2/pull/7548

PR #7548, opened 2021-11-15
#7516: fixed usage of CORS for loading WMS tiled images in Cesium WMS…

mbarto:
…Layer implementation

## Description

Cross origin requests handling has changed in Cesium 1.42 and following, in comparison to 1.17, so some requests to WMS services were failing after the upgrade, because they were not correctly recognized as cross-origin (handled by ajax requests), but as normal images. This was causing a security error when tiles had to drawn on the WebGL context.

To avoid the security error we must ensure those image are requested through ajax. A way to force that when needed is to set a tileDiscardPolicy on the imagery provider. By using the NeverTileDiscardPolicy implementation, we ensure we don't have other side effects (this implementation does nothing). There are other ways to change this behaviour, such as setting headers on the request, but none of them is as safe as using NeverTileDiscardPolicy.

We have also introduced a tileDiscardPolicy option on cesium layers so that if for any reason a project needs to switch back to loading tiles as images for specific services, it can by setting "none" in this option.

Note also that tileDiscardPolicy can also be used to set a real policy, for example to check for missing/broken tiles and load an alternative image, or send events.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#7586 
#7516 

Security error on some WMS service layers with tiled configuration.

**What is the new behavior?**

No more security errors on WMS services layers with tiled configuration.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/map/cesium/plugins/WMSLayer.js
+++ b/web/client/components/map/cesium/plugins/WMSLayer.js
@@ -104,7 +104,9 @@ function wmsToCesiumOptions(options) {
         }),
         // #7516 this helps Cesium to use CORS requests in a proper way, even when headers are not
         // present in the Resource
-        tileDiscardPolicy: options.noCors ? undefined : new Cesium.NeverTileDiscardPolicy(),
+        tileDiscardPolicy: options.tileDiscardPolicy === "none" ?
+            undefined :
+            (options.tileDiscardPolicy ?? new Cesium.NeverTileDiscardPolicy()),
         credit,
         subdomains: urls,
         layers: options.name,
--- a/web/client/components/map/cesium/plugins/WMSLayer.js
+++ b/web/client/components/map/cesium/plugins/WMSLayer.js
@@ -102,6 +102,9 @@ function wmsToCesiumOptions(options) {
             headers,
             proxy: proxy && new WMSProxy(proxy) || new NoProxy()
         }),
+        // #7516 this helps Cesium to use CORS requests in a proper way, even when headers are not
+        // present in the Resource
+        tileDiscardPolicy: options.noCors ? undefined : new Cesium.NeverTileDiscardPolicy(),
         credit,
         subdomains: urls,
         layers: options.name,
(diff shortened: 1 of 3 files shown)


offtherailz (commented):
I wasn't able to understand how "noCors" option, that I suppose is false by default,  fixes existing maps. Anyway I tried and it is this.
I must have missed something...

mbarto:
@offtherailz 

"We introduced a noCors flag on layers options if, for any reason a project needs to switch back to loading tiles as images for specific services."

The noCors flag can be used to force loading wihout a CORS request (via a simple image object, instead of xhr or fetch). I don't foresee any need for actual maps, If you think it's confusing, we can just remove it.

web/client/components/map/cesium/plugins/WMSLayer.js
```diff
@@ -102,6 +102,9 @@ function wmsToCesiumOptions(options) {
             headers,
             proxy: proxy && new WMSProxy(proxy) || new NoProxy()
         }),
+        // #7516 this helps Cesium to use CORS requests in a proper way, even when headers are not
+        // present in the Resource
+        tileDiscardPolicy: options.noCors ? undefined : new Cesium.NeverTileDiscardPolicy(),
```
offtherailz, line 107: let's use options.tileDiscardPolicy :

- tileDiscardPolicy = undefined -> NeverTileDiscardPolicy -- comment the reason of using this as default  in this line, as you did, linking the issue and the permalink to the code, so it is clear why we force it. 
- tileDiscardPolicy = 'none' -> undefined -- use defaults of cesium
Open for other customizations, as object with configuration: 
https://cesium.com/downloads/cesiumjs/releases/1.17/Build/Documentation/DiscardMissingTileImagePolicy.html

Permalink to code of cesium: https://github.com/CesiumGS/cesium/blob/48c8abaca6257b568edc8c992ede5f2c487cdd38/Source/Scene/ImageryProvider.js#L366

mbarto:
@offtherailz I have updated the PR

offtherailz (approved):

offtherailz:

PR #7570, opened 2021-11-17
[Backport 2021.02.xx] Fix for annotations crash because of invalid geometry (#7508)

alexander-fedorenko:
Fix for annotations crash because of invalid geometry (#7508)
PR:
#7531 
#7563

PR #7541, opened 2021-11-12
Geometries become invisible when Marker or Text element is being added #7511

alexander-fedorenko:
## Description
All annotation geometries become invisible when user adds new Marker or Text to the map and before coordinates of new geometry are set.
This PR resolves this issue by partially rolling back changes introduced by #7345

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**
#7511 

**What is the new behavior?**
All geometries stay visible when user adds Marker or Text and before new geometry gets proper coordinate.

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

Issue #7511, opened 2021-11-08
 Annotation plugin - geometries become invisible when Marker or Text element is being added


alexander-fedorenko:
## Description
Geometries on the map become invisible when new Text or Point element is being added to the annotation.

### Case 1: Geometries become invisible while you create new marker or text.
1. Create new annotation or edit existing one.
2. Create new geometry, e.g. polygon or line.
3. Deselect geometry, check that it's visible on the map.
4. Create new point or text geometry.

*Expected Result*
All existing geometries displayed on the map while text or point is in edit mode.

*Current Result*
All annotation geometries become invisible.

https://user-images.githubusercontent.com/4226620/140523717-d8261a82-8738-4ab9-8da7-9ff92db9b074.mp4

### Case 2: Attempt to edit coordinates of any existing geometry ends up with an error if there is a marker or point with empty coordinate.
1. Create new annotation or edit existing one.
2. Create new geometry, e.g. polygon or line.
3. Deselect geometry, check that it's visible on the map.
4. Create new point or text geometry. Skip coordinates setup. 
5. Deselect geometry.
6. Select it again. Add coordinate but keep lat/lon empty and invalid.
7. Deselect geometry.
8. Select polygon and try to edit any coordinate with arrows at the right side of the input.

*Expected Result*
Polygon coordinates are updated.

*Current Result*
App crashes.

https://user-images.githubusercontent.com/4226620/140523753-0ee97461-4e78-48b6-bdfb-949dbc87b884.mp4

- [x] Not browser related

## Other useful information

None for now

alexander-fedorenko:
This is regression bug introduced by #7393
Changes there is one-line fix and this fix can be safely reverted since after #7508 app will make sure that geometry coordinates are valid before giving user opportunity to save data.
It might need additional investigation to check if initializing Marker and Text with geometry = null is a proper way of doing things.

thga489:
salut

PR #7573, opened 2021-11-17
[Backport 2021.02.xx] Geometries become invisible when Marker or Text element is being added #7511

alexander-fedorenko:
Geometries become invisible when Marker or Text element is being added #7511
All logic is already backported via #7570
PR:
#7541 
#7557


MV88 (approved):

MV88:
@ElenaGallo please test in QA

Issue #7577, opened 2021-11-18
User Guide - Add new editing capabilities for annotations


ElenaGallo:
## Description

Update the [Annotations section](https://mapstore.readthedocs.io/en/latest/user-guide/annotations/) section with this #7508

*Documentation section involved*
- [x] User Guide
- [ ] Developer Guide

PR #7581, opened 2021-11-19
User Guide - Note on invalid vertices on annotations

ElenaGallo:
## Description

Add note on invalid vertices on Annotations section

## Other useful information
Related to #7577  
Read the docs [link](https://elena-mapstore-docs.readthedocs.io/en/7577_userguide/user-guide/annotations/)


MV88 (changes requested):
@ElenaGallo can you change the two images:
- docs/user-guide/img/annotations/invalid_vertex.jpg
- docs/user-guide/img/annotations/polygon_annotation_drawing.gif

to show also other valid geometries that are greyed out when you have an invalid vertex?

please, mention also that you can still delete or zoom other geometries while in this invalid situation, maybe in the gif

Also what if we use an mp4 instead of a gif?
can you create it and try using it inside read the docs?

we could use something like
```
<video width="320" height="240" autoplay>
  <source src="docs/user-guide/img/annotations/polygon_annotation_drawing.mp4" type="video/mp4">
</video>
```

ElenaGallo:
@MV88 Using mp4 is an excellent solution and I will definitely talk about it with Tobia but, for the moment, I would leave the gif so everything is more uniform.

MV88 (approved):

Issue #7585, opened 2021-11-22
Refactor Print plugin as a plugin container


mbarto:
## Description

MapStore printing dialog plugin is currently configured as a static list of controls.

To enable plugging new / different controls in projects we want to:

 - refactor the Print plugin/component as a plugins container
   - [x] items property support, with different target containers (left-panel, left-panel-accordion, etc.)
   - [x] NON singleton items support in PluginsUtils.getPluginsItems (same plugin / different id)
 - refactor actual controls (all or some of them) into plugins and inject them by configuration
   - [x] generic TextInput plugin (used for title, description, etc.)
   - [x] Layout plugin (layout sheet and options)
   - [x] LegendOptions plugin
   - [x] Resolution plugin
   - [x] MapPreview plugin
 - to retain backward compatibility, the default configuration will inject the actual controls
   - [x] standardItems in plugins/print/index.js

Connected to https://github.com/geosolutions-it/mapfish-print/issues/39

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #7596, opened 2021-11-23
#7585: refactor of the Print plugin as a plugins container

mbarto:
## Description

This PR is a refactor of the Print plugin, to allow for custom UI plugins to be included from projects.
Most of the printing dialog widgets have been refactored as subplugins of the Print plugin.

Some ancillary work has been included to improve usability:
 - support for different implementations of the same plugin in PluginsUtils.getPluginItems
 - support of testing lazy plugins with pluginsTestUtils.getLazyPluginForTest
 - getByXPath function (in Print-test.jsx to be eventually moved into a utility file) to allow a more react-testing-library like approach for testing
 - support for custom parameters (set in the spec params object) in PrintUtils.getMapfishPrintSpecification

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#7585 

#7585 

**What is the new behavior?**

## Breaking change
 - [x] No


offtherailz (commented):
It's a little scary to me to have too many plugins by default. We already had some doubts with timeline + playback. 

Can I suggest to use plugin injection as replacement for defaults components when you have a specific target. 
E.g. when a plugin as a target like the following in the `containers`: 
```
{ target:  "TextInput"
|        "Layout"
|        "LegendOptions"
|        "Resolution"
|        "MapPreview"
|        "Option"
```
The  (documented) convention of the plugin is that these target are replacements for default components.
I think this solution is more maintainable on the core.
What do you think?

mbarto:
@offtherailz how would you add a custom plugin in a specific position? For example between title and description in the current printing dialog?
I like using the target as a replacement, it gives more flexibility. Do you think we should also handle removal of a "default" widget ?

offtherailz:
> @offtherailz how would you add a custom plugin in a specific position? For example between title and description in the current printing dialog?
In this case I should replace the whole panel (TextInput??).
I think you can be able to:
- Replace default panels
- Add some panels

Moreover I read on mailing list they was asking for adding a button to save on cloud the pdf. So future we may also allow to add some buttons here and there. 


> Do you think we should also handle removal of a "default" widget ?
Now I think they are decoupled so every panel is idependent and I think it makes sense, Maybe injecting an null component for the target may work?

mbarto:
@offtherailz to summarize the changes you suggest should be:
 - leave the default (connected) components untouched in plugins/print/index.js
 - change Print.jsx to allow replacements (or removal) of those components via plugins with a documented target (removal happens if the plugin has a null component definition)
 - allow addition of custom plugins in predefined targets (e.g. left-panel, right-panel, etc.) as already done by this PR

Is it correct?

offtherailz:
Yes @mbarto , correct. Hope it makes the plugin more flexible and maintainable.

mbarto:
@offtherailz I have updated the PR. The build is currently failing due to webpack/nodejs issues (I guess)

web/client/product/plugins.js
```diff
@@ -124,7 +125,13 @@ export default {
...
         ZoomAllPlugin: require('../plugins/ZoomAll').default,
         ZoomInPlugin: require('../plugins/ZoomIn').default,
-        ZoomOutPlugin: require('../plugins/ZoomOut').default
+        ZoomOutPlugin: require('../plugins/ZoomOut').default,
+        PrintTextInputPlugin: require('../plugins/print/TextInput').default,
+        PrintLayoutPlugin: require('../plugins/print/Layout').default,
+        PrintLegendOptionsPlugin: require('../plugins/print/LegendOptions').default,
+        PrintResolutionPlugin: require('../plugins/print/Resolution').default,
+        PrintMapPreviewPlugin: require('../plugins/print/MapPreview').default,
+        PrintOptionPlugin: require('../plugins/print/Option').default
```
offtherailz, line 134: My idea was not to include any plugin, but only allow to override by plugin injection, on demand. Having this big set of plugin to configure or customize a single plugin is a problem. Potentially makes the size of the plugins list explode. 

I asked to have a plugin like : 
```javascript 
{
    "name": "PluginCustomization",
    "containers": {
            "Print": {
                    target: "left-panel-accordion", // I suggested to have specific targets to replace specific components.
                    position:  "" 
                    Component: MyCustomization,
            }
    }
}
```

web/client/plugins/Print.jsx
```diff
@@ -74,29 +98,86 @@ import { isInsideResolutionsLimits } from '../utils/LayersUtils';
...
+ *   "name": "Null",
+ *   "override": {
+ *       "Print": {
+ *          "target": "name",
+ *          "position": 1
+ *       }
+ *   },
+ *   "cfg": {
+ *   }
+ * }
```
offtherailz: Instead of having 3 examples and a lot of configurations and code, I suggest to use one single plugin, with no component, using the multi-target support. 
Please try it. 

```suggestion
// PrintCustomizations.js
const MyCustomPanel = () => <div>Hello, I am a custom component</div>
const MyCustomLayout = connect(mapStateToProps, dispatchToProps)(CustomLayoutSelector);
export default  createPlugin('PrintCustomizations',
  containers: [
  // this entry allows to add a new panel between text input and accordion panel
   {
       Print: [{ // this entry add a panel 
            target: "left-panel",
            position: 1.5,
            Component: MyCustomPanel
}, 
// this entry allows to
{
            target: "layout",
            Component: MyCustomLayout
}, {
           target: "map-preview", // To remove one component, simply create a component that returns null.
           Component: () => null 
}]);

```

web/client/plugins/Print.jsx
```diff
@@ -35,6 +35,30 @@ import { isInsideResolutionsLimits } from '../utils/LayersUtils';
...
+ *  - layout (left-panel-accordion, position: 1)
+ *  - legendoptions (left-panel-accordion, position: 2)
+ *  - resolution (right-panel, position: 1)
+ *  - mappreview (right-panel, position: 2)
+ *  - defaultbackgroundignore (right-panel, position: 3)
+ *  - submit (buttons, position: 1)
+ *  - printpreview (preview-panel, position: 1)
+ *
+ * To remove a widget, you have to include a Null plugin with the desired target.
+ * You can use the position to sort existing and custom items.
```
offtherailz, line 60: You can customize `Print` plugin by creating one custom plugin (or more) that modifies the existing components with your ones. You can configure this plugin in localConfig.json as usual.

offtherailz (changes requested):
It's ok, now I understood what you did. 
- [ ] Please look ad my comments, I think that applying the changes everything will seems easier to configure and maintain
- [ ] Pleas for targets use names with dashes for everything, for constiency, as you did for `left-panel` e.g. `printoptions` --> `print-options`, `defaultbackgroundignore` -> `default-background-ignore`

tdipisa:
This PR has been merged without an approval @mbarto. @offtherailz please check it as soon as you can.

offtherailz:
My review was on 2 points: 
- [x] Please look ad my comments, I think that applying the changes everything will seems easier to configure and maintain
- [x] For targets use names with dashes for everything, for constiency, as you did for `left-panel` e.g. `printoptions` --> `print-options`, `defaultbackgroundignore` -> `default-background-ignore` <-- I did it with my comments

These are missing: 
- [ ] You can avoid to use `createPlugin` for every single panel in `print` folder, using directly the component (that is optionally connected to the state). <-- this seems not to be done, but it is a simplification. 
- [ ] This comment - related to the avoid of using `createPlugin` with an example and some documentation  https://github.com/geosolutions-it/MapStore2/pull/7596#discussion_r759310572

These changes are not mandatory for the functionality itself, but simplify a lot the current solution, that is very tricky. 
If you missed this point in error @mbarto can you open a separate issue for that?
If otherwise you are against these requests for any reason (time, technical issues or other) please feel free to reply, but don't ignore them.

mbarto:
> These are missing:
> 
> * [ ]  You can avoid to use `createPlugin` for every single panel in `print` folder, using directly the component (that is optionally connected to the state). <-- this seems not to be done, but it is a simplification.

If you look at the code you will see that each panel exports both a component and a plugin (as a default).
The component is used in index.js (this I suppose is what you were asking for), while the plugin can be used in project (and it will be used for the client I am doing this work for). Those plugins definition are not included in the product plugins.js (as you requested).

> * [ ]  This comment - related to the avoid of using `createPlugin` with an example and some documentation  [#7585: refactor of the Print plugin as a plugins container #7596 (comment)](https://github.com/geosolutions-it/MapStore2/pull/7596#discussion_r759310572)

I didn't commit your suggestion directly, but if you look again at the example, if has been replaced by something very similar to yours, but verified and tested. 

> These changes are not mandatory for the functionality itself, but simplify a lot the current solution, that is very tricky. If you missed this point in error @mbarto can you open a separate issue for that? If otherwise you are against these requests for any reason (time, technical issues or other) please feel free to reply, but don't ignore them.

That said, feel free to remove my admin commit rights from the project if you prefer.

offtherailz:
> If you look at the code you will see that each panel exports both a component and a plugin (as a default).
The component is used in index.js (this I suppose is what you were asking for), while the plugin can be used in project (and it will be used for the client I am doing this work for). Those plugins definition are not included in the product plugins.js (as you requested).

So all that `createPlugins` can be removed. They are not used and they are not documented in print plugin, It is only confusing. 

Or there is some reason to keep them?

> I didn't commit your suggestion directly, but if you look again at the example, if has been replaced by something very similar to yours, but verified and tested.

Great, I was confused by the double export (I said it was only confusing).

> That said, feel free to remove my admin commit rights from the project if you prefer.

Please don't misunderstand me, I mean you may ask to me if you was in a hurry or what else. We would find a way to clarify and maybe split this clean up in 2. 
I'd say yes, for sure. 
Merging without telling anything only confuses more, and for async work we need trust and clarity.

Issue #7506, opened 2021-11-05
Incorrect handling of wildcards in FilterLayer


offtherailz:
## Description
like/ilike filters wildcards in filter panel are the same of Feature grid, that uses "*" as a wildcard (Passed to WFS). 
Anyway when the filter is used in filter layer plugin, the filter is converted to CQL_FILTER where the wildcard is "%". The converter transforms a filter like `"STATE_NAME" - "LIKE"  - "T*"` in `STATE_NAME "ILIKE '%T*%"` where * should be a %.

## How to reproduce
- Add to the map the gs:us_states layer. (us states)
-  Open filter panel
- Insert an attribute filter "STATE_NAME" - "LIKE"  - "T*"
- Apply the filter

*Expected Result*
I see on the map only "Texas" and "Tennessee" states

*Current Result*
I don't see any state.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

offtherailz:
One thing. like / ilike now adds automatically the wildcard at the beginning and at the end of the search text.
I'm thinking if makes sense to avoid this when a wildcard is present in the string (that means the user wants to explicitly put the wildcards when he likes. 
This may help in case the user want to search "all the strings that starts with" or "all the strings that end with" and so on. 
@tdipisa what do you think?

tdipisa:
>So if the user types `Tex` it will search for `*Tex*`, if types `Tex*` search for `Tex*`
@tdipisa what do you think? Maybe a separate issue for it.

Yes, and to discuss also.

PR #7703, opened 2021-12-29
#7506 Proper handling of wildcards in FilterLayer

alexander-fedorenko:
## Description
Proper handling of wildcards in FilterLayer.
Added utility to process wildcards in the string based on operator used. "like" & "ilike" operators will get wildcards converted to the CQL format where "%" is used for multiple character substitution and "_" is used for a single character.
Following rules are applied:
- WPS requests will not have filter value unconditionally wrapped with ** anymore. There is a check if wildcard is presented in a value. Wrapping will happen only if there are no wildcards in the string.
- CQL: % and _ characters will be escaped with backslash before conversion.
- CQL: * and . will be converted into % and _ only in case if they are not prepended with !
- CQL: !* and !. will be converted into * and .
- CQL: value will be wrapped with %% only if no wildcards are passed into condition. "Te!*s!.t" will be wrapped with %% as there are no wildcards (both cases are escaped with !). "Te*s.t" won't be wrapped.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7506

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/FilterUtils.js
+++ b/web/client/utils/FilterUtils.js
@@ -30,6 +30,15 @@ export const cqlToOgc = (cqlFilter, fOpts) => {
 import { get, isNil, isUndefined, isArray, find, findIndex, isString, flatten } from 'lodash';
 let FilterUtils;
 
+const wrapValueWithWildcard = (value, condition) => {
+    return condition(value) ? '*' + value + '*' : value;
+};
+
+const wrapIfNoWildcards = (value) => {
+    const wildcards = value.match(/(?<!!)[*.]/);
+    return !wildcards?.length;
+};
+
 export const escapeCQLStrings = str => str && str.replace ? str.replace(/\'/g, "''") : str;
 
 export const checkOperatorValidity = (value, operator) => {
@@ -163,7 +172,7 @@ export const  ogcStringField = (attribute, operator, value, nsplaceholder) => {
                     propertyTagReference[nsplaceholder].startTag +
                         attribute +
                     propertyTagReference[nsplaceholder].endTag +
-                    "<" + nsplaceholder + ":Literal>*" + value + "*</" + nsplaceholder + ":Literal>"
+                    "<" + nsplaceholder + ":Literal>" + wrapValueWithWildcard(value, wrapIfNoWildcards) + "</" + nsplaceholder + ":Literal>"
                 );
         }
     }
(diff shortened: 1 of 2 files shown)


web/client/utils/FilterUtils.js
```diff
@@ -834,21 +834,57 @@ export const cqlArrayField = function(attribute, operator, value) {
...
+ * All other operators will receive value as-is, no wildcard processing is applied
+ * @param value
+ * @param operator
+ * @returns {string}
+ */
+export const processCqlWildcards = function(value, operator) {
+    const escapedQuotes = escapeCQLStrings(value);
+    const startCharIsWildcard = value.startsWith('*');
+    const endCharIsWildcard = value.endsWith('*');
+    const startsWithWildcard = startCharIsWildcard && !endCharIsWildcard;
```
offtherailz, line 850: Starts or ends was only examples. Like filter can contain * everywhere.
We need to follow the same convention in the tooltip:

> use * for any number of any char
> use . for a single char
> use ! to escape above two (* and .) 

WFS requests in fact are sent using this convention (Try to use the advanced filter in feature grid):
```
<ogc:PropertyIsLike matchCase="false" wildCard="*" singleChar="." escapeChar="!">
```
The filter, when no wildcard is used is generated like that (inserting `| STATE_NAME | ilike | a |`)

```
<ogc:PropertyName>STATE_NAME</ogc:PropertyName><ogc:Literal>*a*</ogc:Literal>
```
image

Accordingly with the [reference](https://docs.geoserver.org/latest/en/user/styling/ysld/reference/filters.html) CQL uses `%` for multi-character and _ for single char. 

So we need to:

- [ ]  produce a CQL_FILTERS that transforms the string of the operators like and ilike like that: 
    - convert `*` and `.` with `%` and `_`
    - not convert when `!.` or `!*`
    - Escapes when `%` or `_" are present, ( Don't remember the escape char, I think `\.` and `\%` will work, but need to test)

Please document this accordingly in this method.

- [ ] I note right now that inserting `| STATE_NAME | ilike | a *|` in advanced filter of feature grid we have a filter like 
```
<ogc:PropertyName>STATE_NAME</ogc:PropertyName><ogc:Literal>*a**</ogc:Literal>
```

That inconsitent with what we declared in the issue. Therefore we should see to fix also this behavior.

offtherailz (changes requested):
Please read my comment inline. Check when the wildcards are applied in the feature grid filter and propose a solution to make filters consistent.

web/client/utils/FilterUtils.js
alexander-fedorenko: I think I misunderstood requirements, sorry.
All cases listed above should be covered with the changes applied by the last commit.

alexander-fedorenko:
@offtherailz please review when it's convenient. I've updated PR description with the new rules applied to the filter value. Thanks!

offtherailz (approved):
:+1: LGTM

Issue #7705, opened 2022-01-04
failure to print WMTS background layers coming from non-geoserver services


landryb:
## Description
this is extracted from georchestra/mapstore2-georchestra#399 as it's not georchestra-project specific and can be easily reproduced on non-georchestra instances such as https://mapstore.geosolutionsgroup.com/mapstore/#/
## How to reproduce
-  add a WMTS service from https://tiles.craig.fr/ortho/service 
- add a layer from that service
- try to print the map

*Expected Result*
we should get a pdf

*Current Result*
java traceback coming from mapfishprint
```
Error on printing: Error while generating PDF:
org.mapfish.print.JsonMissingException: attribute [spec.layers[1].style] missing
    at org.mapfish.print.utils.PJsonObject.getString(PJsonObject.java:56)
    at org.mapfish.print.map.readers.WMTSMapReader.<init>(WMTSMapReader.java:78)
```

the error is slightly different if the WMTS layer is configured as a background layer in the map loaded at startup, reproduced with https://wxs.ign.fr/essentiels/geoportail/wmts?SERVICE=WMTS&REQUEST=GetCapabilities which chokes on missing `requestEncoding` parameter because there's not default to KVP.

```
Error while generating PDF:
org.mapfish.print.JsonMissingException: attribute [spec.layers[0].requestEncoding] missing
	at org.mapfish.print.utils.PJsonObject.getString(PJsonObject.java:56)
	at org.mapfish.print.map.readers.WMTSMapReader.<init>(WMTSMapReader.java:75)
```

- [x] Not browser related

## Other useful information
most of my findings/time spent analysing it is in georchestra/mapstore2-georchestra#399

a potential fix (to be improved) is this:
```
--- a/web/client/utils/PrintUtils.js
+++ b/web/client/utils/PrintUtils.js
@@ -460,9 +460,9 @@ export const specCreators = {
                 ...dimensionParams,
                 "matrixIds": matrixIds,
                 "matrixSet": tileMatrixSetName,
-                "style": layer.style,
+                "style": layer.style || 'default',
                 "name": layer.name,
-                "requestEncoding": layer.requestEncoding === "RESTful" ? "REST" : layer.requestEncoding,
+                "requestEncoding": layer.requestEncoding === "RESTful" ? "REST" : layer.requestEncoding || 'KVP',
                 "opacity": getOpacity(layer),
                 "version": layer.version || "1.0.0"
             };
```
[...]

PR #7706, opened 2022-01-04
Fix printing WMTS background layers loaded at startup (#7705)

landryb:
default requestEncoding to KVP

## Description
try fixing some of the issues printing WMTS layers

 - [x] Bugfix

## Issue

#7705 

**What is the new behavior?**

printing a WMTS background layer loaded at startup works

## Breaking change
 - [x] No

## Other useful information

note that this is only part of the fix, it doesnt fix printing WMTS layers coming from non-geoserver services such as mapproxy, for which `getDefaultStyleIdentifier` returns null.

### DIFF
--- a/web/client/utils/PrintUtils.js
+++ b/web/client/utils/PrintUtils.js
@@ -561,7 +561,7 @@ export const specCreators = {
                 "matrixSet": tileMatrixSetName,
                 "style": layer.style,
                 "name": layer.name,
-                "requestEncoding": layer.requestEncoding === "RESTful" ? "REST" : layer.requestEncoding,
+                "requestEncoding": layer.requestEncoding === "RESTful" ? "REST" : layer.requestEncoding || "KVP",
                 "opacity": getOpacity(layer),
                 "version": layer.version || "1.0.0"
             };
--- a/web/client/utils/WMTSUtils.js
+++ b/web/client/utils/WMTSUtils.js
@@ -107,7 +107,7 @@ export const getDefaultStyleIdentifier = layer =>{
                 castArray(layer.Style)
                     // the identifier content value is needed
                     .map(l => l["ows:Identifier"]));
-        else {
+        } else {
             return head(
                 castArray(layer.Style)
                     // default is identified by XML attribute isDefault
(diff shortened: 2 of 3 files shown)


landryb:
fwiw, even if eslint chokes on my style (and the error makes no sense to me):
```
/home/runner/work/MapStore2/MapStore2/web/client/utils/WMTSUtils.js
Error:   110:16  error  Unnecessary 'else' after 'return'  no-else-return
```
i've tested the resulting war locally built by backporting those 3 commits to ms2-geor, and that fixes the two distinct wmts printing issues i was seeing. There are probably more here and there, but from my side that's already an improvement.

tdipisa:
@diegovargasg thank you for your contribution. A reviewer has been assigned, we will review this PR as soon as possible. I've just noticed few failing tests that maybe you want to check in the meantime.

landryb:
> I've just noticed few failing tests that maybe you want to check in the meantime.

fwiw that's what i had put in my previous comment, i have no idea why eslint chokes on this else. Or maybe the else should just be dropped completely, but those coding rules feel a bit strange.

offtherailz:
This is the detail of the rule:

https://eslint.org/docs/rules/no-else-return

If the "if" blocks contains a return, the else is an unnecessary block.  
It took a little to me too to understand. ;)

Example:

```js
if (condition) {
  return x;
} else {
  // do y
}
return z;
```
is the same of 

```js
if (condition) {
  return x;
}
// do y
return z;
```

So the else is unnecessary and it may complicate the code reading.

offtherailz (approved):
Thank you very much for contributing.
I tested with the WMTS services you provided and it works. I don't see any problem applying this changes. 
I added a unit test for you. Let's wait the checks to merge

offtherailz:

You can test you can now print layers from these two WMTS services
https://tiles.craig.fr/ortho/service
https://wmts.craig.fr/ortho/service

ElenaGallo:
Test passed, @offtherailz please backport to stable branch.

landryb:
thanks @offtherailz for the merge !

to fully test the default value for `requestEncoding` one should test printing layers coming from https://wxs.ign.fr/essentiels/geoportail/wmts?SERVICE=WMTS&REQUEST=GetCapabilities which isnt mapproxy (and is also extensively used in ms2-georchestra instances)

tiles & wmts.craig.fr are aliases for our mapproxy service.

offtherailz:
Thank you @landryb  for the hint, i improved my test description and verified that it worked

PR #7723, opened 2022-01-12
#7708 Allow latin chars in username and password

offtherailz:
## Description
This PR allows to use UTF-8 chars in username and password, by:
- Updating axios to fix issues that was causing #1882 and we fixed by limiting the allowed chars
- encoding properly the username (the [fix](https://github.com/axios/axios/issues/1446#issuecomment-1007299309) in axios do not include username, unfortunately ) 
- Fixing unit tests
- Fixed also minor issue in localization of error message.

**Please check if the PR fulfills these requirements**
- [ ] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**

#<issue>

**What is the new behavior?**

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

Issue #7708, opened 2022-01-07
Usernames and password with no-latin chars can not login


offtherailz:
## Description
You can create users with names with no-latin chars in the username, but you can not login as that users.
no-latin chars in passwords are denied too. 
We should allow to use no-latin chars in both username and password.

### Username 
-  login as admin
- create a user with a no latin char in the name (e.g. `test҂`)
- logout
- try to login as the user created

*Expected Result*
You can login

*Current Result*
You are not allowed to login (the prompt says a not specified login error). 

### Password 
-  login as admin
- create a user with a no latin char in the password (e.g. `test҂`) --> you was not allowed by the UI because of a limitation in chars.
- logout
- try to login as the user created

*Expected Result*
You can login
You can reset password succesfully with no-latin chars too.

*Current Result*
You are not allowed to create the user

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information
[This](https://github.com/axios/axios/issues/1446#issuecomment-1007299309) is the cause of the  issue. (note, the update of axios do not solve the issue, see my comment about it).

A fix for it can be the following:
```diff
diff --git a/web/client/api/GeoStoreDAO.js b/web/client/api/GeoStoreDAO.js
index c5260e8d6..751c1407a 100644
--- a/web/client/api/GeoStoreDAO.js
+++ b/web/client/api/GeoStoreDAO.js
@@ -154,7 +154,7 @@ const Api = {
         let authData;
         return axios.post(url, null, this.addBaseUrl(merge((username && password) ? {
             auth: {
-                username: username,
+                username: unescape(encodeURIComponent(username)),
                 password: password
             }
         } : {}, options))).then((response) => {
@@ -169,7 +169,7 @@ const Api = {
             }, options)));
         }).then((response) => {
             return { ...response.data, ...authData};
-        });
+        }).catch(e => console.log(e));
     },
```

offtherailz:
here in [this](https://github.com/geosolutions-it/MapStore2/compare/master...offtherailz:latin_username?expand=1) branch I did some tests:
- updated axios (not mandatory for latin username, only for password)
- removed some (maybe) unnecessary code
[...]

PR #7744, opened 2022-01-17
#7708 Support for login with UTF-8 Username and password

offtherailz:
## Description
This PR is a restore of  #7723 with necessary fixes to avoid revert. See #7723 for details and discussion. 

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7708

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/observables/wfs.js
+++ b/web/client/observables/wfs.js
@@ -195,7 +195,7 @@ export const getJSONFeature = (searchUrl, filterObj, options = {}) => {
     return Rx.Observable.defer(() =>
         axios.post(queryString, data, {
             timeout: 60000,
-            headers: { 'Accept': 'application/json', 'Content-Type': 'application/json' }
+            headers: { 'Accept': 'application/json', 'Content-Type': `application/xml` }
         }))
         .let(interceptOGCError)
         .map((response) => workaroundGEOS7233(response.data, getPagination(filterObj, options), options.totalFeatures));
--- a/web/client/utils/EncodeUtils.js
+++ b/web/client/utils/EncodeUtils.js
@@ -1,0 +1,31 @@
+/**
+ * Convert an UTF-8 into a byte representation. It usefult to avoid the `btoa` Unicode problem
+ * A more detailed explaination about this can be found here:
+ *
+ * @param {string} string string to convert
+ * @returns the UTF8 string converted in bytes
+ */
+export function encodeUTF8(string) {
+    return unescape(encodeURIComponent(string));
+}
+export function decodeUTF8(string) {
+    return decodeURIComponent(escape(string));
+}
+/**
+ * This utility function allows to encode UTF-8 strings in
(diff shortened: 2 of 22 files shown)


web/client/observables/wfs.js
```diff
@@ -195,7 +195,7 @@ export const getJSONFeature = (searchUrl, filterObj, options = {}) => {
     return Rx.Observable.defer(() =>
         axios.post(queryString, data, {
             timeout: 60000,
-            headers: { 'Accept': 'application/json', 'Content-Type': 'application/json' }
+            headers: { 'Accept': 'application/json', 'Content-Type': `application/xml` }
```
MV88, line 198: why this?

web/client/utils/EncodeUtils.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Convert an UTF-8 into a byte representation. It usefult to avoid the `btoa` Unicode problem
+ * A more detailed explaination about this can be found here:
```
MV88, line 3: here where?

web/client/utils/EncodeUtils.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param {string} string string to convert
+ * @returns the UTF8 string converted in bytes
+ */
+export function encodeUTF8(string) {
```
MV88, line 8: could be useful to have unit tests for this utils?

web/client/utils/EncodeUtils.js
MV88: this one 
https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem 
?

MV88 (changes requested):
some comments here and there

web/client/observables/wfs.js
offtherailz: Here there was a mistake. Content is an XML but content-type in the heder was "application/json". This was not causing any issue, because probably the content-type is not evaluated by GeoServer for the POST request, assuming it is the standard XML as for the protool. 
Anyway in the updated axios interpret Content-Type of type "json" as a side effect. The body is forwarded as a json string so the final body will be `"<xml ...>"` with `"` at start and end, that is invalid for GeoServer.

web/client/utils/EncodeUtils.js
offtherailz: yes I missed the link 
```suggestion
 * A more detailed explaination about this can be found here: https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem
```

web/client/utils/EncodeUtils.js
offtherailz: sure. I tested it in the action but a test here is better. 
Added a test file right now

MV88 (approved):
LGTM

MV88:
@ElenaGallo please test DEV

Issue #7809, opened 2022-01-31
failure to print WMS layers coming from mapproxy


landryb:
## Description
similar to #7705, but for WMS protocol

## How to reproduce

- go to https://mapstore.geosolutionsgroup.com/mapstore/#
-  add wms service from https://tiles.craig.fr/pci/service
- add `cadastre` layer
- zoom to 'Clermont Ferrand'
- try to print pdf

*Expected Result*
the pdf should contain data from WMS layer

*Current Result*
the image shows `Invalid request: invalid tile size (use 256x256)`, probably because in the POST data ms2 sends `TILED=true` (added by https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/PrintUtils.js#L490).

the postdata sent to mfprint has this:
```
{
    "baseURL": "https://tiles.craig.fr/pci/service?",
    "opacity": 1,
    "singleTile": false,
    "type": "WMS",
    "layers": [
        "cadastre"
    ],
    "format": "image/png",
    "styles": [
        ""
    ],
    "customParams": {
        "TRANSPARENT": true,
        "TILED": true,
        "EXCEPTIONS": "application/vnd.ogc.se_inimage",
        "scaleMethod": "accurate",
        "ENV": ""
   }
}
```

The WMS request sent by mfprint to mapproxy looks like this:
https://tiles.craig.fr/pci/service?BBOX=335973.47%2C5827852.5%2C339748.2%2C5832086.0&WIDTH=713&HEIGHT=800&format_options=dpi%3A96&map_resolution=96&FORMAT=image%2Fpng&STYLES=&ENV=&LAYERS=cadastre&EXCEPTIONS=application%2Fvnd.ogc.se_inimage&TILED=true&SRS=EPSG%3A3857&REQUEST=GetMap&VERSION=1.1.1&SERVICE=WMS&scaleMethod=accurate&TRANSPARENT=true

which reproduces the error. Without `TILED=true` the query succeeds.

The same layer successfully prints when being added/queried as WMTS since #7705.

- [x] Not browser related

From my understanding of WMS, if using `TILED=true` the `HEIGHT` and `WIDTH` parameters should match the server tile size (which at that point mapstore knows, since it uses it to display the layer?).

@offtherailz, any idea ?

landryb:
i would be tempted to remove `TILED=yes` from `PrintUtils.js` but i suppose if its there it's for a geoserver-vendor specific behaviour or passing the request to gwc ?

[...]

PR #7811, opened 2022-02-01
drop TILED arg from wms printing queries (#7809)

landryb:
TILED is vendor-specific, and triggers exceptions with mapproxy
when requesting HEIGHT & WIDTH that dont match the cache tile size or format that doesnt match the cache format.

## Description
Fix printing WMS layers coming from mapproxy

 this can be tested before/after when adding a WMS layer:

- from https://tiles.craig.fr/ortho/service - warning, single tile option needs to be used in layer properties otherwise layer wont load/display, server returns `Invalid request: invalid tile format, use mixed` because of default `TILED=true` behaviour - if removed as done by this PR, then printing the ortho layer works
- from https://tiles.craig.fr/pci/service - triggers `Invalid request: invalid tile size (use 256x256)` exception serverside.

 - [x] Bugfix

## Issue

#7809 - printing WMS layers coming from mapproxy fails because `TILED=true` triggers exceptions, either because the requested `HEIGHT` and `WIDTH` dont match the cache tile size or the requested format doesnt match the cache format.

**What is the new behavior?**
printing WMS layers coming from mapproxy succeeds, tested with both ortho and pci.

## Breaking change
 - [x] MAYBE
 - [ ] No

Maybe, because from my understanding `TILED=true` triggers a vendor-specific behaviour in geoserver/geowebcache ?

Ideally, `TILED=true` should only be added if the remote is geoserver/gwc, but at that point i have no idea if that's an information we can get (and trust) from capabilities document.

## Other useful information

Interoperability is hard.

### DIFF
--- a/web/client/utils/PrintUtils.js
+++ b/web/client/utils/PrintUtils.js
@@ -487,7 +487,6 @@ export const specCreators = {
             ],
             "customParams": addAuthenticationParameter(PrintUtils.normalizeUrl(layer.url), assign({
                 "TRANSPARENT": true,
-                "TILED": true,
                 "EXCEPTIONS": "application/vnd.ogc.se_inimage",
                 "scaleMethod": "accurate",
                 "ENV": generateEnvString(spec.env)


offtherailz (approved):
Generally speaking, tiled option in GeoServer is used to get WMS requests from tile cache, if they match the gridset. Otherwise the response is provided anyway, with some special headers specifying the esit and reason of cache failure
I understood that MapProxy uses the same parameter name with a different behavior, and rises exception so it can be removed (it is only a cache support optimization). 
Anyway the layers has also a tiled parameter, that is used on map. 
If you had to disable it for the layer, maybe is a good idea to read it from the layer object and use it if present (and / or true).

landryb:
> Anyway the layers has also a tiled parameter, that is used on map. If you had to disable it for the layer, maybe is a good idea to read it from the layer object and use it if present (and / or true).

Well, the tiled parameter is on by default on all layers (and i havent found a way to disable that per server in the ms2 config), and it works fine when displaying layers that dont have a '*mixed*' cache format (eg the layer from https://tiles.craig.fr/pci/service). `TILED=true` is used along `HEIGHT=256` and `HEIGHT=256` so mapproxy is fine with that, which isnt the case when querying the server for the image to print (which might be 'against' TILED spec, which is a vendor extension anyway..).

For layers that use a `*mixed*` cache format (eg layers from https://tiles.craig.fr/ortho/service), using `TILED=true` (so, the default) breaks layer display (and also printing).

tdipisa (changes requested):
I don't think we can approve this PR as it is @offtherailz. I completely understand the @landryb point of view and of course we can probably do something to improve the interoperability with MapProxy in this case. 
In order to find an automatic way to set the TILED property to true/false depending on the layer source is something we have to investigate. Therefore a dedicated issue should be opened for this. For the moment, and especially for 2022.01.00, what we can do is just to make it configurable globally (default for MapStore is true). What do you think @offtherailz?

landryb:
well you already have a way to know if you're talking to geoserver in https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/LayersUtils.js#L25 - though i think it'd be much nicer to check for vendor extensions in capabilities. Fwiw my geoserver is on https://ids.craig.fr/wxs/ for various reasons, it's a bit awkward to expect geoserver to be at `/geoserver/`...

As for TILED itself, making it globally configurable would be a valid workaround for now :)

tdipisa:
> well you already have a way to know if you're talking to geoserver in https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/LayersUtils.js#L25 - though i think it'd be much nicer to check for vendor extensions in capabilities. Fwiw my geoserver is on https://ids.craig.fr/wxs/ for various reasons, it's a bit awkward to expect geoserver to be at `/geoserver/`...
> 

I can understand your concern about this but unfortunately, as far as I know, there isn't a clean/stable way implemented so far in GeoServer (eg. a kind of declaration or something similar) to understand that the capabilities are effectively coming from a GeoServer. Therefore some development GeoServer side should be necessary for this (at least for new GeoServer versions). 
One way could be also to delegate the user to declare this through the MS UI (for example when the user define a source in the Catalog tool). Long story short, we haven't made a decision on this yet.
Feel free anyway to contribute to this if you want/can (even if only with your ideas) and thank you so much in any case for your feedback.

> As for TILED itself, making it globally configurable would be a valid workaround for now :)

Do you have in mind or do you have the possibility to contribute for this? Otherwise I'm going to close this PR for now. Thank you again.

landryb:
> > As for TILED itself, making it globally configurable would be a valid workaround for now :)
> 
> Do you have in mind or do you have the possibility to contribute for this? Otherwise I'm going to close this PR for now. Thank you again.

I dont think i have enough knowledge of the platform right now to implement that in a short timeframe.. but that might be something we can fund.

landryb:
> there isn't a clean/stable way implemented so far in GeoServer (eg. a kind of declaration or something similar) to understand that the capabilities are effectively coming from a GeoServer.

another possibility would be to look for `GEOSERVER` under `/Service/KeywordList/Keyword` in `GetCapabilities` as it's a default config in geoserver and i doubt many service administrators change that..

landryb:
> > there isn't a clean/stable way implemented so far in GeoServer (eg. a kind of declaration or something similar) to understand that the capabilities are effectively coming from a GeoServer.
> 
> another possibility would be to look for `GEOSERVER` under `/Service/KeywordList/Keyword` in `GetCapabilities` as it's a default config in geoserver and i doubt many service administrators change that..

replying to myself on a closed issue, also checking for the presence of `updateSequence` attribute to the toplevel `WMS_Capabilities` XML element would be another idea to figure out if the remote is geoserver.

Issue #7789, opened 2022-01-27
Custom Colors Settings screen size problem


ElenaGallo:
When the third value is added, a scroll bar appears, although it is not necessary. It creates problems when entering the class value and class label. See the gif below:

![2](https://user-images.githubusercontent.com/56537133/150777470-5b0e3f5f-ef5d-4e39-86ca-28e45ccfa0a5.gif)

**How to Reproduce**
- Open a map
- Select a layer on TOC
- Click on Widget > Chart > Bar Chart
- Cutomize it and select CUSTOM color
- Click on Edit 
- Select an attribute
- Click on New entry after (x2)
- Enter for the third class a Value and a Label

## Other usefull information
It happens with a 1920x1080 screen

Issue #7784, opened 2022-01-26
Further fixes are needed for the new charts classifier


tdipisa:
## Description

1. The autocomplete field for classification values takes too long to load and, in certain cases, the loading is triggered simultaneously for different fields. 

![autocomplete.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/9dafc4b7-c534-474a-8837-0c2845075e22)

2. The classification attribute is maintained in the WPS request even after switching to a solid color.

![solid_color_wps.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/c8ebc25e-306d-4afc-ad76-ec917505d241)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

ale-cristofori:
the PR for this issue is https://github.com/geosolutions-it/MapStore2/pull/7830

Issue #7770, opened 2022-01-25
Fixes for issues found within testing in dev for colour coded classes charts 


ale-cristofori:
## Description
This ticket is created to collect all issues found while testing issue #7734 
and are documented [here](https://github.com/geosolutions-it/MapStore2/issues/7734#issuecomment-1020014284)

1_ When adding a new entry above, the new Class Color and Class Label are added above, but the existing Class Value remains in the same position

2_ When there are few slices in the pie chart the separation line is not present in the preview. 

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #7830, opened 2022-02-04
#7784 #7789 fixes for new charts classifier

alexander-fedorenko:
## Description
- Debouncing/throttling after "distinctUntilChanged" caused unwanted effect when request was re-executed even if value, currentPage or attribute were not changed.
- Removed classification attribute from store once user switch to color ramp other than custom. This will remove classification attribute from WPS request in such case.
- Corrected CSS for loading spinner.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7784 

**What is the new behavior?**
1. Corrected behavior of throttling and debouncing for "createPagedUniqueAutompleteStream". Debouncing after "distinctUntilChanged" caused unwanted effect when field fired data re-fetching even when it should be distinct, e.g. dropdown toggle while value is empty or value is set and user toggles dropdown.
2. Removed classification attribute value from the store once user switch from custom color ramp to predefined. It also removes classification attribute from WPS request, though user will have to re-define classification field if color ramp is switched to custom again.
3. Added overflow: hidden to the spinner; in some cases spinner was causing scrollbar to appear in the modal.

## Breaking change
 - [x] No

## Other useful information

### DIFF
--- a/web/client/components/widgets/builder/wizard/common/WPSWidgetOptions.jsx
+++ b/web/client/components/widgets/builder/wizard/common/WPSWidgetOptions.jsx
@@ -227,6 +227,9 @@ export default ({
                                                     ...(classification ? { classification: formatAutoColorOptions(classification) } : {} ),
                                                     defaultCustomColor: defaultCustomColor ?? '#0888A1'
                                                 });
+                                                if (!v.custom) {
+                                                    onChange("options.classificationAttribute", undefined);
+                                                }
                                             }}/>
                                     </Col>
                                 </FormGroup>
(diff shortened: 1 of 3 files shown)


alexander-fedorenko:
@tdipisa 
1. Time to load autocomplete values is not caused by something on the frontend, it rely on backend response speed. I've tried it on dev instance in the evening - it was taking 1.5-2.5 seconds to load, while in the morning it is something around 100-150ms. 
As for the simultaneous requests for several fields - unfortunately there is no way to cancel request upon dropdown toggle from open to closed because of the widget that is in use now. It fires "toggle" event multiple times after user clicks on arrow glyph to get options, cancelling data load request on toggle to closed state makes widget load no options at all. Also data fetch requests are distinct and fire once per change of classification attribute, results page or value in the input. Cancelling distinct request will break workflow and field will become unusable until user change value.
2. I've noticed there was a bug that caused data to reload in conditions when it should keep existing options. I've described it above in PR description, section about new behavior.

offtherailz (approved):

ale-cristofori:
This issue was created to solve issue n 1 and n 2 separately from this original issue #7770

Issue #7661, opened 2021-12-15
Error in 3D when terrain is available


offtherailz:
## Description

An error occours when we try to use a BIL terrain model. QA and STABLE doesn't seems to be affected. 
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/dem3d

https://qa-mapstore.geosolutionsgroup.com/mapstore/#/context/dem3d

## How to reproduce

-  Open this context https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/dem3d (the context is attached)
- Switch to 3D

*Expected Result*
You see the dem (screenshot from qa, same tested on stable)
![image](https://user-images.githubusercontent.com/1279510/146162933-5654f322-7523-443b-8c79-07f1bfbd246c.png)

*Current Result*
Error: 
```
An error occurred while rendering. Rendering has stopped.
TypeError: e.throttleRequestByServer is not a function
TypeError: e.throttleRequestByServer is not a function
    at Object.e.when.e.defined.t.ready.e.defined.t.getHeightmapTerrainDataArray (https://dev-mapstore.geosolutionsgroup.com/mapstore/dist/mapstore2.js?c3654326ac83b2bb94d2:2:1165814)
    at n.e.when.e.defined.t.ready.s.requestTileGeometry (https://dev-mapstore.geosolutionsgroup.com/mapstore/dist/mapstore2.js?c3654326ac83b2bb94d2:2:1166618)
    at m (https://cesium.com/downloads/cesiumjs/releases/1.42/Build/Cesium/Cesium.js:537:25427)
    at g (https://cesium.com/downloads/cesiumjs/releases/1.42/Build/Cesium/Cesium.js:537:25540)
    at m.processLoadStateMachine (https://cesium.com/downloads/cesiumjs/releases/1.42/Build/Cesium/Cesium.js:537:27192)
    at E (https://cesium.com/downloads/cesiumjs/releases/1.42/Build/Cesium/Cesium.js:537:29606)
    at Function.b.processStateMachine (https://cesium.com/downloads/cesiumjs/releases/1.42/Build/Cesium/Cesium.js:538:3135)
    at q.loadTile (https://cesium.com/downloads/cesiumjs/releases/1.42/Build/Cesium/Cesium.js:539:2107)
    at M (https://cesium.com/downloads/cesiumjs/releases/1.42/Build/Cesium/Cesium.js:539:20873)
[...]

Issue #7349, opened 2021-09-22
Main dependencies check


tdipisa:
@offtherailz
As we have done with #7233 for branch 2021.02.xx, we have to do the same for master branch:

- [x] print-lib geosolutions-2.0-SNAPSHOT (nothing to do)
- [x] http-proxy 1.3-SNAPSHOT @offtherailz 
- [x] geostore 1.9-SNAPSHOT (#7829) @offtherailz 
- [x] mapstore-services 1.4-SNAPSHOT @offtherailz 

For 2022.01.00 we need:

- [x] print-lib geosolutions-2.0.1(nothing to do)
- [x] http-proxy 1.3.0 (to release) @taba90 
- [x] GeoStore 1.8.0 (nothing to do)
- [x] mapstore-services 1.3 (to release) @offtherailz

tdipisa:
Hi @Mauro 
The next major release of MapStore is currently scheduled for mid March. If you have changes/upades to provide for mapfish-print, I think it would be good to agree now on how to release them as part of a new geosolutions version of the library. Now we are using this: https://github.com/geosolutions-it/mapfish-print/tree/geosolutions-2.0.x

Do you have updates for this? I'm asking because I haven't seen yet a PR on mapfish-print and I see all issues still in progress in the kanban. Will you be able to deliver them all in time to allow testing in master for the major release? We have also to establish what should be the new geosolutions version of mapsfish-print lib.

I'm available to have a quick chat with you to discuss/agree on this.

tdipisa:
@taba90 we need you to release some dependencies for 2022.01.00 (see description). We need these new releases done and available in mvn repo by tomorrow.

taba90:
@tdipisa @offtherailz:
http-proxy 1.3.0 has been released.
According to our call the 1.9.0 release for geostore is no more needed, I've restored the version of master branch back to 1.9-SNAPSHOT.

tdipisa:
Thank you @taba90

offtherailz:
This doesn't need any test from @ElenaGallo . JFI

PR #7831, opened 2022-02-04
#7661 updated localConfig.js proxyUrl useCors array

tonym3nsa:
## Description

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7661 

**What is the new behavior?**

3D terrain is rendering
## Breaking change
 - [x] No

### DIFF
--- a/web/client/configs/localConfig.json
+++ b/web/client/configs/localConfig.json
@@ -10,7 +10,8 @@
           "https://gs-stable.geo-solutions.it/geoserver",
           "https://gs-stable.geo-solutions.it:443/geoserver",
           "http://gs-stable.geo-solutions.it/geoserver",
-          "http://gs-stable.geo-solutions.it:443/geoserver"
+          "http://gs-stable.geo-solutions.it:443/geoserver",
+          "https://georchestra.geo-solutions.it/geoserver"
         ]
     },
     "geoStoreUrl": "rest/geostore/",


web/client/configs/localConfig.json
```diff
@@ -10,7 +10,8 @@
           "https://gs-stable.geo-solutions.it/geoserver",
           "https://gs-stable.geo-solutions.it:443/geoserver",
           "http://gs-stable.geo-solutions.it/geoserver",
-          "http://gs-stable.geo-solutions.it:443/geoserver"
+          "http://gs-stable.geo-solutions.it:443/geoserver",
+          "https://georchestra.geo-solutions.it/geoserver"
```
MV88, line 14: i don't think we want to include this geoserver among the cors urls in the official mapstore repo
could you explain the cause of the issue and why you ended up with this fix?

MV88: @tonym3nsa please reply for the above

tonym3nsa: @MV88 sorry, I must have missed this. The 3D map was unable to render as described [here](https://github.com/geosolutions-it/MapStore2/issues/7661). From my investigations, it seems like an issue caused by CORS - that's why I added the URLs there

MV88: i'll better check this tomorrow, but we do not need to include every geoserver url for every client in the official repo
eventually you include it in the dedicated project if needed. otherwise it may be a problem of the proxy or something else

@offtherailz what do you think?

offtherailz: I think we should use the internal proxy in a transparent way as much as possible. 
Even if CORS is a useful optimization, we can not rely on it. The old version worked without this setting, so we need to keep the same functionality. 
Try to understand how the old worked, how this functionality disappeared during the updated, and how to restore it.

tonym3nsa: @MV88 @offtherailz Well noted

tdipisa:
@tonym3nsa @offtherailz @MV88 let's block this for a moment. MS master is currently using a broken http-proxy version. I would like to test it again after the dependency version update expected with #7349.

tonym3nsa:
@tdipisa the 3D version of the map seems to work fine after you restored the **http-proxy.** Should I go ahead and close this? Or is there anything else I need to note?

tdipisa:
Thank you @tonym3nsa let's put the issue in test and ask @ElenaGallo to confirm with a final test.

tonym3nsa:
@ElenaGallo kindly test https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/dem3d from issue #7661. It should work without the above fix

ElenaGallo:
@tonym3nsa you are right, the problem is no longer present on DEV. We can close this PR.

Issue #5325, opened 2020-05-22
Improve User Guide with Info about WPSs used by MS


tdipisa:
## Description

Include in documentation (user guide) some warning sections that inform about GeoServer WPS processes used by MapStore (Attribute Table, Widgets and Query Panel sections).

*Documentation section involved*
- [x] User Guide
- [ ] Developer Guide

PR #7840, opened 2022-02-07
fix WPS process name in documentation

landryb:
`gs:PageUnique` doesnt exist, that's `gs:PagedUnique` - thats on top of  #5325


tdipisa:
@landryb good catch, thank you.

tdipisa (approved):
@landryb good catch, thank you.

Issue #7773, opened 2022-01-25
Graticule layers improvements, for openlayers and printing


mbarto:
## Description

The current implementation of the graticule layer has some limitations:
 - only supports lat - lon grids
 - does not support labels
 - cannot be printed

We would like to:
 - support metric grids, at least for mercator (EPSG:3857) and UTM projections (we won't implement full UTM grids (https://www.dmap.co.uk/utmworld.htm), for simplicity we will assume a regular grid, without the irregular zones (e.g. 31, 32, etc.)
 - add optional labels for x and y lines, with configurable styling
 - add printing support

The openlayers Graticule layer only supports lat - lon grids, so we will need to either:
 - extend / override it with a MapStore implementation
 - use another library (e.g. https://github.com/Viglino/ol-ext)

Connected to https://github.com/geosolutions-it/mapfish-print/issues/35

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

offtherailz:
@tdipisa this is not included by default as a panel, should we include it?

offtherailz:
We should include it as a panel, instead of plugin, to avoid to reconfigure all the projects depending.
This requires a small work on client. Same for the other task

ElenaGallo:
@tdipisa I can not test the **PrintGraticule** plugin on MS because it is not present in the `plugins.js` file

ElenaGallo:
Missing documentation and tests will be done with #8596

PR #7841, opened 2022-02-07
#7773: ol Graticule layer support for different projections + print support

mbarto:
## Description

Improvements to the graticule layer type:
 - new implementation for openlayers, compatible with the standard one, that delegates to MapGridsUtils most of the work, so that the logic is reusable for maps and printing
 - added support for grids in different projections (ol one only supports lat-lon grids)
 - added support for printing

As a bonus:
 - a plugin for printing is included, that allows adding a graticule to the printed map
 - various fixes / improvements to vector styling to support what is needed by the graticule

Some of the fixes need the latest mapfish-print print-lib.jar (from master) that contains the following fixes / improvements:
 - support for halo in labels / text (outlineLabelColor, outlineLabelWeight)
 - support for labels / text rotation
 - support for UTF-8 chars in labels / text

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#7773 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/grids/GridLabelsUtils.js
+++ b/web/client/utils/grids/GridLabelsUtils.js
@@ -1,0 +1,128 @@
+import isFunction from "lodash/isFunction";
+
+/**
+ * Returns the modulo of a / b, depending on the sign of b.
+ *
+ * @param {number} a Dividend.
+ * @param {number} b Divisor.
+ * @return {number} Modulo.
+ */
+function modulo(a, b) {
+    var r = a % b;
+    return r * b < 0 ? r + b : r;
+}
+
+/**
+ * @param {number} number Number to be formatted
+ * @param {number} width The desired width
+ * @param {number=} optPrecision Precision of the output string (i.e. number of decimal places)
+ * @returns {string} Formatted string
+ */
+function padNumber(number, width, optPrecision) {
+    var numberString = optPrecision !== undefined ? number.toFixed(optPrecision) : '' + number;
+    var decimal = numberString.indexOf('.');
+    decimal = decimal === -1 ? numberString.length : decimal;
+    return decimal > width ? numberString : new Array(1 + width - decimal).join('0') + numberString;
+}
+
+/**
+ * @param {string} hemispheres Hemispheres.
+ * @param {number} degrees Degrees.
+ * @param {number=} optFractionDigits The number of digits to include
+ *    after the decimal point. Default is `0`.
+ * @return {string} String.
+ */
+const degreesToStringHDMS = (hemispheres) => (degrees) => {
+    var normalizedDegrees = modulo(degrees + 180, 360) - 180;
+    var x = Math.abs(3600 * normalizedDegrees);
+
+    var deg = Math.floor(x / 3600);
+    var min = Math.floor((x - deg * 3600) / 60);
+    var sec = x - (deg * 3600) - (min * 60);
+    sec = Math.ceil(sec);
+
+    if (sec >= 60) {
+        sec = 0;
+        min += 1;
+    }
+
+    if (min >= 60) {
+        min = 0;
+        deg += 1;
+    }
+
+    return deg + '\u00b0 ' + padNumber(min, 2) + '\u2032 ' +
+      padNumber(sec, 2, 0) + '\u2033' +
+      (normalizedDegrees === 0 ? '' : ' ' + hemispheres.charAt(normalizedDegrees < 0 ? 1 : 0));
+};
+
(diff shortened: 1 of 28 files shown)


web/client/utils/grids/GridLabelsUtils.js
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Returns the modulo of a / b, depending on the sign of b.
+ *
+ * @param {number} a Dividend.
+ * @param {number} b Divisor.
+ * @return {number} Modulo.
+ */
+function modulo(a, b) {
+    var r = a % b;
+    return r * b < 0 ? r + b : r;
```
offtherailz, line 12: ((a % b ) + b ) % b should do the same, I think :smile:  No problem to do it more explicit. 
ref: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder

offtherailz (changes requested):
Hi @mbarto . Here my first comments. Sorry if I don't finish but I need more information to contiue reviewing it. 
- [ ] Please check and fix warning for linter camel case. in `Graticule.js`. 
- [ ] It is needed to enhance the legacy styler instead to enhancing the new style rules ?
- [ ] In order to test this I need also some fix on mapfish print? I ask you before starting testing and finalizing the review.  I see you are using several unicode chars  in code, and one of the point was that. 
- [ ] I see several usages of `detached` layer in openlayers and cesium implementation you did over the time, but I don't have a definition of that attribute. Can you explaing (and maybe document in a comment what it means, so the usage is clear for the future. 

I have to examinate and test the PR more in detail, but you can start fixing these and answering these questions, so I can better understand your work.

mbarto:
@offtherailz the fixes to mapfish-print are in the master branch of https://github.com/geosolutions-it/mapfish-print, and the corresponding jar has already been deployed to geosolutions maven repository.

offtherailz:
Great, so it should be already deployed on mapstore-dev, I will test on it while waiting for the other points. Thank you @mbarto for the info.

offtherailz:
@mbarto I added the points after testing in the review. Nice feature.

mbarto:
@offtherailz 

> Hi @mbarto . Here my first comments. Sorry if I don't finish but I need more information to contiue reviewing it.
> 
> * [ ]  Please check and fix warning for linter camel case. in `Graticule.js`.

Fixed

> * [ ]  It is needed to enhance the legacy styler instead to enhancing the new style rules ?

The rationale behind this choice is the following:  I tried to use the same path used for vector layers styling, that for openlayers uses VectorStyle.getStyle.
This one is already able to recognize the style type and use it properly (if you use a geostyle style, it should work); currently I see two different paths: either is a geostyler style, or a legacy style; annotations and drawing tools live on a different world, to my understanding, and are not integrated properly.

The story is a bit different for printing, where a specific set of transformers are used to build an ol2 style object (used by mapfish-print) from the vector style(s), and these transformers at the moment support only legacy style or annotation styles. This means that a geostyler style should not be usable, at the moment, for printing.

That said, I heard that an attempt to rationalize vector styling should be under way. In that case, you should consider having a general purpose converter (a la VectorStyle.getStyle) also for vector printing.

> * [x]  In order to test this I need also some fix on mapfish print? I ask you before starting testing and finalizing the review.  I see you are using several unicode chars  in code, and one of the point was that.
> * [ ]  I see several usages of `detached` layer in openlayers and cesium implementation you did over the time, but I don't have a definition of that attribute. Can you explaing (and maybe document in a comment what it means, so the usage is clear for the future.

I added some comments in the code. A detached layer is a layer that is not added/removed to the map via map.addLayer / map.removeLayer, so we need to work in different way with them. Both cesium and openlayers include some layers that need to be used this way.

Some refactoring on this part would probably make this concept more clear, I agree.

> 
> # Printing

> 
> * [ ]  In openlayers, I have an error:
>   Error:

Should be fixed now.
> 
> * [ ]  In cesium looks good. Maybe the default text is not easy to read (small), but acceptable.
>   image
> 

Take into account that the Cesium implementation is untouched and not part of the scope of this work.

> # Mix of the two
> I tried to print a map with both graticule turned on and layer graticule in the map. This is the final result. Not sure the color are the same, but I think this is not intended to be supported isn't it?

No, I don't think it make sense to have the Graticule both from the map and as an option in the printing panel.

offtherailz (approved):
Everything is clear and correct for me. 
We can merge. Thank you for your work.

Issue #1945, opened 2017-06-21
Allow property selection for export in FeatureGrid


offtherailz:
We should allow the user to select the attributes he wants to export, even geometry.

PR #7845, opened 2022-02-07
#1945 Make FeatureGrid attributes visibility selector counted when user exports grid data

alexander-fedorenko:
## Description
This PR suggests changes needed to make FeatureGrid attributes visibility selector counted when user exports grid data.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#1945 

**What is the new behavior?**
WFS export will properly use list of hidden attributes to load data only for visible columns.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/download/FeatureEditorButton.jsx
+++ b/web/client/components/data/download/FeatureEditorButton.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
-import Button from '../featuregrid/toolbars/TButton';
+import withHint from "../featuregrid/enhancers/withHint";
+import TButtonComp from "../featuregrid/toolbars/TButton";
+const TButton = withHint(TButtonComp);
 
-
-export default ({disabled, results = [], mode, isDownloadOpen, onClick = () => {}}) => <Button
+export default ({disabled, results = [], mode, isDownloadOpen, onClick = () => {}}) => <TButton
     id="download-grid"
     keyProp="download-grid"
     tooltipId="featuregrid.toolbar.downloadGridData"
--- a/web/client/epics/layerdownload.js
+++ b/web/client/epics/layerdownload.js
@@ -106,7 +106,7 @@ const hasOutputFormat = (data) => {
 };
 
 const getWFSFeature = ({ url, filterObj = {}, layerFilter, downloadOptions = {}, options } = {}) => {
-    const { sortOptions, propertyName: pn } = options;
+    const { sortOptions, pn } = options;
 
     const data = mergeFiltersToOGC({ ogcVersion: '1.0.0', addXmlnsToRoot: true, xmlnsToAdd: ['xmlns:ogc="http://www.opengis.net/ogc"', 'xmlns:gml="http://www.opengis.net/gml"'] }, layerFilter, filterObj);
 
@@ -131,6 +131,7 @@ const getDefaultSortOptions = (attribute) => {
 const getFirstAttribute = (state)=> {
     return state.query && state.query.featureTypes && state.query.featureTypes[state.query.typeName] && state.query.featureTypes[state.query.typeName].attributes && state.query.featureTypes[state.query.typeName].attributes[0] && state.query.featureTypes[state.query.typeName].attributes[0].attribute || null;
 };
+
 const wpsExecuteErrorToMessage = e => {
     switch (e.code) {
     case 'ProcessFailed': {
@@ -246,7 +247,8 @@ export const startFeatureExportDownload = (action$, store) =>
             filterObj: action.filterObj,
             layerFilter,
             options: {
-                pagination: !virtualScroll && get(action, "downloadOptions.singlePage") ? action.filterObj && action.filterObj.pagination : null
+                pagination: !virtualScroll && get(action, "downloadOptions.singlePage") ? action.filterObj && action.filterObj.pagination : null,
+                pn: action.downloadOptions.propertyName
             }
         })
             .do(({ data, headers }) => {
@@ -265,7 +267,8 @@ export const startFeatureExportDownload = (action$, store) =>
                     layerFilter,
                     options: {
                         pagination: !virtualScroll && get(action, "downloadOptions.singlePage") ? action.filterObj && action.filterObj.pagination : null,
-                        sortOptions: getDefaultSortOptions(getFirstAttribute(store.getState()))
+                        sortOptions: getDefaultSortOptions(getFirstAttribute(store.getState())),
+                        pn: action.downloadOptions.propertyName
                     }
                 }).do(({ data, headers }) => {
                     if (headers["content-type"] === "application/xml") { // TODO add expected mimetypes in the case you want application/dxf
--- a/web/client/epics/layerdownload.js
+++ b/web/client/epics/layerdownload.js
@@ -108,12 +108,12 @@ const hasOutputFormat = (data) => {
 };
 
 const getWFSFeature = ({ url, filterObj = {}, layerFilter, downloadOptions = {}, options } = {}) => {
-    const { sortOptions, pn } = options;
+    const { sortOptions, propertyNames } = options;
 
     const data = mergeFiltersToOGC({ ogcVersion: '1.0.0', addXmlnsToRoot: true, xmlnsToAdd: ['xmlns:ogc="http://www.opengis.net/ogc"', 'xmlns:gml="http://www.opengis.net/gml"'] }, layerFilter, filterObj);
 
     return getXMLFeature(url, getFilterFeature(query(
-        filterObj.featureTypeName, [...(sortOptions ? [sortBy(sortOptions.sortBy, sortOptions.sortOrder)] : []), ...(pn ? [propertyName(pn)] : []), ...(data ? castArray(data) : [])],
+        filterObj.featureTypeName, [...(sortOptions ? [sortBy(sortOptions.sortBy, sortOptions.sortOrder)] : []), ...(propertyNames ? [propertyName(propertyNames)] : []), ...(data ? castArray(data) : [])],
         { srsName: downloadOptions.selectedSrs })
     ), options, downloadOptions.selectedFormat);
 
@@ -254,7 +254,7 @@ export const startFeatureExportDownload = (action$, store) =>
             layerFilter,
             options: {
                 pagination: !virtualScroll && get(action, "downloadOptions.singlePage") ? action.filterObj && action.filterObj.pagination : null,
-                pn: propertyNames
+                propertyNames
             }
         })
             .do(({ data, headers }) => {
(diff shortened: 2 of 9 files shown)


alexander-fedorenko:
Right now this feature is blocked by the limitation on GeoServer side. There are two WPS processes that can be used to achieve desired workflow when resulting layer data is filtered in a way that part of the attributes are removed from the output: "gs:Query" and "vec:Query". Both of them doing same things but first one is marked as deprecated. 

"vec:Query" has following description: "Queries a feature collection using an optional filter and an **optional list of attributes to include**. Can also be used to convert feature collection format." The issue is that it doesn't accept list of attributes in any possible format.
- Comma-separated list results into exception on process execution:

> Process failed during execution java.lang.NullPointerException: PropertyDescriptor is null - did you request a property that does not exist? PropertyDescriptor is null - did you request a property that does not exist? 

image


- Passing each attribute name in a separate wps:Input tag makes process finish successfully, but in such case it take only one last attribute passed with the wps:Input tag. Resulting file contains only one attribute in this case.
```
...
    <wps:Input>
      <ows:Identifier>attribute</ows:Identifier>
      <wps:Data>
        <wps:LiteralData>STATE_ABBR</wps:LiteralData>
      </wps:Data>
    </wps:Input>
    <wps:Input>
      <ows:Identifier>attribute</ows:Identifier>
      <wps:Data>
        <wps:LiteralData>LAND_KM</wps:LiteralData>
      </wps:Data>
    </wps:Input>
...
```

tdipisa:
@taba90 Can we have some assistance from you on this?

MV88:
@tdipisa i think @offtherailz is a more suitable person to review this
let me know your thoughts

tdipisa:
> @tdipisa i think @offtherailz is a more suitable person to review this let me know your thoughts

@MV88 this is in blocked for now

offtherailz:
I did a few attempts and I guessed that the correct format for multi-attributes is this:
```
<?xml version="1.0" encoding="UTF-8"?><wps:Execute version="1.0.0" service="WPS" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.opengis.net/wps/1.0.0" xmlns:wfs="http://www.opengis.net/wfs" xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:gml="http://www.opengis.net/gml" xmlns:ogc="http://www.opengis.net/ogc" xmlns:wcs="http://www.opengis.net/wcs/1.1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 http://schemas.opengis.net/wps/1.0.0/wpsAll.xsd">
  <ows:Identifier>vec:Query</ows:Identifier>
  <wps:DataInputs>
    <wps:Input>
      <ows:Identifier>features</ows:Identifier>
      <wps:Reference mimeType="text/xml" xlink:href="http://geoserver/wfs" method="POST">
        <wps:Body>
          <wfs:GetFeature service="WFS" version="1.0.0" outputFormat="GML2" xmlns:gs="https://gs-stable.geo-solutions.it/geoserver/geoserver">
            <wfs:Query typeName="gs:us_states"/>
          </wfs:GetFeature>
        </wps:Body>
      </wps:Reference>
    </wps:Input>
    <wps:Input>
      <ows:Identifier>attribute</ows:Identifier>
      <wps:Data>
        <wps:LiteralData>STATE_NAME</wps:LiteralData>
      </wps:Data>
    </wps:Input>
    <wps:Input>
      <ows:Identifier>attribute</ows:Identifier>
      <wps:Data>
        <wps:LiteralData>STATE_FIPS</wps:LiteralData>
      </wps:Data>
    </wps:Input>
    <wps:Input>
      <ows:Identifier>attribute</ows:Identifier>
      <wps:Data>
        <wps:LiteralData>WATER_KM</wps:LiteralData>
      </wps:Data>
    </wps:Input>
    <wps:Input>
      <ows:Identifier>attribute</ows:Identifier>
      <wps:Data>
        <wps:LiteralData>the_geom</wps:LiteralData>
      </wps:Data>
    </wps:Input>
  </wps:DataInputs>
  <wps:ResponseForm>
    <wps:RawDataOutput mimeType="text/xml; subtype=wfs-collection/1.0">
      <ows:Identifier>result</ows:Identifier>
    </wps:RawDataOutput>
  </wps:ResponseForm>
</wps:Execute>
```

It doesn't seems to be possible to create this request in demo ->  WPS request builder, I'm guessing this is a limitation or a bug of the current builder. Maybe @aaime knows more about it. Not sure. 

Anyway generating the request like this should work.

offtherailz:
I should put this PR in draft, because it is not ready to be delivered, as far as I understood (and because it is blocked)

alexander-fedorenko:
@offtherailz Thanks for reviewing. You're right, your request (as well as mine in the last sample of the code) is correct. The problem was that I didn't notice initially that gs:Download method transform geometry attribute name to "coordinates" somewhere internally when geojson format is selected. I was using "extractGeometryAttributeName" method to include geometry attribute into output, which caused a problem because such attribute was not available in the data that is feed to the vec:Query process.
I will do minor changes and then it can be tested.

alexander-fedorenko:
@offtherailz Thanks for reviewing. 
You're right, your request (as well as mine in the last sample of the code) is correct. 
The problem was in usage of geoJson format in the intermediate gs:Download process. In this case output is a FeatureCollection, coordinates are not the part of attributes list anymore, so vec:Query process cannot get them from the passed list and they are always ignored. I changed intermediate format to WFS collection and it works flawlessly. Please review :)

offtherailz:
I tested it and I think that we should do it better. Download process has been created to manage big sets of data. With this solution I think you are downloading the data, then resending them to the service.

This is not optimal and for big data sets can be problematic. Anyway I think there is a way to optimize 
 by using [process chaining](https://docs.geoserver.org/stable/en/user/services/wps/processes/chaining.html). 
I never used it but I think it can be the solution. Downloading and re-uploading the features to implement this features is not doable for the kind of dataset we need to manage with this tool.

alexander-fedorenko:
Switched vec:Query process to use reference as an input instead of XML.
Chain will work in a following mode:
gs:DownloadEstimator -> result -> gs:Download -> reference -> vec:Query -> reference. Result can be downloaded by user on demand via widget.
@offtherailz

web/client/selectors/featuregrid.js
```diff
@@ -23,6 +23,7 @@ export const getLayerById = getLayerFromId;
 export const getTitle = (layer = {}) => layer.title || layer.name;
 export const selectedLayerIdSelector = state => get(state, "featuregrid.selectedLayer");
 export const getCustomAttributeSettings = (state, att) => get(state, `featuregrid.attributes[${att.name || att.attribute}]`);
+export const getCustomAttributesSettings = (state) => get(state, `featuregrid.attributes`);
```
MV88, line 26:
```suggestion
export const customAttributesSettingsSelector = (state) => get(state, `featuregrid.attributes`);
```

web/client/utils/FeatureGridUtils.js
```diff
@@ -308,3 +308,26 @@ export const updatePages = (result, { endPage, startPage } = {}, { pages, featur
...
+/**
+ * Process custom attributes settings of feature grid and returns array of feature attributes to be used in a query
+ * to limit results of WFS "getFeature" request
+ * undefined - all attributes to be fetched
+ * array - listed attributes to be fetched
+ * @param {array} attributes complete list of attributes available for export, see attributesSelector
+ * @param {object} customAttributesSettings object containing information about deactivated attributes, see getCustomAttributesSettings
+ * @returns undefined|array
+ */
+export const getAttributesList = (attributes, customAttributesSettings) => {
```
MV88, line 321: can this be unit tested?

web/client/components/data/download/FeatureEditorButton.jsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
-import Button from '../featuregrid/toolbars/TButton';
+import withHint from "../featuregrid/enhancers/withHint";
+import TButtonComp from "../featuregrid/toolbars/TButton";
+const TButton = withHint(TButtonComp);
```
MV88, line 4: why this change?

MV88: i guess it has introduced this
image

MV88 (changes requested):
Please check a few comments

testing this locally with no wps   (direct download it works)

web/client/components/data/download/FeatureEditorButton.jsx
alexander-fedorenko: Yes, tooltip was missing, but I'm wondering why you don't have this string locally? 
image

web/client/utils/FeatureGridUtils.js
alexander-fedorenko: Added test with different scenarios.

web/client/selectors/featuregrid.js
alexander-fedorenko: Thanks, renamed

web/client/epics/layerdownload.js
```diff
@@ -246,7 +253,8 @@ export const startFeatureExportDownload = (action$, store) =>
             filterObj: action.filterObj,
             layerFilter,
             options: {
-                pagination: !virtualScroll && get(action, "downloadOptions.singlePage") ? action.filterObj && action.filterObj.pagination : null
+                pagination: !virtualScroll && get(action, "downloadOptions.singlePage") ? action.filterObj && action.filterObj.pagination : null,
+                pn: propertyNames
```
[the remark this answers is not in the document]
alexander-fedorenko: I just didn't change what was working before me :wink:
Anyways, yes, that's a good point, replaced it with "propertyNames"

alexander-fedorenko:
Please skip this problem with readthedocs, I just don't want to merge master back to this branch cause I'm not sure it won't mess up with the changelog. This issue was fixed on master afaik by you @MV88
Please re-review and let me know if the problem with locale availability is still occurs for that tooltip.

MV88 (approved):

MV88:
@ElenaGallo please test

ElenaGallo:
@alexander-fedorenko sometime it is not possible to download data, in any format and it always happens if I select the attributes I want to export. See the gif attached:


https://user-images.githubusercontent.com/56537133/164026936-e67d6817-fcac-41e7-ba99-2d04d36193ac.mov


https://user-images.githubusercontent.com/56537133/164026964-0c4ae1d4-de40-429c-99f1-a4df2cd06779.mov

tdipisa:
@alexander-fedorenko @ElenaGallo we are too close to the release, therefore I'm going move to the next one (2022.01.02).

alexander-fedorenko:
@ElenaGallo I tried to reproduce problems you had, unfortunately I was not able to do so. As per the screencasts you've shared it really looks like there was some problem with GeoServer running process on any step during file generation. It didn't return meaningful error message in your case, so the only way to trace the error is to reproduce error one more time and check GeoServer logs, or catch a period when problem appears and debug it locally with configured proxy to the dev instance of GeoServer.

ElenaGallo:
Hi @alexander-fedorenko, this issue has been fixed [here](https://github.com/geosolutions-it/MapStore2/issues/7932) so I move this pr to test.

tdipisa:
@ElenaGallo let us know if we can backport to 2022.01.xx.

ElenaGallo:
Test passed, @alexander-fedorenko please backport to 2022.01.xx. Thanks

Issue #7106, opened 2021-07-15
Remove floppy icon also from map editing inside geostories


giohappy:
As done in https://github.com/geosolutions-it/MapStore2/issues/6879 it would be great to remove the save step when editing a map inside a geostory. 

![image](https://user-images.githubusercontent.com/1280027/147103157-b5f94ae3-8797-4348-ac3d-1fe61f5e345d.png)

An example of a particularly unintuitive operation is the change of map zoom and center. This doesn't need an interaction with the editor panel so the user expects to save the map configuration when saving the geostory.

PR #7859, opened 2022-02-10
#7106 Make map in GeoStory updated on the fly without need to save it…

alexander-fedorenko:
## Description
Make map in GeoStory updated on the fly without need to save it by "Save" button explicitly.
- Removed save button from toolbar.
- Click on close glyph will not trigger discard modal. Instead of it all the changes will be kept as is.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other: small changes in feature configuration/workflow

## Issue

#7106 

**What is the new behavior?**
Map configuration is updated with no need to use save button.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/geostory/common/enhancers/__tests__/map-test.jsx
+++ b/web/client/components/geostory/common/enhancers/__tests__/map-test.jsx
@@ -93,12 +93,10 @@ describe("geostory media map component enhancers", () => {
     it('withToolbar it correctly sets buttons', (done) => {
 
         const actions = {
-            toggleEditing: () => {},
             onReset: () => {},
             toggleAdvancedEditing: () => {}
         };
 
-        const SpyToggleEditing = expect.spyOn(actions, 'toggleEditing');
         const SpyOnReset = expect.spyOn(actions, 'onReset');
         const SpyToggleAdvancedEditing = expect.spyOn(actions, 'toggleAdvancedEditing');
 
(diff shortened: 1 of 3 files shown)


alexander-fedorenko:
One particular issue that I can think of without having explicit save button:
- Open geostory with map that allows to pan; change map viewport (move around to another place);
- Click on "configure map" button, change settings of zoom, panning, but do not pan to another viewport.
- Close map configuration edit mode and save geostory.

User may expect that viewport changed prior to activation of the edit mode will be saved along with new settings of the map, but it won't. New viewport and zoom are written to the config only when panning or zoom happen in edit mode.

Previously such situation was impossible because "Save" button was activated only after some changes of pan/zoom/config in edit mode, so user was not able to save configuration without having panning action.

@tdipisa  what do you think? Should we make viewport position of the map saved into configuration automatically once user enters configuration edit mode? Or we should reset viewport to the stored configuration once user switch to edit mode?

dsuren1:
@alexander-fedorenko 
Code looks fine. Will merge this PR once we have a feeback from @tdipisa for the question above.

tdipisa:
@alexander-fedorenko I understand what you mean. To be coherent with the current UI/UX, I think the second option is the right one and so:

> should reset viewport to the stored configuration once user switch to edit mode?

alexander-fedorenko:
@tdipisa With the recent update map viewport will be moved to the last saved configuration when "Edit map configuration" tool is toggled on.
@dsuren1 please review when it's convenient

dsuren1 (approved):
LGTM

dsuren1:
@ElenaGallo Kindly test this issue in DEV once deployed and let us know if it's good to be backported to 2022.01.xx

Issue #7849, opened 2022-02-08
wrong getCapabilities request sent for csw


MV88:
## Description

this issue is coming from https://github.com/geosolutions-it/MapStore2/issues/7800
Looking in the DEV tool, a wrong request to /undefined?service=WMS&version=&request=GetCapabilities is also performed.

## How to reproduce

- open [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/10358)
- open the Catalog tool and select "_CSW Catalog Geo Bretagne"_
- do a search

*Expected Result*

no errors

*Current Result*

wrong request to /undefined?service=WMS&version=&request=GetCapabilities

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

Issue #7690, opened 2021-12-22
Configure visibility limits globally and allow support for CSW services


dsuren1:
## Description

- Allow default onfiguration of visibility limits globally
- Add support for CSW services too
- `autoSetVisibilityLimits` should be false by default (Mapstore)
- Default configuration via MetadataExplorer plugin

Ref: https://github.com/georchestra/mapstore2-georchestra/issues/424#issuecomment-998668124

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Enhancement

PR #7899, opened 2022-03-01
wrong getCapabilities request sent for csw

DavidQuartz:
## Description

This PR fixes wrong requests for some CWS GetCapabalities

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7849 

**What is the new behavior?**

The correct base url is used in the GetCapabilities request

## Breaking change
 - [x] No

### DIFF
--- a/web/client/api/__tests__/CSW-test.js
+++ b/web/client/api/__tests__/CSW-test.js
@@ -108,6 +108,30 @@ describe('Test correctness of the CSW APIs', () => {
             }
         });
     });
+
+    it('does not include capabilities when no parsedUrl in getRecords', (done) => {
+        API.getRecords('base/web/client/test-resources/csw/getRecordsNoWMS.xml', 1, 1).then((result) => {
+            try {
+                expect(result).toBeTruthy();
+                expect(result.records[0].capabilities).toNotExist();
+                done();
+            } catch (ex) {
+                done(ex);
+            }
+        });
+    });
+
+    it('includes capabilities when parsedUrl exists in getRecords', (done) => {
+        API.getRecords('base/web/client/test-resources/csw/getRecordsResponseDC.xml', 1, 1).then((result) => {
+            try {
+                expect(result).toBeTruthy();
+                expect(result.records[0].capabilities).toBeTruthy();
+                done();
+            } catch (ex) {
+                done(ex);
+            }
+        });
+    });
 });
 
 describe('workspaceSearch', () => {
--- a/web/client/api/__tests__/CSW-test.js
+++ b/web/client/api/__tests__/CSW-test.js
@@ -113,7 +113,7 @@ describe('Test correctness of the CSW APIs', () => {
         API.getRecords('base/web/client/test-resources/csw/getRecordsNoWMS.xml', 1, 1).then((result) => {
             try {
                 expect(result).toBeTruthy();
-                expect(result.records[0].capabilities).toNotExist();
+                expect(result.records[0].capabilities).toBeFalsy();
                 done();
             } catch (ex) {
                 done(ex);
@@ -121,8 +121,8 @@ describe('Test correctness of the CSW APIs', () => {
         });
     });
 
-    it('includes capabilities when parsedUrl exists in getRecords', (done) => {
-        API.getRecords('base/web/client/test-resources/csw/getRecordsResponseDC.xml', 1, 1).then((result) => {
+    it('obtains parsedUrl from dc:uri and gets capabilities', (done) => {
+        API.getRecords('base/web/client/test-resources/csw/getRecordsWithDcURI.xml', 1, 1).then((result) => {
             try {
                 expect(result).toBeTruthy();
                 expect(result.records[0].capabilities).toBeTruthy();
--- a/web/client/api/CSW.js
+++ b/web/client/api/CSW.js
@@ -286,8 +286,11 @@ var Api = {
                                                     dc[elName] = finalEl;
                                                 }
                                             }
+                                            const URIs = dc.references.length > 0 ? dc.references : dc.URI;
                                             if (!_dcRef) {
-                                                _dcRef = dc.references.length > 0 ? dc.references : dc.URI;
+                                                _dcRef = URIs;
+                                            } else {
+                                                _dcRef = _dcRef.concat(URIs);
                                             }
                                             obj.dc = dc;
                                         }
--- a/web/client/api/CSW.js
+++ b/web/client/api/CSW.js
@@ -300,7 +300,7 @@ var Api = {
                                 result.records = records;
                                 const { value: _url } = _dcRef?.find(t => t.scheme === 'OGC:WMS' || t.protocol === 'OGC:WMS') || {}; // Get WMS URL from references
                                 const [parsedUrl] = _url && _url.split('?') || [];
-                                return addCapabilitiesToRecords(parsedUrl, result);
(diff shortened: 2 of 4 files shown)


MV88 (changes requested):
@DavidQuartz it seems you have not tested your fix with catalog service specified  in the issue.

the problem still exist because here the parsedUrl is still undefined
image

image


so the fix is around here
https://github.com/geosolutions-it/MapStore2/blob/41230cbf4941050b13af5dc236e0cd031faf80d2/web/client/api/CSW.js#L298-L300

@offtherailz @tdipisa what should we do if the DC_URI has no match for
```js 
t.scheme === 'OGC:WMS' || t.protocol === 'OGC:WMS' 
```

MV88:
> @MV88 If I understood well this is the part that look for a WMS URL in CSW, to add the capabilities URL If I'm correct, if not found (or if a URL is not found for any other of the protocols supported) it should be ignored.

yeah because now, at some point though a getCapabilities request is done anyway
here
https://github.com/geosolutions-it/MapStore2/blob/6e72ede2febfe4e954457b7751c966465f9126c1/web/client/api/WMS.js#L114-L139

so are you saying we should prevent this to happen when we miss the capabilites url?

offtherailz:
Why the request is performed? 
You linked the API function, but only the caller and the reason of this call can suggest what should be the backup bahavior in this case. 
I understand that anyway some "WMS stuff" is done, but It is strange, without recognizing any WMS service URL in the CSW record. I suggest to investigate on this path. Feel free t o contact me

MV88:
@DavidQuartz let me know if you need clarification for the above comments

DavidQuartz:
> 

@MV88 @offtherailz basically, we should avoid making the request in cases where there is no WMS service URL in the CSW record?

MV88:
> @MV88 @offtherailz basically, we should avoid making the request in cases where there is no recognised WMS service URL in the CSW record?

Yes, start investigating this and eventually find a way to avoid it since is not necessary
before you dig into the solution report it here the cause of this

DavidQuartz:
@MV88 @offtherailz after investigation I realised the DC_URIs were not being adequately supplied to `_dcRef` here:
```js 
 if (!_dcRef) {
     _dcRef = dc.references.length > 0 ? dc.references : dc.URI;
 }
```
The WMS service URL is sometimes not included in the first batch of URIs. To fix that, an else block has been included to ensure the continuous supply of DC_URIs like so:
```js 
 if (!_dcRef) {
     _dcRef = dc.references.length > 0 ? dc.references : dc.URI;
 } else {
    const URIs = dc.references.length > 0 ? dc.references : dc.URI;
    _dcRef = _dcRef.concat(URIs);
}
```

In this case we are sure to have all URIs from which to find a WMS service url.

MV88:
> The WMS service URL is sometimes not included in the first batch of URIs.

what does this means exactly?

DavidQuartz:
> > The WMS service URL is sometimes not included in the first batch of URIs.
> 
> what does this means exactly?

I was referring the first run in this loop https://github.com/geosolutions-it/MapStore2/blob/master/web/client/api/CSW.js#L213

`_dcRef` is `undefined` only on the first run, so having only 1 if block to assign URIs to it was not adequate.

MV88:
> _dcRef is undefined only on the first run, so having only 1 if block to assign URIs to it was not adequate.

in that case it can happen that all layers have no WMS uri defined so applying only the else is not a final solution

<hr>

>  const URIs = dc.references.length > 0 ? dc.references : dc.URI;
 if (!_dcRef) {
     _dcRef = URIs;
 } else {
    _dcRef = _dcRef.concat(URIs);
}

with this if the first time is defined with wrong URIs and it never enter the else because already defined
so i would cahnge the check to something like this in the lop 

```js
if(!_dcref) {
  _dcref = dc.references.length > 0 ? dc.references : dc.URI;
  // check for parsed url, if undefined reset _dref to null
  // no else clause
}
```
what do you think ?

<hr>

i was thinking that you should avoid to run addCapabilitiesToRecords if parsedUrl is undefined

if you do this. what happens?

offtherailz:
I checked 10 minutes and I found that `addCapabilitiesToRecords` has been added by @dsuren1 to support visilbility limits in CSW.

https://github.com/geosolutions-it/MapStore2/blob/fbf88b055fe6768101b14acd6ccfccbbbaec7fe6/web/client/api/CSW.js#L87-L93

So this requests should not be done in case it is not present. Moreover IMHO this we should improve this to do the request to the capabilities **only** On layer add, not on CSW page loading.

CSW is often used to optimize with pagination when we have heavy getCapabilities. Always doing getCapabilities, even if with a cache, together with CSW makes this strategy not feasable anymore. 

So two options: 
- FIx this to avoid any request if no WMS reference is found. Open an issue to move the request on layer add
- Fix directly the problem moving the action from record loading to add layer (we already have some requests like this, for time dimension and so on, on layer add).

MV88:
> * FIx this to avoid any request if no WMS reference is found. Open an issue to move the request on layer add
> * Fix directly the problem moving the action from record loading to add layer (we already have some requests like this, for time dimension and so on, on layer add).
> 
> @tdipisa @MV88 decide the strategy accordingly to your timings

I guess we have no time for option 2 which in another moment would have been my choice

@tdipisa 
let's go for option 1 if you agree

dsuren1:
@tdipisa 
I think option 1 should be fine. With respect to #7690, when WMS reference is not present, we won't have capabilities info added, which is valid

tdipisa:
@DavidQuartz I think you can proceed at your convenience with option one then. Thank you so much @dsuren1.

MV88 (changes requested):
@DavidQuartz 
- [x] please add unit test for this PR

- one for parsedUrl that is truthy
- one for parsedUrl  that is falsy

they must check capabilities inside records

web/client/api/__tests__/CSW-test.js
```diff
@@ -108,6 +108,30 @@ describe('Test correctness of the CSW APIs', () => {
             }
         });
     });
+
+    it('does not include capabilities when no parsedUrl in getRecords', (done) => {
+        API.getRecords('base/web/client/test-resources/csw/getRecordsNoWMS.xml', 1, 1).then((result) => {
+            try {
+                expect(result).toBeTruthy();
+                expect(result.records[0].capabilities).toNotExist();
```
MV88, line 116: you can use toBeFalsy() here

web/client/api/__tests__/CSW-test.js
```diff
@@ -108,6 +108,30 @@ describe('Test correctness of the CSW APIs', () => {
...
+                done(ex);
+            }
+        });
+    });
+
+    it('includes capabilities when parsedUrl exists in getRecords', (done) => {
+        API.getRecords('base/web/client/test-resources/csw/getRecordsResponseDC.xml', 1, 1).then((result) => {
+            try {
+                expect(result).toBeTruthy();
+                expect(result.records[0].capabilities).toBeTruthy();
```
MV88, line 128: this check is the same for **getRecords update capabilities**
can you do something more specific or check values

in this case we can merge these two unit tests

MV88 (changes requested):

MV88 (approved):
LGTM

tdipisa:
@ElenaGallo please test this in DEV and let us know if we can backport to 2022.01.xx

ElenaGallo:
Test passed, @DavidQuartz please backport this to 2022.01.xx. Thanks

Issue #7853, opened 2022-02-09
3D Zoom to via share URL


tonym3nsa:
## Description

Update the Share Tool to provide the possibility to zoom to a specified location of the Cesium map through specific 3D viewer parameters.
The URL must contain the data necessary to replicate the exact position in cesium, so camera position (x,y,z) and orientation ( heading, pitch, roll )

**Mockup**

![146767459-42a18bfb-4344-4ce0-a789-063ab7f8923b.jpg](https://images.zenhubusercontent.com/61236737eb6eac6b03da89b8/78aa07be-a5c4-47be-8355-546321702eae)

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #7913, opened 2022-03-07
Issue #7853; 3D zoom to map via share URL - fix after test

tonym3nsa:
## Description

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7853 

**What is the new behavior?**

- The 3D map zooms to the location as provided in the share URL
- The share form accepts negative values for roll and pitch

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/share/SharePanel.jsx
+++ b/web/client/components/share/SharePanel.jsx
@@ -151,12 +151,8 @@ class SharePanel extends React.Component {
             !isEqual(this.props.isVisible, newProps.isVisible)) {
             this.initializeDefaults(newProps);
         }
-        if (this.props?.viewerOptions?.orientation) {
-            this.setState({
-                heading: convertRadianToDegrees(newProps.viewerOptions?.orientation?.heading),
-                pitch: convertRadianToDegrees(newProps.viewerOptions?.orientation?.pitch),
-                roll: convertRadianToDegrees(newProps.viewerOptions?.orientation?.roll)
-            });
+        if (!isEqual(this.props?.viewerOptions?.orientation, newProps.viewerOptions?.orientation)) {
+            this.initializeDefaults(newProps);
         }
     }
 
@@ -198,9 +194,9 @@ class SharePanel extends React.Component {
             defaultLoaded: isVisible,
             isCenterAndZoomDefault,
             isMarkerAndZoomDefault,
-            heading: viewerOptions?.orientation?.heading,
-            pitch: viewerOptions?.orientation?.pitch,
-            roll: viewerOptions?.orientation?.roll
+            heading: convertRadianToDegrees(viewerOptions?.orientation?.heading),
+            pitch: convertRadianToDegrees(viewerOptions?.orientation?.pitch),
+            roll: convertRadianToDegrees(viewerOptions?.orientation?.roll)
         });
     }
 
--- a/web/client/components/share/SharePanel.jsx
+++ b/web/client/components/share/SharePanel.jsx
@@ -25,7 +25,7 @@ import {
     Tooltip
 } from 'react-bootstrap';
 import Message from '../../components/I18N/Message';
-import { join, isNil, isEqual, inRange, isEmpty } from 'lodash';
+import { join, isNil, isEqual, inRange, isEmpty, pick, omit } from 'lodash';
 import { removeQueryFromUrl, getSharedGeostoryUrl, CENTERANDZOOM, BBOX, MARKERANDZOOM, SHARE_TABS } from '../../utils/ShareUtils';
 import { getLonLatFromPoint, convertRadianToDegrees, convertDegreesToRadian } from '../../utils/CoordinatesUtils';
 import { getMessageById } from '../../utils/LocaleUtils';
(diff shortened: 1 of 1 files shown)


web/client/components/share/SharePanel.jsx
```diff
@@ -151,6 +151,13 @@ class SharePanel extends React.Component {
             !isEqual(this.props.isVisible, newProps.isVisible)) {
             this.initializeDefaults(newProps);
         }
+        if (this.props?.viewerOptions?.orientation) {
```
dsuren1, line 154: I think we can move this block inside `initializeDefaults` by adding a condition to check `!isEqual(this.props?.viewerOptions?.orientation, newProps.viewerOptions?.orientation)` invoking `initialzeDefaults` func

dsuren1 (changes requested):
@tonym3nsa 
Looks fine. Kindly update the PR with this minor change.

web/client/components/share/SharePanel.jsx
tonym3nsa: well noted. Thanks

web/client/components/share/SharePanel.jsx
```diff
@@ -151,6 +151,9 @@ class SharePanel extends React.Component {
             !isEqual(this.props.isVisible, newProps.isVisible)) {
             this.initializeDefaults(newProps);
         }
+        if (!isEqual(this.props?.viewerOptions?.orientation, newProps.viewerOptions?.orientation)) {
+            this.initializeDefaults(newProps);
+        }
```
dsuren1, line 156: Why not have this block here with another `||` operator https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/share/SharePanel.jsx#L151? We could avoid a separate if block.

dsuren1 (changes requested):
@tonym3nsa 
Could you check this please

dsuren1 (approved):
LGTM

dsuren1:
@ElenaGallo Kindly test this fix once deployed to DEV

ElenaGallo:
@tonym3nsa there is a regression, now the Min and Max values on Heading, Roll and Pitch allow values beyond the specified limit range
value

tonym3nsa:
@ElenaGallo I can't seem able to recreate the issue; can you please let me know how you entered the figures?

ElenaGallo:
@tonym3nsa you are right, the problem only occurs on the second attempt to enter the value.


3d

tdipisa:
> @ElenaGallo I can't seem able to recreate the issue; can you please let me know how you entered the figures?

The same for me @ElenaGallo

tonym3nsa:
@ElenaGallo I might have seen what the issue is. It happens when you type fast. Providing a  fix right now

web/client/components/share/SharePanel.jsx
tonym3nsa: @dsuren1 I think I might have to revert to the original implementation. After refactoring to this, the inputted values go beyond the specified range in certain cases.

tdipisa:
@ElenaGallo @tonym3nsa (FYI @dsuren1) It seems to me one problem can be the following. @tonym3nsa (@dsuren1) let me know your thoughts.

Here I'm in a situation (without typing anything in the fields) where values in form seems wrong:

2022_03_08_12_36_31_MapStore_HomePage

and they don't match the ones in the URL (**which anyway shows exactly what I'm trying to share**):

https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/cesium/10358?center=-102.56493882780978,32.27312307744788&zoom=4.636726959224782&heading=6.2471550569244325&pitch=-1.4768349282484725&roll=6.282676334180332
 
image

Below the steps to reproduce (sometimes I can reproduce and sometimes not):

https://demo.geo-solutions.it/share/tobia/3d_cesium.gif

dsuren1:
@tdipisa 
This url https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/cesium/10358?center=-102.56493882780978,32.27312307744788&zoom=4.636726959224782&heading=6.2471550569244325&pitch=-1.4768349282484725&roll=6.282676334180332 has incorrect roll (360), probably created before the fixes, not sure.

Can you try the same from the share panel, by supplying incorrect orientation values, because `onBlur` of the field, the validation kicks in and resets the incorrect value in the field. So it shouldn't be possible to create an incorrect share url with orientations, if not then this needs to be fixed.

tdipisa:
> This url https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/cesium/10358?center=-102.56493882780978,32.27312307744788&zoom=4.636726959224782&heading=6.2471550569244325&pitch=-1.4768349282484725&roll=6.282676334180332 has incorrect roll (360), probably created before the fixes, not sure.
> 
> Can you try the same from the share panel, by supplying incorrect orientation values, because `onBlur` of the field, the validation kicks in and resets the incorrect value in the field. So it shouldn't be possible to create an incorrect share url with orientations, if not then this needs to be fixed.

@dsuren1 as you can see from the video, I've used the URL as it is provided by the Share tool. I disn't change anything in the form, I've simply changed the orientation in Cesium and this sometimes generate wrong values for Roll. In addition, all values in the form are different from the ones in the URL.

tonym3nsa:
@tdipisa the values in the URL are in radian whereas the values in the SharePanel are in degrees - that is what they don't match. What I do is get the value from Cesium and convert it into degrees in the share panel. At times, during the back and forth conversions of the float, instead of getting a whole number such as  for example - 360 degrees, the conversion causes the figure to have more decimal places - thus, in your example, instead of 360 degrees, we have 359.97....

tdipisa:
> @tdipisa the values in the URL are in radian whereas the values in the SharePanel are in degrees - that is why they don't match. What I do is get the radian value from Cesium and convert it into degrees in the share panel. At times, during the back and forth conversions of the float, instead of getting a whole number such as for example - 360 degrees, the conversion causes the figure to have more decimal places - thus, in your example, instead of 360 degrees, we have 359.97....

I understand, thank you for the explanation.

Issue #7890, opened 2022-02-28
Align print-lib to the latest version


mbarto:
## Description

After merging https://github.com/geosolutions-it/mapfish-print/pull/49, some changes to the application context will be needed to match requirements of the updated print-lib library (that has been aligned with the latest official 2.1.5 version).

In particular, the following should be added to all the instances of applicationContent-print.xml in the repo (both for the product and standard project):

```xml
<bean id="threadResources" class="org.mapfish.print.ThreadResources">
        <property name="connectionTimeout" value="30000"/>
        <property name="socketTimeout" value="30000" />
        <property name="globalParallelFetches" value="200"/>
        <property name="perHostParallelFetches" value="30" />
</bean>
...
<bean id="metricRegistry" class="com.codahale.metrics.MetricRegistry" lazy-init="false"/>
<bean id="healthCheckRegistry" class="com.codahale.metrics.health.HealthCheckRegistry" lazy-init="false"/>
<bean id="loggingMetricsConfigurator" class="org.mapfish.print.metrics.LoggingMetricsConfigurator"  lazy-init="false"/>
<bean id="jvmMetricsConfigurator" class="org.mapfish.print.metrics.JvmMetricsConfigurator" lazy-init="false"/>
<bean id="jmlMetricsReporter" class="org.mapfish.print.metrics.JmxMetricsReporter" lazy-init="false"/>
```

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

tdipisa:
Thank you @mbarto for the update. The migration guidelines should be updated accordingly then. It would be good to start testing the new version in DEV soon if we want to release this on 2022.01.00, the release deadline is quite close.

offtherailz:
I had to revert the PR because it caused the problem listed.
See details here: https://github.com/geosolutions-it/MapStore2/pull/7891#issuecomment-1060455482

PR #7942, opened 2022-03-09
#7890: align applicationContext-print.xml to work with upgraded mapfish-print library

mbarto:
## Description

Upgrades print-lib to use geosolutions-2.1-SNAPSHOT.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: library upgrade

## Issue
#7890 

MapStore won't start with the upgraded print-lib.

**What is the new behavior?**

MapStore starts correctly with the upgraded print-lib.

## Breaking change
 - [ ] No

All projects built with the printing profile enabled will need to update the applicationContext-print.xml file as explained in the migration notes.

### DIFF
--- a/docs/developer-guide/mapstore-migration-guide.md
+++ b/docs/developer-guide/mapstore-migration-guide.md
@@ -153,7 +153,7 @@ Downstream project should update following configurations:
 ### Upgrading the printing engine
 The mapfish-print based printing engine has been upgraded to align to the latest official 2.1.5 in term of functionalities.
 
-An update to the MapStore printing engine context file (applicationContext-print.xml) is needed for all projects built with the printing profile enabled. The following sections should be added to the file:
+An update to the MapStore printing engine context file (`applicationContext-print.xml`) is needed for all projects built with the printing profile enabled. The following sections should be added to the file:
 
 ```diff
 <bean id="configFactory" class="org.mapfish.print.config.ConfigFactory"></bean>
--- a/docs/developer-guide/mapstore-migration-guide.md
+++ b/docs/developer-guide/mapstore-migration-guide.md
@@ -150,6 +150,29 @@ Downstream project should update following configurations:
 
 - This step is needed only for custom project with a specific `publicPath` different from the default one. In this case you may need to specify what folder deliver the  cesium build ( by default `dist/cesium`). To do that, you can add the  `cesiumBaseUrl` parameter in the webpack dev and prod configs to the correct location of the cesium static assets, widgets and workers folder.
 
+### Upgrading the printing engine
+The mapfish-print based printing engine has been upgraded to align to the latest official 2.1.5 in term of functionalities.
+
+An update to the MapStore printing engine context file (applicationContext-print.xml) is needed for all projects built with the printing profile enabled. The following sections should be added to the file:
+
+```diff
+<bean id="configFactory" class="org.mapfish.print.config.ConfigFactory"></bean>
++<bean id="threadResources" class="org.mapfish.print.ThreadResources">
++    <property name="connectionTimeout" value="30000"/>
++    <property name="socketTimeout" value="30000" />
++    <property name="globalParallelFetches" value="200"/>
++    <property name="perHostParallelFetches" value="30" />
++</bean>
+
+<bean id="pdfOutputFactory" class="org.mapfish.print.output.PdfOutputFactory"/>
++
++<bean id="metricRegistry" class="com.codahale.metrics.MetricRegistry" lazy-init="false"/>
++<bean id="healthCheckRegistry" class="com.codahale.metrics.health.HealthCheckRegistry" lazy-init="false"/>
++<bean id="loggingMetricsConfigurator" class="org.mapfish.print.metrics.LoggingMetricsConfigurator"  lazy-init="false"/>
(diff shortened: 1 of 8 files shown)


offtherailz (changes requested):
We have a big list of exclusions. They are all jars coming from the printing module?
I'm afraid the conflict comes in the maven-war-plugin, but porting and maintaining them across all projects  can become a mess.

It is difficoult to exclude them in directly in the printing modiule/printing bundle? Otherwise we need to document and explain a lot of things. 

Please let me know, tree options: 
- [ ] Exclude the jars directly in the printing modules (printing bundle, printing modiule)
- [ ] If the first are not possible, exclude them in the printing profile(s), so at least they are self contained. 
- [ ] If neither of the options before is possible and we absolutely need this long list of packagingExcludes, we need to document 
   - [ ] Commenting them. so it is clear where they are coming from
   - [ ] Add documentation to migration guidelines

Feel free to ask questions or clarify something if I said something wrong.

mbarto:
> We have a big list of exclusions. They are all jars coming from the printing module?
No, the ones coming from printing are the ones explicitly excluded in the printing profile as subdependencies of mapfish-print.
The other ones are other duplicates I found along the way, and thought it would be good to do some cleaning. Indeed it would be good to clean up things every now and then.

>I'm afraid the conflict comes in the maven-war-plugin, but porting and maintaining them across all projects can become a mess.
This is why we tried to sponsor the new projects architecture, to automate migrations as much as possible, instead of having to document them, but that project, as you know, didn't receive enough love to replace the current one.

> It is difficoult to exclude them in directly in the printing modiule/printing bundle? Otherwise we need to document and explain a lot of things.
I don't think this is possible with current projects, because they have their own pom.xml and that's exactly what we were trying to avoid with the NEW projects.

> Please let me know, tree options:
> 
> * [ ]  Exclude the jars directly in the printing modules (printing bundle, printing modiule)
> * [ ]  If the first are not possible, exclude them in the printing profile(s), so at least they are self contained.
> * [ ]  If neither of the options before is possible and we absolutely need this long list of packagingExcludes, we need to document
We don't absolutely need them, I can remove the excludes that are not in the printing profile, as I said above I added them as a bonus, but if they are just triggering more work on my side, I am happy to remove them from the PR.

>   * [ ]  Add documentation to migration guidelines
I can add this, limited to the excludes in the printing profile.

offtherailz:
Thank you for the clarification.

**If** the problem is solved by the exclusions in mapfish-print dependency, it is ok for me. 

People need to:
- Update mapfish-print
- Being able to update project with mapfish print, possibly without knowing every single detail

Is essential to not have to solve dependencies problems, at least for standard project, in every single project. 
So please add to the migration guidelines also the part regarding the exclusions necessary for this fix. 

About the clean-up I thank you a lot. If you don't explain it in the PR I assume they are part of the fix. 

These things are very delicate so, if they are not essential for this fix, I'd do them in a separate PR. It is very important to be able to account the changes to a specific bug/improvement.  
If you agree I'd open a separate issue for it. 
The best solution should be to exclude the dependencies from the overlay they are coming from, so It's easier for us to account each of them. But I understand this requires time. In any case, this work goes also on the war scheleton for the new project, so it is not thrown away.

mbarto:
@offtherailz, I pushed an update, removing additional exclusions and (as a bonus) fixing the printingbundle profile too.

> **If** the problem is solved by the exclusions in mapfish-print dependency, it is ok for me.

The problem you notified should be solved by the above mentioned exclusions, but since the failure only happens on Linux, I would suggest you to test on this platform (I could only test Windows, where it was working before and it is still working now). I can't test eventual other regressions not happening on Windows.

> * Update mapfish-print
> * Being able to update project with mapfish print, possibly without knowing every single detail

The migration guide has been enriched with info on updating the projects pom.xml properly.

> It is essential to not have to solve dependencies problems, at least for standard project, in every single project. So please add to the migration guidelines also the part regarding the exclusions necessary for this fix.

If it is essential, I would suggest to prioritize adopting the new projects architecture.

> These things are very delicate so, if they are not essential for this fix, **I'd do them in a separate PR**. It is very important to be able to account the changes to a specific bug/improvement. If you agree I'd open a separate issue for it. 

You can do it, I cannot guarantee I will have time to redo the cleanup work again soon.

offtherailz:
Thank you Mauro. Unfortunately as I said in the PR, I wasn't able to replicate even on my desktop linux env too. 
It was happening only on dev environment, and it was blocking all the tests on MapStore, for this reasons I had to revert. 

A general clean-up is needed, as you said. I'll take note of your work in a separate issue so we can create a separate PR. 
Thank you for patience, but I think we should be able to exaclty identify the issue and the solution, in order to not have the same problem in the future. 

I'm not around tomorrow, and I can not verify it is working, so I'd merge this PR monday @tdipisa

offtherailz:
One last question @mbarto , here you excluded `guava`, `httpclient`, `httpcore` and `xerces` from you previous PR.
I didn't get the reason. 

https://github.com/geosolutions-it/MapStore2/pull/7942/files/7bea7f60e4fac376d4b571da6b748776f00a776f..6075d9dd0accc58120760ff4ed7eeb149339ecb2#diff-f9bdfc59d49013ad57196e3afa6e1eb15fe1277662806a0b66650e5f8ed05255L332-L343

Issue #7997, opened 2022-03-22
Improve migration guidelines 


MV88:
## Description

While working on an [update](https://github.com/geosolutions-it/MapStore2-C040/issues/538) of a project  to latest 2022.01.xx 
I encountered a few missing parts, or typos in the recent migration guidelines

here is a list of things

- [ ] some project may define old versions of Jackson package which is not compatible with new spring version (5.3.9), 
so we suggest to remove them and use the one springs brings in

- [ ] remove jackson version declaration for new projects [here](https://github.com/geosolutions-it/MapStore2/blob/31cc4847b973b3b11514ee6270f38459941a4217/project/standard/templates/backend/pom.xml#L22-L30) and [here](https://github.com/geosolutions-it/MapStore2/blob/31cc4847b973b3b11514ee6270f38459941a4217/project/custom/templates/backend/pom.xml#L22-L30)

- [x] another problem i found was that in the migration guidelines things are not so clear
in particular when it specifies two different versions of servlet-api

![image](https://user-images.githubusercontent.com/11991428/159299893-3ec1ec7d-cb0a-4a7e-be64-b744ffef5880.png)
here the correct version is 3.1.0 and the artifactId is javax.servlet-api and this is not included in the steps

- [ ] the printing engine update is not backported yet on 2022.01.xx hence it should be included in these migration guidelines (from 2021.02.00 to 2022.01.00)

- [x] another point that I don't understand why it specifies the snapshot version instead of fixed ones
![image](https://user-images.githubusercontent.com/11991428/159300871-176c29cf-8b76-4f46-8b09-2e4e1bbd8786.png)
in the projects usually we use fixed versions hence the guide should be amended. please correct me if I'm wrong

- [x] the order of migration guidelines is not correct
![image](https://user-images.githubusercontent.com/11991428/159444556-756fe9f2-e8b5-487d-acab-e3f9c491598a.png)

- [ ] this is more of a questions, but why we list these dependency update if we do not use directly in the java-services , but in geostore? (CXF, JPA, Hibernate, hibernate-generic-dao, h2   )

*Documentation section involved*
- [ ] User Guide
- [x] Developer Guide

PR #8031, opened 2022-03-25
[Backport 2022.01.xx] #7997 Fixed migration guidelines

offtherailz:
[Backport 2022.01.xx] #7997 Fixed migration guidelines


MV88 (changes requested):
I've requested changes to PR on master, so those changes will need to be backported here too

offtherailz:
It should be aligned

Issue #8029, opened 2022-03-25
Save button disappear on map Import


tdipisa:
## Description

Save button disappear on map Import.

## How to reproduce

![map_import.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/bef52267-e320-4707-8ad1-3a0a2198888b)

*Expected Result*

If the user is authenticated with edit permission on the current map, then the Save button should be maintained in the Option menu.

*Current Result*

The Save button disappear in the Option menu.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

This but affects DEV, QA and STABLE

PR #8044, opened 2022-03-29
Issue #8029 - Save button disappear on map Import

ale-cristofori:
Fixes the bug including `fileName` and `mapId` when loading
the json and wmc file, moreover triggers the mapInfo action
to record mapSave conditionals variables to make the save button appear

## Description

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

Loading a map configuration either in `json` or `wmc` from an already saved map (not a new one) when the loaded map finished loading, overwriting the pre-saved one the button `Save` disappears from the burger menu plugin. As a consequence it is not possible to overwrite a pre-saved map with a new configuration.
#8029 

**What is the new behavior?**
Loading a map configuration either in `json` or `wmc` from an already saved map (not a new one) when the loaded map finished loading, overwriting the pre-saved one the button `Save` is in the burger menu plugin. It is now possible to overwrite a pre-saved map with a new configuration.

recording_1

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/import/dragZone/enhancers/__tests__/useFiles-test.js
+++ b/web/client/components/import/dragZone/enhancers/__tests__/useFiles-test.js
@@ -24,7 +24,6 @@ describe('useFiles enhancer', () => {
         setTimeout(done);
     });
     it('useFiles rendering with map', (done) => {
-
         const actions = {
             loadMap: (conf, mapId, zoomToExtent ) => {
                 expect(conf).toExist();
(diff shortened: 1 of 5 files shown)


web/client/reducers/config.js
```diff
@@ -69,7 +69,9 @@ function mapConfig(state = null, action) {
             }),
             mapConfigRawData: { ...action.config }
         };
-        newMapState.map = assign({}, newMapState.map, {mapId: action.mapId, size, version: hasVersion ? action.config.version : 1});
+        // if map is loaded from an already saved map keep the same id
+        const mapId = action.config?.fileName && state.map.present.mapId ? state.map.present.mapId : action.mapId;
```
MV88, line 73: - [ ]  I do not think there be the case where  action.config?.fileName is falsy and state.map.present.mapId is truthy,
but the opposite should be possible when you import on a new map, so please test this use case

- [ ] and check also if it is already present the case where the mapId is taken from the action

- [ ] when you first load a map `action.config?.fileName` is undefined and `state.map` is null therefore it does not causes a problem, but i would make it more robust the second half of the condition by adding optional chain operator ?.  ` =>  state.map?.present?.mapId`

web/client/components/import/dragZone/enhancers/useFiles.js
```diff
@@ -9,21 +9,26 @@ import { checkIfLayerFitsExtentForProjection } from '../../../../utils/Coordinat
...
                         ...map,
                         map: {
                             ...map.map,
                             zoom: map.map.zoom || zoom,
                             center: map.map.center || center
                         }
-                    }, null, !map.map.zoom && (map.map.bbox || {bounds: map.map.maxExtent}));
+                    }, mapId ? mapId : null, !map.map.zoom && (map.map.bbox || {bounds: map.map.maxExtent}));
+                    // keeps mapinfo of pre-existing map if present (to keep save map overwrite)
+                    if (mapId && fileName) {
```
MV88, line 29: I think that using fileName as a check here is not necessary, but we can keep it as a more strong check

web/client/components/import/dragZone/enhancers/__tests__/useFiles-test.js
```diff
@@ -57,6 +57,45 @@ describe('useFiles enhancer', () => {
...
+                expect(conf.map).toExist();
+                expect(conf.map.bbox).toExist();
+                expect(conf.map.center).toExist();
+                expect(conf.map.zoom).toExist();
+                expect(mapId).toExist();
+                expect(zoomToExtent).toBe(false);
+                done();
+            },
+            loadMapInfo: (mapId, info) => {
+                expect(mapId).toBe("10");
```
MV88, line 73: are you sure this check are done?

because when the done() in loadMap is called the test stops, so please make sure they are called for example put a 
```js 
expect(false).toBeTruthy();
```

in this way you are sure that the test would fail, and maybe we can move here the call to done() 
also when you move here the done add the same tricky check on the loadMap callback to see if everything works as expected

MV88 (changes requested):
@ale-cristofori hi
I've left a couple of comments,
let me know if you need some clarifications

web/client/components/import/dragZone/enhancers/__tests__/useFiles-test.js
ale-cristofori: I modified the test and checked the coverage, this now looks like it is executed

web/client/components/import/dragZone/enhancers/useFiles.js
ale-cristofori: If it is no harm having the extras check, I'd leave it there

web/client/reducers/config.js
MV88: > Added one test when an external config file is loaded and current map has an id

thank you

MV88 (changes requested):

MV88:
@ElenaGallo please test

ElenaGallo:
Test passed, @ale-cristofori please backport to 2021.01.xx.

ale-cristofori:
> 2021.01.xx.

@ElenaGallo, did you mean 2022.01.xx ?

ElenaGallo:
@ale-cristofori yes, sorry

PR #8163, opened 2022-05-04
Restore snapshots for 2022.01.xx

MV88:
## Description

Restored snapshot versions for maven modules

@offtherailz please keep a note on templates files 
should we restore snapshots also on other dependencies like geostore proxy ecc?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## [2022.01.01](https://github.com/geosolutions-it/MapStore2/tree/v2022.01.01) (2022-05-03)
+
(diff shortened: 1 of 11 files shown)


offtherailz (approved):
Good. @MV88 No we don't need to do it. We keep the dependency fixed while we don't have any important update to do, on stable branch.

MV88:
> Good. @MV88 No we don't need to do it. We keep the dependency fixed while we don't have any important update to do, on stable branch.

therefore I need to provice a change for this PR, please wait to merge this

MV88:
@offtherailz 
PR updated, please review again

PR #8169, opened 2022-05-05
8086 Bugfixes for layout changes

alexander-fedorenko:
## Description
This PR resolves issues found after testing
https://github.com/geosolutions-it/MapStore2/pull/8085#issuecomment-1118586740

1.  Sidebar now will be active in map viewer available on context creation/edit.  Original issue appeared because burger menu became not available in "desktop" list of enabled plugins in localConfig.json. Map viewer available on "Map configuration" step of context creation/edit is using "viewer" array of plugins from localConfig.json that will fall back to "desktop" when "viewer" is not available. That's the case for default version of the file, moreover, if we add "viewer" array there, it will also affect /viewer/ path and Burger menu will return there. @offtherailz @tdipisa Could you check this point just to be sure that this is correct and applicable approach in this case?
2. Corrected classes applied to the map catalog and identify container when they are open. That was the reason of incorrect offset for catalog at the map viewer that opens in modal (e.g. in geostory)
3. Cesium viewer has no option to customize compass position except of css. Therefore I've added sidebar flag to the map plugin, this flag is true when sidebar plugin is instantiated on the page. Cesium map implementation uses this flag to apply "data-sidebar" attribute to the wrapper. Additional offset for compass is applied if this data attribute exists.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

https://github.com/geosolutions-it/MapStore2/pull/8085#issuecomment-1118586740

**What is the new behavior?**
As per PR description 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/contextcreator/ContextCreator.jsx
+++ b/web/client/components/contextcreator/ContextCreator.jsx
@@ -17,8 +17,9 @@ import ConfigureMap from './ConfigureMapStep';
 import ConfigureThemes from './ConfigureThemes';
 import {CONTEXT_TUTORIALS} from '../../actions/contextcreator';
 /**
- * Filters plugins and applies overrides.
- * The resulting array will filter the pluginsConfig returning only the ones present in viewerPlugins.
+ * Merges plugins "cfg" from pluginsConfigs and applies "overrides" from viewerPlugins.
+ * The resulting array will return viewerPlugins with "cfg" applied directly from the pluginsConfigs and optionally overridden
+ * with "overrides" from viewerPlugins.
  * If some viewerPlugin is an object, it can contain a special entry "overrides". If so, the configuration here
  * will override the ones in the original plugin config.
  * Actually overrides are supported only for "cfg" values.
@@ -46,17 +47,20 @@ import {CONTEXT_TUTORIALS} from '../../actions/contextcreator';
  *               "cfg": { activateQueryTool: false, otherOptions: true }
  *           },
  *   ```
+ * NOTE: Logic has changed to support custom list of plugins to be active on the map configuration step of context wizard.
+ * The final result will match previous implementation except that it will also contain plugins that are missing
+ * in "viewer"/"desktop" array inside localConfig.json -> plugins
  * @param {array} pluginsConfigs array of plugins (Strings or objects) to override
  * @param {array} viewerPlugins list of plugins to use
  */
 export const pluginsFilterOverride = (pluginsConfigs, viewerPlugins) => {
-    return pluginsConfigs.map(p => {
-        const pName = isObject(p) ? p.name : p;
+    return viewerPlugins.map((viewerPlugin) => {
+        const pName = isObject(viewerPlugin) ? viewerPlugin.name : viewerPlugin;
         // find out
-        const viewerPlugin = find(viewerPlugins, vp => {
-            return pName === (isObject(vp) ? vp.name : vp);
+        const p = find(pluginsConfigs, plugin => {
+            return pName === (isObject(plugin) ? plugin.name : plugin);
         });
-        if (viewerPlugin) {
+        if (p) {
             if (isObject(viewerPlugin) && viewerPlugin.overrides) {
                 const newP = isObject(p) ? p : { name: p };
                 const cfg = {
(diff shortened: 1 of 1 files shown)


MV88:
@alexander-fedorenko i noticed that here in toc toolbar it is missing add layer button, even if catalog is closed
image

MV88:
Is this already tracked?
image

cesium navigation gets covered a little

offtherailz:
Same issue for the addLayer is for the BackgroundSelector (add background functionality)

ElenaGallo:
Hi @alexander-fedorenko, I don't know if it's a problem, but when the burger menu plugin is placed in a context and the catatologist or gfi panel is opened a piece of search bar is displayed below.

https://user-images.githubusercontent.com/56537133/167108991-53bd7f74-c3fe-46be-8eb6-c25a9a636023.mov

web/client/components/map/cesium/Map.jsx
```diff
@@ -327,7 +329,7 @@ class CesiumMap extends React.Component {
         }) : null;
         const ErrorPanel = this.props.errorPanel;
         return (
-            <div id={this.props.id}>
+            <div id={this.props.id} data-sidebar={this.props.sidebar ? true : null}>
```
allyoucanmap, line 332: Map components and plugins should not be aware of the sidebar plugin.
In my opinion we could keep the 40px margin always without detect if the burger menu is active or not.
So please remove this data and the related prop.

web/client/plugins/Map.jsx
```diff
@@ -408,6 +410,7 @@ class MapPlugin extends React.Component {
                     zoomControl={this.props.zoomControl}
                     onResolutionsChange={this.props.onResolutionsChange}
                     errorPanel={ErrorPanel}
+                    sidebar={this.props.sidebar}
```
allyoucanmap, line 413: Please remove this data and the related prop.

web/client/plugins/map/selector.js
```diff
@@ -15,6 +16,7 @@ export default createStructuredSelector({
     projectionDefs: projectionDefsSelector,
     map: mapSelector,
     mapType: mapTypeSelector,
+    sidebar: sidebarIsActiveSelector,
```
allyoucanmap, line 19: So please remove the sidebar prop.

web/client/themes/default/less/cesium.less
```diff
@@ -28,4 +28,8 @@
     text-shadow: none;
     display: table-cell;
     vertical-align: middle;
-}
\ No newline at end of file
+}
+
+#map[data-sidebar] .cesium-viewer .compass {
```
allyoucanmap, line 33: I would remove the [data-sidebar] and apply always the 40px
the results without sidebar and burgermenu as an acceptable margin
image

allyoucanmap (changes requested):
> Sidebar now will be active in map viewer available on context creation/edit. Original issue appeared because burger menu became not available in "desktop" list of enabled plugins in localConfig.json. Map viewer available on "Map configuration" step of context creation/edit is using "viewer" array of plugins from localConfig.json that will fall back to "desktop" when "viewer" is not available. That's the case for default version of the file, moreover, if we add "viewer" array there, it will also affect /viewer/ path and Burger menu will return there. @offtherailz @tdipisa Could you check this point just to be sure that this is correct and applicable approach in this case?

We have an inconsistency with the sidebar due to the login position where the step 2 have the login inside the sidebar instead of the omnibar.
@tdipisa @alexander-fedorenko @offtherailz
if you don't bother about this we could fix later anyway we need to align this side bar in all viewer as I mentioned in the previous PR review https://github.com/geosolutions-it/MapStore2/pull/8085#pullrequestreview-956757238

image

image

image

image

web/client/themes/default/less/cesium.less
alexander-fedorenko: Reverted commit and added static position change regardless of sidebar availability

tdipisa:
@ElenaGallo please complete tests of this and #8085 in DEV

ElenaGallo:
Test passed, @alexander-fedorenko please backport this pr on the dedicated georchestra branch. Thanks

Issue #8206, opened 2022-05-18
Timeline - Disable Time interval for a layer with 'instants' 


ElenaGallo:
Related to this https://github.com/geosolutions-it/MapStore2/issues/7958#issue-1168680480 when the timeline of the layer with 'instants' (ex: Meteorite_Landings_from_NASA_Open_Data_Portal) the Time interval snap point should
be disabled because selecting the End, the animation option does not work.

**How to Reproduce**
- Open [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/39497)
- Follows the steps in the gif below:
![connseerioni_unmiss.gif](https://images.zenhubusercontent.com/5da5b30f66de7b0001e4e77c/9e960439-32df-486b-b8e4-6162a0672e56)

**Current Result**
The animation does not work

tdipisa:
@ale-cristofori  according to what we saw together it is needed to review how the End radio is managed across different layers with different time domain type (range/instants) because not only the animation is not working (something that could be also acceptable as a short term functionality to be reviewed later) but also the snapping when the End radio is selected has problems (and this need to be fixed somehow)

ale-cristofori:
Double checking with @taba90 there might be something not quite right BE side.

An example with `time_vlaanderen` dataset and this request

http://localhost:8081/proxy/?url=https%3A%2F%2Fvlaanderen-dev-mapstore.geosolutionsgroup.com%2Fgeoserver%2Fvlaanderen%2Fgwc%2Fservice%2Fwmts%3Fservice%3DWMTS%26version%3D1.0.0%26request%3DGetDomainValues%26tileMatrixSet%3DEPSG%253A4326%26layer%3Dtimeline_vlaanderen%26domain%3Dtime%26fromValue%3D2020-12-31T23%253A00%253A00.000Z%26sort%3Dasc%26limit%3D1%26fromEnd%3Dfalse

Should at least return 1 result since we have at least 1 interval starting on 31/12/2020. No domains are returned instead, this made the enabling/disabling of the radio button not consistent and that's why we decided to remove the conditional disabling.

@taba90 is having a look, hopefully the solution is BE and we will be able to re-implement the conditional disabling of the radio button.

ale-cristofori:
The main issue for the inconsistent behaviour here is the need to implement a logic to enable/disable the radio button (start/end) when a layer made of just instants time occurrences is selected. 



The logic as is as:
[...]

PR #8229, opened 2022-05-25
#8206 Separated workflow for setting data as interval

alexander-fedorenko:
## Description
Implemented separated workflow for setting data as interval
code refactoring for multidim getDomainValues observable

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8206 

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/__tests__/playback-test.js
+++ b/web/client/epics/__tests__/playback-test.js
@@ -7,24 +7,15 @@
 */
 
 import expect from 'expect';
-
-// import { testEpic, addTimeoutEpic } from './epicTestUtils';
-import { testEpic } from './epicTestUtils';
-// import { UPDATE_METADATA, STOP, play, stop, FRAMES_LOADING, SET_FRAMES, SET_INTERVAL_DATA } from '../../actions/playback';
-import { UPDATE_METADATA, STOP, play, stop, FRAMES_LOADING, SET_FRAMES } from '../../actions/playback';
+import {testEpic} from './epicTestUtils';
+import {UPDATE_METADATA, STOP, play, stop, FRAMES_LOADING, SET_FRAMES, SET_INTERVAL_DATA} from '../../actions/playback';
 
 import {
     retrieveFramesForPlayback,
     playbackStopWhenDeleteLayer,
-    playbackCacheNextPreviousTimes
+    playbackCacheNextPreviousTimes,
+    setIsIntervalData
 } from '../playback';
-
-// import {
-//     retrieveFramesForPlayback,
-//     playbackStopWhenDeleteLayer,
-//     playbackCacheNextPreviousTimes,
-//     setIsIntervalData
-// } from '../playback';
 
 import DOMAIN_VALUES_RESPONSE from 'raw-loader!../../test-resources/wmts/DomainValues.xml';
 import DOMAIN_INTERVAL_VALUES_RESPONSE from 'raw-loader!../../test-resources/wmts/DomainIntervalValues.xml';
@@ -299,17 +290,17 @@ describe('playback Epics', () => {
             }
         });
     });
-    // it.only('setIsIntervalData', done => {
-    //     testEpic(addTimeoutEpic(setIsIntervalData, 88), 1, selectLayer("testLayerId"), ([action]) => {
-    //         mock.onGet('MOCK_DOMAIN_VALUES').reply(200, DOMAIN_INTERVAL_VALUES_RESPONSE);
-    //         try {
-    //             const { type, timeIntervalData } = action;
-    //             expect(type).toBe(SET_INTERVAL_DATA);
-    //             expect(timeIntervalData).toBe(true);
-    //             done();
-    //         } catch (e) {
-    //             done(e);
-    //         }
-    //     }, ANIMATION_MOCK_STATE);
-    // });
+    it('setIsIntervalData', done => {
+        mock.onGet('MOCK_DOMAIN_VALUES').reply(200, DOMAIN_INTERVAL_VALUES_RESPONSE);
+        testEpic(setIsIntervalData, 1, selectLayer("playback:selected_layer"), ([action]) => {
+            try {
+                const { type, timeIntervalData } = action;
+                expect(type).toBe(SET_INTERVAL_DATA);
+                expect(timeIntervalData).toBe(true);
+                done();
+            } catch (e) {
+                done(e);
+            }
+        }, ANIMATION_MOCK_STATE);
+    });
 });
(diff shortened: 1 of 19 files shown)


web/client/epics/__tests__/playback-test.js
```diff
@@ -290,4 +290,17 @@ describe('playback Epics', () => {
             }
         });
     });
+    it('setIsIntervalData', done => {
+        mock.onGet('MOCK_DOMAIN_VALUES').reply(200, DOMAIN_INTERVAL_VALUES_RESPONSE);
+        testEpic(setIsIntervalData, 1, selectLayer("playback:selected_layer"), ([action]) => {
```
MV88, line 295: @alexander-fedorenko is this working even without setTimeout wrapper? aren't there some async task to perform inside the observables

MV88 (changes requested):
@alexander-fedorenko 

there is something wrong here

When you open the timeline and the options the snap tool is disabled for ft1 layer, but it should be enabled. then if you click on it in the timeline layer list gets enabled

https://user-images.githubusercontent.com/11991428/170472177-9e6036dd-ed1f-4b13-9494-759ff10f7682.mp4


I used the map present in the issue description
this is the local link http://localhost:8081/?debug=true#/viewer/openlayers/39497

tdipisa:
@MV88 thank you for the review but let's clarify the use case first of all with @ale-cristofori who implemented this since @alexander-fedorenko has been involved only for unit tests. It is probably a small bug (I hope), we will see later together with @ale-cristofori if involve @alexander-fedorenko

web/client/epics/__tests__/playback-test.js
alexander-fedorenko: @MV88 maybe it works because of the mock adapter used for axios, I can't say for sure. But yes, it works for this epic and for all the other epics that utilize same observable.

web/client/epics/playback.js
```diff
@@ -346,23 +348,33 @@ export const playbackCacheNextPreviousTimes = (action$, { getState = () => { } }
...
         });
+
+/**
+ * Get domains with a slight buffer to detect whether the layer consists of
+ * instants/point or intervals/bars time value. The results is used to
+ * disable/enable the radio buttons to snap to start/end of time interval
+ */
+export const setIsIntervalData = (action$, { getState = () => { } } = {}) =>
+    action$.ofType(SELECT_LAYER, SET_CURRENT_TIME)
+        .filter(({type, layerId}) => (type === SELECT_LAYER && layerId))
```
MV88, line 367:
```suggestion
      .filter(({type, layerId}) => (type === SET_CURRENT_TIME || (type === SELECT_LAYER && layerId)))
```

MV88 (changes requested):
@ale-cristofori 
- [ ] When you toggle off the layer in TOC of the selecter layer the check on Snap to guide layer needs to redone

web/client/epics/playback.js
```diff
@@ -346,23 +349,47 @@ export const playbackCacheNextPreviousTimes = (action$, { getState = () => { } }
...
+ * In case a layer in the timeline is unselected from the TOC the timeline
+ * settings for snapping for layers are toggled off, this is to avoid persistence
+ * of selection of a non-visible layer on the timeline state, causing inconsistencies
+ * in case of mixed (point/interval) time based layers
+ */
+export const switchOffSnapToLayer = (action$, { getState = () => { } } = {}) =>
+    action$.ofType(CHANGE_LAYER_PROPERTIES)
+        .filter(({newProperties, layer}) => {
+            const selectedLayer = selectedLayerSelector(getState());
+            return (newProperties?.visibility !== undefined && !newProperties.visibility && selectedLayer === layer);
```
MV88, line 389: I think this condition should be enough
```suggestion
            return ( !newProperties.visibility && selectedLayer === layer);
```

please apply it

MV88 (changes requested):
just one thing

MV88 (approved):
LGTM

MV88:

tdipisa:
@ElenaGallo functional tests for this issue have the priority. Thank you.

ElenaGallo:
Hi @tdipisa @ale-cristofori @alexander-fedorenko , these issues seem to have been introduced with this latest PR:

1_ When the user clicks on play, the affected layer is deselected

https://user-images.githubusercontent.com/56537133/170954217-f7474395-3160-4e2b-a069-918721594b16.mov

2_ There is the possibility to click the play button when no layers are selected

https://user-images.githubusercontent.com/56537133/170955903-b441e640-8354-4a62-a416-711320ef01da.mov

3_When playing the time slide with the time interval snap to the end point, the animation does not work as expected (it seems that the animation starts from the start point of the interval and sometimes the animation goes backwards)

https://user-images.githubusercontent.com/56537133/170954847-bc0a2e04-9fc6-4a3d-a0a4-eebbfcd80d54.mov

MV88:
> @MV88, issue n1, I think the lastest change to the code introduced issue n 1. checking for falsy produces different result than checking for undefined see this example https://stackblitz.com/edit/js-lxg6cz?file=index.js, could we revert the change maybe ?

yes I agree, we can also add a check to see if the timeline is visible

ale-cristofori:
@MV88, so if I understood correctly, the change 

```javascript
    import { isVisible } from '../selectors/timeline';
    //
    return (newProperties?.visibility !== undefined &&
        selectedLayer === layer &&
        // check if timeline component is visible
        isVisible(getState()));
```

is this what you mean?

tdipisa:
@ElenaGallo 
> @ElenaGallo issue n2, if I understood it correctly, this is the expected behaviour, animations (play button enabled) will work even if no layers are selected, the bar will snap to the frame duration set in the Timeline Settings, the fact that you do not see the cog button enabled is because you need to click on stop to stop the animation before setting up the animation step (which is disabled when snapping is enabled)

Look also at QA. It is the normal behavior when the snapping is not active.

MV88:
> @MV88, so if I understood correctly, the change
> 
> ```js
>     import { isVisible } from '../selectors/timeline';
>     //
>     return (newProperties?.visibility !== undefined &&
>         selectedLayer === layer &&
>         // check if timeline component is visible
>         isVisible(getState()));
> ```
> 
> is this what you mean?

yes

ElenaGallo:
> @ElenaGallo
> 
> > @ElenaGallo issue n2, if I understood it correctly, this is the expected behaviour, animations (play button enabled) will work even if no layers are selected, the bar will snap to the frame duration set in the Timeline Settings, the fact that you do not see the cog button enabled is because you need to click on stop to stop the animation before setting up the animation step (which is disabled when snapping is enabled)
> 
> Look also at QA. It is the normal behavior when the snapping is not active.

Yes, the point 2 is not a bug.

Issue #7778, opened 2022-01-26
Charts out of synch when saving the map


tdipisa:
## Description

Saving the charts (only when saving) they are no longer in synch with the current viewport

## How to reproduce

- Bug reproduced using [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/10358)
- See current result below for the steps to reproduce

*Expected Result*

If the widget is connected to the map, it should be always in sync with the current map viewport at every pan/zoom.

*Current Result*

![widget_chart_sync_issue.gif](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/e005c02a-b07e-4a28-ad53-c16275a120bb) 

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

tdipisa:
@alexander-fedorenko it seems your commits for this issue are breaking somehow the Sharing tool with the camera orientation properties. See [here](https://github.com/geosolutions-it/MapStore2/pull/8064#issuecomment-1139439523)

PR #8255, opened 2022-05-30
#7778 Fix for regression bug that was preventing cesium to properly use orientation on map load

alexander-fedorenko:
## Description

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
As it was stated https://github.com/geosolutions-it/MapStore2/issues/7778#issuecomment-1141177328 here original implementation is breaking one of the aspects of the sharing tool. Orientation on cesium map load is not triggered correctly, zooming/panning to the coordinates passed in url is not happening.
This PR changes the way how `bbox` is preserved upon map saving and saving as.

#7778
Orientation action is not triggered correctly for Cesium viewer

**What is the new behavior?**
No bbox calculation. Bbox is taken from the existing state and map view update is not triggered.
Map orientation works as expected.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/config.js
+++ b/web/client/epics/config.js
@@ -21,15 +21,13 @@ import {
     loadMapConfig,
     loadMapInfo
 } from '../actions/config';
-import {changeMapView, zoomToExtent} from '../actions/map';
+import {zoomToExtent} from '../actions/map';
 import Persistence from '../api/persistence';
 import { isLoggedIn, userSelector } from '../selectors/security';
 import { projectionDefsSelector } from '../selectors/map';
 import {loadUserSession, USER_SESSION_LOADED, userSessionStartSaving, saveMapConfig} from '../actions/usersession';
 import {userSessionEnabledSelector, buildSessionName} from "../selectors/usersession";
 import {getRequestParameterValue} from "../utils/QueryParamsUtils";
-import {getBbox} from "../utils/MapUtils";
-import {mapTypeSelector} from "../selectors/maptype";
 
 
 const prepareMapConfiguration = (data, override, state) => {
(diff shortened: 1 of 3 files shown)


web/client/epics/config.js
```diff
@@ -160,12 +161,16 @@ export const zoomToMaxExtentOnConfigureMap = action$ =>
...
-            const bbox = getBbox(center, zoom);
-            return changeMapView(center, zoom, bbox, size, mapStateSource, projection, viewerOptions, resolution);
+        .switchMap(({config}) => {
+            const mapType = mapTypeSelector(store.getState());
+            if (mapType !== 'cesium') {
+                const { center, zoom, size, mapStateSource, projection, viewerOptions, resolution } = (config?.map ?? {});
+                const bbox = getBbox(center, zoom);
+                return Observable.of(changeMapView(center, zoom, bbox, size, mapStateSource, projection, viewerOptions, resolution));
+            }
+            return Observable.empty();
```
allyoucanmap, line 173: I would avoid to add a new epic to solve the issue if was introduced of because it's adding a side effect affecting all the MAP_CONFIG_LOADED action, in addition is manipulating the state of the map by using changeMapView an action that usually is triggered by the map itself while panning or zooming.
This epic was introduced to solve this issue https://github.com/geosolutions-it/MapStore2/issues/7778 related to a missing update of chart if they are connected to the map viewport (bbox in the state). While reviewing the PR I noticed that once a user save the map the bbox in the state is undefined and this is causing the issue.
If we look the save workflow we noticed that we are replacing the current map content with the new one returned by the save response and this content is not including the bbox information

https://github.com/geosolutions-it/MapStore2/blob/ac760dfd4c4fc5a8702dcf218a9b1f87fd6bfa6d/web/client/epics/maps.js#L340

I would suggests following changes:

- [ ] remove this epic
- [ ] merge the saved map with the map in the state so we include also the current missing information

example

```js
function mergeMapState(state, data) {
    return {
        ...data,
        map: {
            ...mapSelector(state), // this contains the current bbox, missing in the response
            ...data?.map
        }
    };
}
```

then

```js
configureMap(mergeMapState(store.getState(), resource.data), rid)
```
I'm not sure if a similar merge map config exists in the code base, if so use the existing one

Note: I tested locally only with Save so please test that everything works also for SaveAs

allyoucanmap (changes requested):

alexander-fedorenko:
@allyoucanmap Thanks for the debugging session!

So during the call we figured out another approach that will keep bbox in-place upon map saving and saving as.
`reducers/config.js`, `MAP_CONFIG_LOADED` will preserve bbox from the state if it exists.

allyoucanmap:
@ElenaGallo please test in dev and let us know if we can backport to the stable branch 2022.01.xx

ElenaGallo:
The regression has been fixed, @alexander-fedorenko please backport to 2022.01.xx

Issue #8211, opened 2022-05-19
Map Switcher for Map widget


dsuren1:
## Description

It is requested to provide a new feature to Map widgets in MapStore dashboards to allow switching from multiple maps in a single Map widgets.

## Acceptance criteria

- [ ] A kind of dropdown on top of the Map widgets allow to switch between different pre-configured maps
- [ ] It will be possible to activate this function and to select the list of maps during the Map widget creation/editing
- [ ] The connection capabilities with other widgets at map widget level will remain the same so that existing functionalities should continue working also when the widget switch over another map selected by the user

**Requirement**:
The Map Switcher will be implemented as part of the Map widget wizard, where there will be an option (maybe in the toolbar) to define a list maps (maps and maps coming from a context) through a map catalog with multiple selection.
If one or more maps are selected, the map switcher will appear automatically.

For what concern the Legend part, to be in line with the current MapStore functional logic widget oriented, the proposal is to enhance the current legend widget by improving its functionalities. The user will be able to control from the legend widget the layer visibility and opacity and it will be also possible to consult the legend as depicted below.

As soon as the user switch to another map using the Map Switcher above, the legend is updated accordingly.

![2022_04_27_16_42_46_Window](https://user-images.githubusercontent.com/1280027/165548733-daa5e556-b0e1-42ac-8844-a03f279df2d6.png)

The purpose of the screenshot above is to give an high level indication of what is requested in terms of functionalities, while it is requested to take into account also the following point.

1. Better decide which is the more convenient place (also for responsive aspects) to put the switcher (widget header/in overlay to the map)
2. Optimize the space as much as possible inside the legend (eg. placing the slider just below the layer title)
3. Include the ability to collapse/expand the legend

**Reference**: https://github.com/geosolutions-it/MapStore2-C040/issues/575

dsuren1:
@tdipisa 
For widget creation of type `MapList`. Is this design approach fine, your thoughts?
![MapList](https://user-images.githubusercontent.com/26929983/169262414-341c8ade-4c45-4243-b9c6-dd71dd4b909d.png)

tdipisa:
@dsuren1 the description of this issue need to be reviewed because it need to take into account what is effectively agreed [here](https://github.com/geosolutions-it/MapStore2-C040/issues/575#issuecomment-1111109756).

For what concern the comment above:
>For widget creation of type map with list. Kindly let me know if this design sits well with the requirement. Your thoughts?

[...]

PR #8276, opened 2022-06-01
#8211: Map switcher and legend widgets

dsuren1:
## Description
This PR adds new feature to dashboard map and legend widgets
1. Map widget now lets user to select multiple maps and configure them with the support of Map switcher
2. Legend widget now allows user to change layer settings (visibility, opacity), legend looks similar to layers in TOC

 - [x] Feature

## Issue

#8211 

**What is the new behavior?**
https://docs.google.com/document/d/1cRGdMZGDhFGSBVjxAkSG__Vyty9FAhIA2grVserH9Xc/edit

legend-ex

wid-add-layer

## Breaking change
 - [x] No

## Other useful information

Reference: https://github.com/geosolutions-it/MapStore2-C040/issues/575


web/client/components/widgets/builder/wizard/map/MapCatalog.jsx
```diff
@@ -0,0 +1,71 @@
...
+const SideGrid = compose(
+    loadingState(({ loading, items = [] }) => items.length === 0 && loading),
+    emptyState(
+        ({ loading, items = [] }) => items.length === 0 && !loading,
+        {
+            title: <Message msgId="catalog.noRecordsMatched" />,
+            style: { transform: "translateY(50%)" }
+        })
+
+)(require('../../../../misc/cardgrids/SideGrid').default);
```
MV88, line 29: please use import statement here

web/client/components/widgets/builder/wizard/map/MapCatalog.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+export default ({
```
MV88, line 31: please let's give a name default export as a general rule

MV88 (changes requested):
@dsuren1 on the code point of view I did not spot big things so my focus was more on the functionality point of view

@tdipisa 
here is a list of things I noticed. probably some of them can be excluded or moved to a different issue

-  [ ] you could potentially select  all the maps you have in the list,
can this be a configurable number, or do we not care?

- [x] sometimes an horizontal scrollbar appears
image

- [x] name of widget is covered if no space is available
image

- [ ] it's not clear that you have to press CTRL to select multiple cards/maps
- [x] I have noticed some connection problems in particular when a  table widget is connect to a map and then if you change the map and then select a map that does not have the layer used by the table among layers

https://user-images.githubusercontent.com/11991428/171450373-dfdd1ce4-6f4d-484c-bc40-7193820d955f.mp4

- [x] do we have to show something if legend is not available?
image
in another case we have a message
image

- [x] is it normal to have legends continuously re-rendered?
https://user-images.githubusercontent.com/11991428/171452793-d0876411-d6c0-46a1-a5c2-e98a1d623458.mp4

dsuren1:
@MV88 
Thanks.

> * [ ]  name of widget is covered if no space is available

_I'm thinking of showing an icon when widget width is small. As we expand the widget when the width is enough to fit map switcher we display the same. Your thoughts?_
image

> * [ ]  it's not clear that you have to press CTRL to select multiple cards/maps

_I thought since layers in TOC follows the same, it doesn't need any explicit helptext. Shall I add something like this?_
image

> * [ ]  I have noticed some connection problems in particular when a  table widget is connect to a map and then if you change the map and then select a map that does not have the layer used by the table among layers
> 
>  sync.connection.problem.mp4 

_I think this scenario is produced this way_
_1. Create map widget_
_2. Create table widget_
_3. Connect table with map_
_4. Delete map widget_ 
_5. Now add new map widget_
_6. Connection doesn't work. Because old connection is not cleared when dependant widget is deleted. Seems to be an existing issue. I will fix this_

> * [ ]  do we have to show something if legend is not available?

_This is due to the layer being `3dtiles`. I will filter the layers accordingly, such that it doesn't display any `3dtiles` layer in legend widget/wizard_

> * [ ]  is it normal to have legends continuously re-rendered?
>   https://user-images.githubusercontent.com/11991428/171452793-d0876411-d6c0-46a1-a5c2-e98a1d623458.mp4

_Possibly coz of [this]( https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/TOC/fragments/legend/Legend.jsx#L57) when `0` is used as `urlIdx`, and url is array it tends to generate a random number causing it to rerender. I will fix this_

dsuren1:
> @dsuren1 thanks for commenting it out
> 
> > > * [ ]    name of widget is covered if no space is available
> > 
> > 
> > _I'm thinking of showing an icon when widget width is small. As we expand the widget when the width is enough to fit map switcher we display the same. Your thoughts?_ image
> 
> it could be a solution, unless you have already implemented we can consider it as a future enhancement

_I have implemented it along other updated in this PR_

> 
> > > * [ ]    it's not clear that you have to press CTRL to select multiple cards/maps
> > 
> > 
> > _I thought since layers in TOC follows the same, it doesn't need any explicit helptext. Shall I add something like this?_ image
> 
> You are right (I did not knew you could to that also in the TOC). Anyway here it is more obvious that a user can select multiple items in the list, so I would specify it, we should also consider that CTRL is not present in MAC OS, please amend it considering this mac os aspect. maybe it could be similar to the previous proposal to use a icon showing a tooltip to explain selection @allyoucanmap what do you think here?
> 
> @dsuren1 So apart from first consideration, have you covered all the rest, so i can re-check right?

_Sure. I can add a tooltip instead of sub text including mac os's command key (⌘)_

MV88:
> Sure. I can add a tooltip instead of sub text including mac os's command key (⌘)

i would change it as the following:

"
for multi-selection 
on windows use CTRL + click
on mac use control key + click
"

MV88 (changes requested):
- [x] how big should be a map widget to see the map switcher?

image

I think you can revert this change since we have little time

- [ ] there are still some cases where dependencies are not correctly updated
see this

https://user-images.githubusercontent.com/11991428/172162929-aeca3db0-4a04-4b2a-8073-624da008ae2d.mp4

dsuren1:
> * [ ]  how big should be a map widget to see the map switcher?
> 
> image
> 
> I think you can revert this change since we have little time

_The width of the widget should be >800px so it can fit the title and switcher_

> 
> * [ ]  there are still some cases where dependencies are not correctly updated
>   see this
> 
>  sync.problem.mp4

_Let me check_

MV88:
> The width of the widget should be >800px so it can fit the title and switcher

then I would make it smaller, like > 400 which includes when widgets is large at least 2 columns, or you revert it to what it was before

MV88 (approved):
LGTM

DavidQuartz:
@tdipisa @dsuren1 the multiple map selection feature 'control + click' does not work on Mac. It triggers a 'right click' effect instead.

Issue #8106, opened 2022-04-14
Include Cesium viewer settings 


allyoucanmap:
## Description

Add the possibility to apply setting of cesium viewer per maps:

- showSkyAtmosphere
- showGroundAtmosphere
- enableFog
- depthTestAgainstTerrain
- light options

## Acceptance criteria

- [ ] 
- [ ]

Issue #8058, opened 2022-04-06
3dtiles layer wrongly overlayed on top of custom terrain


landryb:
## Description
following up to #7924, testing a 3d tiles layer on top of a custom terrain provider the overlaying looks wrong:

![image](https://user-images.githubusercontent.com/3421760/161950803-6d3a8ec9-2254-445c-a122-12a4253d00ac.png)

## How to reproduce

- configure a custom terrain using this in localConfig.json:
```
          "terrainProvider": {
              "type": "cesium",
              "url": "https://3d.craig.fr/terrain",
              "requestVertexNormals": true
          }
```
- load 3dtiles layer from https://3d.craig.fr/datasets/Aurillac/3dtiles/tileset.json

*Expected Result*

it should display fine like below:
![image](https://user-images.githubusercontent.com/3421760/161951488-3c0277b7-605c-4f5e-b4f0-620dc886442d.png)

this screenshot comes from https://3d.craig.fr/viewer/ which is a simple cesium viewer displaying our 3dtiles on top of our own terrain server, select 'Aurillac' layer in the 3rd menu on the left.

*Current Result*

the terrain/osm background layer comes on top of the 3dtiles in some areas.

if using the default `"ellipsoid"` terrain (without another layer configured as BIL terrain provider), then the 3dtiles layer *floats* above the ground.

![image](https://user-images.githubusercontent.com/3421760/161952823-248debb9-80d6-4938-a60a-d3454f13b023.png)

## Other useful information

something wrong in the z-index for the 3dtiles layer ?

landryb:
changing the layer ordering in the layertree to put a WMS/WMTS layer on top/below the 3d tiles layer (to force updating the z-index/cesium layer order?) doesn't change anything wrt this strange display behaviour. (in the screenshot below, 'test' is the 3dtiles layer)

![image](https://user-images.githubusercontent.com/3421760/161957113-033e2440-2f72-40f8-bf48-44e442e6c94e.png)

tdipisa:
@landryb Does the terrain you are using have the same elevation profile as the 3D tile you are loading?
If not, in layer setting/display tab you have also the ability to tune the height offset
![image](https://user-images.githubusercontent.com/1280027/162007440-7a8fe6d4-f51d-4a8f-aefd-f151e9fd6a8f.png)

landryb:
[...]

PR #8280, opened 2022-06-03
8106 cesium viewer settings

weberjavi:
## Description

This PR creates `MapSettings` component that is used in the `Map` plugin inside the `Settings` `container`. With this approach whenever a user uses the `Map` plugin the settings window will display the available settings for each map type.
In this first iteration only `cesium` map type settings are available.

 - [x] Feature

## Issue

No option to update cesium scene options whatsoever. Some of them might be critical for the proper display of 3D layers on top of the terrain (see #8058 as an example).

#8106 

**What is the new behavior?**

Some cesium scene configuration options have been added to the settings panel (displayed only when the selected map type is cesium). 
The added options (`showSkyAtmosphere`, `showGroundAtmosphere`, `enableFog`, `depthTestAgainstTerrain`) are all of `boolean` type. 

## Breaking change
 - [x] No

## Other useful information
In order to provide some configuration to the scene lighting a different type of form inputs need to be designed/implemented. [Scene lights](https://cesium.com/learn/cesiumjs/ref-doc/Light.html) could update its color and intensity but some sort of color picker and numeric input should be used.


web/client/plugins/map/mapsettings/MapSettings.jsx
```diff
@@ -0,0 +1,87 @@
...
+    isCesium,
+    updateConfigAction
+}) => {
+    const [mapOptions, setMapOptions] = useState({});
+
+    useEffect(() => {
+        if (map) {
+            setMapOptions({
+                ...ConfigUtils.getConfigProp("defaultMapOptions")[mapType],
+                ...map.mapOptions
```
allyoucanmap, line 43: Here is missing the Map.cfg.mapOptions default configuration If compared with options merge inside the Map plugin. We could apply this change inside the Settings plugin

```diff
        return Object.keys(settingsFirst)
            .filter(this.isEnabled)
            .map((setting) => settingsFirst[setting])
-           // TODO: here every item (item.tool) we emit should have a "key" property
-           .concat(this.props.items.map((item) => item.tool))
+           .concat(this.props.items.map((item) =>
+               castArray(item.tool)
+                   .map((tool, idx) => cloneElement(tool, { ...item.cfg, key: `${item.name}-${idx}` })))
+           )
            .concat(
                Object.keys(settingsLast)
                    .filter(this.isEnabled)
                    .map((setting) => settingsLast[setting])
            );
```

- cloneElement from react
- castArray from lodash

with item.cfg spreat to the tool props we will get mapOptions from cfg inside this component so we can merge it with the other map options. Note the mapOptions from cfg is grouped with mapType

web/client/actions/map.js
```diff
@@ -31,6 +31,7 @@ export const MOUSE_OUT = 'MOUSE_OUT';
 export const MAP_PLUGIN_LOAD = 'MAP:MAP_PLUGIN_LOAD';
 export const ORIENTATION = 'MAP:ORIENTATION';
 export const UPDATE_MAP_VIEW = 'MAP:UPDATE_MAP_VIEW';
+export const UPDATE_SCENE_CONFIG = 'MAP:UPDATE_SCENE_CONFIG';
```
allyoucanmap, line 34: Could we call this `UPDATE_MAP_OPTIONS` instead? Looking to the reducer it seems not strictly related to the scene so I would suggest to keep this action related to all the map options.

web/client/actions/map.js
```diff
@@ -254,6 +255,11 @@ export const updateMapView = (data) => ({
     data
 });
 
+export const updateSceneConfig = (configUpdate) => ({
```
allyoucanmap, line 258: following the previous comment also this should be changed to updateMapOptions

web/client/components/map/cesium/Map.jsx
```diff
@@ -190,6 +190,19 @@ class CesiumMap extends React.Component {
...
+        }
+        if (prevProps && (!isEqual(this.props.mapOptions.showGroundAtmosphere, prevProps?.mapOptions?.showGroundAtmosphere))) {
+            this.map.scene.globe.showGroundAtmosphere = this.props.mapOptions.showGroundAtmosphere;
+        }
+        if (prevProps && (!isEqual(this.props.mapOptions.enableFog, prevProps?.mapOptions?.enableFog))) {
+            this.map.scene.fog.enabled = this.props.mapOptions.enableFog;
+        }
+        if (prevProps && (!isEqual(this.props.mapOptions.depthTestAgainstTerrain, prevProps?.mapOptions?.depthTestAgainstTerrain))) {
+            this.map.scene.globe.depthTestAgainstTerrain = this.props.mapOptions.depthTestAgainstTerrain;
+        }
```
allyoucanmap, line 205: Please update this branch with the latest master and verify if we need to add map.scene.requestRender(); after this scene updates to visualize the changes in the viewer

related to https://github.com/geosolutions-it/MapStore2/commit/3df6f48a01f34b4dbd9ef444a0a30b6ae819187f

web/client/components/map/cesium/Map.jsx
```diff
@@ -190,6 +190,19 @@ class CesiumMap extends React.Component {
             };
             this.setView(position);
         }
+
+        if (prevProps && (!isEqual(this.props.mapOptions.showSkyAtmosphere, prevProps?.mapOptions?.showSkyAtmosphere))) {
```
allyoucanmap, line 194: I'm not sure we need is equal for this options comparison, shouldn't be all boolean? same for these others new props

allyoucanmap (changes requested):

web/client/components/map/cesium/Map.jsx
weberjavi: It is not needed, it seems that the scene is updated (and rendered) right after property reassignment.

allyoucanmap (approved):

allyoucanmap:

PR #8283, opened 2022-06-03
7966 cesium embed map params fix for point 1

ale-cristofori:
## Description
Referring to this PR https://github.com/geosolutions-it/MapStore2/pull/8064
for this issue
https://github.com/geosolutions-it/MapStore2/issues/7966

two bugs were found within testing. This PR addresses the first in the list of of bugs described in this comment https://github.com/geosolutions-it/MapStore2/pull/8064#issuecomment-1139439523

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features) / as part of the previous PR

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

#7966 

**What is the current behavior?**
Please refer to the first bug highlighted in this comment
https://github.com/geosolutions-it/MapStore2/pull/8064#issuecomment-1139439523

**What is the new behavior?**
The embedded map will retain both the 3D options for Cesium and no errors will be shown on the console when the embedded map in cesium  loads. The loading spinner will disappear when the map finishes loading.
recording_1

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

Issue #7966, opened 2022-03-15
3D Zoom for embedded map


ElenaGallo:
Now the Cesium plugin is not enable for the embedded map but, when the user enable the 3D viewer parameters, the embedded code report the heading, pitch and roll parameters, see in the image below:

![parametri](https://user-images.githubusercontent.com/56537133/158405120-d90c168b-1517-49e2-b403-76daa6e59cee.png)

The **solutions** can be:
**1_** Add a Cesium plugin for the embedded map
**2_** Remove the heading, pitch and rol parameters on the embedded code

**How to Reproduce**
- Open a map
- Click on 3D 
- Click on Options > Share
- Click on Embed tab
- Enable the Advanced options
- Enable the Add center and zoom

ale-cristofori:
@tdipisa From what I can see in the code about the options @ElenaGallo suggested

1. This is the implementation of an enhancement, which has partially be implemented, there is the UI part to do it but nothing happens when the map link to embed it is made.

2. Removing the the cesium parameters `heading` `pitch` and `rol` is ok but is reverting the (small piece) of work that has been done.

On a side note, I see that exporting a map in Cesium has no effects, if the map is open again it defaults to a 2d viewer (either OL or Leaflet) I am not sure if it is related to this work but I imagine that we would want to store some sort of additional parameter in the embed url to tell MapStore to visualise the map in 3d mode. 

@tdipisa Could these assumptions be discussed before starting the development please?

tdipisa:
> @tdipisa From what I can see in the code about the options @ElenaGallo suggested
> 
> 1. This is the implementation of an enhancement, which has partially be implemented, there is the UI part to do it but nothing happens when the map link to embed it is made.
> 2. Removing the the cesium parameters `heading` `pitch` and `rol` is ok but this is  basically reverting the (small piece) of work that has been done on the UI.
> 
> On a side note, I see that exporting a map in Cesium has no effects, if the map is open again it defaults to a 2d viewer (either OL or Leaflet) I am not sure if it is related to this work but I imagine that we would want to store some sort of additional parameter in the embed url to tell MapStore to visualise the map in 3d mode.

The fact is that by default the 3D plugin is not present in the embedded map configuration. For this reason the Share URL for the embed part is normally pointing to the 2D map. This is not so correct as you can imaging: if I'm sharing a map in 3D mode the share URL should be in 3D mode. 
 
> @tdipisa Could these assumptions be discussed before starting the development please?

[...]

PR #8291, opened 2022-06-07
backport 2022.01.xx - #7966 cesium embed map params fix for point 1 (#8283)

ale-cristofori:
backport 2022.01.xx - #7966 cesium embed map params fix for point 1 (#8283)

### DIFF
--- a/web/client/epics/queryparams.js
+++ b/web/client/epics/queryparams.js
@@ -11,7 +11,7 @@ import { LOCATION_CHANGE } from 'connected-react-router';
 import {get, head, isNaN, includes, toNumber, isEmpty, isObject, isUndefined, inRange, every, has, partial} from 'lodash';
 import url from 'url';
 
-import {zoomToExtent, ZOOM_TO_EXTENT, CLICK_ON_MAP, changeMapView, CHANGE_MAP_VIEW, orientateMap} from '../actions/map';
+import {zoomToExtent, ZOOM_TO_EXTENT, CLICK_ON_MAP, changeMapView, CHANGE_MAP_VIEW, orientateMap, INIT_MAP} from '../actions/map';
 import { ADD_LAYERS_FROM_CATALOGS } from '../actions/catalog';
 import { SEARCH_LAYER_WITH_FILTER, addMarker, resetSearch, hideMarker } from '../actions/search';
 import { TOGGLE_CONTROL, setControlProperty } from '../actions/controls';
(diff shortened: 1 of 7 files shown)


tdipisa:
@ale-cristofori can you please check and fix failing tests?

web/client/epics/queryparams.js
```diff
@@ -27,15 +27,16 @@ import {mapSelector} from '../selectors/map';
...
+import {getCesiumViewerOptions, getParametersValues, getQueryActions} from "../utils/QueryParamsUtils";
 import {mapProjectionSelector} from "../utils/PrintUtils";
 import {updatePointWithGeometricFilter} from "../utils/IdentifyUtils";
 
 /*
 it maps params key to function.
 functions must return an array of actions or and empty array
 */
-const paramActions = {
+export const paramActions = {
```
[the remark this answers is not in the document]
allyoucanmap: We could mock what we need for the test directly inside the test file

web/client/utils/QueryParamsUtils.js
```diff
@@ -83,3 +83,56 @@ export const postRequestLoadValue = (name, storage = sessionStorage) => {
...
+ * where each is single key is the parameter and the value is the parameter value
+ * mapping is based on an object that maps each query string param to a redux action
+ * @param {object} paramActions objects that maps each parameter to its respective action to trigger
+ * @param {object} state the application state
+ * @returns {object} { param: value } KVP object
+ */
+export const getParametersValues = (paramActions, state) => (
+    Object.keys(paramActions)
+        .reduce((params, parameter) => {
+            const value = getRequestParameterValue(parameter, state);
```
allyoucanmap, line 99: I'm not sure yet but the failing tests could be related to use of sessionStorage in the getRequestParameterValue function that result different in the master env. The only way to verify that is to debug it locally and check if the sessionStorage has different content when it reach the failing unit tests comparing naster with 2022.01.xx branches. We should ensure that the unit test contains all the needed data and does not rely on configuration composed in other unit tests

allyoucanmap (changes requested):

tdipisa:
@ale-cristofori @allyoucanmap considering the above review, I'm going to exclude the issue from the release since it is scheduled for tomorrow. @ale-cristofori it would be good to fix what requested as soon as you can through a new PR for master and then updating this PR accordingly.

web/client/utils/QueryParamsUtils.js
ale-cristofori: It is ok for the investigation. If the root cause is the use of `session storage` I may need the help of the dev who wrote that method in the first place (@alexander-fedorenko)

ale-cristofori: from `master` session storage is:

```JSON 
{"queryParams":"{"featureinfo":{"lat":5,"lng":5,"filterNameList":[]},"zoom":3,"center":"41,0"}"}
```

ale-cristofori: @allyoucanmap @tdipisa as discussed, the bug that made tests fail was due to lines of code missing from ported branch, those have been reinstated, diffing with `master` (where these lines are there) and @allyoucanmap saw the result and he was happy with it. 

As for the `export const paramActions ...` this will probably require further discussion with other devs to consider the best place where to place `paramActions` object, the epics file where it is now might not be the most appropriate file, it is probably better to move the login the `QueryParamUtils` file.

allyoucanmap:
Closing because it was not included in latest minor release. Next release will be a major so it will include this fix already on master

Issue #8319, opened 2022-06-08
POST approach of passing query parameters is not working


alexander-fedorenko:
## Description
POST approach to pass query parameters to the map viewer is not working on master branch https://github.com/geosolutions-it/MapStore2/commit/2cecdde2abffdaa85d11e9de38ba7c4f94c0feab

## How to reproduce
- create new html file using example from documentation: https://mapstore.readthedocs.io/en/latest/developer-guide/map-query-parameters/#post-request
- open html file in browser and set following parameters in form values:
method: POST
MapStore Base URL: https://dev-mapstore.geosolutionsgroup.com/mapstore/
map:
```json
{"version":2,"map":{"projection":"EPSG:900913","units":"m","center":{"x":1250000,"y":5370000,"crs":"EPSG:900913"},"zoom":5,"maxExtent":[-20037508.34,-20037508.34,20037508.34,20037508.34],"layers":[{"type":"osm","title":"Open Street Map","name":"mapnik","source":"osm","group":"background","visibility":true}]}}
```
page: ../../#/viewer/openlayers/new
center: -1.652,48.111
zoom: 18

*Expected Result*
Map zoomed and centered using values from parameters

*Current Result*
Default map viewport is loaded

Issue does not affect QA instance.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #8320, opened 2022-06-09
#8319 Making POST method of getting query parameters work again on master

alexander-fedorenko:
## Description
Fix for regression by using single epic to process query parameters.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8319 
This regression was introduced by changes in #7966. It appears because `switchMapType` epic at ` epics/queryparams.js` is running before `readQueryParamsOnMapEpic` where query parameters are processed and actions are dispatched. It means that prior to the fix query parameters values in sessionStorage were removed before they read by the epic that process parameters.

**What is the new behavior?**
All epics that process query parameters are combined into single one. Parameters reading and processing happens only once per outer stream run. Parameters obtained from sessionStorage are available for inner streams,

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/queryparams.js
+++ b/web/client/epics/queryparams.js
@@ -163,15 +163,12 @@ export const readQueryParamsOnMapEpic = (action$, store) => {
                 .do(() => {skipProcessing = false;})
         ))
         .switchMap(() => {
-            const state = store.getState();
-            const map = mapSelector(state);
-            const parameters = getParametersValues(paramActions, state);
-            const cesiumViewerOptions = getCesiumViewerOptions(parameters, map);
-
+            const parameters = getParametersValues(paramActions, store.getState());
             return Rx.Observable.merge(
                 action$.ofType(INIT_MAP)
                     .take(1)
                     .switchMap(() => {
+                        const cesiumViewerOptions = getCesiumViewerOptions(parameters, mapSelector(store.getState()));
                         if (cesiumViewerOptions) {
                             skipProcessing = true;
                             return Rx.Observable.of(changeMapType('cesium'));
--- a/web/client/epics/queryparams.js
+++ b/web/client/epics/queryparams.js
@@ -9,7 +9,6 @@
 import * as Rx from 'rxjs';
 import { LOCATION_CHANGE } from 'connected-react-router';
 import {get, head, isNaN, includes, toNumber, isEmpty, isObject, isUndefined, inRange, every, has, partial} from 'lodash';
-import url from 'url';
 
 import {zoomToExtent, ZOOM_TO_EXTENT, CLICK_ON_MAP, changeMapView, CHANGE_MAP_VIEW, orientateMap, INIT_MAP} from '../actions/map';
 import { ADD_LAYERS_FROM_CATALOGS } from '../actions/catalog';
@@ -31,6 +30,21 @@ import { changeMapType } from '../actions/maptype';
 import {getCesiumViewerOptions, getParametersValues, getQueryActions} from "../utils/QueryParamsUtils";
 import {mapProjectionSelector} from "../utils/PrintUtils";
 import {updatePointWithGeometricFilter} from "../utils/IdentifyUtils";
+
+/**
+ * Semaphore function to skip epic processing under specific conditions
+ * @param sem$
+ * @param start
+ * @param condition
+ * @return {external:Observable}
+ * */
+const semaphore = (sem$, start = true, condition = c=>c) =>
+    stream$ =>
+        stream$.withLatestFrom(
+            sem$.startWith(start)
+        )
+            .filter(([, s]) => condition(s))
+            .map(([e]) => e);
 
 /*
 it maps params key to function.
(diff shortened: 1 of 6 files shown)


web/client/epics/queryparams.js
```diff
@@ -167,7 +167,7 @@ export const switchMapType = (action$, store) =>
                 .switchMap(() => {
                     const state = store.getState();
                     const map = mapSelector(state);
-                    const parameters = getParametersValues(paramActions, state);
+                    const parameters = getParametersValues(paramActions, state, true);
```
allyoucanmap, line 170: After a discussion with @offtherailz we noticed that the url parameters are managed by 3 different epics:

- introduced to change map type on location change

https://github.com/geosolutions-it/MapStore2/blob/3456d2d3b9857b0ef319b6859bfcce7fcecaf6d4/web/client/epics/queryparams.js#L162-L177

- introduced to change map orientation in 3D view, this one duplicated the getParametersValues that is currently causing the regression

https://github.com/geosolutions-it/MapStore2/blob/3456d2d3b9857b0ef319b6859bfcce7fcecaf6d4/web/client/epics/queryparams.js#L232-L253

- this should be the only epic that manage the query params but right now it has not effect on 3D viewer because it does not trigger a LAYER_LOAD action

https://github.com/geosolutions-it/MapStore2/blob/3456d2d3b9857b0ef319b6859bfcce7fcecaf6d4/web/client/epics/queryparams.js#L139-L152

Change proposals:

- [ ] use only the `readQueryParamsOnMapEpic` epic with the following updates:
  - listen for LOCATION_CHANGE
  - get the parameters with getParametersValues and switch to 3D or 2D based on params pitch/roll/heading or event better be more explicit and use a map type params that can be provided only via embed
  - due the fact that 3d viewer does not trigger LAYER_LOAD we should add a new action if does not exist (eg MAP_READY) common to all the map types and  listen for a this one action of the map
  - dispatch all the returned action including new logic for the map orientation
- [ ] remove `checkMapOrientation` beacuse will be included in the `readQueryParamsOnMapEpic` epic
- [ ] remove `switchMapType` beacuse will be included in the `readQueryParamsOnMapEpic` epic

We should verify at which level a map ready should be initialized, if at componentDidMount or if the map library support a map.on('load', () => {})

These are proposal so please @offtherailz @alexander-fedorenko take a look and tell me what you think before proceed

allyoucanmap (changes requested):

web/client/epics/queryparams.js
alexander-fedorenko: `checkMapOrientation` is not causing regression because it's not reading values from the storage (sessionStorage in this particular case), it only parses search string of the url which is ok.
Regression was caused by having multiple epics reading values from storage and by the fact that first reading removes the value to prevent unwanted scenario when another map is getting its content/viewport modified or replaced, or when reloading of the page cause same actions triggered.
Merging `switchMapType` into `readQueryParamsOnMapEpic` will work, but map viewer switch will happen too late, after user already see map viewport & some/all data loaded.

allyoucanmap: > checkMapOrientation is not causing regression because it's not reading values from the storage (sessionStorage in this particular case), it only parses search string of the url which is ok.

@alexander-fedorenko @offtherailz 

Now I don't understand why sometimes we have particular cases and others not. Shouldn't all the query parameters managed that the same process? Then checkMapOrientation is listening for map change view that were triggered on every call, we temporary added .take(1) but imo we should merge this to the main readQueryParamsOnMapEpic epic

> Merging switchMapType into readQueryParamsOnMapEpic will work, but map viewer switch will happen too late, after user already see map viewport & some/all data loaded.

I was thinks something like this in my proposal, I did not tested and I'm not sure if it will work

```js
export const readQueryParamsOnMapEpic = (action$, store) =>
    action$.ofType(LOCATION_CHANGE)
        .switchMap(() => {
            const map = mapSelector(state);
            const parameters = getParametersValues(paramActions, state);
            const queryActions = getQueryActions(parameters, paramActions, state);
            const cesiumViewerOptions = getCesiumViewerOptions(parameters, map);
            return Observable.concat(
                ...(cesiumViewerOptions
                    ? [Observable.of(changeMapType('cesium'))]
                    : []),
                action$.ofType(MAP_READY)
                    .take(1)
                    .switchMap(() => {
                        return head(queryActions) 
                            ? Observable.of(...queryActions) 
                            : Observable.empty(); 
                    })
            );
        });
```

allyoucanmap: still `action$.ofType(CHANGE_MAP_VIEW).take(1)` seems a workaround, the action CHANGE_MAP_VIEW should not be used as listener imho

offtherailz: yes, it is a workaround because we don't have a specific action that tells us when we can perform that specific operation. I think we can live with it, if it didn't notice particular problems, and if we need to fix it quickly.

A general solution for it looks complicated to me. 


If you have a simpler solution, please propose. This is how I imagine the correct solution: 

### Idea for more well defined solution

- First we need to identify what we need specifically to trigger query parameters actions.

I think we need for sure the map json config to be loaded, in order to apply map parameter to the current map. 
From the code of 3D seems that we have to make sure that the map lib itself has been loaded too, but I'm not sure for that. 

Basically we need an **selector** that tell us "the map is ready, so we can trigger the actions needed from query parameters".
Let's imagine it is 
```javascript
const mapReady = ({ready}) => ({type: MAP_READY, ready}),
```
sets a part in the state ready to true/false. let's imagine the selector checks `isMapReadySelector`.

!!! note: 
    Anyway we have to properly do this and verify when and why trigger this MAP_READY (e.g. when maptype changes? on unmount have to be set to false or undefined)

 - Then, to read the query params, ideally we need a sequence like this:

- wait for any LOCATION_CHANGE 
- Check if some parameters are present
    - The map is ready --> continue
    - The map is not ready --> wait for MAP_READY and re-check the condition, until map is ready
 - perform the actions for the parameters found.


BTW:
If you want to make the scheleton for future hardening of this we can add an epic like this:

```javascript
action$.ofType(CHANGE_MAP_VIEW).take(1).map(mapReady())
```
That will be sooner removed in favor of an event triggered by the map plugin

alexander-fedorenko: @allyoucanmap Concatenation works fine, here is the snippet I used and it works well:
```js
const map = mapSelector(state);
const parameters = getParametersValues(paramActions, state);
const queryActions = getQueryActions(parameters, paramActions, state);
const cesiumViewerOptions = getCesiumViewerOptions(parameters, map);

return Rx.Observable.concat(
  action$.ofType(INIT_MAP)
      .take(1)
      .switchMap(() => {
          if (cesiumViewerOptions) {
              return Rx.Observable.of(changeMapType('cesium'));
          }
          return Rx.Observable.empty();
      }),
  action$.ofType(LAYER_LOAD)
      .take(1)
      .switchMap(() => {
          return head(queryActions)
              ? Rx.Observable.of(...queryActions)
              : Rx.Observable.empty();
      })
);
```

@offtherailz @allyoucanmap suggested approach is fine, but I have a feeling that proper implementation will need a separate ticket and some time for investigation:
- `LAYER_LOAD` event is used as a trigger for parameters processing on a purpose. As far as I remember this is related to "actions" parameter processing: there are some actions that are applied to the loaded layer data and they will fail if parameters processing starts earlier.
- Cesium viewer did not process query parameters at all (prior to these recent changes that added support of cesium viewer orientation based on query parameters), there is no `LAYER_LOAD` action there, I saw discussion with [example](https://stackoverflow.com/questions/67882437/how-to-listen-the-load-event-of-cesium-terrain/67883291#67883291) how to check that data is queued for loading, but it's not clear if it's possible to distinguish between different layers.
- Cesium implementation of zoom, bbox, center and other viewport-positioning parameters processing will be different.

I suggest to:
- Fix regression with the approach already committed
- Combine three epics into single one `readQueryParamsOnMapEpic`, it's ok to have `checkMapOrientation` merged too because CHANGE_MAP_VIEW always happen after LOCATION_CHANGE.
- Open separate ticket with proposal for proper implementation and extending query parameters processing support for cesium viewer.

alexander-fedorenko: @tdipisa what do you think?

tdipisa: @offtherailz @allyoucanmap at the end of this long discussion we have to take a decision to solve the problem we have in master now considering also that on mid July we will release 2022.02.00 and the problem must be fixed. If you need to discuss this with me feel free to schedule a call.

tdipisa: @offtherailz @allyoucanmap any news for my above comment? I saw a new commit of @alexander-fedorenko but I didn't receive any feedback before.

web/client/epics/queryparams.js
```diff
@@ -131,50 +145,64 @@ export const paramActions = {
...
-        );
+export const readQueryParamsOnMapEpic = (action$, store) => {
+    let skipProcessing = false;
+    return action$.ofType(LOCATION_CHANGE)
+        .let(semaphore(
+            action$.ofType(LOCATION_CHANGE)
+                .map(() => !skipProcessing)
+                .startWith(true)
+                .do(() => {skipProcessing = false;})
+        ))
```
offtherailz, line 163:
```suggestion
        // this stop / start listening for une LOCATION_CHANGE event if `skipProcessing` is true, useful when this action is triggered by switching map-type
        .let(semaphore(
            action$.ofType(LOCATION_CHANGE)
                .map(() => !skipProcessing)
                .startWith(true)
                .do(() => {skipProcessing = false;})
        ))
```

allyoucanmap:
@ElenaGallo please test in dev, thanks

Issue #8101, opened 2022-04-14
Support of Multiple Terrains in Cesium


tdipisa:
## Description

It is not currently possible to select different Terrain layers in a single map.  The current terrain support is based on a global configuration used for all 3D maps or context maps. It should be possible to define multiple Terrain layers to switch between them in a single Cesium map. 

## Acceptance criteria

- [x] The work should be based on the DDS/BIL support using raster data from GeoServer. 
- [x] Include a new layer type for Terrain layers
- [x] Investigate on the configuration design for a future UI Terrain Switcher 
- [x] Check positioning behaviours of entities or 3D tiles over a terrain.
- [x] Check configuration to apply different terrain per context

landryb:
> * The work should be based on the DDS/BIL support using raster data from GeoServer.

are you also planning support for `Cesium.CesiumTerrainProvider` ? it works just fine and is the 'native' terrain provider in cesium.. and as such, it's *much* faster than the DDS/BIL hack used by GeoserverTerrainProvider.

granted, one then needs to self-generate a terrain cache with ctb-tile (or use a 3rd-party provider?), its not as easy as dropping a DEM raster in geoserver, but it's much much better.

tdipisa:
@landryb 

>are you also planning support for Cesium.CesiumTerrainProvider ?

Yes, we are doing it too.

PR #8323, opened 2022-06-09
Create terrain layer type

weberjavi:
## Description

This PR creates a dedicated layer type that allows the customization of the [terrain provider](https://cesium.com/learn/cesiumjs/ref-doc/TerrainProvider.html) for the cesium type of map.
Decoupling the `terrain` layer from the `wms` layer allows for a better separation of concerns avoiding the complexity on further terrain providers support.
The new implementation adds support for [cesium terrain provider](https://cesium.com/learn/cesiumjs/ref-doc/CesiumTerrainProvider.html) on top of the already supported `DDL/BIL` format.

 - [x] Feature

## Issue

On the current implementation the `DDS/BIL` layers are handled inside the `wms` layer type and are used as `terrain` data only if they have the `useForElevation` attribute set to true. No other terrain provider is supported.
#8101 

**What is the new behavior?**
The new implementation sets the ground for further developments of terrain providers on its own layer. As first take away the cesium terrain provider is also supported allowing third party sources of terrain data.
The new implementation introduces an update on how the terrain is defined on the global configuration (although the old configuration would still be supported through the `wms` layer type).
The most straightforward way to make use of this implementation right away is to create different map configurations per context with different terrain layers inside the `additionalLayers` attribute on the `locallConfig.json` that can be edited  by clicking the Edit Plugin Configuration button on the map plugin:

imagen
Read the docs regarding the elevation 

## Breaking change
 - [ ] No

## Other useful information

Additional testing and investigation has been performed as part of this issue. 

- By treating the terrain as a layer it could be dinamically updated. This would allow having a set of `terrain` layers from which the user could select, behaving as the layers on the `background` group on the current implementation. Further design and development would need to be done.
- With this new implemenation a layer could be added to the `layers` array on the map configuration.
- The shape of the JSON obect in the map layer config would be:

`DDS/BIL provider`
```json
{ 
"type": "terrain",
"title": "terrain testing",
"terrainProvider": "bil",
"url": "https://gs-stable.geo-solutions.it/geoserver/wms",
"format": "application/bil16",
"name": "sf:sfdem",
"littleendian": false,
"visibility": true
}
```

`cesium terrain provider`
```json
{
"type": "terrain",
"title": "terrain testing",
"terrainProvider": "cesium",
"url": "https://api.maptiler.com/tiles/terrain-quantized-mesh-v2/?key={yourApiKey}",
"littleendian": false,
"visibility": true
}
```

`ellipsoid`
```json
{
"type": "terrain",
"title": "terrain ellipsoid",
"terrainProvider": "ellipsoid",
"littleendian": false,
"visibility": true
}
```

- It updates the `terrainProvider` for the whole scene, so only one terrainLayer could be displayed at once. **In order to support several DTMs at the same time some investigation would need to be performed**. Some initial ideas would be to investigate the merging of different height map imagery server side (and consume it as only one height info source client side) or treating the different DTM meshes as 3D obects client side.

- **With the current implementation both vector layers and 3d layers do not follow the terrain profile**. This is barely noticeable if the `terrainProvider` is the `ellipsoid` but  quite evident if a DTM based mesh is used. Also depending on the value of `depthTestAgainstTerrain` some features might render invisible (hidden under the terrain). To bypass this issue we would need to use the terrain height on the coordinates of the features to define their initial `z` position. 
  Some initial approach could be taken from [here](https://gis.stackexchange.com/questions/194334/line-layer-seems-to-be-floating-and-shifted-after-adding-terrain-effect-in-cesiu) and [here](https://gis.stackexchange.com/questions/189392/3d-osm-buildings-seems-to-be-floating-and-shifted-after-adding-terrain-effect-in)


web/client/components/map/cesium/plugins/TerrainLayer.js
```diff
@@ -0,0 +1,106 @@
...
+const createLayer = (options, map) => {
+    map.terrainProvider = undefined;
+    let terrainProvider;
+    switch (options.terrainProvider) {
+    case 'bil': {
+        terrainProvider = new BILTerrainProvider(wmsToCesiumOptionsBIL(options));
+        break;
+    }
+    case 'cesium': {
+        terrainProvider = new Cesium.CesiumTerrainProvider(options);
```
allyoucanmap, line 72: Please review this structure as discussed to persist the property in the saved/eported map (LayerUtils saveLayer function [reference](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/LayersUtils.js#L583-L584

```js
CesiumTerrainProvider({
   url: options.url,
   credits: //,
  ...options.options
})
```

web/client/components/map/cesium/plugins/TerrainLayer.js
```diff
@@ -0,0 +1,106 @@
...
+        }),
+        littleEndian: options.littleendian || false,
+        layerName: options.name
+    };
+}
+
+const createLayer = (options, map) => {
+    map.terrainProvider = undefined;
+    let terrainProvider;
+    switch (options.terrainProvider) {
```
allyoucanmap, line 66: We need to decide if to use `terrainProvider` or `provider`, I would use provider because is already persisted and terrain seems redundant

web/client/components/map/cesium/plugins/TerrainLayer.js
```diff
@@ -0,0 +1,106 @@
...
+            map.terrainProvider = new Cesium.EllipsoidTerrainProvider();
+        },
+        setVisible: () => {}
+    };
+};
+
+const updateLayer = (layer, newOptions, oldOptions, map) => {
+    if (newOptions.securityToken !== oldOptions.securityToken
+    || oldOptions.credits !== newOptions.credits
+    || oldOptions.provider !== newOptions.provider) {
```
allyoucanmap, line 97: this should be aligned with the key we decide for the provider based on previous comment

docs/developer-guide/maps-configuration.md
```diff
@@ -1133,6 +1137,64 @@ i.e.
...
+
+The `terrain` layer of `cesium` type allows using Cesium terrain format compliant services (like Cesium Ion resources or [MapTiler meshes](https://cloud.maptiler.com/tiles/terrain-quantized-mesh-v2/)).
+
+```json
+{
+    "type": "terrain",
+    "title": "terrain cesium",
+    "terrainProvider": "cesium",
+    "url": "https://api.maptiler.com/tiles/terrain-quantized-mesh-v2/?key={yourApiKey}",
+    "littleendian": false,
```
allyoucanmap, line 1174: Please review the properties based on the new structure and remove the one that are not needed (eg littleendian is not used by cesium terrain)

docs/developer-guide/maps-configuration.md
```diff
@@ -1133,6 +1137,64 @@ i.e.
...
+```json
+{
+    "name": "Map",
+    "cfg": {
+        "additionalLayers": [{
+            "type": "terrain",
+            "title": "terrain cesium",
+            "terrainProvider": "cesium",
+            "url": "https://api.maptiler.com/tiles/terrain-quantized-mesh-v2/?key={yourApiKey}",
+            "littleendian": false,
```
allyoucanmap, line 1190: same as previous comment

allyoucanmap (changes requested):

web/client/components/map/cesium/plugins/TerrainLayer.js
```diff
@@ -0,0 +1,106 @@
...
+        littleEndian: options.littleendian || false,
+        layerName: options.name
+    };
+}
+
+const createLayer = (options, map) => {
+    map.terrainProvider = undefined;
+    let terrainProvider;
+    switch (options.terrainProvider) {
+    case 'bil': {
```
allyoucanmap, line 67: bil should be changed to wms, because the  the provider is wms with format application/bil16

allyoucanmap (changes requested):

web/client/components/map/cesium/plugins/TerrainLayer.js
```diff
@@ -0,0 +1,117 @@
...
+function wmsOptionsMapping(config) {
+    let url = config.url;
+    const headers = getAuthenticationHeaders(url, config.securityToken);
+    return {
+        url: new Cesium.Resource({
+            url,
+            headers,
+            proxy: getProxy(config)
+        }),
+        layerName: config.name
```
allyoucanmap, line 58:
```suggestion
        littleEndian: config.littleendian || false,
        layerName: config.name
```

allyoucanmap (changes requested):

allyoucanmap:
@ElenaGallo please test this in dev. We should try to use the new terrain type instead of the old one used with wms to apply a terrain provider to a map. I'm going to prepare some data to test it

Issue #8082, opened 2022-04-12
MapStore doesn't normalize longitude when editing geometries


giohappy:
## Description
MapStore repeats the world map horizontally because it's a useful feature for the visualization of global data. However, it also permits the editing of geometries outside the valid WGS84 extent (-180<lon<180 ).
This raises an exception in Geoserver when the geometries are saved since they have invalid longitudes.

## How to reproduce
- open a map with an editable vector layer
- move either to the right or to the left of the central map, beyond the DateLine (Pacific) 
- edit a feature's geometry and save
- an error from Geoserver is returned telling that the geometry has coordinates outside the valid extent

*Expected Result*
Coordinates outside the valid extent should be normalized

*Current Result*
Coordinates are not normalized

MV88:
@barema4 we have added a layer that you can edit in our gs-stable geoserver version, so for just when you work on this you can remvoe the proxy configuration used and us directly the default bproxy settings (remove changes to webpack-config.js)

then you can use this layer **regioni_test** with this user 
username: barema 
pwd: barema

anyway this is the map you can search when you are logged
![image](https://user-images.githubusercontent.com/11991428/169547407-4b3b89eb-1bef-4df2-bf5a-d9b8e85b2743.png)

barema4:
Thanks for the update

tdipisa:
@barema4 this layer should work now in editing **_developers:regioni_test_**

barema4:
@tdipisa  @MV88  the layer is now working.

giohappy:
Any update on this @barema4 @MV88 @tdipisa

giohappy:
@tdipisa no specific client. This was discovered inside a project for a client but the fix is not a requirement for it.

tdipisa:
At the moment we don't have resources to employ on this to review the implementation and finalize it. I would like to prevent the risk that the current bugged implementation finish in the next major release. Therefore @MV88 it should be better to revert it from both master and 2022.02.xx.

MV88:
the commit to revert is only this one https://github.com/geosolutions-it/MapStore2/tree/cbb5a9d9eb1c621f67d309f90a288c40714ce4c9

the pr is this one https://github.com/geosolutions-it/MapStore2/pull/8512

PR #8325, opened 2022-06-10
#8082 normalize longitude

barema4:
## Description
* This PR normalizes the longitude incase it is outside the range of (-180, 180)

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
* #8082
* Longitude coordinates are not normalized
**What is the new behavior?**
* Coordinates outside the valid extent  are  normalized

## Breaking change
 - [x] No


web/client/reducers/draw.js
```diff
@@ -49,7 +50,10 @@ function draw(state = initialState, action) {
             currentStyle: action.currentStyle
         });
     case GEOMETRY_CHANGED:
-        return assign({}, state, {tempFeatures: action.features});
+        let newData = action.features;
+        const normalizedData = newData[0].geometry.coordinates.map(i => i.map(x => x.map((item) => [normalizeLng(item[0]), item[1]])));
```
MV88, line 54: when a geometry is changes you are not sure about the type of the geomtry you are changing. 
It can be a Point a LineString a Polygon or a MultiPolygon etc.

so you cannot take for granted that these nested mapping will work always and you have to add more checks before proceeding

Other than this units testing are missing and this is a perfect PR to add them. 
please add a unit test for each type of geometry
- Point
- LineString
- Polygon
- MultiPoint
- MultiLineString
- MultiPolygon

MV88 (changes requested):
this will work only in certain scenarios, please add unit tests and more checks for the different type of geometries,
maybe with the aid of a switch or auxiliary functions you can write

web/client/reducers/draw.js
barema4: Alright @MV88 , Let me add them

barema4: Done writing unit tests @MV88

MV88 (changes requested):
Please fix the values used in tests
you can take examples from [here](https://en.wikipedia.org/wiki/GeoJSON)

web/client/reducers/__tests__/draw-test.js
```diff
@@ -86,6 +85,153 @@ describe('Test the draw reducer', () => {
...
+                someProp: "someValue"
+            }
+        };
+        let testAction = {
+            type: GEOMETRY_CHANGED,
+            features: [feature1]
+        };
+        let state = draw({}, testAction);
+        expect(state.tempFeatures).toExist();
+        expect(state.tempFeatures[0]).toBe(feature1);
```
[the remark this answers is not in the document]
MV88: you have not changes this, and other similar checks

web/client/reducers/draw.js
```diff
@@ -49,7 +50,21 @@ function draw(state = initialState, action) {
...
+        let newData = action.features;
+        if (newData[0].geometry.type === 'Point') {
+            const normalizedPoint = [normalizeLng(newData[0].geometry.coordinates[0]), newData[0].geometry.coordinates[1]];
+            newData[0].geometry.coordinates = normalizedPoint;
+        } else if (newData[0].geometry.type === 'LineString') {
+            const normalizedLineString =  newData[0].geometry.coordinates.map((item) => [normalizeLng(item[0]), item[1]]);
+            newData[0].geometry.coordinates = normalizedLineString;
+        } else if (newData[0].geometry.type === 'MultiPoint') {
+            const normalizedMultiPoint =  newData[0].geometry.coordinates.map((item) => [normalizeLng(item[0]), item[1]]);
+            newData[0].geometry.coordinates = normalizedMultiPoint;
```
MV88, line 62: here you can put together these checks
```suggestion
        } else if (newData[0].geometry.type === 'LineString' || newData[0].geometry.type === 'MultiPoint) {
            const normalizedCoord =  newData[0].geometry.coordinates.map((item) => [normalizeLng(item[0]), item[1]]);
            newData[0].geometry.coordinates = normalizedCoord ;
```

web/client/reducers/draw.js
```diff
@@ -49,7 +50,21 @@ function draw(state = initialState, action) {
...
+        } else {
+            const normalizedData = newData[0].geometry.coordinates.map(i => i.map(x => x.map((item) => [normalizeLng(item[0]), item[1]])));
+            newData[0].geometry.coordinates = normalizedData;
+        }
```
MV88, line 66: Polygon and Multilinestring has similar structure so you can check them together

but MultiPolygon should have one more level

web/client/reducers/__tests__/draw-test.js
```diff
@@ -86,6 +85,155 @@ describe('Test the draw reducer', () => {
...
+        expect(state.tempFeatures.length).toEqual(1);
+    });
+    it('MultiPolygon GEOMETRY_CHANGED', () => {
+        let feature3 = {
+            type: "Feature",
+            geometry: {
+                type: "MultiPolygon",
+                coordinates: [
+                    [
+                        [ [102.0, 2.0], [103.0, 2.0], [103.0, 3.0], [102.0, 3.0], [102.0, 2.0] ]
```
MV88, line 143: please use values that are not normalized so we test everything works fine
for example
here

```suggestion
                        [ [-202.0, 2.0], [-303.0, 2.0], [-103.0, 3.0], [302.0, 3.0], [182.0, 2.0] ]
```

web/client/reducers/draw.js
```diff
@@ -49,7 +50,21 @@ function draw(state = initialState, action) {
             currentStyle: action.currentStyle
         });
     case GEOMETRY_CHANGED:
-        return assign({}, state, {tempFeatures: action.features});
+        let newData = action.features;
```
MV88, line 53: since you mutate the newData later maybe is better to clone it before
```suggestion
        let newData = assign({},action.features);
```

MV88 (changes requested):
@barema4 
there is still something not correct.

- [ ] first I would like you to move th geoemtry in a non normalized status so that the check is done correctly
- [ ] then I would like to not use toBe function for comparing objects , so please amend your tests

MV88 (changes requested):

web/client/reducers/__tests__/draw-test.js
```diff
@@ -84,8 +83,173 @@ describe('Test the draw reducer', () => {
...
+            }
+        };
+        let expectedPolygon = [ [ [ 170, 10 ], [ 168, 45 ], [ -164.00000000000003, 40 ], [ 161.99999999999997, 20 ], [ 160, 10 ] ], [ [ -160, 30 ], [ -150, 35 ], [ 139.99999999999997, 20 ], [ -130.00000000000003, 30 ] ] ];
+        let testAction = {
+            type: GEOMETRY_CHANGED,
+            features: [feature1]
+        };
+        let state = draw({}, testAction);
+        expect(state.tempFeatures).toExist();
+        expect(state.tempFeatures[0]).toEqual(feature1);
```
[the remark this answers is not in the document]
barema4: Alright @MV88 , I get it, Let me work on it.

web/client/utils/ImmutableUtils.js
```diff
@@ -82,6 +85,29 @@ export const arrayDelete = (path, condition, object) => {
     }
     return object;
 };
+
+export const normalizeGeomerty = (state, action) => {
```
MV88, line 89: ImmutabilityUtils is a set of function that handles Immutability
please remove reducer related stuff from this utils file.

what I intended with my previous meeting was to reuse the set function exported  here
which I can see you are not doing 

I'll try to be clearer:
instead of doing 
```js
let newData = assign({}, action.features);
```
we can do
```js 
    let newCoords = [];
    if (action.features[0].geometry.type === 'Point') {
        newCoords = [normalizeLng(newData[0].geometry.coordinates[0]), newData[0].geometry.coordinates[1]];
        return set("tempFeatures", set("features[0].geometry.coordinates", newCoords, action), state);
        
        or         
       return assign({}, state, {tempFeatures: set("features[0].geometry.coordinates", newCoords, action)});
```
please double check this and do the same changes to the various cases
let me know if you need clarifications

web/client/reducers/__tests__/draw-test.js
```diff
@@ -8,6 +8,7 @@
 import expect from 'expect';
 
 import draw, {defaultSnappingConfig} from '../draw';
+import { normalizeGeomerty } from '../../utils/ImmutableUtils';
```
MV88, line 11: please remove this
it was good as you did before

if you apply changes to the reducers you will notice 
that you will have to fix tests were you do the check
expect(state.tempFeatures[0]).toEqual(feature1);

MV88 (changes requested):

MV88 (approved):

Issue #7928, opened 2022-03-08
Mobile - Extra page on GFI panel


ElenaGallo:
## Description

Some time on GFI panel there is a page with this message: `no features were found` and a blank page.

![MOBILE TRUE](https://user-images.githubusercontent.com/56537133/157256232-8a5c4e82-67a7-405f-920c-f6f0261dff94.gif)

## How to reproduce

-  Open a map on mobile
- Click e feature 
- Scroll the page on GFI panel

*Current Result*

On GFI panel there is a page with this message: `no features were found` and a blank page.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

buhao75:
Hello!

Sorry for cross posting but could you have a short look at this post of me in the developers group: https://groups.google.com/g/mapstore-developers/c/cqNUL_-n5jg
Might this be related to this issue or this one: https://github.com/geosolutions-it/MapStore2/pull/7867

It took me quite a while to figure it out and if it is a bug it would be cool if it would be solved.

Thanks and all the best,
Michael

PR #8332, opened 2022-06-15
7928 Redundant panels inside identify results on mobile & swiping

alexander-fedorenko:
## Description
Identify panel allows to switch between results of map info requests using select element that contains layers that DO have data in a certain point user clicked/hovered.
In mobile mode user is able to swipe left or right to switch between available layers easily.

The problem is that in mobile view there were redundant empty panels appearing upon swipe from time to time as it described in the ticket connected to this PR.

The reason of such behavior was in the way how data was rendered inside the container that allows to swipe:
- All the responses (in other words - all the layers) got a panel in results list regardless of response validity.
- Index was set to the first valid response, meaning that swipeable container "moved" viewport to make it show valid result.

It was working quite well on desktop, but on mobile it still allowed to swipe back and forth to the invalid results because they, eventually, were rendered as empty panels inside of swipe container.

This PR suggest following change to prevent the issue:
1. Layer selector will list ONLY layers where response was valid.
2. Swipe container will render panel ONLY per valid response.
3. Indexing of the select options and panels will change accordingly, 0 for the first valid result and so on.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#7928 

**What is the new behavior?**
As per PR description. No redundant panels that user can swipe to in mobile view.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/identify/DefaultViewer.jsx
+++ b/web/client/components/data/identify/DefaultViewer.jsx
@@ -77,7 +77,7 @@ class DefaultViewer extends React.Component {
     getResponseProperties = () => {
         const validator = this.props.validator(this.props.format);
         const responses = this.props.responses.map(res => res === undefined ? {} : res); // Replace any undefined responses
-        const validResponses = this.props.renderValidOnly ? validator.getValidResponses(responses) : responses;
+        const validResponses = validator.getValidResponses(responses);
         const invalidResponses = validator.getNoValidResponses(this.props.responses);
         const emptyResponses = this.props.requests.length === invalidResponses.length;
         const currResponse = this.getCurrentResponse(validResponses[this.props.index]);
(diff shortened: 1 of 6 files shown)


dsuren1 (changes requested):
Hi @alexander-fedorenko 

This PR contradicts with Identify tool's existing feature of displaying the first response in the panel rather than first index (desktop)
**Ref**
https://github.com/geosolutions-it/MapStore2/issues/6724

GFI tool has some unique behavior in desktop and mobile view.

Kindly limit the fix to mobile view, as we don't want any regression on desktop view. Thanks!

alex-fko:
@dsuren1 Kindly review it at your convenience. Approach is updated to not affect the changes introduced by #6724. It will pick up first loaded response as default selection just like before.
I've updated description of PR to describe new (hopefully simplified) approach.

dsuren1 (changes requested):
Fix works fine in mobile but desktop shows some artifacting effect (shown in gif, the page slides/changes few times until settles on the target page)
gfi_artifacting

dsuren1:
Closing this PR. A fix will be given as part of new PR

Issue #7952, opened 2022-03-10
Smarter authentication request interceptor


offtherailz:
## Description
If you use relative paths like "/geoserver/" and "http://test/geoserver/" with the authkey, the current authenticationRules need two different rules:
- one for the relative URL: `"urlPattern": "^\\/geoserver\\/.*",`
- one with the absolute url `http(s)?\\:\\/\\/test(\\:443|\\:80)?\\/geoserver\\/.*` 

We have to investigate if we can improve the behavior to recognize somehow the localhost and add a single rule, or a placeholder for localhost requests.

## Acceptance criteria

- [ ] 
- [ ]

alexander-fedorenko:
@offtherailz Would it be sufficient to use combined regex that will be valid for both relative and absolute url?
```
/(http(s)?\:\/\/gs-stable\.geo-solutions\.it(\:443|\:80)?\/geoserver\/.*)|(^\/geoserver\/.*)/
```
Example to test: https://regex101.com/r/ZzAiOC/1

offtherailz:
Yes, I though about it. 
The problem of this solution is that you have to imagine every possible domain the application is deployed on. This will need rules for test environment, production and so on. And they will be forwarded to the production environment. 
Using data directory may solve the problem, by setting rules by environment on the server.
Anyway, adding a "currentPath" property, alternative to "urlPattern", that is a shortcut to check paths of the current domain/ protocol , may simplify life in several projects, that usually have different deployment in test/pre-prod/prod but same shape for relative paths.

PR #8335, opened 2022-06-16
#7952 Added pathPattern property support into authenticationRules

alexander-fedorenko:
## Description
This PR provides minor modification to the functions that return authentication rule and authentication method based on URL and rule URL pattern. It adds "pathPattern" property that can be optionally defined for rule and it will be used as a second alternative way to match rule against passed URL.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: Small improvement

## Issue

#7952 

**What is the new behavior?**
It's possible to define "pathPattern" for the rule that will me used to check for a match of "path" part of the URL.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/SecurityUtils.js
+++ b/web/client/utils/SecurityUtils.js
@@ -15,6 +15,7 @@ import isNil from "lodash/isNil";
 import isArray from "lodash/isArray";
 
 import {setStore as stateSetStore, getState} from "./StateUtils";
+import {urlParts} from "./URLUtils";
 
 /**
  * Stores the logged user security information.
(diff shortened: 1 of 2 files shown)


offtherailz (changes requested):
- [ ] It should match only current protocol-host-port - fail otherwise. The test seems to demonstrate the opposite. It works for every domain. The purpose of this is to easy migrate reuse same configuration on different domains, when the applications authenticated are deployed on same domain

For instance a path rule `\/geoserver\/` should match mapstore is running on `https://mysite:8123/mapstore/` and the request is for a request to `/geoserver/` or `https://mysite:8123/geoserver` but fail for `https://othersite:8123/geoserver`. While the original rule match both cases. So we can have a simple rule without the risk to send authkey to all other geoservers around the world

- [ ] Add proper documentation about this new parameter

tdipisa:
Out of date. PR is anyway connected to the issue so it can be reopened if needed.

Issue #8523, opened 2022-08-30
Burger menu not visible in DEV for context


tdipisa:
## Description

Its seems that due to recent updates in DEV the burger menu is no longer visible in DEV if configured for a context

## How to reproduce

- Open [this context](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/espacesvertsdrennes) in edit mode
- In the plugin wizard add the burger menu
- Save and then open the context again

*Expected Result*

The sidebar disappear and the burger menu is available and functional

*Current Result*

A sidebar covering the burger menu with only the measurement tool is present.

![image.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/82586923-c1ed-475a-be25-fd72e5df0f37)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

The tool works well in QA.

Related comment is https://github.com/geosolutions-it/MapStore2/pull/8530#issuecomment-1235240725

alexander-fedorenko:
Not a regression, see description here https://github.com/geosolutions-it/MapStore2/issues/8463#issuecomment-1232954389
Requires a minor fix in sidebar menu visibility logic.

tdipisa:
Relevant comment from @allyoucanmap in [#8530](https://github.com/geosolutions-it/MapStore2/pull/8530#issuecomment-1235240725) to resume the work for this issue as part of the 2022.02.01 iteration.

offtherailz:
@allyoucanmap need to set the alwaysRender flag in the plugin to make it work?

allyoucanmap:
@offtherailz this PR has already the Measure plugin updated with alwaysRender
https://github.com/geosolutions-it/MapStore2/pull/8754/files#diff-434d509b048e12e9e1c4c92944d208355b10faf3f2528ff363b5113199ffb6d7R203

offtherailz:
Sorry my fault. I filtered commits for the review and then I forgot to reset before doing my comment.

Issue #8463, opened 2022-08-03
Save button no longer present when saving a map from a context


tdipisa:
## Description

The 'Save' button is missing from the viewer option menu for maps saved from a context
## How to reproduce

-  Login to MS
- Open [this](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/espacesvertsdrennes) context 
- Click on "Save as" and save a map from that context
- Open the saved map 

*Expected Result*

Both "Save" and "Save as" buttons should be visible in the option menu (or Sidebar if configured)

*Current Result*

only "Save as" is present

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MV88:
this issue has another problem which can be a regression from #8488  
![image](https://user-images.githubusercontent.com/11991428/186941408-b0b13053-73bb-40f3-9f44-3ce57c4e3e0b.png)

the measurement is inserted into SideBarMenu overriding BurgerMenu one

anyway i can workaround it by disabling burgermenu entry by default

MV88:
The reason why this is not rendered is that the check made on selector of plugin is done too early and the map has not yet being loaded. In particular that context uses the sidebar and this happens only when using a context that has a sidebar inside

Also this problem with the sidebar is there since 5 may 2022 
I've checked it by testing it locally

I noticed that using a new context this is issue not happening,
especially if the context does not include burger menu like **8463complete**
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/8463complete/40562

but it happens with old contexts like **espacesvertsdrennes** that includes burger menu as optional
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/espacesvertsdrennes/40562
![image](https://user-images.githubusercontent.com/11991428/186950120-7f6a5985-3ac6-4e5d-9a25-a45806a1a436.png)

investigation continues

alexander-fedorenko:
@tdipisa 
I've tried to reproduce desired behavior and get "Save" button available for the map saved from context in last 4 or 5 releases, but I'm getting the same result regardless of the version - only "save as" is available. 

@MV88 
[...]

PR #8530, opened 2022-08-31
#8523 Extended condition of sidebar menu visibility

alexander-fedorenko:
## Description
This PR adds extra check to make sure that sidebar menu is hidden whenever burger menu is rendered on the page, regardless of the length of items to render in sidebar.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8523 

**What is the new behavior?**
Sidebar menu will be hidden whenever burger menu is rendered on the page.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/SidebarMenu.jsx
+++ b/web/client/plugins/SidebarMenu.jsx
@@ -100,7 +100,7 @@ class SidebarMenu extends React.Component {
         const { hidden } = this.state;
         const burgerMenuEnabled = this.props.state?.controls?.burgermenu?.enabled;
         const visibleElements = this.visibleItems('sidebar').length;
-        visibleElements && prevProps.isActive === false && onInit();
+        visibleElements.length && !burgerMenuEnabled && prevProps.isActive === false && onInit();
 
         if ((visibleElements === 0 || burgerMenuEnabled) && !hidden) {
             onDetach();
--- a/web/client/plugins/SidebarMenu.jsx
+++ b/web/client/plugins/SidebarMenu.jsx
@@ -102,7 +102,7 @@ class SidebarMenu extends React.Component {
         const { hidden } = this.state;
         const burgerMenuIsActive = this.props.burgerMenuIsActive;
         const visibleElements = this.visibleItems('sidebar').length;
-        visibleElements.length && !burgerMenuIsActive && prevProps.isActive === false && onInit();
+        visibleElements && !burgerMenuIsActive && prevProps.isActive === false && onInit();
 
         if ((visibleElements === 0 || burgerMenuIsActive) && !hidden) {
             onDetach();
--- a/web/client/plugins/SidebarMenu.jsx
+++ b/web/client/plugins/SidebarMenu.jsx
@@ -98,13 +98,14 @@ class SidebarMenu extends React.Component {
     componentDidUpdate(prevProps) {
         const { onInit, onDetach } = this.props;
         const { hidden } = this.state;
+        const burgerMenuEnabled = this.props.state?.controls?.burgermenu?.enabled;
         const visibleElements = this.visibleItems('sidebar').length;
         visibleElements && prevProps.isActive === false && onInit();
 
-        if (visibleElements === 0 && !hidden) {
+        if ((visibleElements === 0 || burgerMenuEnabled) && !hidden) {
             onDetach();
             this.setState((state) => ({ ...state, hidden: true}));
-        } else if (visibleElements > 0 && hidden) {
+        } else if (visibleElements > 0 && !burgerMenuEnabled && hidden) {
             onInit();
             this.setState((state) => ({ ...state, hidden: false}));
         }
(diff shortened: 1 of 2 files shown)


allyoucanmap (changes requested):
This PR adds a selector to the Side menu to ensure to make that not visible but I see another issue if compared with QA.
Why QA has also login/home tools while DEV show them for a brief moment and then it hides them?
QA screenshot
image

DEV screenshot

image

I would like to understand what's the difference between the two implementation QA vs DEV so could you please verify if the behaviour changes if you import BurgerMenu and SidebarMenu as static plugin?
I wouldn't expect to apply an additional selector to SidebarMenu but I would expect a fix on dynamic import.

alexander-fedorenko:
@allyoucanmap
This behavior indeed can be fixed by converting both BurgerMenu and SidebarMenu to static plugins. But I can't say for sure the exact reason why this change fixes it. QA is also affected by this issue (tested on QA branch locally), but it seems that these steps happens earlier there and they kept out of the user sight.
The reason behind those items appear in sidebar and then getting hidden is the fact that Home, Login plugins have same priority for OmniBar and SidebarMenu, but they should disappear from Omnibar to prevent duplicates once Sidebar is on the page.
The sequence is following: Page Loaded and plugins are getting mounted -> Sidebar is mounted and it is dispatching action to set flag in controls state -> Burger Menu is mounted and it is dispatching action to set flat that it is active -> Sidebar deactivates because BurgerMenu is on the page.
All these cross-dependencies are added to support seamless work of existing/old contexts.

I'm not sure if it's the best approach, but I think it will be cleaner to keep currently used set of plugins somewhere in state and keep it updated whenever new set of plugins is going to be rendered on the page (e.g. when PluginsContainer receives new set of plugins). This way both SidebarMenu and BurgerMenu can rely on information that will be available prior to components being mounted. @allyoucanmap what do you think?

allyoucanmap:
> This behavior indeed can be fixed by converting both BurgerMenu and SidebarMenu to static plugins. 

If the **main bundle size does not change** and If it could be fixed by having the same behaviour of QA reverting this two plugins to static import I'll go for it for the moment. I don't think this two plugins are importing additional dependencies so it should be safe to add them as static (please double check with a bundle analyzer)

> I'm not sure if it's the best approach, but I think it will be cleaner to keep currently used set of plugins somewhere in state and keep it updated whenever new set of plugins is going to be rendered on the page (e.g. when PluginsContainer receives new set of plugins). This way both SidebarMenu and BurgerMenu can rely on information that will be available prior to components being mounted. @allyoucanmap what do you think?

I'm going to link your comment in a separate issue but probably we will need to review how all the plugins are managed inside the context viewer. I would prefer to not add too much interdependencies between plugins.

alexander-fedorenko:
@allyoucanmap with the last change sidebar should stop appear for a moment when both sidebar & burger menu are added into the context.

allyoucanmap:
@tdipisa this issue is related to the new Measure tool configuration (I missed this https://github.com/geosolutions-it/MapStore2/issues/8463#issuecomment-1232954389) so is currently affecting master and we can move this to 2022.02.01.

In the Measure tool I removed priority because I needed to render a component always inside the map and noticed that was working if none of the container has priority defined.
So to fix this we need to restore the priorities and make sure that also component included in the Map container is always rendered.
There was a proposal to add a property called `alwaysRendered` in the container configuration to force the rendering and skip the priority check. Example:

```diff
export default createPlugin('Measure', {
    component: MeasurePlugin,
    containers: {
        BurgerMenu: {
            name: 'measurement',
            position: 9,
            panel: false,
            help: <Message msgId="helptexts.measureComponent"/>,
            tooltip: "measureComponent.tooltip",
            text: <Message msgId="measureComponent.Measure"/>,
            icon: <Glyphicon glyph="1-ruler"/>,
            action: () => setControlProperty("measure", "enabled", true),
            doNotHide: true,
+         priority: 2
        },
        SidebarMenu: {
            name: 'measurement',
            position: 9,
            panel: false,
            help: <Message msgId="helptexts.measureComponent"/>,
            tooltip: "measureComponent.tooltip",
            text: <Message msgId="measureComponent.Measure"/>,
            icon: <Glyphicon glyph="1-ruler"/>,
            action: toggleControl.bind(null, 'measure', null),
            toggle: true,
            toggleControl: 'measure',
            toggleProperty: 'enabled',
            doNotHide: true,
+         priority: 1
        },
        Map: {
            name: 'Measure',
            Tool: MeasureSupport,
            doNotHide: true,
+         alwaysRendered: true
        }
    },
    reducers: {
        measurement
    },
    epics: measurementEpics
});
```
Adding `alwaysRendered` means a change to the plugin architecture api, specifically to the [priority filter](https://github.com/geosolutions-it/MapStore2/blob/8d438d9ea2a0256115c34dfafc2680e6ae107224/web/client/utils/PluginsUtils.js#L332-L333) adding a check for this proposed property

tdipisa:
This issue is not affecting QA and a new PR will be open for 2022.02.01 to provide a better fix as described above.

Issue #8543, opened 2022-09-04
Tutorial plugin is not initialized properly when load map, dashboard, geoStory


alexander-fedorenko:
## Description
This is a regression caused by turning tutorial plugin into module plugin. Tutorial is not shown on page load like it was before, click on tutorial item in burger menu or sidebar menu does nothing.

## How to reproduce

- Go to map, dashboard or geoStory and refresh the page using Ctrl+F5.
- Click on tutorial item in menu or sidebar.

*Expected Result*
Tutorial pops-up

*Current Result*
Nothing happens

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #8544, opened 2022-09-04
#8543 Make tutorial compatible with being module plugin.

alexander-fedorenko:
## Description
There is a problem caused by the fact that plugins reducers are loaded after LOCATION_CHANGE action dispatched. It makes plugin epic first run skipped. This PR corrects this behavior

- Improved REDUCERS_LOADED action to make it pass list of added reducers as a payload.
- Updated switchTutorialEpic to make it actually triggered once tutorial reducer is added & one of the subsequent actions dispatched.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8543 

**What is the new behavior?**
Epic is not skipped, tracked actions list is extended to work correctly both on initial load & on location changes.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/StateUtils.js
+++ b/web/client/utils/StateUtils.js
@@ -15,6 +15,18 @@ import isEmpty from 'lodash/isEmpty';
 import {BehaviorSubject, Subject} from 'rxjs';
 import {normalizeName} from "./PluginsUtils";
 
+export const REDUCERS_LOADED = 'REDUCERS_LOADED';
+
+/**
+ * Action that should be dispatched whenever new reducers are added
+ * @param reducers
+ * @returns {{reducers, type: string}}
+ */
+export const reducersLoaded = (reducers) => ({
+    type: REDUCERS_LOADED,
+    reducers
+});
+
 /**
  * Returns a list of standard ReduxJS middlewares, augmented with user ones.
  *
@@ -338,6 +350,6 @@ export const augmentStore = ({ reducers = {}, epics = {} } = {}, store) => {
     Object.keys(reducers).forEach((key) => {
         persistedStore.storeManager.addReducer(key, reducers[key]);
     });
-    persistedStore.dispatch({type: 'REDUCERS_LOADED'});
+    persistedStore.dispatch(reducersLoaded(reducers));
     persistedStore.storeManager.addEpics('notMutable', wrapEpics(epics));
 };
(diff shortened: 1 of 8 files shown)


web/client/utils/StateUtils.js
```diff
@@ -15,6 +15,18 @@ import isEmpty from 'lodash/isEmpty';
 import {BehaviorSubject, Subject} from 'rxjs';
 import {normalizeName} from "./PluginsUtils";
 
+export const REDUCERS_LOADED = 'REDUCERS_LOADED';
+
+/**
+ * Action that should be dispatched whenever new reducers are added
+ * @param reducers
```
MV88, line 22:
```suggestion
 * @param {string[]} reducers list of reducers loaded
```

MV88 (changes requested):
- [ ] I noticed then you use tutorial in map and then you open either a new or old dashboard the tutorial there it does not start
this was happening on last release and on QA

https://user-images.githubusercontent.com/11991428/188470783-b0f13da6-d4df-4ab6-a1ed-0277b796aeea.mp4


there is also something weird, it seems it is continuing the map tutorial
if you instead go directly to a dashboard it starts the correct tutorial

this is not affecting story

@tdipisa do we need to pen a dedicated issue for this?

tdipisa:
>@tdipisa do we need to pen a dedicated issue for this?

The problem seems  to be effectively connected to this bug and therefore it should be solved as part of this PR. Can you please check @alexander-fedorenko?

alexander-fedorenko:
@MV88 Thanks for review! All listed requests for changes are applied.

Issue with incorrect behavior of Tutorial was related to the `LOCATION_CHANGE` event and to the fact that outer stream in `switchTutorialEpic` was muted at some pages upon movement from one route with tutorial plugin active to another. As a result in some cases `LOCATION_CHANGE` was intercepted by epic on one route, while inner stream and one of the intercepted actions ran on another page. Inner stream was relying on information obtained from outer action and route information was incorrect. Changing logic to use state value of router instead solved the issue.

I cross-checked code of another plugins where epics intercepting `LOCATION_CHANGE` and didn't see any other place where similar situation can appear. Issue can only appear if epic returns nested stream and rely on information from outer action.

MV88 (approved):
LGTM

MV88:

ElenaGallo:
Test passed @alexander-fedorenko please backport to 2022.02.xx. Thanks

Issue #8483, opened 2022-08-12
App Context Manager issues


ElenaGallo:
## Description

The Application context manager has problems to work on DEV.
- On new context, the Show Tutorial on the wizard steps does not work 
- When modifying an existing context the changes cannot be saved, the Save button at the end of the wizard does not work

## How to reproduce

-  Login as admin 
- Open the app context manager and try to edit and create contexts as described above
 

*Expected Result*

It is possible to start a tutorial though the Show Tutorial for all the steps of the wizard.
It is possible to save changes on a context

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

It seems a regression introduced after https://github.com/geosolutions-it/MapStore2/pull/8457. The tool works well in QA.

tdipisa:
@ElenaGallo this seems to be solved for me in DEV. Please double check.

ElenaGallo:
@alexander-fedorenko 

**2_** After testing the #8483  I found this:

For the second steps of the wizard, the tutorial start at the beginning but then when I click on Show Tutorial button the tutorial does not start and for the third steps of the wizard, the tutorial does not start and when I click on Show Tutorial button the tutorial does not start. 

## How to reproduce
See the gif below: 

https://user-images.githubusercontent.com/56537133/188417165-a6aa5978-9e89-49e9-abea-3225fcf2ef95.mov

Issue #8481, opened 2022-08-12
The MAP EDITOR INLINE is not visible 


ElenaGallo:
## Description

On geostory The MAP EDITOR INLINE is not visible
  
## How to reproduce

-  Login as admin
- Click on new map
- Click on Change Media Source
- Click on Maps tab
- Click on +

or
- Open a geostory
- From the map background, click on Edit map
- Click on Open advanced map editor

*Current Result*

The MAP EDITOR INLINE is not visible 
![map editor](https://user-images.githubusercontent.com/56537133/184393914-206225be-196a-4c39-b69e-baee6e20043a.png)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

It seems a regression introduced after #8457. The tool works well in QA.

tdipisa:
@ElenaGallo this seems to be solved for me in DEV. Please double check.

ElenaGallo:
@alexander-fedorenko

**1_** After testing the #8481 I found this:

I can not save the map from the MAP EDITOR INLINE
  
## How to reproduce

-  Login as admin
- Click on new geostory
- Click on Change Media Source
- Click on Maps tab
- Click on +
- Click on OK


*Current Result*
The Ok button is not clickable

PR #8550, opened 2022-09-05
#8483 Properly mute epics with nested PluginsContainer

alexander-fedorenko:
## Description

We have faced similar type of  issue (https://github.com/geosolutions-it/MapStore2/issues/8481#issuecomment-1237022102 & https://github.com/geosolutions-it/MapStore2/issues/8483#issuecomment-1237022941) when some plugins stops working after map is rendered.
Further investigation shown that this is related to nested instances of `PluginsContainer` that is wrapped with `useModulePlugins` hook.

Short loopback on how module plugins are working (considering functionality affected by this issue):
- Plugins are loaded on-demand, only those are on the rendered page.
- When user change route and new set of plugins have to be rendered on the page, new plugins will be loaded. already loaded plugins that are missing on the page will have their epics muted, plugins that are got back to the page will have their epics unmuted.

Issue is caused by the fact that nested instance of `PluginsContainer` has different subset of plugins, part of plugins loaded in parent `PluginsContainer` does not intersect. This makes logic of 'useModulePlugins' hook to force such plugn's epics being muted because it thinks that they are not on the page anymore (whilst they just missing in plugins passed to the nested instance of container and rendered by outer instance).

Proposed PR provides a really simple fix that should solve problem for similar case: multiple `PluginsContainer` components on the same page while path in URL is not changed. It adds a check to see if path has changed since the last run of the code that mutes epics. If path is not changed - epics mute logic won't be triggered, while unmute logic will run as always.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8483
#8481

**What is the new behavior?**
Plugins load by nested `PluginsContainer` does not cause mute of epics.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/hooks/__tests__/useModulePlugins-test.js
+++ b/web/client/hooks/__tests__/useModulePlugins-test.js
@@ -33,18 +33,15 @@ describe('useModulePlugins hook', () => {
             rootEpic: () => {}
         }
     };
-    const Component = ({onLoaded}) => {
+    const Component = ({onLoaded = () => {}, plugins, config}) => {
         const { plugins: loadedPlugins, pending } = useModulePlugins({
-            pluginsEntries: {
-                ExamplePlugin: toModulePlugin('Example', () => import('../../test-resources/module-plugins/dummy'))
-            },
-            pluginsConfig: [
-                'ExamplePlugin'
-            ]
+            pluginsEntries: plugins,
+            pluginsConfig: config
         });
         if (!pending) onLoaded(loadedPlugins);
         return false;
     };
+
     beforeEach((done) => {
         setStore(store);
         document.body.innerHTML = '<div id="container"></div>';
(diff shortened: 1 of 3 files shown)


web/client/hooks/useModulePlugins.js
```diff
@@ -119,14 +121,20 @@ function useModulePlugins({
...
         const store = getStore();
+        const urlQuery = url.parse(window.location.href, true).hash;
         if (store.storeManager) {
+            // some plugins like ContextCreator have edge-case of this scenario:
+            // there are two PluginsContainer components, whereas inner one load another list of plugins, like mapViewer
+            // on second step of context creator. To make it properly work, we need to check if router.location.path has
+            // changed and do not mute epics unless that's the case. Epics can be unmuted with no limitations though.
             Object.keys(pluginsCache).forEach((plugin) => {
-                if (!pluginsKeys.includes(plugin)) {
+                if (!pluginsKeys.includes(plugin) && urlQuery !== location) {
```
MV88, line 131: can you add a unit test?

MV88 (changes requested):
PR looks good

would be possible to add a unit test?

web/client/hooks/useModulePlugins.js
alexander-fedorenko: Added two tests to cover cases when epics should be muted (route changed) and not.

MV88: ty, can you check why pr is failing?

alexander-fedorenko: Problem resolved

web/client/hooks/__tests__/useModulePlugins-test.js
```diff
@@ -33,34 +33,130 @@ describe('useModulePlugins hook', () => {
...
+
+        ReactDOM.render(<Component plugins={plugins} config={pluginsConfig} />, document.getElementById('container'));
+        setTimeout(() => {
+            window.history.pushState('page2', 'Title', '/mapstore/#/newPage');
+            ReactDOM.render(<Component
+                plugins={{ExamplePlugin: toModulePlugin('Example', () => import('../../test-resources/module-plugins/dummy'))}}
+                config={['ExamplePlugin']} />, document.getElementById('container'));
+        }, 200);
+        setTimeout(() => {
+            expect(true).toBe(false);
```
MV88, line 159: so this is not failing because you call done before and the test was to make sure mute epics was called, right??

MV88 (approved):

web/client/hooks/__tests__/useModulePlugins-test.js
alexander-fedorenko: Yes. Otherwise, if it doesn't go inside storeManager.muteEpics() - it will fail with timeout

MV88:
@ElenaGallo testable in DEV

ElenaGallo:
Test passed, @alexander-fedorenko please backport to stable branch.

Issue #8580, opened 2022-09-12
Contexts plugin should register context manager epics since contexts grid on home page relies on them


alexander-fedorenko:
## Description
There is a problem when it's impossible to open context edit wizard from home page if contexts are listed there in the grid. This is a regression caused by the changes related to module plugins, to be precise - epics responsible for a few actions upon grid elements are not active on the home page (either not registered or muted).

## How to reproduce
- Activate "Contexts" plugin for the home page.
- Login and create a context
- Go to home page, refresh the page
- Try to open context edit wizard using pencil icon over the context grid item

*Expected Result*
Context editor is open

*Current Result*
Nothing happens

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #8583, opened 2022-09-13
#8580 Unmute epics on plugin registration

alexander-fedorenko:
## Description
This PR fixes couple of issues:
1. withExtensions hook was not passing store as a third argument to the `loadExtensions` method. This caused error logs and could potentially lead to incorrect registration of extension's reducers or epics.
2. App can have several plugins registering same epics and using side effects maintained by them (e.g. context manager and contexts grid on the home page). Flow of epics mute/unmute operations sometimes can cause the case when epic was muted while another plugin sharing same epic is not yet loaded. In this case epic will be muted mistakenly.
Example: Open contexts manager and refresh the page to make app have registered plugins only specific to this page. Pencil icon will allow you to edit context (as an admin or if it's permitted). Click on the home icon to jump back home. At this point, prior to the load of new plugins, app will mute epic that adds side effect that opens context wizard once you click on pencil. Even after `Contexts` extension is loaded - epic will still be muted.
This PR adds correction to this behavior by making an attempt to unmute all the epics that plugin/extension uses.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8580 
- Errors in log after saving the context if app has extensions registered.
- Incorrectly muted epics in some cases.

**What is the new behavior?**
See description

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/app/withExtensions.js
+++ b/web/client/components/app/withExtensions.js
@@ -78,7 +78,8 @@ function withExtensions(AppComponent) {
                     if (action.type === LOAD_EXTENSIONS) {
                         this.loadExtensions(
                             ConfigUtils.getConfigProp('extensionsRegistry'),
-                            (plugins, translations) => this.onPluginsLoaded(plugins, translations, store)
+                            (plugins, translations) => this.onPluginsLoaded(plugins, translations, store),
+                            store
                         );
                     }
                     if (action.type === PLUGIN_UNINSTALLED) {
@@ -164,6 +165,7 @@ function withExtensions(AppComponent) {
                                 reducersList.push(name);
                             });
                             store.storeManager.addEpics(impl.name, impl?.epics ?? {});
+                            store.storeManager.unmuteEpics(impl.name);
                             const pluginDef = {
                                 [pluginName + "Plugin"]: {
                                     [pluginName + "Plugin"]: {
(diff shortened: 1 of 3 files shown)


web/client/components/app/withExtensions.js
```diff
@@ -164,6 +165,7 @@ function withExtensions(AppComponent) {
                                 reducersList.push(name);
                             });
                             store.storeManager.addEpics(impl.name, impl?.epics ?? {});
+                            store.storeManager.unmuteEpics(impl.name);
```
allyoucanmap, line 168: just a question: do we need to unmute also if epics is empty?
I see that `useModulePlugins.js` is using

```js
if (size(impl.epics)) {
    store.storeManager.addEpics(impl.name, impl.epics);
    store.storeManager.unmuteEpics(impl.name);
}
```

is there a different behaviour here?

alexander-fedorenko: No, it will behave identically. There is an extra check inside unmuteEpics to see if plugin has epics registrations, otherwise there is nothing to iterate. An overhead would be in an extra run of normalizeName(), nothing else.

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test on dev and let us know when we can backport, thanks

ElenaGallo:
Test passed on DEV, @alexander-fedorenko please backport. Thanks

tdipisa:
Works also for me, @alexander-fedorenko please proceed with the backport

Issue #8591, opened 2022-09-19
Wrong class position in legend and chart


MV88:
## Description

In [this dashboard](https://mappe.comune.genova.it/MapStore2/#/dashboard/1000018941) (see the _Anagrafe_ chart),
there is wrong order, data traces are not respecting custom labels ordering

## How to reproduce

- open that dashboard
- check legends ordering

*Expected Result*

![image](https://user-images.githubusercontent.com/11991428/191003029-fbc9e647-8ef9-49be-88cf-0552d801206a.png)

*Current Result*

there is a strangely wrong position of the _yellow_ class color (with label _oltre 66_) in both chart and legend as it is possible to see from the screenshot below:
![2022-09-09 10_07_47-GeoPortale - Comune di Genova.png](https://images.zenhubusercontent.com/5c9ca89d6607b6793c66ec35/81979cea-7c0a-433d-8cce-020fe93854e6)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #8592, opened 2022-09-19
#8591 Fix chart legend ordering matching colorCatagories order

MV88:
## Description

Now charts shows unordered data based on classification label list for "value" type
dashboard with id [dev-41396](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/dashboard/41396) can be a dashboard to be used for testing

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8591

**What is the new behavior?**

Data are now sorted based on custom classification label order

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/charts/WidgetChart.jsx
+++ b/web/client/components/charts/WidgetChart.jsx
@@ -10,7 +10,7 @@ import React, { Suspense } from 'react';
 import { sameToneRangeColors } from '../../utils/ColorUtils';
 import { parseExpression } from '../../utils/ExpressionUtils';
 import LoadingView from '../misc/LoadingView';
-import { every, includes, isNumber, isString, union, orderBy } from 'lodash';
+import { every, includes, isNumber, isString, union, orderBy, groupBy, find, flatten } from 'lodash';
 const Plot = React.lazy(() => import('./PlotlyChart'));
 
 export const COLOR_DEFAULTS = {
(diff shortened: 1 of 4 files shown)


dsuren1 (changes requested):
@MV88 
As we discussed, please find below are the issues identified

- [ ] Default class label and color is missing in the chart widget
- [ ] Secondly, the sort order is matching only when the chart is stacked

image

MV88:
>  Default class label and color is missing in the chart widget

yes this was missed and I'll provide a quick fix for it

> Secondly, the sort order is matching only when the chart is stacked

I'm not sure this was requested as part of requirements, since the PR targets a specific use case, when there are more data grouped by a common attribute, when you have more records inside each bar 
for example
image

I'll check if it is possible to cover also this scenario but, this requires some more time 
@tdipisa can I take half a day more to do this point?


BTW there was no mention on what to do with default class label, so with data not matched from custom color palette ,
so I'm putting them ad the end of the bar (chart starts from the bottom)

MV88:
Nevermind I think I have found a quick way to obtain this result aswell if data are not grouped by
image
going to push an update if tests pass

tdipisa:
> Nevermind I think I have found a quick way to obtain this result as well if data are not grouped by image going to push an update if tests pass

Nothing was requested for this scenario but, since you found a quick way to provide it, it is well accepted of course.

MV88:
image

dsuren1 (changes requested):
@MV88 
I found two instances where the order of classification is not matching the legend order

- It seems to work on chart based on classification on a string attribute. However with the range color classification it doesn't
 
2022-09-22 12_45_14-MapStore2

- Also it doesn't work on stacked where the x and y attributes are different

2022-09-22 12_24_47-MapStore2

This behavior needs to be decided upon

tdipisa:
@dsuren1 since @MV88 is not available now I kindly ask you to complete this issue according to the review. Please check the code and let me know the estimate. Issue also assigned to you @dsuren1.

dsuren1:
@MV88 

> * It seems to work on chart based on classification on a string attribute. However with the range color classification it doesn't

Looking at the code, we don't have support for range classification at the moment. We need to factor in `range classification` for legend ordering to fix this one

For the stacked issue, regardless of the attributes, if the data stacks in multiple along x, we encounter this scenario (attached sample map above) where the ordering is not maintained, we need to enhance the current implementation to factor in this scenario along with range classification to fit this use case.

tdipisa:
Moving in blocked for now waiting for clarification.

alex-fko:
So the root cause of all the issues with incorrect order of the legend is how Plotly library renders legend. Legend elements will inherit order of corresponding data entries appearance (https://community.plotly.com/t/customizing-the-order-of-legends/12668).

With the applied changes both ranged and plain value classifications work fine, for all the examples previously mentioned, though now it will only reorder based on classification attribute; it will not make an attempt to reorder entries in a sub-group when X-axis values are grouped (see `DATASET_5_ORDERED` data sample to see how its changed), otherwise we end up with the case when such nested reordering breaks legend order when not all classifications are represented in the first group.

dsuren1 (approved):

dsuren1:
@ElenaGallo 
Kindly test this change in DEV once deployed and let us know if it can be backported to 2022.01.xx-c040 and eventually to 2022.02.xx

ElenaGallo:
Test passed on DED, @dsuren1 please backport. Thanks

tdipisa:
@alex-fko you can proceed with the backport. Thank you so much guys.

tdipisa:
@alex-fko please backport also to 2022.02.xx. Thank you.

Issue #8597, opened 2022-09-21
Simplifying query params handling 


tdipisa:
## Description

The MapStore query params handling needs to be reviewed a bit with the aim to make it more simple. The accepted proposal is available [here](https://github.com/geosolutions-it/MapStore2/wiki/%5BProposal%5D-MapStore-query-parameters-simplification). 

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #8617, opened 2022-09-23
#8597 Simplifying query params

alex-fko:
## Description
Part of the query parameters have a complex syntax and may be hard to use by end-user.
This PR updates existing and add new query parameters with simple syntax.

- `featureInfo` parameter got a simple syntax: `#/viewer/openlayers/new?featureInfo=38.72,-95.625` where lat and lng are comma-separated values.

- `mapInfo` parameter allows to use `SEARCH_LAYER_WITH_FILTER` action upon specific layet to get feature info based on filter criteria:
 `#/viewer/openlayers/new?addLayers=tiger:poly_landmarks;gs_stable_wms&mapinfo=tiger:poly_landmarks&mapInfoFilter=CFCC='H41'`
`mapInfoFilter` in this case is a supplementary parameter to pass cql filter in it.
Previously it was not possible to trigger `SEARCH_LAYER_WITH_FILTER` on a layer that is added via query parameters. Now if both `addLayers` and `mapInfo` parameters present in a query string, app will compare name of layer in `mapInfo` and will delay search execution whenever layers is also listed in `addLayers`. Example above will delay search execution because `tiger:poly_landmarks` is in both parameters.

- `addLayers` is equivalent of `ADD_LAYERS_FROM_CATALOGS` action: `#/viewer/openlayers/new?addLayers=tiger:poly_landmarks;gs_stable_wms,tiger-ny;gs_stable_wms&layerFilters=CFCC='H41'`
`layerFilters` is a supplementary parameter to pass cql filter applied upon layer addition.

- `background` allows to dynamically add background to the map and activate it. Supports default backgrounds provided by static service defined in `localConfig.json` (`default_map_backgrounds`) as well as other layers:
`#/viewer/openlayers/new?background=Sentinel`
`#/viewer/openlayers/new?background=tiger-ny;gs_stable_wms`
Please note that default service for `background` is `default_map_backgrounds`. Based on the selection criteria for this service it is enough to pass desired layer name even partially, e.g. `background=Sen`, it will use first layer matching search criteria.

All parameters from now on support both `camelCase` and `lowercase`, e.g. `addLayers` and `addlayers`.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8597 

**What is the new behavior?**
Existing parameters syntax is updated according to proposal.
New parameters are added (`mapInfo`, `background`, `addLayers`).
Added support of having `mapInfo` processor applied to the dynamically loaded layer (via `addLayers`)

## Breaking change
 - [x] No

### DIFF
--- a/docs/developer-guide/map-query-parameters.md
+++ b/docs/developer-guide/map-query-parameters.md
@@ -205,9 +205,13 @@ If so, it will postpone search to ensure that layer is added to the map. Otherwi
 immediately.
 
 GET: `#/viewer/openlayers/new?addLayers=layer1|service&mapinfo=layer1&mapInfoFilter="BB='cc'"`
-Where `layer1` is layer name. `service` is the service name providing layer data. Service name is optional,
-service selected by default will be used if it is omitted.
-`mapInfoFilter` is a cql filter applied to the layer.
+
+Where : 
+
+- `layer1` is layer name. 
+- `service` is the service name providing layer data. Service name is optional. If missing the default.
+- `mapInfoFilter` is a cql filter applied to the layer.
+
 
 ### Background
 
@@ -216,6 +220,7 @@ Supports default backgrounds provided by static service defined in `localConfig.
 as other layers:
 
 `#/viewer/openlayers/new?background=Sentinel`
+
 `#/viewer/openlayers/new?background=layer1|service`
 
 Where `Sentinel` & `layer1` are layer names, `service` is the service name providing layer data. Service name is optional,
--- a/docs/developer-guide/map-query-parameters.md
+++ b/docs/developer-guide/map-query-parameters.md
@@ -196,7 +196,7 @@ In the example above:
 - `layer1` and `layer2` are layer names;
 - `service` is the service identifier of the catalog.
 If no service is provided, the default service will be used.
-- `layerFilters` is a list of cql filters delimited by a semicolon (`;`) to apply to the corresponding layer in the same position of the `addLayers` parameter
+- `layerFilters` is a list of cql filters to apply to the corresponding layer in the same position of the `addLayers` parameter, separated by `;`
 
 ### MapInfo
 
--- a/docs/developer-guide/map-query-parameters.md
+++ b/docs/developer-guide/map-query-parameters.md
@@ -187,12 +187,16 @@ GET: `#/viewer/openlayers/config?bbox=8,8,53,53`
 
 This is a shortened syntax for `CATALOG:ADD_LAYERS_FROM_CATALOGS` action described down below.
 
-GET: `#/viewer/openlayers/config?addLayers=tiger:layer1|service,layer2&layerFilters="MM='nn'"`
+GET: `#/viewer/openlayers/config?addLayers=layer1;service,layer2&layerFilters=MM='nn'`
 
-Where `layer1` and `layer2` are layer names. `service` is the service name providing layer data. Service name is optional,
-service selected by default will be used if it is omitted.
-`layerFilters` is a comma-separated list of cql filters to apply to the corresponding layer.
-E.g. `layerFilters="MM='nn'","CC='bb'"`
+`addLayers` parameter is a comma separated list of `<layerName>;<service>` (`service` is optional, and if present is separated
+from the layerName by a `;`.
+
+In the example above:
+- `layer1` and `layer2` are layer names;
+- `service` is the service identifier of the catalog.
+If no service is provided, the default service will be used.
+- `layerFilters` is a comma-separated list of cql filters to apply to the corresponding layer in the same position of the `addLayers` parameter
 
 ### MapInfo
 
@@ -204,9 +208,9 @@ In this case search execution will be postponed up to the moment when layer is a
 If so, it will postpone search to ensure that layer is added to the map. Otherwise, in case of no matches, search will execute
 immediately.
 
-GET: `#/viewer/openlayers/new?addLayers=layer1|service&mapinfo=layer1&mapInfoFilter="BB='cc'"`
+GET: `#/viewer/openlayers/new?addLayers=layer1;service&mapinfo=layer1&mapInfoFilter=BB='cc'`
 
-Where : 
+Where:
 
 - `layer1` is layer name. 
 - `service` is the service name providing layer data. Service name is optional. If missing the default.
@@ -221,10 +225,13 @@ as other layers:
 
 `#/viewer/openlayers/new?background=Sentinel`
 
-`#/viewer/openlayers/new?background=layer1|service`
+`#/viewer/openlayers/new?background=layer1;service`
 
-Where `Sentinel` & `layer1` are layer names, `service` is the service name providing layer data. Service name is optional,
+Where:
+- `Sentinel` & `layer1` are layer names,
+- `service` is the service name providing layer data. Service name is optional,
 `default_map_backgrounds` will be used if it is omitted.
+
 According to the implementation of default service, it is enough to pass desired layer name even partially, e.g. `background=Sen`, 
 it will use the closest layer name match in this case.
 
--- a/docs/developer-guide/map-query-parameters.md
+++ b/docs/developer-guide/map-query-parameters.md
@@ -187,7 +187,7 @@ GET: `#/viewer/openlayers/config?bbox=8,8,53,53`
 
 This is a shortened syntax for `CATALOG:ADD_LAYERS_FROM_CATALOGS` action described down below.
 
-GET: `#/viewer/openlayers/config?addLayers=layer1;service,layer2&layerFilters=MM='nn'`
+GET: `#/viewer/openlayers/config?addLayers=layer1;service,layer2&layerFilters=MM='nn';Aa='bb'`
 
 `addLayers` parameter is a comma separated list of `<layerName>;<service>` (`service` is optional, and if present is separated
 from the layerName by a `;`.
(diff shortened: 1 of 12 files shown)


offtherailz (changes requested):
I tested the functionality and it works quite well. 

I noticed something that should be improved in doc and specification, moreover some syntax issue that we didn't noticed during the planning. 


- [ ] You used in the example the separator `|` instead of `;`. I see the application manage both. I think we should manage only one to avoid confusion or possible overlaps with reserved chars. both `;` and `|` do not seems to be used in CQL, but `|` is less common to find in many keyboards, while `;` is easier. For this reason I should use only `;`.  Anyway we have some problems with some special chars, so see comments below. 
- [ ] I noticed during the review that both "layer names" and "cql_filters" may contain `,` and cql_filter can contain `"` chars. Here we are using these chars to separate values.

@alex-fko followed the specification correctyl  but we have to decide what to do to avoid issues.

In particular
    - `,` in layer names can be send to WMS to require more than one layers in the same call overlapped
    - '," in cql_filter are used in filters `IN` 
    - `"` in cql_filter are used to specifiy attribute names and it is needed when these names have speces. E.g. the attribute filter FEATURE SIZE = 1 is not valid. While "FEATURE_SIZE" = 1 is the correct syntax.

### Commas for layer names

I think we can simply not support this for the moment. As an arternative I should use `|` to separate layers and `;` to separate parts of the same layer. 

### Commas in the cql_filters

If we change the separator for layers to |, we can follow the same approach and separate cql_filters with `|`, avoiding the issues with filters with commas. 

### Double quotes in cql_filters

Now, I see `"` in the specification used only to wrap the filters, but it doesn't seem to be necessary for a first inspection. Can we remove them?

## ... So summarising

I propose to change: 
- `|` to separate layers and cql_filters 
- `; ` to separate layer and service
- `"` remove initial and final `"` in the filter, that are unnecessary and makes the syntax more complex. 

@tdipisa @alex-fko please let me know your opinion about it.

- [ ] I also added some changes to the doc in general, to clarify more the usage and to correct generated doc indentation.  please apply my changes (and check them on the generated doc).

tdipisa:
> I tested the functionality and it works quite well.
> 
> I noticed something that should be improved in doc and specification, moreover some syntax issue that we didn't noticed during the planning.
> 
> * [ ]  You used in the example the separator `|` instead of `;`. I see the application manage both. I think we should manage only one to avoid confusion or possible overlaps with reserved chars. both `;` and `|` do not seems to be used in CQL, but `|` is less common to find in many keyboards, while `;` is easier. For this reason I should use only `;`.  Anyway we have some problems with some special chars, so see comments below.
> * [ ]  I noticed during the review that both "layer names" and "cql_filters" may contain `,` and cql_filter can contain `"` chars. Here we are using these chars to separate values.
> 
> @alex-fko followed the specification correctyl but we have to decide what to do to avoid issues.
> 
> In particular - `,` in layer names can be send to WMS to require more than one layers in the same call overlapped - '," in cql_filter are used in filters `IN` - `"` in cql_filter are used to specifiy attribute names and it is needed when these names have speces. E.g. the attribute filter FEATURE SIZE = 1 is not valid. While "FEATURE_SIZE" = 1 is the correct syntax.
> 
> ## Proposed changes
> I propose to change:
> 
> * `|` to separate layers and cql_filters. `|` is not present in [ecql specification](https://github.com/geotools/geotools/blob/main/modules/library/cql/ECQL.md), while ";"  is persent, and for consistency with the syntax, we should use the same separator for layers and filters.
> * `; ` to separate layer name and service
> * `"` remove initial and final `"` in the filters in the original proposal, that are unnecessary and overlap with the cql_sintax.
> 
> @tdipisa @alex-fko please let me know your opinion about it.
> 
> * [ ]  I also added some changes to the doc in general, to clarify more the usage and to correct generated doc indentation.  please apply my changes (and check them on the generated doc).

@offtherailz let's have a call tomorrow to check this together please.

tdipisa:
> ## Proposed changes
> I propose to change:
> 
> * `|` to separate layers and cql_filters. `|` is not present in [ecql specification](https://github.com/geotools/geotools/blob/main/modules/library/cql/ECQL.md) (or at least I couldn't find it), while ";"  is persent, and for consistency with the syntax, we should use the same separator for layers and filters.

`|` is not so common for these purposes and I don't think `;` can generate problems in the use cases we need to cover. Consider that also GeoServer allow CQL filters separation using `;` 

`https://gs-stable.geosolutionsgroup.com/geoserver/gs/wms?service=WMS&version=1.1.0&request=GetMap&layers=gs%3Any_landmarks,gs:ny_roads&CQL_FILTER=LANAME=%27Central%20Park%27;NAME=%27West%20Dr%27&bbox=-74.047185%2C40.679648%2C-73.90782%2C40.882078&width=528&height=768&srs=EPSG%3A4326&styles=&format=image/png`

As far as the separator for layers is concerned I don't see any reason to use `,` that is the most common separator. It is not a problem in general to use different separators if needed for different query params while it is a problem instead to allow two different separators for the same param :-) that it is effectively confusing.

> * `; ` to separate layer name and service

ok with this, only `;` is requested as per [original proposal](https://github.com/geosolutions-it/MapStore2/wiki/%5BProposal%5D-MapStore-query-parameters-simplification).

> * `"` remove initial and final `"` in the filters in the original proposal, that are unnecessary and overlap with the cql_sintax.

Ok, if it is not necessary we can remove `"`

alex-fko:
@tdipisa @offtherailz
I found `|` as separator to be more prominent in query string making it a little bit more readable because semicolon `;` is getting lost surrounded by letters. Anyway, as discussed above, I removed it in favor of keeping only one type of separator: `;`.

So now:
`;` - is used as a delimiter for layer name and service;
`,` - is used as a delimiter for `layer;service` definitions & for cql filters;

Double quotes are removed from examples in documentation.
Documentation is updated.

offtherailz:
@alex-fko cql_filter syntax contains `,` so we should separate filters with ";" (as GeoServer does).

web/client/utils/QueryParamsUtils.js
```diff
@@ -222,28 +231,97 @@ export const paramActions = {
...
     },
-    heading: () => {
+    addLayers: (parameters, state) => {
+        const layers = parameters.addLayers;
+        if (typeof layers === 'string') {
+            const parsed = layers.split(',');
+            if (parsed.length) {
+                const defaultSource = selectedServiceSelector(state);
+                const pairs = parsed.map(el => el.split(";"));
+                const layerFilters = (parameters.layerFilters ?? '').split(',') ?? [];
```
offtherailz, line 286: we should use ";". "," is reserved in cql.

docs/developer-guide/map-query-parameters.md
```diff
@@ -177,6 +183,58 @@ GET: `#/viewer/openlayers/config?marker=0,0&zoom=5`
...
+GET: `#/viewer/openlayers/config?addLayers=layer1;service,layer2&layerFilters=MM='nn'`
+
+`addLayers` parameter is a comma separated list of `<layerName>;<service>` (`service` is optional, and if present is separated
+from the layerName by a `;`.
+
+In the example above:
+- `layer1` and `layer2` are layer names;
+- `service` is the service identifier of the catalog.
+If no service is provided, the default service will be used.
+- `layerFilters` is a comma-separated list of cql filters to apply to the corresponding layer in the same position of the `addLayers` parameter
```
offtherailz, line 199:
```suggestion
- `layerFilters` is a list of cql filters to apply to the corresponding layer in the same position of the `addLayers` parameter, separated by `;`
```

offtherailz (changes requested):
Please use ";" as separator for cql_filters and update the examples in the PR to test as asked.

`,` is reserved for cql

`;` is used, but only to separate conditions ( filter expressions ) that is exactly the usage we are having now. Also geoserver allows this this way.

Reference: 
https://github.com/geotools/geotools/blob/main/modules/library/cql/ECQL.md

web/client/utils/QueryParamsUtils.js
alex-fko: Thanks, updated it

docs/developer-guide/map-query-parameters.md
```diff
@@ -187,7 +187,7 @@ GET: `#/viewer/openlayers/config?bbox=8,8,53,53`
 
 This is a shortened syntax for `CATALOG:ADD_LAYERS_FROM_CATALOGS` action described down below.
 
-GET: `#/viewer/openlayers/config?addLayers=layer1;service,layer2&layerFilters=MM='nn'`
+GET: `#/viewer/openlayers/config?addLayers=layer1;service,layer2&layerFilters=MM='nn';Aa='bb'`
```
offtherailz, line 190:
```suggestion
GET: `#/viewer/openlayers/config?addLayers=layer1;service,layer2&layerFilters=attributeLayer1='value';attributeLayer2='value2'`
```

offtherailz (approved):
Please update the description (and/or the issue) to make it testable from the test team.

alex-fko:
Description is updated, here is the complex example that might be useful to test changes on dev:
`#/viewer/openlayers/new?addLayers=unesco:Unesco_point;gs_stable_wms&layerFilters=cod_unesco='IT_1025'&background=Sen&mapInfo=unesco:Unesco_point&mapInfoFilter=cod_unesco='IT_1025'`

tdipisa:
@ElenaGallo please test this in DEV and let us know if we can backport to 2022.02.xx

ElenaGallo:
Test passed on DEV, @alex-fko please backport to 2022.02.xx. Thanks

Issue #8621, opened 2022-09-28
Map widgets in mobile view


alexander-fedorenko:
## Description
Widgets are not available for the map on mobile currently, there is a referenced issue https://github.com/geosolutions-it/MapStore2-C040/issues/605 requesting such enhancement.

## Acceptance criteria

- [ ]  Implement a mobile widgets layout to visualize one card at the time. This included all the improvement needed on the widget view, introduce the concept of selection of a single widget at the time and adjustment of the theme
- [ ] Review the widget header to include additional tool needed for the mobile layout. This include also all the logic to switch selected widget. The expected tool in the widgets header are:
        - collapse (minus icon)
        - expand to full screen (resize arrows)
        - go to previous widget (left arrow)
        - selectable title. Select that list all the available widgets
        - go to next widget (right arrow)
        - list of tool available for the specific widget

## Other useful information
Reference: https://github.com/geosolutions-it/MapStore2-C040/issues/605

PR #8622, opened 2022-09-28
#8621 Widgets on map mobile view

alex-fko:
## Description
This PR provides all the necessary changes to make map widgets work properly on mobile (as well as tablets).

- It updates grid layout to show single widget.
- It replaces widget title with an extra controls (select box and arrows) to allow user to toggle currently active widget.
- It simplifies behavior of widgets tray allowing only to collapse and expand widget; ability to toggle visibility of specific widget is hidden.
- Pinning, editing are deactivated for mobile view.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8621 

**What is the new behavior?**
Widgets are available for mobiles and tablets

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/widgets/widget/MobileControls.jsx
+++ b/web/client/components/widgets/widget/MobileControls.jsx
@@ -1,0 +1,56 @@
+import React, { useMemo } from 'react';
+import Select from 'react-select';
+import { Glyphicon } from 'react-bootstrap';
+import classnames from 'classnames';
+
+import Button from "../../../components/misc/Button";
+
+const MobileControls = ({ options }) => {
+    const maxIndex = options.dropdownWidgets.length - 1;
+    const activeIndex = options.dropdownWidgets.findIndex(el => el.id === options.activeWidget.id);
+    const nextIndex = activeIndex === maxIndex ? 0 : activeIndex + 1;
+    const prevIndex = activeIndex === 0 ? maxIndex : activeIndex - 1;
+    return (
+        <div className="widget-selector">
+            {options.showMobileNavigation ?
+                <Button
+                    className="previous-widget btn-sm no-border"
+                    onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[prevIndex])}
+                >
+                    <Glyphicon
+                        glyph="arrow-left"/>
+                </Button> : null
+            }
+            <Select
+                openOnFocus
+                clearable={false}
+                searchable={false}
+                value={{ id: options.activeWidget.id, label: options.activeWidget.title}}
+                onChange={(w) => options.setActiveWidget(options.dropdownWidgets.find(el => el.id === w.value))}
+                options={useMemo(() =>
+                    options.dropdownWidgets.map(w => ({value: w.id, label: w.title, active: w.id === options.activeWidget.id})),
+                [options.dropdownWidgets])}
+                optionComponent={({option, onSelect}) => <div
+                    className={classnames({
+                        "Select-option": true,
+                        "is-focused": option.active
+                    })}
+                    onPointerDown={(e) => onSelect(option, e)}
+                >
+                    {option.label}
+                </div>}
+            />
+            {options.showMobileNavigation ?
+                <Button
+                    className="next-widget btn-sm no-border"
+                    onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[nextIndex])}
+                >
+                    <Glyphicon
+                        glyph="arrow-right"/>
+                </Button> : null
+            }
+        </div>
+    );
+};
+
+export default MobileControls;
--- a/web/client/components/widgets/widget/SingleWidgetControls.jsx
+++ b/web/client/components/widgets/widget/SingleWidgetControls.jsx
@@ -5,14 +5,14 @@ import classnames from 'classnames';
 
 import Button from "../../../components/misc/Button";
 
-const MobileControls = ({ options }) => {
+const SingleWidgetControls = ({ options }) => {
     const maxIndex = options.dropdownWidgets.length - 1;
     const activeIndex = options.dropdownWidgets.findIndex(el => el.id === options.activeWidget.id);
     const nextIndex = activeIndex === maxIndex ? 0 : activeIndex + 1;
     const prevIndex = activeIndex === 0 ? maxIndex : activeIndex - 1;
     return (
         <div className="widget-selector">
-            {options.showMobileNavigation ?
+            {options.showArrowsNavigation ?
                 <Button
                     className="previous-widget btn-sm no-border"
                     onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[prevIndex])}
@@ -40,7 +40,7 @@ const MobileControls = ({ options }) => {
                     {option.label}
                 </div>}
             />
-            {options.showMobileNavigation ?
+            {options.showArrowsNavigation ?
                 <Button
                     className="next-widget btn-sm no-border"
                     onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[nextIndex])}
@@ -53,4 +53,4 @@ const MobileControls = ({ options }) => {
     );
 };
 
-export default MobileControls;
+export default SingleWidgetControls;
--- a/web/client/components/widgets/widget/SingleWidgetControls.jsx
+++ b/web/client/components/widgets/widget/SingleWidgetControls.jsx
@@ -12,15 +12,13 @@ const SingleWidgetControls = ({ options }) => {
     const prevIndex = activeIndex === 0 ? maxIndex : activeIndex - 1;
     return (
         <div className="widget-selector">
-            {options.showArrowsNavigation ?
-                <Button
-                    className="previous-widget btn-sm no-border"
-                    onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[prevIndex])}
-                >
-                    <Glyphicon
-                        glyph="arrow-left"/>
-                </Button> : null
-            }
+            <Button
+                className="previous-widget btn-sm no-border"
+                onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[prevIndex])}
+            >
+                <Glyphicon
+                    glyph="arrow-left"/>
+            </Button>
             <Select
                 openOnFocus
                 clearable={false}
@@ -40,15 +38,13 @@ const SingleWidgetControls = ({ options }) => {
                     {option.label}
                 </div>}
             />
-            {options.showArrowsNavigation ?
-                <Button
-                    className="next-widget btn-sm no-border"
-                    onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[nextIndex])}
-                >
-                    <Glyphicon
-                        glyph="arrow-right"/>
-                </Button> : null
-            }
+            <Button
+                className="next-widget btn-sm no-border"
+                onPointerDown={() => options.setActiveWidget(options.dropdownWidgets[nextIndex])}
+            >
+                <Glyphicon
+                    glyph="arrow-right"/>
+            </Button>
         </div>
     );
 };
--- a/web/client/plugins/Widgets.jsx
+++ b/web/client/plugins/Widgets.jsx
@@ -38,6 +38,7 @@ import editOptions from './widgets/editOptions';
 import autoDisableWidgets from './widgets/autoDisableWidgets';
 
 const RIGHT_MARGIN = 55;
+const singleWidgetLayoutBreakpoint = 600;
 import { widthProvider, heightProvider } from '../components/layout/enhancers/gridLayout';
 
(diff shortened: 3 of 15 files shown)


web/client/components/widgets/widget/MobileControls.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { useMemo } from 'react';
+import Select from 'react-select';
+import { Glyphicon } from 'react-bootstrap';
+import classnames from 'classnames';
+
+import Button from "../../../components/misc/Button";
+
+const MobileControls = ({ options }) => {
```
allyoucanmap, line 8: I would prefer a different name because even we are using this on mobile mainly it seems to have an effect also on desktop after shrinking the screen width ( eg. WidgetSelector, WidgetInlineSelector, ... )

web/client/plugins/Widgets.jsx
```diff
@@ -37,13 +53,17 @@ compose(
             getMaximizedState,
             dependenciesSelector,
             (state) => mapLayoutValuesSelector(state, { right: true}),
-            (id, widgets, layouts, maximized, dependencies, mapLayout) => ({
+            state => state.browser && state.browser.mobile,
```
allyoucanmap, line 56: I think we have already an isMobile selector anyway I'm wondering if we should rely only on the mobile selector, we already know that if a plugin is listed in the plugins.mobile list of localConfig this will be available only on mobile device. We could use a cfg only on mobile to enable the single widget visualization.

web/client/plugins/Widgets.jsx
```diff
@@ -160,6 +251,7 @@ class Widgets extends React.Component {
  * @memberof plugins
  * @name Widgets
  * @class
+ * @prop {boolean} [showMobileNavigation] show arrows to toggle between available widgets in mobile view. True by default.
```
allyoucanmap, line 254: Why do we need an option to disable/enable only the arrows?

web/client/themes/default/less/widget.less
```diff
@@ -154,6 +154,57 @@
...
+            }
+
+            .previous-widget {
+                left: calc(25% - 15px);
+                transform: translateX(-50%);
+            }
+            .next-widget {
+                right: calc(25% - 15px);
+                transform: translateX(+50%);
+            }
```
allyoucanmap, line 205: Why do we need to move this two buttons? I think it will be better to let them follow the flex layout default behavior

web/client/themes/default/less/widget.less
```diff
@@ -154,6 +154,57 @@
...
+                        }
+                    }
+                    .Select-input {
+                        height: 24px;
+                    }
+                }
+                .Select-menu-outer {
+                    margin-top: 2px;
+                }
+            }
```
allyoucanmap, line 191: Do we need all these changes on the select component?

alex-fko:
@allyoucanmap It's not interfering with desktop layout. `state.browser.mobile` is not affected by window width change, it relies on browser user agent. Therefore `isTablet` and `isMobile` are additional checks to precisely determine whether mobile browser is running on tablet or mobile.
Layout behave exactly the same way on dev, switching between multi column and single column, it's probably somehow related to the width checks but I'm not sure if changing this is in a scope of a task.
both `isTablet` and `isMobile` are needed just to make widgets usage experience better on tablets where it's slightly more free space to show widgets, therefore they can be displayed in desktop-like mode with some limitations (e.g. no edit mode & limited grid width etc.).

web/client/components/widgets/widget/MobileControls.jsx
alex-fko: It sounds reasonable to change the name, I like WidgetInlineSelector, but it doesn't effect desktop, otherwise select list would be visible there on shrinking browser window.

web/client/plugins/Widgets.jsx
alex-fko: This is an option, but even in this case we need to distinguish between Mobiles where viewport is usually below 600px (400+px actually) and tablets where it's more than 600px (checked here https://screensiz.es/tablet). For tablets it makes sense to have several columns because there are more space to utilize by widgets, especially in landscape mode

web/client/plugins/Widgets.jsx
alex-fko: It doesn't make sense to allow to disable select list completely, navigation between widgets is not possible without it. While arrows is an additional navigation method (to be honest I prefer it more than a select list) that could be disabled to free up some space for select list. Maybe we should allow to toggle between these two navigation methods or use both of them?

web/client/themes/default/less/widget.less
alex-fko: This is done to make it's possible to have select list dropdown as wide as its possible without breaking its default behavior. If I limit select width, internal menu element is limited by this width, making dropdown too narrow to include long words without a breaks.
Arrows with transform:
Screenshot_31

Without transform:
Screenshot_32

TLDR; to free up some space for dropdown menu

web/client/themes/default/less/widget.less
alex-fko: At least for the height, yes, widget height is 28px while react-select default style is 34px and in this version of react-select that's the only way to control its height (by rewriting it in a few places). Rest of styles are for its proper work with arrow navigation including `transform` property.

alex-fko:
@allyoucanmap Please re-review. As discussed during the call:
- Code is simplified and now activates single widget view based on window width OR if there is a mobile browser user agent.
- Naming changes;
- Simplified styling for single widget layout, widget header.

Regarding grid layout changing from multi-column to single-column and back to multi-column:
- Grid has breakpoints that are activated based on grid width (not viewport width).
- There was a condition like "if screen width is > 800px, then add an offset 500px from the left" and this effectively was shrinking available space for widgets down to 300px when viewport is 801px.
- So 0-600px width: 1 column layout based on viewport width; 600-800px - 4 column layout based on grid width; 801-960px - 1 column due to grid width <480px; 960px and up - 4 columns.

My concern is about the changes related to this. I simplified this condition to not use left offset at all, regardless of the screen width, it removed unwanted layout changes, but now widgets can be overlapped by editor panel. But that was the case previously as well, for screen sizes with width 800px and less widgets were overlapped in some cases. We can minimize this behavior by checking if left panel is open and dynamically change grid width by adding offset from the left ONLY if it won't hit the breakpoint that changes layout to single column. @allyoucanmap what do you think?

web/client/components/widgets/widget/SingleWidgetControls.jsx
```diff
@@ -0,0 +1,56 @@
...
+import Button from "../../../components/misc/Button";
+
+const SingleWidgetControls = ({ options }) => {
+    const maxIndex = options.dropdownWidgets.length - 1;
+    const activeIndex = options.dropdownWidgets.findIndex(el => el.id === options.activeWidget.id);
+    const nextIndex = activeIndex === maxIndex ? 0 : activeIndex + 1;
+    const prevIndex = activeIndex === 0 ? maxIndex : activeIndex - 1;
+    return (
+        <div className="widget-selector">
+            {options.showArrowsNavigation ?
```
allyoucanmap, line 15: please remove `options.showArrowsNavigation` is not requested in the issue

web/client/components/widgets/widget/SingleWidgetControls.jsx
```diff
@@ -0,0 +1,56 @@
...
+                    className={classnames({
+                        "Select-option": true,
+                        "is-focused": option.active
+                    })}
+                    onPointerDown={(e) => onSelect(option, e)}
+                >
+                    {option.label}
+                </div>}
+            />
+            {options.showArrowsNavigation ?
```
allyoucanmap, line 43: please remove options.showArrowsNavigation is not requested in the issue

web/client/plugins/Widgets.jsx
```diff
@@ -60,10 +81,16 @@ compose(
     compose(
         heightProvider({ debounceTime: 20, closest: true, querySelector: '.fill' }),
         widthProvider({ overrideWidthProvider: false }),
-        withProps(({width, height, maximized, mapLayout} = {}) => {
+        withProps(({ isMobileAgent }) => {
```
allyoucanmap, line 84: please try to apply this patch:

- review of width/height breakpoints
- move the singleWidgetLayoutBreakpoint to a prop

[resize.zip](https://github.com/geosolutions-it/MapStore2/files/9703862/resize.zip)

web/client/plugins/Widgets.jsx
```diff
@@ -160,6 +242,7 @@ class Widgets extends React.Component {
  * @memberof plugins
  * @name Widgets
  * @class
+ * @prop {boolean} [showArrowsNavigation] show arrows to toggle between available widgets in mobile view. True by default.
```
allyoucanmap, line 245: please remove this cfg

web/client/plugins/widgets/WidgetsTray.jsx
```diff
@@ -23,7 +23,7 @@ import BButton from '../../components/misc/Button';
 import {mapLayoutValuesSelector} from "../../selectors/maplayout";
 
 const Button = tooltip(BButton);
-
+const singleWidgetLayoutBreakpoint = 600;
```
allyoucanmap, line 26: provide this breakpoint as a prop with default value 600.

web/client/plugins/widgets/WidgetsTray.jsx
```diff
@@ -80,14 +80,17 @@ class WidgetsTray extends React.Component {
...
     };
     static defaultProps = {
         enabled: true,
         items: [],
         expanded: false,
         setExpanded: () => { },
-        layout: {}
+        layout: {},
+        isMobile: false,
+        isTablet: false
```
allyoucanmap, line 93: please remove this default props

web/client/plugins/widgets/WidgetsTray.jsx
```diff
@@ -126,6 +134,11 @@ export default compose(
         hasCollapsedWidgets: widgets.filter(({ collapsed } = {}) => collapsed).length > 0,
         hasTrayWidgets: widgets.length > 0
     })),
+    withProps(({ isMobileAgent }) => {
+        return {
+            isSingleWidgetLayout: isMobileAgent || window.innerWidth <= singleWidgetLayoutBreakpoint
+        };
```
allyoucanmap, line 140: this should have a similar logic we applied in the Widgets plugin, see the resize.patch. We should avoid to use  window.innerWidth and use a width detector (eg the HOC `import { withResizeDetector } from 'react-resize-detector';` )

allyoucanmap (changes requested):

web/client/plugins/Widgets.jsx
allyoucanmap: I would prefer to remove this option for the moment as mentioned in the next review

web/client/plugins/Widgets.jsx
alex-fko: removed

web/client/plugins/Widgets.jsx
alex-fko: thanks, applied the patch

web/client/components/widgets/widget/SingleWidgetControls.jsx
alex-fko: removed

web/client/components/widgets/widget/SingleWidgetControls.jsx
alex-fko: removed

web/client/plugins/widgets/WidgetsTray.jsx
alex-fko: provided via the patch you've posted, thanks

web/client/plugins/widgets/WidgetsTray.jsx
alex-fko: Unfortunately react-resize-detector didn't work for WidgetsTray (undefined width and height), while react-container-dimensions as a HOC/enhancer worked well.

ElenaGallo:
@alex-fko a few notes after testing the issue:

- If there is only one widget the arrows to move through the widgets are enabled

*How to Reproduce*
[Example map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/41726)

https://user-images.githubusercontent.com/56537133/194564389-2dca8eb2-7215-494a-975a-fbdf5ed1afa5.mov

- If there are more than one widget the collapse button does not work as expected

*How to Reproduce*
[Example map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/openlayers/10358)

https://user-images.githubusercontent.com/56537133/194564791-a7eb1b7f-a852-4fbf-b774-f9a14e8977f9.mov

alex-fko:
Thanks @ElenaGallo, I'll raise PR with fixes for these two problems asap.

Issue #8644, opened 2022-09-30
Simplified query params issues 


tdipisa:
## Description

Issues found during functional testing in DEV

*Expected Result*

https://github.com/geosolutions-it/MapStore2/wiki/%5BProposal%5D-MapStore-query-parameters-simplification

*Current Result*

`?featureinfo=42.212,-108.369

working

`?addLayers=gs:us_states`

working

`?addLayers=gs:us_states&bbox=-117,43,-101,49` 

working

`?addLayers=gs:us_states&center=-106.458,45.699&zoom=5&mapInfo=gs:us_states&mapInfoFilter=STATE_FIPS=30`

Identify not triggered

`?addLayers=gs:us_states&bbox=-117,43,-101,49&mapInfo=gs:us_states&mapInfoFilter=STATE_FIPS=30`

not working, identify not showing up (without _bbox_ works)

`?addLayers=gs:spearfish_roads,gs:spearfish_bugsites;gs_stable_csw,gs:spearfish_archsites;gs_stable_csw,gs:spearfish_restricted&mapInfo=gs:spearfish_restricted&mapInfoFilter=cat=3`

working

`?addLayers=gs:spearfish_roads,gs:spearfish_bugsites;gs_stable_csw,gs:spearfish_archsites;gs_stable_csw,gs:spearfish_restricted&mapInfo=gs:spearfish_restricted&mapInfoFilter=cat=3&bbox=-103.882,44.374,-103.618,44.501`

Identify not working

`?addLayers=gs:spearfish_roads,gs:spearfish_bugsites;gs_stable_csw,gs:spearfish_archsites;gs_stable_csw,gs:spearfish_restricted&mapInfo=gs:spearfish_restricted&mapInfoFilter=cat=3&layerFilters=cat='4',str1='Beetle site',str1='Bob Miller',cat='3'`

_layerFilters_ is not processed as expected if multiple filters are present (one for each layer in the list) it works only if one filter is defined

`?addLayers=gs:spearfish_roads,gs:spearfish_bugsites;gs_stable_csw,gs:spearfish_archsites;gs_stable_csw,gs:spearfish_restricted&mapInfo=gs:spearfish_restricted&mapInfoFilter=cat=3&layerFilters=cat='4'&background=gs:spearfish_dem;gs_stable_wms` 

not working as expected _background_ only works if the source is specified while it should work also without specifying it with the default source

[...]

PR #8646, opened 2022-10-02
#8644 Resolving race condition that was preventing mapInfo parameter to work in some cases

alex-fko:
## Description
This PR solves the issue with `mapInfo` query parameter described here https://github.com/geosolutions-it/MapStore2/issues/8644

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8644 
`mapInfo` parameter wasn't working under some conditions that led to `LAYER_LOAD` action appear before `ADD_LAYER` we need.

**What is the new behavior?**
`takeUntil` now checks for a `layerId` passed into `LAYER_LOAD` action to make a match with the layer name.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/search.js
+++ b/web/client/epics/search.js
@@ -335,5 +335,5 @@ export const delayedSearchEpic = (action$) =>
                 .switchMap(() => {
                     return Rx.Observable.of(searchLayerWithFilter({layer: name, "cql_filter": cqlFilter}));
                 })
-                .takeUntil(action$.ofType(LAYER_LOAD));
+                .takeUntil(action$.ofType(LAYER_LOAD).filter(({ layerId }) => layerId.startsWith(name)));
         });
--- a/web/client/epics/search.js
+++ b/web/client/epics/search.js
@@ -10,41 +10,48 @@ import * as Rx from 'rxjs';
 import toBbox from 'turf-bbox';
 import pointOnSurface from '@turf/point-on-surface';
 import assign from 'object-assign';
-import { sortBy, isNil } from 'lodash';
+import {isNil, sortBy} from 'lodash';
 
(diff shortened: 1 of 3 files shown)


web/client/epics/search.js
```diff
@@ -335,5 +335,5 @@ export const delayedSearchEpic = (action$) =>
                 .switchMap(() => {
                     return Rx.Observable.of(searchLayerWithFilter({layer: name, "cql_filter": cqlFilter}));
                 })
-                .takeUntil(action$.ofType(LAYER_LOAD));
+                .takeUntil(action$.ofType(LAYER_LOAD).filter(({ layerId }) => layerId.startsWith(name)));
```
offtherailz, line 338: Why `takeUntil(LAYER_LOAD)` ? 
It looks to me like it stops emitting when layer starts to load. I think it is a wrong check to do.
I'd use a `take(1)` instead, after filter by name. 
`takeUntil` can stop it at all and if you want to make sure that this flow is not pending on future changes, use `takeUntil` with page change.

Moreover I think this can be more an epic of the queryParams itself, so I think the action and the epic should be moved in `epics (actions) /queryparams.js`

Add a unit test to check this issue.

alex-fko: Thanks @offtherailz, you're right, `takeUntil` didn't work correctly, `take(1`) works good with extended test coverage.
I've kept changes inside search epics/actions just because this feature doesn't make sense without `Search` plugin being active; another required epic simply won't be available to fulfill workflow in this case. Besides, adding it to `queryparams` epics will require epic to be added as one of the `appEpics` inside `main.jsx`; we don't have `queryparams` actions right now at all.

offtherailz (approved):

tdipisa:
@ElenaGallo please test this in DEV and let us know.

Issue #8306, opened 2022-06-07
Legend component for WFS/Vector layers


allyoucanmap:
## Description

We could implement a component that use the json encoding style to generate a legend client side for WFS/Vector layers

The available symbolizer at the moment are:
| Symbolizer | Expected Symbol in Legend  | Available properties |
| ------------- | ------------- | ------------- |
| Polygon | square with description | fill and outline |
| Line | diagonal line | stroke and dash (line cap and join ?) |
| Icon | image icon | icon image with opacity (rotation and size?) |
| Mark | image icon | see function getWellKnownNameImageFromSymbolizer in VectorStyleUtils.js
| Text | we skip this for the moment | |

Sample structure of the expected symbols in the legend based on the previous table
![image](https://user-images.githubusercontent.com/19175505/193536746-4769da3b-4c0a-4015-9fcc-31f90c5086c2.png)

## Acceptance criteria

- [ ] Component to show the lengend inside the TOC for layer of type vector or wfs

PR #8657, opened 2022-10-04
#8306 TOC - legend for WFS and Vector layers

belom88:
## Description
Generate legend for WFS and Vector layers from the json encoding style

 - [ ] Bugfix
 - [X] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8306 

## Breaking change
 - [x] No

## Screenshots

Screenshot from 2022-10-04 17-33-04
Screenshot from 2022-10-04 17-36-38


web/client/components/TOC/DefaultLayer.jsx
```diff
@@ -118,12 +119,15 @@ class DefaultLayer extends React.Component {
...
-                    {this.props.activateLegendTool ?
+                    {this.props.activateLegendTool && this.props.node.type === 'wms' &&
                         <Row>
                             <Col xs={12}>
                                 <WMSLegend node={this.props.node} currentZoomLvl={this.props.currentZoomLvl} scales={this.props.scales} language={this.props.language} {...this.props.legendOptions} />
                             </Col>
-                        </Row> : null}
+                        </Row>}
+                    {this.props.activateLegendTool && ['wfs', 'vector'].includes(this.props.node.type) &&
+                        <WFSLegend rules={this.props.node.style.body.rules}/>
```
allyoucanmap, line 129: Instead to pass the rules array we should pass the style object (`this.props.node.style`) so the Legend component could:
- if the format is geostyler use directly the style rules
- if the style is not supported don't render the legend

In the future we could extend the possibility to convert from different format to geostyler style inside the Legend component

web/client/components/TOC/DefaultLayer.jsx
```diff
@@ -164,8 +168,12 @@ class DefaultLayer extends React.Component {
                 : null);
     }
     renderNode = (grab, hide, selected, error, warning, isDummy, other) => {
-        const isEmpty = this.props.node.type === 'wms' && !this.props.activateLegendTool && !this.props.showFullTitleOnExpand
-        || this.props.node.type !== 'wms' && !this.props.showFullTitleOnExpand;
+        const isEmpty = !(
+            this.props.showFullTitleOnExpand
+            || this.props.activateLegendTool
+                && this.props.node.type === 'wms'
+                || ['wfs', 'vector'].includes(this.props.node.type) && this.props.node.style.format === 'geostyler'
```
allyoucanmap, line 175: following the previous comment I would propose to remove the `this.props.node.style.format === 'geostyler'` check and include `3dtiles` type. Please do a test with these maps `#/viewer/cesium/39079` (batched 3d model) and `#/viewer/cesium/39241` (point cloud)
```
 || ['wfs', 'vector', '3dtiles'].includes(this.props.node.type)
```

web/client/components/TOC/DefaultLayer.jsx
```diff
@@ -118,12 +119,15 @@ class DefaultLayer extends React.Component {
...
                     {this.props.showFullTitleOnExpand ? <Row><Col xs={12} className="toc-full-title">{this.getTitle(this.props.node)}</Col></Row> : null}
-                    {this.props.activateLegendTool ?
+                    {this.props.activateLegendTool && this.props.node.type === 'wms' &&
                         <Row>
                             <Col xs={12}>
                                 <WMSLegend node={this.props.node} currentZoomLvl={this.props.currentZoomLvl} scales={this.props.scales} language={this.props.language} {...this.props.legendOptions} />
                             </Col>
-                        </Row> : null}
+                        </Row>}
+                    {this.props.activateLegendTool && ['wfs', 'vector'].includes(this.props.node.type) &&
```
allyoucanmap, line 128: based on the previous comment
```
['wfs', 'vector', '3dtiles']
```

web/client/components/TOC/fragments/MarkIcon.jsx
```diff
@@ -0,0 +1,36 @@
...
+    const [iconData, setIconData] = useState(null);
+    useEffect(() => {
+        if (!symbolizer) {
+            return;
+        }
+        async function getIconData() {
+            const data = await getWellKnownNameImageFromSymbolizer(symbolizer);
+            setIconData(data);
+        }
+        getIconData();
```
allyoucanmap, line 22: please use the promise instead o async/await, we currently did not setup babel to support them and in general we are not using them inside react components.
I would suggest to do something like this to apply icon data only if the component is still mounted

```js
const [iconData, setIconData] = useState(null);
const isMounted = useRef();

useEffect(() => {
    isMounted.current = true;
    return () => {
        isMounted.current = false;
    };
}, []);

useEffect(() => {
    getWellKnownNameImageFromSymbolizer(symbolizer)
        .then((data) => {
            if (isMounted.current) {
                setIconData(data);
            }
        });
}, [symbolizer]);
```

web/client/components/TOC/DefaultLayer.jsx
```diff
@@ -22,6 +22,7 @@ import ToggleFilter from './fragments/ToggleFilter';
 import tooltip from '../misc/enhancers/tooltip';
 import localizedProps from '../misc/enhancers/localizedProps';
 import { isInsideResolutionsLimits } from '../../utils/LayersUtils';
+import WFSLegend from './fragments/WFSLegend';
```
allyoucanmap, line 25: Looking at the code I think the legend is more related to the style rather than the layer style so I propose to change the name WFSLegend to `VectorStyleLegend` where vector refer to all features rendered client side

web/client/components/TOC/fragments/WFSLegend.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+        }
+    };
+
+    const renderRules = () => {
+        return rules.map((rule) => {
+            return (<div className="wfs-legend-rule" key={rule.ruleId}>
+                <div className="wfs-legend-icon">{renderIcon(rule.symbolizers[0])}</div>
+                <span>{rule.name || 'No Label'}</span>
+            </div>);
```
allyoucanmap, line 75: I think we should remove the no label string and adopt one of this options:
- keep only the marker without text similar to GeoServer GetLegendGraphic
- hide completely the legend rule if label is missing

what do you think @tdipisa ?

web/client/components/TOC/DefaultLayer.jsx
belom88: We need to check the format here to hide collapsible button (.toc-legend)

allyoucanmap: ok understood, we can keep `this.props.node?.style?.format === 'geostyler'` here but I would pass the style object to the Legend component anyway

allyoucanmap: as discussed we can skip the `3dtiles` for the moment,
@belom88 thanks for checking

web/client/components/TOC/DefaultLayer.jsx
allyoucanmap: as discussed we can skip the `3dtiles` for the moment,
@belom88 thanks for checking

web/client/components/TOC/fragments/WFSLegend.jsx
tdipisa: @allyoucanmap I think the first, but if you can show me maybe it is better.

web/client/components/TOC/DefaultLayer.jsx
```diff
@@ -118,12 +119,15 @@ class DefaultLayer extends React.Component {
...
                     {this.props.showFullTitleOnExpand ? <Row><Col xs={12} className="toc-full-title">{this.getTitle(this.props.node)}</Col></Row> : null}
-                    {this.props.activateLegendTool ?
+                    {this.props.activateLegendTool && this.props.node.type === 'wms' &&
                         <Row>
                             <Col xs={12}>
                                 <WMSLegend node={this.props.node} currentZoomLvl={this.props.currentZoomLvl} scales={this.props.scales} language={this.props.language} {...this.props.legendOptions} />
                             </Col>
-                        </Row> : null}
+                        </Row>}
+                    {this.props.activateLegendTool && ['wfs', 'vector', '3dtiles'].includes(this.props.node.type) &&
```
allyoucanmap, line 128:
```suggestion
                    {this.props.activateLegendTool && ['wfs', 'vector'].includes(this.props.node.type) &&
```

allyoucanmap (changes requested):

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test on dev, thanks

Issue #8588, opened 2022-09-16
Charts Switcher for Chart widget


dsuren1:
## Description

That's something similar to what implemented for the Map Switcher in [#575](https://github.com/geosolutions-it/MapStore2-C040/issues/575). It is requested to provide a new feature to Chart widgets in MapStore dashboards to allow switching between multiple charts in a single Chart widget.

## Acceptance criteria

- [ ] A kind of dropdown on top of the Chart widget allow to switch between different pre-configured charts. the same implemented for the Map Switcher
- [ ] It will be possible to activate this function and to select the list of charts during the Chart widget creation/editing
- [ ] The connection capabilities with other widgets at chart widget level will remain the same so that existing functionalities should continue working also when the widget switch over another chart selected by the user
- [ ] The Chart Switcher will be implemented as part of the Chart widget wizard, where there will be an option, a kind of plus button maybe in the toolbar, to define a list charts. The button, if clicked, will bring the user again at the first step of the wizard where it is possible to choose the layer and configure a chart for it. If more than one chart are created following this procedure a dropdown appears also in he wizard to allow the selection of the chart to see change its configuration
- [ ] If one or more charts are selected, the chart switcher will appear automatically both in the wizard and in the widget in the same way of the what implemented fir [#575](https://github.com/geosolutions-it/MapStore2-C040/issues/575)

## Other useful information
**Ref**: https://github.com/geosolutions-it/MapStore2-C040/issues/624

dsuren1:
**Revised steps in multi-chart widget**
https://github.com/geosolutions-it/MapStore2-C040/issues/624#issuecomment-1259094866

PR #8684, opened 2022-10-12
#8588 - Chart switcher for Chart widget

dsuren1:
## Description
This PR adds new feature to dashboard chart widgets where the user can configure multi charts and configure them in a widget

 - [x] Feature

## Issue

#8588 

**What is the new behavior?**
User can configure multi charts by selecting multiple layers and configure charts options individually

2022-10-10 16_48_14-MapStore HomePage

## Breaking change
 - [x] No


web/client/plugins/widgetbuilder/ChartLayerSelector.jsx
```diff
@@ -0,0 +1,88 @@
...
+import CatalogComp from './Catalog';
+
+const Button = tooltip(ButtonRB);
+
+const Catalog = compose(
+    branch(
+        ({catalog} = {}) => !catalog,
+        connect(createSelector(selectedCatalogSelector, catalog => ({catalog})))
+    )
+)(CatalogComp);
```
allyoucanmap, line 34: I'm seeing the use of recompose a lot in this PR,
do we really need it also in new components like this one?
please replace it with a react api only solution if possible

allyoucanmap (changes requested):
I did not review the code in depth yet but before to proceed with that I would like to highlight these problems, I'll resume the code review as soon as these are solved

Note: the first 3 bugs are probably related to multiple chart that should not be enabled in map widgets

- [ ] existing chart widgets on map shows an empty list of layers 
image

- [ ] editing charts on maps allows to switch charts type but once changed it's not possible to proceed to next step
image

- [ ] the application crashes after clicking on the plus button while editing on a chart widget inside a map
image

- [ ] the chart preview loses filter after switching between layers

https://user-images.githubusercontent.com/19175505/195566727-3da1abdf-aec4-4b60-92b0-e392e5ae297d.mp4

- [x] the new added layer is not one selected in the UI (maybe is by design but shouldn't be better if the new added layer is automatically selected in the editor to allow the user to fill all the input without manually switch to the new one?). Steps:
  - edit an a chart
  - add a new layer and selected one from the catalog
  - the selected layer is still the previous one and not the new one

- [ ] UI change proposal (left current PR, right proposal cc @tdipisa )

image

dsuren1:
@allyoucanmap 
Thanks for the review.
Except for the change in UI proposal, I have pushed a missing commit that should fix rest of the issues mentioned above. Kindly review at your convenience.

tdipisa:
@dsuren1 @allyoucanmap 

>UI change proposal (left current PR, right proposal cc @tdipisa )

The UI seems more clean in the one proposed by @allyoucanmap, I like it. @dsuren1 what is the effort to use it instead of the current one you provided?

dsuren1:
> > UI change proposal (left current PR, right proposal cc @tdipisa )
> 
> The UI seems more clean in the one proposed by @allyoucanmap, I like it. @dsuren1 what is the effort to use it instead of the current one you provided?

@tdipisa 
A little more than 1 point, which makes it a 2 including implementation, code and style cleanup, unit tests etc

tdipisa:
> > > UI change proposal (left current PR, right proposal cc @tdipisa )
> > 
> > 
> > The UI seems more clean in the one proposed by @allyoucanmap, I like it. @dsuren1 what is the effort to use it instead of the current one you provided?
> 
> @tdipisa 1 point including implementation, code and style cleanup, unit tests etc

@dsuren1 please proceed.

dsuren1:
@tdipisa

_Updated chart wizard layout_  
@allyoucanmap  Kindly review at your convenience. 

image

web/client/translations/data.it-IT.json
```diff
@@ -1922,6 +1929,12 @@
                 "infoOnHide": "Espandi il widget per vedere il selettore della mappa",
                 "subTitle": "<p>Per la selezione multipla,<br>su Windows usa CTRL + clic su <br> e su Mac usa control + clic</p>"
             },
+            "chartSwitcher": {
+                "noResults": "Nessun livello trovato",
+                "selectLabel": "Strato",
```
allyoucanmap, line 1934:
```suggestion
                "selectLabel": "Layer selezionato",
```

allyoucanmap (changes requested):
- [ ] I did not spot where this is happening in the code but it seems there is an issue after adding a new layer to chart that it is still using the attributes of previous layer (SUB_REGION attribe in the error is a States layer property)
https://user-images.githubusercontent.com/19175505/196222661-31a62780-1bfe-484d-bc53-7a5634ad2661.mp4
- [x] probably the screenshots for the doc needs to be updated based on the new changes on UI

dsuren1:
> * [ ]  I did not spot where this is happening in the code but it seems there is an issue after adding a new layer to chart that it is still using the attributes of previous layer (SUB_REGION attribe in the error is a States layer property)
>   https://user-images.githubusercontent.com/19175505/196222661-31a62780-1bfe-484d-bc53-7a5634ad2661.mp4

@allyoucanmap 
In this example, the widget used is connected to a table and when adding a new layer to the chart widget, it uses the filter applied on the table (sync) and results in the following error. Adding a layer with similar attributes should work ex: `USA Population`. This behavior is observed when connected widget is a table.

tdipisa:
@dsuren1 

>@tdipisa Kindly let us know the option that you think is best suited for the user

As discussed in the call the good solution should be that one:

>One could be to prevent to add layer with explanation if a filter is applied (least favorable)

We should effectively prevent the user to reach the chart config section if the layer doesn't match the schema in that particular case (widget connected to a parent table). 

@dsuren1 please let me know how much effort this check requires.

dsuren1:
> As discussed in the call the good solution should be that one:
> 
> > One could be to prevent to add layer with explanation if a filter is applied (least favorable)
> 
> We should effectively prevent the user to reach the chart config section if the layer doesn't match the schema in that particular case (widget connected to a parent table).
> 
> @dsuren1 please let me know how much effort this check requires.

It requires a little more than 1 pt (approx. 2 pts) including cleanup, unit tests etc

dsuren1:
@tdipisa @allyoucanmap 
Following error will be displayed upon selecting a layer with unmatched attributes with dependant table widget
image

tdipisa:
@dsuren1 nice, it looks good. Thank you. @allyoucanmap you can review I think at your convenience.

ElenaGallo:
Hi @dsuren1, some notes after my testing (you can used [this](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/dashboard/41908) dashboard):

- [ ] Editing Pie and Bar charts with custom color allows to switch charts type but once changed to Line Chart the custom color on Pie and Bar charts are no longer enable

https://user-images.githubusercontent.com/56537133/197206968-26dc302e-692f-4c9d-a342-1b441cfbe4f5.mov

- [ ] After inserting the filter to one of the two configured charts, opening the filter again, it is no longer visible

https://user-images.githubusercontent.com/56537133/197207317-ca615a22-142d-47fc-92bc-66394b226905.mov

- [ ] After inserting the filter to one of the two configured charts, the second chart is no longer visible in the preview

https://user-images.githubusercontent.com/56537133/197207451-8e378e99-c2c9-400e-8dc1-e40cd1675658.mov

- [ ] Once a new Pie Chart has been added to the list, the legend is visible even if the Display Legend option is not enable

https://user-images.githubusercontent.com/56537133/197207687-84373bf9-202d-4703-bde5-1fcabc6883dc.mov

- [ ] @tdipisa In the Configure Widget Info, wouldn't it be better not to have the Layers list if the title and description can only be one for all the charts?

https://user-images.githubusercontent.com/56537133/197207826-44d4cce8-7b09-494a-9fdb-b2199c6e9ea8.mov

- [ ] If I add a new widget to a map, it is not visible on the map.

https://user-images.githubusercontent.com/56537133/197207944-f9519c69-c80d-4ed8-99c4-55600906924e.mov

dsuren1:
Hi @ElenaGallo 
Thanks.

> * [ ]  Editing Pie and Bar charts with custom color allows to switch charts type but once changed to Line Chart the custom color on Pie and Bar charts are no longer enable

_That's because the color classification is not available for line chart and hence when switching the data is overriden or removed. This is existing behavior (QA for reference)_

> * [ ]  After inserting the filter to one of the two configured charts, opening the filter again, it is no longer visible

> * [ ]  After inserting the filter to one of the two configured charts, the second chart is no longer visible in the preview
 
> * [ ]  Once a new Pie Chart has been added to the list, the legend is visible even if the Display Legend option is not enable

_I will check these ones_
 
> * [ ]  If I add a new widget to a map, it is not visible on the map.

_I couldn't reproduce this issue. The widget panel is enabled below, which probably means the chart widget is created but hidden behind the TOC. Could you please check this?_

ElenaGallo:
Hi @dsuren1, 

> _I couldn't reproduce this issue. The widget toolbar is enabled below, which probably means the chart widget is created but hidden behind the TOC. Could you please check this?_

you are right, the widget is behind the TOC

tdipisa:
@dsuren1 I've also noticed that when I try to create a new chart on a existing widget using the same layer, that layer is disabled in the list while it should be enabled since it should be possible to select the same layer for a different charts configurations in the same widgets.

image

dsuren1:
@tdipisa 
> To be honest, I don't remember now the reason of this [here](https://github.com/geosolutions-it/MapStore2-C040/issues/624#issuecomment-1259094866).
> 
> > In the layer catalog, the previously selected layers will be disabled while giving user the option to select one or more layers on top of already selected layers [Screenshot-2]
> 
> Was there a specific constraints to prevent the use to select the same dataset for a different chart configuration?

_Not that I can think of, during our brainstorming we decided go with disabling when adding new one on top of existing. I can modify it to allow user to add same layer during selection_

> It would be also good to allow the user to customize the chart name in dropdown instead of leaving the layer name. Let me know the effort please.

_I think I should be able to complete by EOD today, this one including fixes for the ones mentioned above_

tdipisa:
@dsuren1 

>I think I should be able to complete by EOD today, this one including fixes for the ones mentioned above

Thank you. In that case please proceed with the above points.

dsuren1:
> @dsuren1
> 
> > I think I should be able to complete by EOD today, this one including fixes for the ones mentioned above
> 
> Thank you. In that case please proceed with the above points.

FYI, so the flow goes like this for editing chart name,
1. User selects more than one layer -> navigates to configure chart
2. Drop down shows `Chart-1`, `Chart-2` etc..
3. An edit button will be present next to dropdown
4. On clicking edit, the dropdown changes to input field to allow user to change the chart name 
5. Click ok next to the field, the dropdown is updated with new chart name

Issue #8686, opened 2022-10-12
Snapshot plugin - refactoring to be compatible with sidebar menu


alex-fko:
## Description
As agreed here https://github.com/geosolutions-it/MapStore2-C040/pull/646#issuecomment-1274354048 `Snapshot` plugin should be slightly refactored.

Acceptance criteria:
- Snapshot plugin work both for burger menu and sidebar;
- Use a separate render with the item only as a button for burgerMenu/sideBarMenu;
- Make you sure that the panel is rendered only when open, not only hidden, it must be not rendered at all because it renders a preview map.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

marthamareal:
Snapshot functionality tested locally with a Jenkins war file.

tdipisa:
@alex-fko can you please backport to 2022.02.xx?

@marthamareal as soon as a PR for a minor release is tested in DEV please remember to ask for the backport to the stable branch.

tdipisa:
@marthamareal please don't close issues for which the backport is still missing.
Thank you @alex-fko for reopening it.

PR #8687, opened 2022-10-12
#8686 Snapshot plugin refactoring

alex-fko:
## Description
Minor refactoring of `Snapshot` plugin to make it compatible with `Toolbar`, `BurgerMenu`, `SidebarMenu`.
This PR also simplifies rendering of plugin by preserving two modes (instead of three before): floating dialog or static panel.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8686 

**What is the new behavior?**
Snapshot plugin supports all three containers; Rendering is simplified to the two modes: floating dialog or static panel.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/mapcontrols/Snapshot/SnapshotPanel.jsx
+++ b/web/client/components/mapcontrols/Snapshot/SnapshotPanel.jsx
@@ -9,14 +9,14 @@
 
 import './css/snapshot.css';
 
-import { isEqual } from 'lodash';
+import {isEqual} from 'lodash';
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Alert, Col, Glyphicon, Grid, Image, Panel, Row, Table } from 'react-bootstrap';
+import {Alert, Col, Glyphicon, Grid, Image, Panel, Row, Table} from 'react-bootstrap';
 
 import ConfigUtils from '../../../utils/ConfigUtils';
 import Button from '../../misc/Button';
-import { DateFormat } from '../../I18N/I18N';
+import {DateFormat} from '../../I18N/I18N';
 import Message from '../../I18N/Message';
 import Dialog from '../../misc/Dialog';
 import Portal from '../../misc/Portal';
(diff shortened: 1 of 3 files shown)


offtherailz (approved):
Now it is visible, when enabled in localConfig.json. Approving.

offtherailz:
This can not be tested on dev because it is not officially enabled. 
@tdipisa should we simply accept it, backport and test on C040?

tdipisa:
@offtherailz we cannot test in c040 untill the PR there is approved and merged. @ElenaGallo I think it is something we can assign to @marthamareal to test locally, then we will update the PR in c040 accordingly after the backport to 2022.02.xx.

offtherailz:
Also local test is a little hard, because of CORS. The better way to test is to install mapstore and geoserver on the same tomcat and add to localConfig.json -->plugins--> desktop the "Snapshot" plugin.

tdipisa:
If so @offtherailz let's enable it temporarily in DEV to allow tests. Simple, why to waste time on complicated solutions? If tests will pass in DEV we will backport in QA testing it in the same way. Finally we will update the c040 accordingly.

offtherailz:
Also on dev may not work because gs-stable is on another server. Let's try it anyway

tdipisa:
@offtherailz let's be pragmatic: we don't need to necessarily to add layers from gs-stable to test the snapshot functionalities. we can do that also locally (imo) with an empty background and a vectorl layer imported for example. Imo @marthamareal you can proceed in this way.

marthamareal:
@alex-fko @offtherailz, I have tested this locally and the functionality works as follows:
- When the configuration is added in the localConfig.json, the snapshot button is available.
- When you click on the snapshot button, a popup opens with the current map view.
- if the view is savable by the plugin, there are no warning in popup and user can save the image. else a warning message is available to inform the user another approach of saving the image.

However, I have noticed that if a user makes a request on a view that is not savable, and error is raised. But if they close the popup and change to a savable view, on clicking the snapshot button, the previous error still exists in the popup. see video
[screen-capture (1).webm](https://user-images.githubusercontent.com/20810749/195880322-7afb9d9c-641c-4b70-9ea0-4b82b496dfe9.webm)

marthamareal:
@alex-fko @offtherailz, I have tested this locally and the functionality works as follows:
- When the configuration is added in the localConfig.json, the snapshot button is available.
- When you click on the snapshot button, a popup opens with the current map view.
- if the view is savable by the plugin, there are no warning in popup and user can save the image. else a warning message is available to inform the user another approach of saving the image.

However, I have noticed that if a user makes a request on a view that is not savable, and error is raised. But if they close the popup and change to a savable view, on clicking the snapshot button, the previous error still exists in the popup. see video
[screen-capture (1).webm](https://user-images.githubusercontent.com/20810749/195880322-7afb9d9c-641c-4b70-9ea0-4b82b496dfe9.webm)

offtherailz:
Yes, unfortunately it is the major defect of this tool. Because the warning happens when the map canvas is tainted (contain resources dot allowed for download by the security policies of the browser). And there is no way to clear this condition without removing the whole map at all and recreate a new one. It is a known issue of the tool (it will be probably improved in the future).
So the problem of the snapshot that didn't open seems to be solved.
Thank you.
I think it can be backported

PR #8700, opened 2022-10-17
#8706 get WFS feature from pick

belom88:
## Description
Get WFS feature info from pick method instead of doing get feature requests

 - [X] Bugfix
 - [X] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8706

**What is the new behavior?**
Use picking feature data in indentify plugin

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/mapinfo/wfs.js
+++ b/web/client/utils/mapinfo/wfs.js
@@ -27,8 +27,8 @@ import assign from 'object-assign';
  * @return {object} an object with `request`, containing request params, `metadata` with some info about the layer and the request, and `url` to send the request to.
  */
 const buildRequest = (layer, { map = {}, point, currentLocale, params, maxItems = 10 } = {}, infoFormat, viewer, featureInfo) => {
-    const { features = [] } = point?.intersectedFeatures?.find(({ id }) => id === layer.id) || {};
-    if (features.length) {
+    if (point?.intersectedFeatures) {
+        const { features = [] } = point?.intersectedFeatures?.find(({ id }) => id === layer.id) || {};
         return {
             request: {
                 features: [...features],
@@ -37,7 +37,10 @@ const buildRequest = (layer, { map = {}, point, currentLocale, params, maxItems 
             metadata: {
                 title: isObject(layer.title)
                     ? layer.title[currentLocale] || layer.title.default
-                    : layer.title
+                    : layer.title,
+                regex: layer.featureInfoRegex,
+                viewer,
+                featureInfo
             },
             url: 'client'
         };
@@ -90,8 +93,8 @@ const getIdentifyGeometry = point => {
 export default {
     buildRequest,
     getIdentifyFlow: (layer, baseURL, defaultParams) => {
-        const { point, features = [], ...baseParams} = defaultParams;
-        if (features.length) {
+        const { point, features, ...baseParams} = defaultParams;
+        if (features) {
             return Observable.of({
                 data: {
                     features
--- a/web/client/utils/mapinfo/wfs.js
+++ b/web/client/utils/mapinfo/wfs.js
@@ -27,6 +27,21 @@ import assign from 'object-assign';
  * @return {object} an object with `request`, containing request params, `metadata` with some info about the layer and the request, and `url` to send the request to.
  */
 const buildRequest = (layer, { map = {}, point, currentLocale, params, maxItems = 10 } = {}, infoFormat, viewer, featureInfo) => {
+    const { features = [] } = point?.intersectedFeatures?.find(({ id }) => id === layer.id) || {};
+    if (features.length) {
+        return {
+            request: {
+                features: [...features],
+                outputFormat: 'application/json'
+            },
+            metadata: {
+                title: isObject(layer.title)
+                    ? layer.title[currentLocale] || layer.title.default
+                    : layer.title
+            },
+            url: 'client'
+        };
+    }
     /* In order to create a valid feature info request
      * we create a bbox of 101x101 pixel that wrap the point.
      * center point is re-projected then is built a box of 101x101pixel around it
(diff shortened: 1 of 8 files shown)


web/client/utils/mapinfo/wfs.js
```diff
@@ -27,6 +27,21 @@ import assign from 'object-assign';
  * @return {object} an object with `request`, containing request params, `metadata` with some info about the layer and the request, and `url` to send the request to.
  */
 const buildRequest = (layer, { map = {}, point, currentLocale, params, maxItems = 10 } = {}, infoFormat, viewer, featureInfo) => {
+    const { features = [] } = point?.intersectedFeatures?.find(({ id }) => id === layer.id) || {};
+    if (features.length) {
```
allyoucanmap, line 31: here we should check if point?.intersectedFeatures is available instead of features.length because even an empty features array is a valid response.
intersectedFeatures represents the property that tell us if we should make or not the WFS requests.

```suggestion
    if (point?.intersectedFeatures) {
```

web/client/utils/mapinfo/wfs.js
```diff
@@ -75,7 +90,14 @@ const getIdentifyGeometry = point => {
 export default {
     buildRequest,
     getIdentifyFlow: (layer, baseURL, defaultParams) => {
-        const { point, ...baseParams} = defaultParams;
+        const { point, features = [], ...baseParams} = defaultParams;
+        if (features.length) {
```
allyoucanmap, line 94: Here we should prevent to make request even if the features array is empty because it could be a valid response.
Please let's remove the empty array fallback and check only if features is provided in the params
```suggestion
        const { point, features, ...baseParams} = defaultParams;
        if (features) {
```

web/client/utils/mapinfo/wfs.js
```diff
@@ -27,6 +27,21 @@ import assign from 'object-assign';
...
+        return {
+            request: {
+                features: [...features],
+                outputFormat: 'application/json'
+            },
+            metadata: {
+                title: isObject(layer.title)
+                    ? layer.title[currentLocale] || layer.title.default
+                    : layer.title
+            },
```
allyoucanmap, line 41: here are missing the metadata available in the other request: regex, viewer and featureInfo; so please add them, thanks

allyoucanmap (commented):
I saw that is a draft PR so I'm guessing some part are missing, here some additional feedback:
- ClickUtils review, is it still needed? Does pick work with transparent features without this? Do we need drillPick also here with queryable property? https://github.com/geosolutions-it/MapStore2/pull/8691/files#diff-14825f064f4138a20fe9c4a2c4222a5be38dbf4ec0450c7ba8df03b7b6f13e65
- queryable should be added to the marker in the layersSelector

belom88:
> ClickUtils review, is it still needed? Does pick work with transparent features without this? 

`scene.pickTranslucentDepth = true;`: For fully transparent marker (eg. rgba(221, 221, 221, 0) this change doesn't work. For translucent marker (eg. rgba(221, 221, 221, 1) this change is not necessary.
`scene.globe.depthTestAgainstTerrain = true;`: with `drillPick` this change confuses feature picking. If the feature is hidden behind the terrain - the marker is set on the terrain. Despite this the feature is picked because terrain is not `queryable` and is filtered after `drillPick`.

> Do we need drillPick also here with queryable property?

`drillPick` is needed to select feature under the marker. In fact the current implementation filters from picking all layers with `queryable` is not truthy.

allyoucanmap:
> drillPick is needed to select feature under the marker. In fact the current implementation filters from picking all layers with queryable is not truthy.

@belom88 I was referring to the fact that [ClickUtils](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/cesium/ClickUtils.js#L27) is still using pick and not drillPick, I think at least this need to be change to get the correct intersected queryable feature position

allyoucanmap:
> scene.pickTranslucentDepth = true;: For fully transparent marker (eg. rgba(221, 221, 221, 0) this change doesn't work. For translucent marker (eg. rgba(221, 221, 221, 1) this change is not necessary.

What happens with a feature partially transparent rgba(221, 221, 221, 0.5)? (the issue was for feature partially transparent)

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test on dev, thanks

ElenaGallo:
@belom88 now it is easier to select a point or a line but if there are two layers one above the other the GFI will only be for the first and not for the second.

*Expected Result*
Used [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/cesium/39593)

https://user-images.githubusercontent.com/56537133/198080998-750d3ec0-315b-4de4-91dc-b1564a9d7a13.mov



*Current Result*

https://user-images.githubusercontent.com/56537133/198080906-084abc24-f042-4106-a594-37c67e0d4679.mov

tdipisa:
Correct if I'm wrong @allyoucanmap but probably it is the expected behavior in 3D: the first feature that intersect it the only one returned in the Identify.

ElenaGallo:
@ElenaGallo I saw that both layers return in the Identify in production. See [this map](https://mapstore.geosolutionsgroup.com/mapstore/#/viewer/cesium/8824)

allyoucanmap:
> Correct if I'm wrong @allyoucanmap but probably it is the expected behavior in 3D: the first feature that intersect it the only one returned in the Identify.

@tdipisa I'm going to open a PR to revert the behavior to be able to get also the features/layers behind the first one.

Once we revert to multiple layer intersection we should take into account following consideration about GFI in 3D environment for future discussion.
The section diagrams below shows two situation:

- camera is perpendicular to surface similar to 2D view. In this case the ability to get all the features also the one behind the first intersected one could work and it will be consistent also with the 2D GFI. This GFI will returns based on the diagram the following layer:
  - 3D feature 2
  - WMS Layer

- camera is tilted. In this case we will intersect the first feature that will return our coordinates for the WMS GFI then the camera ray will take all the feature intersected behind. This GFI will returns based on the diagram the following layer:
  - 3D feature 2
  - 3D feature 1
  - WMS Layer
  - WFS Layer 

image

tdipisa:
@allyoucanmap all above are good points. I would like to create a new issue for further discussion with the above comment. Please, can you do that and assign it to me? Thank you.

PR #8727, opened 2022-10-24
#8729 Map views plugin implementation

allyoucanmap:
## Description

This PR introduces a new tool to configure and navigate views in the map viewer. The configuration of this tool is saved inside the map config under the `mapViews` key, here an example of structure:

```js
{
    mapViews: {
        active: true,
        selectedId: 'view.id.01',
        views: [
            {
                id: 'view.id.01',
                title: 'Title',
                description: '<p>Description</p>',
                duration: 10,
                flyTo: true,
                center: {
                    longitude: 8.93690091201193,
                    latitude: 44.39522451776296,
                    height: -0.0022900843616703204
                },
                cameraPosition: {
                    longitude: 8.93925651181738,
                    latitude: 44.38698231953802,
                    height: 655.705914040523
                },
                zoom: 17.89659156734602,
                bbox: [
                    8.920925393119584,
                    44.39084055670365,
                    8.948118718933738,
                    44.40554444092288
                ],
                mask: {
                    enabled: true,
                    resourceId: 'resource.id.01',
                    inverse: true,
                    offset: 10000
                },
                terrain: {
                    clippingLayerResourceId: 'resource.id.02',
                    clippingPolygonFeatureId: 'feature.id.01',
                    clippingPolygonUnion: true
                },
                globeTranslucency: {
                    enabled: true,
                    fadeByDistance: false,
                    nearDistance: 500,
                    farDistance: 50000,
                    opacity: 0.5
                },
                layers: [
                    {
                        id: 'layer.id.01',
                        visibility: true,
                        opacity: 0.5
                    },
                    {
                        id: 'layer.id.04',
                        visibility: true,
                        clippingLayerResourceId: 'resource.id.02',
                        clippingPolygonFeatureId: 'feature.id.01',
                        clippingPolygonUnion: false
                    }
                ]
            }
        ],
        resources: [
            {
                id: 'resource.id.01',
                data: {
                    type: 'vector',
                    name: 'mask',
                    title: 'Mask',
                    id: 'layer.id.02'
                }
            },
            {
                id: 'resource.id.02',
                data: {
                    type: 'wfs',
                    url: '/service/wfs',
                    name: "clip",
                    title: 'Clip',
                    id: 'layer.id.03'
                }
            }
        ]
    }
}
```

The UI shows a new button in the sidebar to activate/deactivate this tool

image

once the tool is activate the user will see a toolbar at the top left of the map:

- authenticated user has access to all editing, navigation and selection tools

image

- anonymous user has access only to navigation and selection tools

image

Note: if the user is anonymous and the views are empty the button on the sidebar will not be visible

The main panel to manage the view are:
- list of map view where is possible to sort, remove or select a view
image
- settings panel where is possible to change configuration of a view
image

 - [x] Feature

## Issue

#8729

**What is the new behavior?**

This PR introduces a new tool to configure and navigate views in the map viewer

## Breaking change
 - [x] No


offtherailz (changes requested):
First of all, very interesting !!!

here my points

## bugs / fixes 
- [x] A localized string missing here:
image
- [x] We must ask confirm before to delete a view, or we may lost hours of work.
- [x] If I turn off 3D you loose the current views (if I added views, they disappear). 
- [x] If I import a map file on an existing map, or if I started editing views and then I import, the views are not updated with the ones from the file. 
- [x] If I import with the TOC open, the TOC overlaps the map views. Otherwise if I open the TOC the map views tool disappears.


## Improvements suggestion
- [x] I'd replace "Show views geometries" with "Show views positions" or "show views positions" (position is the name of the label). An info tool tooltip icon that says "Shows on the map the position of the camera for all the views". 
- [x] I should replace "show settings"  in "edit", with a pencil icon. Show/hide settings is not so intuitive at least for me. Maybe it is ok like this, but it should at least show an active state when editing (I see it is dark, but it is not visible and so it is not so clear I'm in edit mode). One idea is to move the "clone" and "edit" button in the view row, near the delete, instead of the top bar. (Anyway I see that this design make the editing faster. You don't have to change and you can edit them by using navigation buttons. But this makes a lot harder to understand the position)
- [x]  I'm able to add a map-view while I'ma editing the current one. This causes to close the current in an unkown state and create a new view in an unknown position. I suggest to disable the button when editing. 
- [x] The add button logic is to add below the selected. I should show this information in the tooltip. "Add a new view below the selected one" or "Add a new" if nothing is selected. Maybe we could move the button add in list.
- [x] It is not so clear that this opens the list of the views. From a button like this I expect to show/hide the description. I should use this for description, and a list button (when eiditable) in the toolbar (instead of the buttons). 
image
- [x] the progress bar is almost invisible. 2px is too small.At list it should be 5px 
This is 5px size (of course we have to adjust the heights accordingly):
image
An alternative, to use more space, is to add it below the views bar.
image
- [x] Instead of Navigation - "Animation" or "Playback". Navigation can be confused with other terms in GIS.
- [x] Fade translucency by distance --> Fade translucency **with** distance. I should use transparency.
          Minimum distance (m)   
          Maximum distance (m)

Here a screenshot about the changes in toolbar I suggest. Of course to review, just as a reference
image


About 2D map

- [x] Having it in 2D is cool too. Anyway it do not have position. A tool like this It should provide the position of the map and hide / show layer to make sense. I think we should add this feature or hide the tool at all in 2D mode
- [x] I have the "Mask" layer anyway in the 2D view, and it should be removed.

tdipisa:
>Here a screenshot about the changes in toolbar I suggest. Of course to review, just as a reference

Leave it as it is now

allyoucanmap:
@ElenaGallo some additional suggestions for the test of this plugin:

- test the responsiveness of the toolbar and panels. Try with 1280 width ×720 height as minimum sizes, for the moment it should work only on desktop
- test if the plugin is working inside contexts

offtherailz (approved):
After the fixes asked for 180 rotation bug and edit/close edit label (@allyoucanmap  fixed them as well as I found during the review), it looks good to me.

I'd prefer to show edit and list tools active in green instead of dark blue (because dark blue is also the color of a focused button, and not so visible). Plus I had the rest of the comments to review, but I see we let them go at this stage.

Apart for that comments, it is ok for me.

@tdipisa and @allyoucanmap  if you read my comment, you can merge the PR. 
Please @allyoucanmap make also @ElenaGallo aware of the current limitations. Having a sample JSON map to try could be good.
And don't forget to document the format of views object in dev guide and describe this plugin in user guide.

tdipisa:
Thank you so much @offtherailz and @allyoucanmap

@ElenaGallo please sync with @allyoucanmap to prepare the functional test once merged.

@offtherailz, @allyoucanmap provided a a last change for the CSS as you requested. Can we effectively merge now?

offtherailz:
Merged. @ElenaGallo, could you please test this on  DEV as well as it is deployed ? Thank you

tdipisa:
@ElenaGallo plese coordinate with @allyoucanmap to test in DEV.

ElenaGallo:
Hi @allyoucanmap, I only noticed two little things during my tests:
**1_** If the TOC (or the catalog)  is open and the Map view is enabled the TOC should close automatically to allow the user to see 

https://user-images.githubusercontent.com/56537133/199505985-15271da0-6714-4ea6-ad6a-c57622caa98f.mov

the Map View toolbar

**2_** If the Map view is enabled and the attribute table is open, it is not possible to close the Map view because the button in the sidebar is not visible

https://user-images.githubusercontent.com/56537133/199506006-d1fd2c4f-0e00-4a62-952b-133560a41070.mov

Issue #8778, opened 2022-11-04
Feature Grid filter by viewport


alex-fko:
## Description
In order to allow the users to quickly explore the data of a particular area, the feature grid should provide the possibility to filter data on the current viewport.

## Acceptance criteria

- [ ]  As an administrator I want to be able to configure a new feature called "filterByViewport". This functionality allow to filter the data in the feature to the features that intersect the map viewport. 
    - [ ] As an administrator I can configure an optional button to turn on/off this functionality.
    - [ ] The visibility and the initial state (on/off) of this feature have to be configured in the feature grid configuration. The 2 configuration are independent (so the 
        - [ ] `filterByViewport` : `true` | `false` for the initial state
        - [ ] `showFilterByViewportTool`: `true` | `false` to display | hide the button.
- [ ] As a user, when this functionality is active, I will see on  grid only the results that intersect the current viewport. As well as the map viewport change due to pan or zoom (manual or triggered by other tools) the result list will update to filter results properly. 
- [ ] As a user, if the button is visible, I can turn on/off the filterByViewport. 
- [ ] The filter should be added in "AND" with all the other filters of the feature grid, and should be probably use the same logic of the quick filters.

## Other useful information
Reference issue: https://github.com/geosolutions-it/npa-cgg/issues/289

PR #8793, opened 2022-11-10
#8778 Feature Grid filter by viewport

alex-fko:
## Description
This PR adds an additional filter type to the feature editor records: dynamic filter by viewport, updated on each map view change.

One important note about implementation: any other spatial filter applied to the request will be added as an "AND" condition.
E.g. when user selects viewport filter via advanced filters or draw rectangle on the map to select features. In all these cases condition will be `Drawn box AND Viewport rectangle must intersect feature` or `Advanced filter viewport AND Dynamic viewport must intersect feature`.

Tool is disabled by default and not available in toolbar. Please use following configuration to activate button and activate tool by default:
```
{
  "name": "FeatureEditor",
  "cfg": {
    "filterByViewport": true,
    "showFilterByViewportTool": true
  }
}
```

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#8778 

**What is the new behavior?**
New tool is available in toolbar once active in plugin configuration.

## Breaking change
 - [x] No


offtherailz (changes requested):
Looks quite good.

- [ ] For the icon please look at this conversation https://github.com/geosolutions-it/MapStore2/pull/8788#issuecomment-1310072554 
- [ ] The filter should be reset together with the other tools (e.g. sync with map). Instead switching from one map to the other I see only it activated. Please find out how the other tools are reset and do the same.
- [x] Make the tool visible
- [x] I should change the text into "Filter data by the current viewport" - "Do not filter data by the current viewport" ("Activate viewport filter" is ambigous and messes up the translations --> they become something like "activate the filter of the view" )

offtherailz:
Hi alexander, the icon has been added here  https://github.com/geosolutions-it/MapStore2/pull/8807 , please apply the change and request the review when you are ready
Thank you very much.

alex-fko:
Thanks @offtherailz, I've applied new icon, all other points are addressed too.

offtherailz (changes requested):
Good. We have only a couple of issues left: 
- [ ] The filter should be disabled in 3D mode
- [ ] The filter should be deactivated (reset) together with the other tools. I see it remains active all the time. This can cause
     - Issues when switching between different contexts (e.g. with this button not visible, but the tool that could stay active from a previous usage in a context where it was present and the user enabled)
     - Issues when switching from 3D to 2D

Please test the two workflow above to verify that the tool is consistent in any case

- [ ] Moreover there is another issue, more a UX problem, because we have in the results also the ones hidden by the table you drag and drop from the bottom. 
Looking at the current API of Openlayers and mapstore, it looks difficout to allow this. 
apparently calculateExtent method from view do not allow to pass a custom padding. The only way I can imagine we can fix it is to synchronize the padding with the current mapLayout state that holds the size in pixels of the tools around. 
Anyway this can easely become tricky and involve other tools that use padding (zoomToExtent/Indentify) and have to be discussed with the team. Do you have any other idea  @alex-fko ?

@tdipisa do you think we can live with this last Ux problem for the moment?

tdipisa:
@offtherailz 

>@tdipisa do you think we can live with this last Ux problem for the moment?

behind certain extents you cannot expect I can provide an useful feedback on a topic simply mentioning me in the comment. That's something you should raise during our periodic sync maybe showing me what you are referring to :)

alex-fko:
@offtherailz "The filter should be deactivated (reset) together with the other tools", I see your point. Each location change should reset value back to the default one defined in plugin configuration.

Regarding bbox calculation and the fact that filter includes part of viewport covered by the panel itself: I think it should be doable by utilizing  GET_COORDINATES_FROM_PIXEL_HOOK: Get coordinates of the ponts for the rectangle not covered with panel. Convert them into bbox and use it in request.

offtherailz:
Using `GET_COORDINATES_FROM_PIXEL_HOOK` may be hard and complicate a lot the things.
Moreover I see from this issue that the way I suggested (by setting the padding in the map) has been fixed and properly works only for OL > 6.6.0  (we are on ol 5.3.0 now).

offtherailz (changes requested):
Looks almost perfect. I noticed this glitch here (filter button that alternate blue and green color): 

https://user-images.githubusercontent.com/1279510/201884874-205036b9-818e-4f80-9659-5e3ed63d2637.mp4

Do you think we can fix it?

alex-fko:
@offtherailz it looks like it's intended, it behaves same way on production now: once there is any filter (not necessarily via advanced search) - this button turns green. In case if there is no advanced filter or, spatial via drawn shape on the map, or simple via inputs above properties - it gets back to inactive blue state.

alex-fko:
@offtherailz thanks, it should not flick now

web/client/utils/FilterUtils.js
```diff
@@ -1119,6 +1119,7 @@ export const isFilterValid = (f = {}) =>
     || (f.simpleFilterFields && f.simpleFilterFields.length > 0)
     || (f.spatialField && f.spatialField.geometry && f.spatialField.operation ||
         isArray(f.spatialField) && findIndex(f.spatialField, field => field.operation && field.geometry) > -1)
+    || (f.viewportFilter && f.viewportFilter.geometry && f.viewportFilter.operation)
```
offtherailz, line 1122: Sorry, is this needed anymore?

offtherailz (changes requested):
Last change: 
- [ ] change mapView !== 'cesium' into a selector called `isFilterByViewportSupported`. Makes everything more explicit and resistent to future evolutions.

offtherailz (approved):

offtherailz:

ElenaGallo:
Test passed on DEV

Issue #8767, opened 2022-11-03
Exclude some layers from timeline 


offtherailz:
## Description
To avoid a lot of layers in the timeline, some layers can be configured to not appear in timeline, even if syncronized with time selected.
If you have a lot of data with the same dates, this will simplify the visualization (10 layers with the same point may show 10 lines in the timeline, all with the same points, the UI is redundant). 

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## Acceptance criteria
- [x] Prepare a mockup first
- [ ]  As a user, if a layer has a time dimension, I can hide it from timeline settings tool. The timeline settings panel has  a new "layers" panel. in this panel you can see the list of layers. Each layer has a checkbox, 
    - [ ] By default every checkbox is checked. This means that the layer is visible in the timeline.
    - [ ] If the checkbox is unchecked the layer disappears from the timeline. 
    - [ ] If I check the checkbox, the layer appears again in the timeline.
    - [ ] The guide layer can not be hidden.
    - [ ] At least one layer must stay on the timeline. 
    - [ ] Anyway the layers will all be in sync with the current time, even if hidden.
    - [ ] If I save the map and reload, the status of the checkboxes must be persisted.

dsuren1:
@offtherailz 
Basic mockup of the layer selector in timeline settings. Kindly let me know if needs some refinement.

![time_layers](https://user-images.githubusercontent.com/26929983/200324908-63ac823b-c693-4e95-9fbe-099a42961552.jpg)

offtherailz:
ok @dsuren1 keep in mint to 
- store the hidden ones list (new layers should be visible by defalut)
- make the it scrollable when the layers when overflows.

dsuren1:
**Updated mockup**
![image](https://user-images.githubusercontent.com/26929983/201946846-14f94d31-8c3b-4bfd-9d6c-6721dba6bb5a.png)

PR #8823, opened 2022-11-16
#8767: Show/hide layers from timeline

dsuren1:
## Description
This PR adds new layers tab to timeline settings. Allows user to show/hide a layer from the timeline 

 - [x] Enhancement

## Issue

#8767 

**What is the new behavior?**
- Show/hide layers from timeline
- Guide layer is disabled or when only one layer present
- Save time layers settings

## Breaking change
 - [x] No


web/client/components/playback/Settings.jsx
```diff
@@ -121,73 +125,108 @@ export default ({
...
+        <Tab eventKey={2} title={<Message msgId="playback.settings.tab.layers" />}
+            style={{
+                height: 250,
+                width: 280,
+                overflowY: "auto",
+                paddingTop: 8
+            }}>
+            {layers.map(({title, id, checked} = {}) => {
+                const checkedLayer = layers?.filter(l=> l.checked);
+                const disable = (checkedLayer.length === 1 && checkedLayer?.[0]?.title === title) || selectedLayer === id;
```
offtherailz, line 210: You should never compare a "title" to guess a "disable" flag.  Please always use layer ID, or if we change the title, we loose the relationship.

web/client/epics/timeline.js
```diff
@@ -546,6 +551,41 @@ export const rangeOnInitSelectLayer = (action$, { getState = () => { } } = {}) =
...
+                // Skip time layer update on initialization
+                // as layer data is updated from map config
+                return $observable;
+            }
+            const layers = layersWithTimeDataSelector(state)?.filter(l => l.visibility);
+            if (!isEmpty(layers)) {
+                let layersSetting = layers.map(layer => {
+                    const id = layer.id;
+                    const timeLayer = timeLayers.find(l => l.id === id);
+                    return {title: layer.title, id, checked: isEmpty(timeLayer) ? true : timeLayer.checked};
```
offtherailz, line 577: You should not store the title of the layer here. Only the `id` and the properties you want to add should be there.
(also a map `{ ["LAYER_ID"]: {props...}}` may make sense.
The property should be "hideInTimeline" or similar, because "checked" is a UX tool condition, not representing the semantics. Moreover I suggest "hide" instead of "show" because "show" (checked) is the default. 
Using hide you don't need to initialize anything. And you may be able to store only the information when the user changes something.. 

-In layer settings, set as unchecked only the items with "hideInTimeline" = true.
- when the user unchecks the item, you can add / update the item in the settings. hideInTimeline: true
- when the user checks again the item, you can update it with hideInTimeline: "false".

web/client/components/playback/Settings.jsx
```diff
@@ -121,73 +125,108 @@ export default ({
...
+                return (
+                    <div className={"layer-setting"}>
+                        <label key={id}>{title}</label>
```
offtherailz, line 213: - [ ] If the title layer is localized (from TOC), the application crashes. Please properly localize the title.

docs/user-guide/timeline.md
```diff
@@ -121,3 +121,14 @@ In order to properly set the Animation Ranger, some controls are available to he
...
 * Extend the animation's green range until it fits the guide layer time extension through the **Fit to selected layer's range** button <img src="../img/button/timeline-fit-layer-range.jpg" class="ms-docbutton"/>
+
+## Layers Setting
+
+The **layers** tab lists all the available time layers present in the map. The user can toggle a layer to be *shown/hidden* in the timeline by clicking on the checkbox next to the layer title.
+
+The layer checkbox is disabled in the following condition
+
+* When the layer is a guide layer
+* When only one layer is present
```
offtherailz, line 132: This detail is already in UI, no need to specify it in documentation.
```suggestion
```

offtherailz (changes requested):
I asked to fix some bugs and labeles, moreover I asked to change the logic and semantics of stored data. 
Please let me know if you have any comment or if I'm wrong with something.

Thank you in advance

tdipisa:
@dsuren1 please let me know how much time you need for the above for a proper schedule.

web/client/configs/localConfig.json
```diff
@@ -486,7 +486,9 @@
             {
               "name": "Timeline",
               "cfg": {
-                "expandLimit": 20
+                "expandLimit": 20,
+                "endValuesSupport": true,
+                "snapRadioButtonEnabled": true
```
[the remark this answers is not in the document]
dsuren1: Look to be from old commit

web/client/components/playback/Settings.jsx
dsuren1: Yes, it's supposed to be layer id

web/client/components/playback/Settings.jsx
dsuren1: Localized now

offtherailz (approved):

PR #8924, opened 2023-01-23
Update User Guide - Images and links

ElenaGallo:
## Description

Update images and links on User Guide


MV88 (changes requested):
- [ ] docs/user-guide/img/managing-users-and-groups/login-win.jpg is not containing home button
- [ ] docs/user-guide/img/managing-users-and-groups/manager.jpg is not containing home button
- [ ] docs/user-guide/img/home-page/new-res.jpg when you use new-res.jpg near "Create new resources" can you specify which element is
- [ ] is this filter icon explained somewhere? 
image
- docs/user-guide/img/background/bck-available.jpg has changed a lot, why ?
image

ElenaGallo:
Hi @MV88, 

> - [ ] docs/user-guide/img/managing-users-and-groups/login-win.jpg is not containing home button
> - [ ] docs/user-guide/img/managing-users-and-groups/manager.jpg is not containing home button
for the speech that is being made in the section where these two images are located, I think it is better that they are screenshots of what can be seen from the Homepage and not from Manage Accounts

> - [ ] is this filter icon explained somewhere? 
image

I open a new issue

> that button although is missing in current Side bar list explanation

It is not listed in the sidebar because it is only found in the sidebar of a context and not by default in the map viewer

MV88:
>  For the speech that is being made in the section where these two images are located, I think it is better that they are screenshots of what can be seen from the Homepage and not from Manage Accounts

ok

> It is not listed in the sidebar because it is only found in the sidebar of a context and not by default in the map viewer

ok

MV88 (approved):

PR #8926, opened 2023-01-23
#7928: Fix - Mobile - Extra page on GFI panel

dsuren1:
## Description
This PR fixes displaying of extra pages in GFI panel in mobile when making a GFI query

 - [x] Bugfix

## Issue

#7928 

**What is the new behavior?**

- **Desktop** - _Click_ and _Hover_ behavior is unaffected
- **Mobile** - Only valid responses pages are displayed and can swipe to view only those (Empty pages are not displayed)

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/identify/DefaultViewer.jsx
+++ b/web/client/components/data/identify/DefaultViewer.jsx
@@ -74,14 +74,15 @@ class DefaultViewer extends React.Component {
     /**
      * Get validation properties of the responses
      */
-    getResponseProperties = () => {
+    getResponseProperties = (renderValidOnly = this.props.renderValidOnly) => {
         const validator = this.props.validator(this.props.format);
         const responses = this.props.responses.map(res => res === undefined ? {} : res); // Replace any undefined responses
-        const validResponses = this.props.renderValidOnly ? validator.getValidResponses(responses) : responses;
+        const validResponses = renderValidOnly ? validator.getValidResponses(responses) : responses;
         const invalidResponses = validator.getNoValidResponses(this.props.responses);
         const emptyResponses = this.props.requests.length === invalidResponses.length;
         const currResponse = this.getCurrentResponse(validResponses[this.props.index]);
         return {
+            responses,
             validResponses,
             currResponse,
             emptyResponses,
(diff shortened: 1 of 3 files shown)


web/client/components/data/identify/DefaultViewer.jsx
```diff
@@ -97,6 +98,20 @@ class DefaultViewer extends React.Component {
         return validator.getValidResponses([response]);
     }
 
+    /**
+     * Helper method to calculate panel index properly on mobile devices.
+     */
+    getPanelIndex = (index, filterIndex = true) => {
+        // Recalculate index value of valid responses when mode is mobile
+        const { responses, validResponses } = this.getResponseProperties(this.props.isMobile);
```
allyoucanmap, line 106: I need some clarifications, Why do we have two different behaviours in mobile and desktop?
Shouldn't both have the same results?
If I understood correctly this is happening because of the swipe interaction, is there a reason to render invalid pages on desktop? if not wouldn't be better to have same behavior in desktop and mobile so to remove always the invalid pages?

dsuren1: The behavior of desktop and mobile is same. Only difference is that in desktop, the user can't swipe to look at empty render pages as the only way to interact is via the layer selector DD field. In mobile, however, the user can swipe to look at container's child which may have empty pages, so to avoid that the panel index is calculated such that user doesn't view these empty pages.

allyoucanmap (approved):

allyoucanmap:

ElenaGallo:
Hi @dsuren1, 
on mobile, an extra page still appears with the note of the features not present on the clicked point.


https://user-images.githubusercontent.com/56537133/214902077-f7299ca5-21b4-4cce-8aab-003cf44b5d90.mov

dsuren1:
> Hi @dsuren1, on mobile, an extra page still appears with the note of the features not present on the clicked point.
> 
>  gfi.mov

Hi @ElenaGallo 
That is not mobile view as you can see sidebar menu. Probably happened due to app not being fully loaded in mobile view. 
In order to avoid this, when you switch to mobile view in devtools, kindly refresh the page so the application reloads based on the emulation applied. Please give it a try and let me know how it goes. Thanks

ElenaGallo:
Hi @dsuren1, you are right, now the test passes. Thanks

Issue #8795, opened 2022-11-10
Attribute table has highlighted boolean empty cells even if they are not changed


allyoucanmap:
## Description

Attribute table has highlighted boolean type empty cells even if they are not changed

https://user-images.githubusercontent.com/19175505/200610039-9663be92-863d-4839-a40f-b58ca0a5c11a.mp4

## How to reproduce

- login in mapstore as admin
- add from catalog the `Types` layer from WMS gs-stable
- edit the layer in with the attribute table
- double click on an Boolean type empty cell to edit it and keep the value empty

*Expected Result*

- The cell should not be highlighted as changed

*Current Result*

- The cell is highlighted as changed

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

mulondo:
This issue looks to be fixed, I can't replicate it on my side


https://user-images.githubusercontent.com/22595454/214548468-5b53e853-ff98-408b-a5d2-b61108ee4d9d.mp4

tdipisa:
@mulondo is is perfectly reproducible, but it seems not with all field types but only text ones.

mulondo:
@tdipisa, with which fields can I test it no?

allyoucanmap:
@mulondo just updated the layer, now the string field is empty so you can test it

![image](https://user-images.githubusercontent.com/19175505/214791647-b1871ac3-23e3-41cf-91b2-5ba39415d3af.png)

mulondo:
Alright let me test it again @allyoucanmap

mulondo:
@allyoucanmap, still I can't reproduce the bug, everything seems fine.


https://user-images.githubusercontent.com/22595454/214799503-bc8d66f0-942c-4740-8d6c-a23347632a37.mp4

mulondo:
@allyoucanmap and @tdipisa, I have seen it. It's on the String and Boolean.

allyoucanmap:
@tdipisa as discussed in the PR the boolean and string issues have been split in two different issue so this is the new one
#8958

PR #8946, opened 2023-02-01
#8795 Attribute table has highlighted boolean empty cells even if they are not changed

mulondo:
## Description
The PR fixes the issue with the attribute table cells being highlighted even if they are not changed

 - [x] Bugfix

## Issue

#8795 

**What is the new behavior?**
The attribute table cells are no longer highlighted are when not changed.

## Breaking change
 - [x] No

## Demo

https://user-images.githubusercontent.com/22595454/215998823-349993ff-7762-4792-9de2-1573888c4151.mp4

### DIFF
--- a/web/client/components/data/featuregrid/renderers/CellRenderer.jsx
+++ b/web/client/components/data/featuregrid/renderers/CellRenderer.jsx
@@ -24,7 +24,7 @@ class CellRenderer extends React.Component {
     }
     render() {
         const isProperty = this.context.isProperty(this.props.column.key);
-        const isModified = (this.props.rowData._new && isProperty) || this.context.isModified(this.props.rowData.id, this.props.column.key);
+        const isModified = (this.props.rowData._new && isProperty);
         const isValid = isProperty ? this.context.isValid(this.props.rowData.get(this.props.column.key), this.props.column.key) : true;
         const className = (isModified ? ['modified'] : [])
             .concat(isValid ? [] : ['invalid']).join(" ");
--- a/web/client/components/data/featuregrid/renderers/CellRenderer.jsx
+++ b/web/client/components/data/featuregrid/renderers/CellRenderer.jsx
@@ -24,7 +24,7 @@ class CellRenderer extends React.Component {
     }
     render() {
         const isProperty = this.context.isProperty(this.props.column.key);
-        const isModified = (this.props.rowData._new && isProperty);
+        const isModified = (this.props.rowData._new && isProperty) || this.context.isModified(this.props.rowData.id, this.props.column.key);
         const isValid = isProperty ? this.context.isValid(this.props.rowData.get(this.props.column.key), this.props.column.key) : true;
         const className = (isModified ? ['modified'] : [])
             .concat(isValid ? [] : ['invalid']).join(" ");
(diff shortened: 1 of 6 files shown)


web/client/components/data/featuregrid/renderers/CellRenderer.jsx
```diff
@@ -24,7 +24,7 @@ class CellRenderer extends React.Component {
     }
     render() {
         const isProperty = this.context.isProperty(this.props.column.key);
-        const isModified = (this.props.rowData._new && isProperty) || this.context.isModified(this.props.rowData.id, this.props.column.key);
+        const isModified = (this.props.rowData._new && isProperty);
```
allyoucanmap, line 27: @mulondo I think we can't solve the issue by removing the [check for modified values](https://github.com/geosolutions-it/MapStore2/blob/e97942967045fef363cd8b8a36a49a36b950f8a5/web/client/components/data/featuregrid/FeatureGrid.jsx#L78-L81) as you are proposing here because this will remove the highlights also for the one already working.

@tdipisa I did an investigation on the reason of the highlighting of changed fields even if there are no visual changes from a UI. The reason seems related to String and Boolean property with `null` value (new created feature has all properties set to null).  When we are changing the value from the UI for string and boolean we are transforming a `null` value to `""` (an empty string) so this is detected has change and highlighted, even if we are rendering `null` and `""` in the same way inside the grid cell.

I think we can solve the problem of the boolean value by adding a fallback to null instead of empty string while we are [forcing the selection](https://github.com/geosolutions-it/MapStore2/blob/e97942967045fef363cd8b8a36a49a36b950f8a5/web/client/utils/FeatureGridEditorUtils.js#L12) but for the string is a bit different. I wondering if from the UI should be clear when a field is `null` because an empty string it could be a valid input for this type of property, if so I think the issue need to be reviewed and blocked for the moment

allyoucanmap (changes requested):

web/client/components/data/featuregrid/renderers/CellRenderer.jsx
tdipisa: I would like to concentrate the effort on the _boolean_ case for now. Then, through a dedicated issue, we could think on a way to graphically distinguish a _Null_ String from an _Empty_ string and maybe apply that graphic enhancement to all Null fields independently on their type: I think we should dedicate some of our time to evaluate the "how to" better.

mulondo: > @mulondo I think we can't solve the issue by removing the [check for modified values](https://github.com/geosolutions-it/MapStore2/blob/e97942967045fef363cd8b8a36a49a36b950f8a5/web/client/components/data/featuregrid/FeatureGrid.jsx#L78-L81) as you are proposing here because this will remove the highlights also for the one already working.
> 
> @tdipisa I did an investigation on the reason of the highlighting of changed fields even if there are no visual changes from a UI. The reason seems related to String and Boolean property with `null` value (new created feature has all properties set to null). When we are changing the value from the UI for string and boolean we are transforming a `null` value to `""` (an empty string) so this is detected has change and highlighted, even if we are rendering `null` and `""` in the same way inside the grid cell.
> 
> I think we can solve the problem of the boolean value by adding a fallback to null instead of empty string while we are [forcing the selection](https://github.com/geosolutions-it/MapStore2/blob/e97942967045fef363cd8b8a36a49a36b950f8a5/web/client/utils/FeatureGridEditorUtils.js#L12) but for the string is a bit different. I wondering if from the UI should be clear when a field is `null` because an empty string it could be a valid input for this type of property, if so I think the issue need to be reviewed and blocked for the moment

@allyoucanmap I am done with that!

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test this fix in dev, thanks

PR #8950, opened 2023-02-03
Update User Guide - Heading levels error

ElenaGallo:
## Description

Fix heading levels error Measurements section


tdipisa:
@ElenaGallo this PR is not correct. Please fix the precise problem without providing same contests already merged in a new clean PR.

Issue #9016, opened 2023-03-15
The edit catalog modal remains open when changing maps


ElenaGallo:
## Description

The edit catalog modal remains open when changing maps

## How to reproduce

-  Follow the steps in the video below:

https://user-images.githubusercontent.com/56537133/225263177-7545adce-dff2-4a44-807e-5b52439b80f8.mov

## Other useful information
It is also present on master

dsuren1:
**Investigation**:
This is caused by catalog `mode` not being reset when navigating to home page. Currently the mode is reset when closing the panel.

Fix is to reset catalog **mode** `onChangeCatalogMode("view")`  when the `CatalogServiceEditor` component is unmounted.

tdipisa:
Can it be a regression introduced recently @dsuren1 ?

dsuren1:
It's not a regression as it also exists in previous versions (ex: tested on v2022.02.02, v2022.02.00)

PR #9028, opened 2023-03-16
#9016: Fix - Catalog panel mode reset on changing map

dsuren1:
## Description
This PR fixes the catalog mode not reset when changing map

 - [x] Bugfix

## Issue
- #9016 

**What is the new behavior?**
The catalog mode is reset to view when the component is unmounted. i.e the mode set to view when changing map

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/catalog/CatalogServiceEditor.jsx
+++ b/web/client/components/catalog/CatalogServiceEditor.jsx
@@ -5,7 +5,7 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import React, {useState} from 'react';
+import React, {useState, useEffect} from 'react';
 import Spinner from "react-spinkit";
 
 import { FormGroup, Form, Col } from "react-bootstrap";
@@ -52,6 +52,9 @@ export default ({
     autoSetVisibilityLimits = false
 }) => {
     const [valid, setValid] = useState(true);
+    useEffect(() => {
+        return () => onChangeCatalogMode("view");
+    }, []);
     return (<BorderLayout
         bodyClassName="ms2-border-layout-body catalog"
         header={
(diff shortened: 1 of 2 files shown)


MV88 (approved):
LGTM

dsuren1:
@ElenaGallo Kindly test it in dev and let us know if it can be backported. Thanks!

dsuren1:
@ElenaGallo 
The fix address only the edit mode of the catalog. 
It seems the scenario reported above happen ever before this fix, example I have tried this is while being in _view_ mode in **QA** 

record


It caused by previous results not cleared, and hence when opening a catalog with selected service different from previous map **ex**: `map 1 -> catalog -> load wms service` now load `map 2 -> catalog and default selected service is csw` so trying to parse records of _wms_ with _csw_, results in invalid results

I can try to fix this in this issue itself. Fix follows a similar approach used this PR. Instead of resetting the view mode, we can dispatch a catalog close action on plugin unload which should trigger a side effect to reset mode and the results.

@tdipisa Kindly let me know if I can proceed with this

tdipisa:
@ElenaGallo the problem happens also in QA indeed.
@dsuren1 what you identified in QA is correct. Anyway, it seems to me the behavior in QA is a bit different than in DEV. Please retry in QA by following the exact steps of @ElenaGallo. 

In any case the core problem and its origin are the same I think in both DEV and QA and we should provide a fix through a dedicated issue. @ElenaGallo please create the new issue and assign it to @dsuren1 for 2023.01.01.

ElenaGallo:
@dsuren1 the new issue is #9036. Thanks

dsuren1:
> I can try to fix this in this issue itself. Fix follows a similar approach used this PR. Instead of resetting the view mode, we can dispatch a catalog close action on plugin unmount which should trigger a side effect to reset mode and the results eventually

@tdipisa 
Sorry I was away and couldn't reply sooner.
Actually we don't need this PR to be backported as applying what stated in this comment should be enough to fix both the catalog mode and invalid catalog result issues. When I mean it fixes catalog mode, I might need to remove the code applied in this PR and add a code to the plugin directly. So backporting this PR is pointless. May we add reference to the new issue instead?

tdipisa:
@dsuren1 

>Actually we don't need this PR to be backported as applying what stated in this comment should be enough to fix both the catalog mode and invalid catalog result issues. When I mean it fixes catalog mode, I might need to remove the code applied in this PR and add a code to the plugin directly. 

ok

>May we add reference to the new issue instead?

ok

Update this PR then and connect it to both issues

dsuren1:
**Note:** 
This PR need not be backported instead it will be added as part of #9036

Issue #9099, opened 2023-04-14
Ability to define aliases for feature attributes


tdipisa:
## Description

This issue is aimed to define a specific support in MS for providing aliases for features attributes. Doing that is obviously possible but it is necessarily to point out the following aspects that need to be considered involved constraints/requirements for this new feature:

1.  The change cannot be done at Attribute Table level because we have to keep consistency across different tools that manage/visualize somehow the feature attributes: Visual Style Editor, Filter Layer tool, Identify tool etc. This means that aliases for feature attributes need to be controlled/centralized in Layers Settings

2. For transparency it is necessary to let the user always know the real attribute name in certain cases. Therefore, as it happen in QGIS for example, the real attribute name can appear in brackets (only inside a tooltip) when you move the mouse over the alias in the Attribute Table header
 
3. Since we are going to control aliases (we are not changing the real attribute name from MapStore), the real attribute name can be obviously used where needed (eg. OGC filters, SLDs, other OGC requests involving charts/widgets) in a transparent way. This means that from UI the user can see/select the alias name while the real name is used for final requests to the remote services

4. The following constrains anyway exist:
   - MapStore cannot have control on Identify formats other than application/json for which it is possible to intervene on the output programmatically to show the alias instead of the real attribute name
   - Inside the Code Style Editor (the style text area) the real attribute name need to be anyway used

5. From Layer Settings can be possible to set aliases to real attributes name through a kind of tabular view where a reload button allow to reload the Feature Type schema on demand (eg. if an attribute name is changed server side) so that the user can set again the alias when needed

6. What already implemented for widgets (https://github.com/geosolutions-it/MapStore2/issues/8559) can be maintained. The widget can inherit aliases defined at Layer Settings level giving the possibility to further change them at widget level. 

7. Each map will have its own setting of aliases

```[tasklist]
### Tasks
- [ ] https://github.com/geosolutions-it/MapStore2/issues/9120
- [ ] https://github.com/geosolutions-it/MapStore2/issues/9122
- [ ] https://github.com/geosolutions-it/MapStore2/issues/9129
- [ ] https://github.com/geosolutions-it/MapStore2/issues/9127
- [ ] https://github.com/geosolutions-it/MapStore2/issues/9128
- [ ] https://github.com/geosolutions-it/MapStore2/issues/9136
- [ ] https://github.com/geosolutions-it/MapStore2/issues/9146
```

## Acceptance criteria

[...]

PR #9125, opened 2023-04-24
Implementation of alias editor and usage in feature grid

offtherailz:
## Description

The current PR provides

- A small refactor of the layer setting panel (for clean up organization)
- An new tab in layer settings, with Fields configuration. It provides the possibility to edit the aliases (optionally localized) #9127 
- The implementation of usage of Alias for the feature grid (part of #9128 )

@allyoucanmap : For the moment, the 3 features above are provided in 3 different commits, for an easier review, hope it helps to have a clear overview of the changes applied

The other implementations for #9099 (Visual Style Editor, Query Panel, Filter Layer tool, Identify, Widgets) are still pending. 
We can evaluate if merge this and then the other implementations as separate pull requests or continue the work on this branch and then merge all the work together.

Anyway I can suggest to start reviewing this work from now, so we can have less files to review at a single time. 

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#<issue>

**What is the new behavior?**

## Breaking change
 - [x] No


offtherailz:
A suggestion from @allyoucanmap to check, is to remove the shadow near the tabs, to have a better visual effect.

image

I'll try to give you the result and check if it works also for the other tabs ( it was already there for some reason I suppose)

web/client/components/TOC/fragments/LayerFields/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+        setError();
+        setLoading(true);
+        loadFields(layer, merge)
+            .then((newFields) => {
+                setLoading(false);
+                setFields(newFields);
+            })
+            .catch((e) => {
+                setLoading(false);
+                setError(e);
```
allyoucanmap, line 74: I think we should check if the component is still mounted to avoid to trigger internal state while the component is not mounted anymore

allyoucanmap (commented):
The code seems good, just some things to check:

- alias is not currently save in the exported or saved map
- should we allow to edit the name of the geometry field?
- about the style here the result without shadows:

image

tdipisa:
> The code seems good, just some things to check:
> 
> * alias is not currently save in the exported or saved map

@offtherailz correct me if I'm wrong but this is part of a dedicated issue mentioned in the task list of [#9099](https://github.com/geosolutions-it/MapStore2/issues/9099#issue-1668414178) that still need to be defined

> * should we allow to edit the name of the geometry field?

I think we can omit the geometry field which is never used in the UI to display its value (eg. in Attribute Table or Identify or other tools)

offtherailz:
- Style

@allyoucanmap can you provide the style changes you used or apply them direcly on the branch?

- Saving

```
correct me if I'm wrong but this is part of a dedicated issue mentioned in the task list of https://github.com/geosolutions-it/MapStore2/issues/9099#issue-1668414178 that still need to be defined
```
Yes you are right, it is part of a dedicated task in the task list, anyway it makes sense to include it in the current PR for consistency. So it is ok. 

```
hide the gemetry
```
 ok i can exclude it from the UI

offtherailz:
@allyoucanmap I fixed the parts you asked. 
Notes: 
- I avoided to remove the shadow from the top it is applied on `ms-header`, and when I tried to customize this component of TOCItemsSettings I had this result.

image

- I added a light border to better separate header and footer of the scrollable content from the content itself.

allyoucanmap (approved):

offtherailz:

ElenaGallo:
Hi @offtherailz, I noticed that:

- the shadow under the settings tabs is still present
shadow

- when Alias is removed, in the attribute table it is not replaced with the name of the attribute but the empty space is left
(only by using the 'Clear customization' button does the attribute name become visible again in the attribute table)
https://github.com/geosolutions-it/MapStore2/assets/56537133/64b23c0b-42b7-4093-bc5d-14231793db77

PR #9131, opened 2023-04-26
Update changelog and versions

offtherailz:
## Description

Update changelog and versions for branch 2023.01.01

Part of #9105


tdipisa (changes requested):
@offtherailz I think we should also include this
https://github.com/geosolutions-it/MapStore2/blob/2023.01.xx/project/standard/templates/web/pom.xml

tdipisa (approved):

PR #9174, opened 2023-05-16
Fix #9172 add flag to always select first layer in toc in identify results

### LINKED ISSUE #9172: Make first layer in TOC be always selected in identify container
## Description

ID 99 – Informations attributaire

Currently, when in MapStore the user clicks on a map containing several layers, if on the clicked point there are several objects from different layers underneath, the information window displays the list of those layers with an order which does not always correspond to the layer order in the map, with the most visible layer on the top.

The requirement is to allow the display order of the layers in the information window as it is in the map.

_In the example below, we can see that information has been found on 3 different layers, by default, the information displayed when the attribute information window is opened relates to the “Petite centrale hydroélectrique” layer. If the order of the layers had been taken into account, the window would have opened on the "Ouvrages principaux" layer because as we can see, the "Ouvrages principaux" layer is located first in relation to the "Small hydroelectric power plant" layer._ 

Image

## Technical solution

The plugin can be embedded as enhancement of the main MapStore project.

The current MapStore behavior considers the performance of the application as the main goal to be achieved, selecting the layer that responded fastest from the drop-down list. The order of layers in the drop-down is already the same as in the TOC.

The plugin will be carried out to allow the requested behavior. To allow both the current and the requested layer order mode to be used, a new option from plugin configuration will be added, to be managed by administrator users. No UI changes are expected for this task.

- [x] Minor changes to existing features
- [ ] Build related changes
- [ ] CI related changes

MV88:
## Description

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#9172

**What is the new behavior?**

when ignorePerformances is set to true this value is edited from cfg, then the first request/layer is selected no matter the validity/presence of the data in the response.

and all layers are listed in the layer selector field

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/identify/IdentifyContainer.jsx
+++ b/web/client/components/data/identify/IdentifyContainer.jsx
@@ -38,7 +38,7 @@ export default props => {
         onClose = () => {},
         responses = [],
         index,
-        ignorePerformances,
+        showAllResponses,
         viewerOptions = {},
         format,
         dock = true,
@@ -134,7 +134,7 @@ export default props => {
                         <LayerSelector
                             responses={responses}
                             index={index}
-                            ignorePerformances={ignorePerformances}
+                            showAllResponses={showAllResponses}
                             loaded={loaded}
                             setIndex={setIndex}
                             missingResponses={missingResponses}
--- a/web/client/components/data/identify/IdentifyContainer.jsx
+++ b/web/client/components/data/identify/IdentifyContainer.jsx
@@ -38,6 +38,7 @@ export default props => {
         onClose = () => {},
         responses = [],
         index,
+        ignorePerformances,
         viewerOptions = {},
         format,
         dock = true,
@@ -133,6 +134,7 @@ export default props => {
                         <LayerSelector
                             responses={responses}
                             index={index}
+                            ignorePerformances={ignorePerformances}
                             loaded={loaded}
                             setIndex={setIndex}
                             missingResponses={missingResponses}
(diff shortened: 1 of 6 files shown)


allyoucanmap (commented):
@MV88 @tdipisa The implementation works fine my only concern is about the name of the cfg proposed here. We are not changing the way we perform the requests so I don't see performances changes but it seem we are forcing to visualize all the available layers even if there is no response. My proposal is to change the name `ignorePerformances` into `displayAllResponses`, `showAllResponses`, `displayAllVisibleLayersResponses`, `includeInvalidResponses` or something more focused on the fact we are listing all the layer in order.

Note: If a layer is not visible is not included in the list, is this expected?

MV88:
> @MV88 @tdipisa The implementation works fine my only concern is about the name of the cfg proposed here. We are not changing the way we perform the requests so I don't see performances changes but it seem we are forcing to visualize all the available layers even if there is no response. My proposal is to change the name `ignorePerformances` into `displayAllResponses`, `showAllResponses`, `displayAllVisibleLayersResponses`, `includeInvalidResponses` or something more focused on the fact we are listing all the layer in order.
> 
> Note: If a layer is not visible is not included in the list, is this expected?

I knew this name was a point of a discussion and I did not changed it on purpose. 
The reason I called it ignorePerformances is because you do not select the first response receive but always the first layer in the list and by doing that we also allow invalid/empty layers to be selected (ignoring invisible layers for which the feature info request is not being sent).

Anyway **showAllResponses** should be good enough, I'll update PR accordingly

allyoucanmap (approved):

MV88:

PR #9184, opened 2023-05-18
Fix #9025 WMS caching with custom scales (custom resolutions strategy from WMTS)

### LINKED ISSUE #9025: WMS caching with custom scales
## Description

GeoServer allows to use tile cache for WMS using the cache generated by GWC. If a request in WMS is aligned to one of the gridsets defined in GeoServer for the given layer/projection, than the tile is returned. 

MapStore partially allows to use tile cache for WMS by generating a gridset aligned to the map resolutions (exponential by default) and using as origin (top-left corner) as the origin of the extent of the current projection.

This allows to match the gridsets of GeoServer only if the scales/resolutions of  MapStore/Tileset are exponential (default for both MapStore and GeoServer).

Anyway, when defining custom scales for the map, this reflects on the layer and aligning the 2 tilesets becomes impossible. 

Even defining a custom tileset the on GeoServer with the same scales do not work

This because the fixed-origin used by GeoServer to generate the tilegrid starts from the bottom left corner, while Openlayers starts from the top-left corner. While in exponential resolutions there was no problem (because we have exactly 1-tile - 2-tiles - 4-tiles -8tiles on the Y axis), and so the top-left origin will be exactly at the top-left corner of a tile, in the case of custom scales/resolutions the top-left corner vary for each zoom level, so having only one origin is not sufficient anymore. 

## Proposed solution

There are several use cases: 

0. Map and GeoServer with exponential resolutions (same origin)
1. Map with custom resolutions, standard exponential resolution on GeoServer (generated by default)
2. Map with custom resolutions, custom resolutions on GeoServer.

Openlayers allows to have a set of resolutions different from the current scales of the map (leaflet do not allow this). Openlayers gets automatically the nearest resolutions to render, scaling the tiles accordingly. 

For this reason I propose to detouch the tilegrid resolutions from the map resolutions. 

We should generate resolutions for the layer by default exponentially, checking the map resolutions only to calculate automatically when to stop. (to fit the minmum the resolutions of the map) *this strategy need to be investigated more*

This allows to generate resolutions on GeoServer exponentially, too and they should always match by default. 

If we want a better quality and corrispondence between the map scales on the map and the ones on the server, we can allow from layer settings to customize tilegrid settings, Customizing origin(s) and resolutions of the gridset) sizes, tileSizes.

This solution offers:
    - The same behivior by the default  
    - Fast configurability and caching also when using custom scales
    - The possibility to customize the gridset for fine work

```[tasklist]
### Tasks
- [ ] Align default projection EPSG:4326, EPSG:3857 extent definitions to the GeoServer one. Allow to overrides extent of default projections from localConfig.json inside the projectioDefinitions property.
- [ ] Review the default tile grid created for the WMS layers in the OpenLayers library using the MapUtils [getResolutionsForProjection](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/MapUtils.js#L218)
- [ ] Allow to use a WMTS strategy to create the TileGrid for WMS layer 
```

## Acceptance criteria

- [ ] Given MapStore uses custom scales for a map and GeoServer provides an exponential tilegrid, tile cache hits on every zoom level for WMS. 
- [ ] As a user, I can use the cache of a custom gridset on GeoServer. Getting the origins from WMTS service and configuring them in MapStore, together with the resolutions, I'm able to hit the GWC cache for WMS also in this case. 

## Other useful information

WMS tile caching in GeoServer is based on TMS standard, starting from the bottom left corner of the gridset extension, keeping the origin fixed for all the resolutions configured. 
WMTS instead uses the top-left corner as reference, but allows to define an origin coordinate from all the scales.

allyoucanmap:
## Description

This PR adds the possibility to apply custom tile grids to WMS layer through a request to a WMTS endpoint generated from a GeoServer WMS layer url. The PR includes following changes:

- add logic to WMS Layer for the OpenLayer implementation to support custom tileGrids
- add two new properties to wms layer configuration: `tileGrids` and `tileGridStrategy`
- add two new icons: `grid-regular` and `grid-custom`
- add buttons and information inline with the WMS cache options setting to toggle the custom strategy

UI Changes:

- Button to enable custom strategy

image

- Info icon shows information about tile grids and highlight the one in use (automatically selected)

image

- Info icon shows information about tile grids with warning if it's not possible to find a tile grid that matches the map projection and selected tile size

image

- warning when there isn't a tile grid available that matches the available projections in MapStore

image

- error message when is not possible to connect to the WMTS endpoint generated from the WMS layer

image

 - [x] Feature

## Issue

#9025

**What is the new behavior?**

It is possible to apply custom grid set configured server side to GeoServer WMS layers

## Breaking change

 - [x] No


web/client/translations/data.it-IT.json
```diff
@@ -118,6 +118,17 @@
                 "noVendor": "No Vendor",
                 "geoserver": "GeoServer"
             },
+            "availableTileGrids": "Tile grids disponibili",
+            "crsId": "Id",
+            "projection": "Proiezione",
+            "tileSize": "Dimensione tile",
+            "tileGridInUse": "Il tile grid in uso è {id}",
+            "disableCustomTileStrategy": "Disabilita la strategia personalizzata per la tile grid",
+            "enableCustomTileStrategy": "Abilita la strategia personalizzata per la tile grid",
```
offtherailz, line 127: Pay attention to the translation, custom tile grids, not custom strategy. This problem should go away when talking about fetching tile grids instead of enabling/disabling strategy.

offtherailz (changes requested):
- [ ] I noticed that also the format and the styles may be provided. In fact if format or style do not match, you have a green light, but the tile cache misses with  message "not a tile cache for requested format" ( I suppose the same for style )
The WMTS provides this list:
```
<Format>image/png</Format>
<Format>image/jpeg</Format>
<Format>image/png8</Format>
```
So maybe we canevaluate to check also this, what do you think? 

Instead style may be more complicated, you can evaluate to skip this checks. 

- [ ] I suggested to simplify the messages, avoiding to talk about strategies. More information can stay in user documentation, I think. Take into accoun that if you consider also formats and/or styles, the strings change.

@tdipisa we we should not forget in doc to mention that 
- MapStore automatically follow the same algorithm of GeoServer to generate the tile grids and use the cache, so most of the cases creating the tile grids with default bounding box and resolution for the tilesizes and the projections used is enough to activate caching. 
- In case you are using custom tilegrids, activating the button will fetch the tilegrids and MapStore will automatically select the best tilegrid if available.

tdipisa:
>@tdipisa we we should not forget in doc to mention that 

@ElenaGallo remember to open an issue for the update of the user guide.

web/client/api/WMTS.js
```diff
@@ -158,7 +158,9 @@ export const getLayerTileMatrixSetsInfo = (url, layerName = '', options) => {
             return {
                 tileMatrixSets,
                 tileMatrixSetLinks,
-                tileGrids
+                tileGrids,
+                styles: castArray(wmtsLayer?.Style || []).map((style) => style['ows:Identifier']),
+                formats: castArray(wmtsLayer?.Format || [])
```
offtherailz, line 163: Update JSDoc with formats

offtherailz (changes requested):
- [ ] bug: Can't turn off the custom tile grid anymore once activated: 

[custom_tile_grid.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/05b479cc-aab1-4160-b166-36fa949ab828)

- [ ] One inline comment about JSDoc to update. 
- [ ] Showing also the list of formats supported by the tile grid is a plus

About user experience, one thing I noticed: 

Now we have a feedback of cache match only in case of custom gridset, but we do not have anything in case of standard gridset, that looks a little strange
It could make sense to check if the default strategy meets the remote gridset. Maybe the refresh button may be a check button that tells if the default strategy fits. 

A more rational workflow could tell the user (if clicks on a proper button) if the standard setup works or no, and if it doesn't, he can decide to use a custom on. A workflow like this looks more rational to me, instead of having 
Use cache with no feedback, and then turn on custom and see that it works or it doesn't.

web/client/translations/data.it-IT.json
```diff
@@ -118,6 +118,20 @@
                 "noVendor": "No Vendor",
                 "geoserver": "GeoServer"
             },
+            "availableTileGrids": "Tile grids disponibili",
+            "crsId": "Id",
+            "projection": "Proiezione",
+            "tileSize": "Dimensione tile",
+            "tileGridInUse": "La tile grid '{id}' corrisponde alla proiezione della mappa corrente e alla dimensione della tila selezionata per questo layer",
```
offtherailz, line 125:
```suggestion
            "tileGridInUse": "La tile grid '{id}' corrisponde alla proiezione della mappa corrente e alla dimensione delle tile selezionate per questo layer",
```

allyoucanmap:

ElenaGallo:
@allyoucanmap from dashboard and geostory settings panel, the "available tile grids" popup always report that it's not possible to find a tile grid that matches the map projection and selected tile size.
(Use [this dashboard](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/dashboard/44358))


https://github.com/geosolutions-it/MapStore2/assets/56537133/4a9358e0-74a4-450b-8c9f-ced62440e978

PR #9196, opened 2023-05-26
Fix #9025 Available tile grids popup always reports mismatch in geostories and dashboards

### LINKED ISSUE #9025: WMS caching with custom scales

allyoucanmap:
## Description

This PR fixes a bug reported in this comment https://github.com/geosolutions-it/MapStore2/pull/9184#issuecomment-1564053808

 - [x] Bugfix

## Issue

#9025

**What is the new behavior?**

The WMS cache info message reports correctly the state of the matching with remote configuration

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/widgets/builder/wizard/map/NodeEditor.jsx
+++ b/web/client/components/widgets/builder/wizard/map/NodeEditor.jsx
@@ -12,6 +12,7 @@ import { NavItem as BSNavItem, Col, Glyphicon, Nav, Row } from 'react-bootstrap'
 
 import Message from '../../../../I18N/Message';
 import tooltip from '../../../../misc/enhancers/tooltip';
+import { VisualizationModes } from '../../../../../utils/MapTypeUtils';
 
 const NavItem = tooltip(BSNavItem);
 
@@ -48,6 +49,8 @@ export default ({
                     settings={settings}
                     retrieveLayerData={onRetrieveLayerData}
                     isLocalizedLayerStylesEnabled={isLocalizedLayerStylesEnabled}
+                    projection={props.map?.projection}
+                    isCesiumActive={props.map?.visualizationMode === VisualizationModes._3D}
                     onChange={(key, value) => isObject(key) ? onUpdateParams(key, realtimeUpdate) : onUpdateParams({ [key]: value }, realtimeUpdate)} />
             ))}</Col>
     </Row>);
--- a/web/client/components/widgets/builder/wizard/map/__tests__/NodeEditor-test.jsx
+++ b/web/client/components/widgets/builder/wizard/map/__tests__/NodeEditor-test.jsx
@@ -10,8 +10,11 @@ import expect from 'expect';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {createSink} from 'recompose';
-
+import { Simulate } from 'react-dom/test-utils';
+import { VisualizationModes } from '../../../../../../utils/MapTypeUtils';
 import NodeEditor from '../NodeEditor';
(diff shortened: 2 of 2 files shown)


offtherailz (approved):

tdipisa:

Issue #9179, opened 2023-05-17
Include editing support to allowed user groups 


tdipisa:
## Description

As discussed and agreed with #7636 it is required to allow certain editing functionalities depending on the groups the user belongs to. 
At the moment the _editingAllowedRoles_ configuration property is available for [FeatureEditor](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.FeatureEditor) and [StyleEditor](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.StyleEditor) tools to allow users to use the editing functionality depending on their role (ADMIN or USER) by disabling the pencil icon if the user belongs to a role not granted.

**It is necessary to extend this system to allow editing functions to users also according to the groups they belong to.**

## Acceptance criteria

- [x] The support for a new configuration property named _editingAllowedGroups_ will be included
- [x] That new property can work together with the existing one. For example:

    ```
    "editingAllowedRoles": [
          "ADMIN",
          "USERS"
     ],
    "editingAllowedGroups": [
          "Group 1",
          "Group 2",
          "Group 3",
          ...
     ],
     ```
- [x] That new property will be supported everywhere the existing _editingAllowedRoles_ is managed in the MS code base (that means for both [FeatureEditor](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.FeatureEditor) and [StyleEditor](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.StyleEditor))
- [x] Both properties should be probably prioritized somehow to properly catch special cases. For example an ADMIN user should be always allowed independently of the group he belongs to. Therefore _editingAllowedGroups_ should be applied only to users with USER role 
[...]

PR #9210, opened 2023-06-05
#9179: Include editing support to allowed user groups

dsuren1:
## Description
This PR adds support to configure `editingAllowedGroups` to `FeatureEditor` and `StyleEditor` plugin

 - [x] Feature

## Issue

- #9179 

**What is the new behavior?**
The edit permission configuration follows the below hierarchy (Applicable for `FeatureEditor` and `StyleEditor` plugin)
1. Allow all user to edit, when editingAllowedRoles has "ALL" configured in it
2. Allow edit, when user role and group belongs to either `editingAllowedRoles` or `editingAllowedGroups` configured for the plugin

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/FeatureEditor.jsx
+++ b/web/client/plugins/FeatureEditor.jsx
@@ -49,12 +49,11 @@ import {isViewportFilterActive} from "../selectors/featuregrid";
   *    }]
   *}
   * ```
-  * @prop {object} cfg.editingAllowedRoles array of user roles allowed to enter in edit mode
-  * @prop {object} cfg.editingAllowedGroups array of user groups allowed to enter in edit mode.
-  * When configured, the property is applicable only when logged-in user's role is other than ADMIN
-  * User edit permission follows the below hierarchy
-  * 1. User's role is ADMIN
-  * 2. User's role is NON-ADMIN, then user's group falls in one of the configured allowed groups
+  * @prop {string[]} cfg.editingAllowedRoles array of user roles allowed to enter in edit mode.
+  * Support predefined ('ADMIN', 'USER', 'ALL') and custom roles. Default value is ['ADMIN'].
+  * Configuring with ["ALL"] allows all users to have access regardless of user's permission.
+  * @prop {string[]} cfg.editingAllowedGroups array of user groups allowed to enter in edit mode.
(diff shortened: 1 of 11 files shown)


offtherailz (changes requested):
Hi @dsuren1 . I tryed to suggest all the changes inline (initially only some review of code base to clarify it more)  but I noticed during testing that one of the things didn't worked as expected.

In particular there is no way of allowing all logged users to be able to edit at the same time, without indicating all the groups of the system.This because the logic of the editing allowed selector for groups requires that the role is in the list.  

I suggested to change this logic in the main selectorCreator used but you may need to test and check it more (and fix the unit tests). 

We should handle and prioritize permissions this way: 

- "editingAllowedRoles": ["ADMIN"], editingAllowedGroups: ["geosolutions"] means all admins and members of geosolutions group.
- "editingAllowedRoles": ["ADMIN", "USER"] means - all logged users (admins or users) so groups in this case are not important. 
So: 
- [x] Please check if the logic i suggested works and the documentation fits this requirement, fixing unit tests 
- [x] I wasn't able to test the styler. Please make you sure that this logic is correctly applied to the styler too.

dsuren1:
Hi @offtherailz 
Thanks for the feedback.

> In particular there is no way of allowing all logged users to be able to edit at the same time, without indicating all the groups of the system.This because the logic of the editing allowed selector for groups requires that the role is in the list.
> 
> I suggested to change this logic in the main selectorCreator used but you may need to test and check it more (and fix the unit tests).
> 
> We should handle and prioritize permissions this way:
> 
> * "editingAllowedRoles": ["ADMIN"], editingAllowedGroups: ["geosolutions"] means all admins and members of geosolutions group.
> * "editingAllowedRoles": ["ADMIN", "USER"] means - all logged users (admins or users) so groups in this case are not 

This deviates from the requirement, the requirement was to allow ADMIN regardless of group the user belongs, and consider the group only when user is NON-ADMIN. For NON-ADMIN, default allowedGroups is `everyone`. With the current implementation, this configuration `editingAllowedRoles: ["ADMIN", "USER"]` will work for all the users of Mapstore.

tdipisa:
@ElenaGallo please test this in DEV taking into account the comment above.

Issue #9180, opened 2023-05-17
Attribute Table in read only mode


tdipisa:
## Description

As discussed and agreed with #7285 it is necessary to have the Attribute Table in read-only as needed for certain layers in TOC.
It is required to have the ability to make the Attribute Table read-only per layer if desired (eg. because a layer has been added to a map for consultation purposes only and not to be edited). For these layers the editing mode of the [Attribute Table](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.FeatureEditor) should be disabled at all. 

## Acceptance criteria

- [x] A new configuration property (of type boolean) will be added for WMS layers in [Layer Settings](https://docs.mapstore.geosolutionsgroup.com/en/v2023.01.01/user-guide/layer-settings/) as indicated [here](https://github.com/geosolutions-it/MapStore2/issues/7285#issuecomment-1549532934)
- [x] If checked (and so True) the editing mode for the Attribute Table will be disabled for the layer (eg. pencil icon not abailable)
- [x] The editing functionality of the Attribute Table ([plugins.FeatureEditor](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.FeatureEditor)) can be reached also from the [Identify](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.Identify) tool if its `cfg.showEdit` property is set to _true_. That case should be managed accordingly (eg. the pencil icon in Identify panel not available) according the value of the new property in Layer Settings
- [x] As indicated [here](https://github.com/geosolutions-it/MapStore2/issues/7285#issuecomment-1425520670) the context editor (who has write permission on the context) can control that option from layers settings of layers added in the context (step 2 of the wizard). The map editor (who has write permission on a map) can do the same at map level. **The others cannot control that option and it should not be visible.**

## Other useful information
See also #9179

tdipisa:
@dsuren1, @ElenaGallo, @offtherailz I have to reopen this since you missed to consider one relevant point I've not included by mistake in the AC list of this issue but it present in the original issue #7285. I've updated the AC list above.

dsuren1:
Hi @tdipisa 
I will update the code to show this option only if user has write permission to map and context.

PR #9227, opened 2023-06-16
#9180: Layer setting - Feature editing option with permission

dsuren1:
## Description
This PR adds validation to layer setting `disableFeaturesEditing` option. The option is shown only if the user has write permission to map/context.

 - [x] Feature

## Issue

- https://github.com/geosolutions-it/MapStore2/issues/9180#issuecomment-1593179308

**What is the new behavior?**
The option is visible only when user has write permission to map/context

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/TOC/fragments/settings/General.jsx
+++ b/web/client/components/TOC/fragments/settings/General.jsx
@@ -40,7 +40,8 @@ class General extends React.Component {
         showTooltipOptions: PropTypes.bool,
         allowNew: PropTypes.bool,
         enableLayerNameEditFeedback: PropTypes.bool,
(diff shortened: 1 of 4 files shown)


offtherailz (changes requested):
This do not work 
- [ ] In context manager map
- [ ] In new map

I think you should use the same logic of mapSave: 
use mapInfoSelector if the id is present use canEdit value, otherwise, it is allowed.

something like  

```
const canEditFeatures = () => {
  const {id, canEdit} = mapInfoSelector(state) 
  return id ? canEdit : true
}

Please check if this works on new map and conext editor map too

offtherailz (approved):

PR #9245, opened 2023-06-21
[Release 2023.01.02] Fixed versions for backend modules

offtherailz:
[Release 2023.01.02] Fixed versions for backend modules


tdipisa (changes requested):
what about package.json and printing [here](https://github.com/geosolutions-it/MapStore2/blob/2023.01.xx/product/pom.xml#L351)?
Maybe it is useful for the next times to have a precise list of files locations where the version need to be changed

tdipisa (approved):

Issue #9241, opened 2023-06-21
Longitudinal profile plugin


MV88:
## Acceptance criteria

- exploit in Mapstore the altitude or depth data present in digital terrain models or bathymetric surveys

INPUT

- [ ] Height data source 
- [ ] linear profile: Three(3) options
-   [ ] Drawing it directly on the map;
-   [ ] Importing it as JSON selected from file system
-   [ ] Importing it as ESRI Shapefile, selected from file system (compressed zip containing shp, shx, dbf and prj files)
-   [ ] Importing it as DXF file, selected from file system, and related coordinate system to be used (from a list)
- [ ] Title of the graphic profile
- [ ] Pitch used (distance): Maximum distance between 2 points along the profile, in meters)
- [ ] After drawing the linear profile or importing it as a file (shp or dxf), the module offers to display and export the profile produced in several ways
- [ ] ADD Option to use bathymetric data sources as height data input, in addition to DTMs from IGN (and lidar)

OUTPUT

- [ ] Graph display
-   [ ] The graph is interactive. Altitude and distance traveled values are dynamically displayed on hover
-   [ ] The x-axis indicates the distance traveled from the starting point of the drawn or imported linear
-   [ ] The y-axis indicates the altitude returned by the height source at the coordinates along the profile.
-   [ ] The y-axis is initialized at 10 m below the minimum altitude encountered on the profile and 10 m above the maximum altitude encountered on the profile
-   [ ] Information on altimetric data source and altimetric reference is specified.
- [ ] Summary display
-   [ ] Elevation data source used
-   [ ] Total linear length of the profile
-   [ ] Positive elevation
-   [ ] Negative elevation
-   [ ] Number of vertex points on which the altimetry was calculated. The number of stitches varies according to the pitch chosen.

Output data - Export

-   [ ] Image format (PNG) of the graph (only chart or also with other info?)
-   [ ] Tabular format (CSV) of the longitudinal profile, specifying for each coordinate of the profile the Z value in addition to the X and Y values. The distance of each point from the starting point is also mentioned.
-   [ ] Printable format (PDF) of the longitudinal profile. The PDF document should include the graphic information as well as the extract from the map allowing you to visualize the line of the profile in 2D (whether the latter has been drawn on the map or imported via an ESRI Shapefile or DXF file)

PR #9253, opened 2023-06-26
#9241 long profile tool

MV88:
## Description

This pr introduce a new plugin called longitudinal plugin that works in combination of a new wps process gs:LongitudinalProfile  that can be found [here](https://github.com/geosolutions-it/MapStore-C265/issues/8#issuecomment-1587246044)

to review this locally i suggest to install that jar in a local geoserver proxying to it and publish sfdem layer too,, you can found it [here](https://github.com/geosolutions-it/MapStore-C265/issues/8#issuecomment-1581174935) 

also by default this is not enabled in main viewer and
can only added in a context

it should also not conflict with the current similar implementation [here ](https://github.com/georchestra/mapstore2-longitudinal-profile)

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#9241

**What is the new behavior?**

Now you can use the longitudinal profile plugin needs wps on geoserver

## Breaking change
 - [x] No


offtherailz (changes requested):
I wasn't able to test the upload and the 3D extensively. Anyway I found the following issues. 

- [ ] Upload file doesn't work. 

[screencast-localhost_8081-2023.06.27-12_31_25.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/f8a05d07-a8fb-4cd2-b30d-83b9b6a17da8)

- [ ] Having 2 ways to edit settings is unuseful, remove one
image

- [ ] The info tab uses Icons, but it is not clear what are the values.  Add at least some labels here, what are the values shown? For instance my first impression was that the arrows indicate max and min altitude, but they indicate the active and passive differences in altitude.

image

- [ ] "Points" string is not localized (see screen above)

- [x] the marker is not aligned in 3d mode

[screencast-localhost_8081-2023.06.27-12_33_00.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/a2c5e7a3-5fbb-4d0c-bac5-4763c94523bc)

- [ ] Maximize do not fit the window, overlaps on the right 

[screencast-localhost_8081-2023.06.27-12_34_04.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/4b22edd8-c7c3-4ee0-bec3-c8c5d13b6dce)

- [ ] When I draw a profile, the geometry is centered without taking into account the panel displacement. Fix this or otherwise is better to not center the geometry at all.

[screencast-localhost_8081-2023.06.27-12_57_09.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/dc0c9ae9-da0d-4c83-892c-ca76c2baa3a6)

- [ ] When I draw outside, I see there is a process failure, but it is not notified to the user. ( Maybe I missed some notifications?

[screencast-localhost_8081-2023.06.27-13_00_04.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/a42dadba-9e76-42fa-a5f9-cb236f8577b8)

- [ ] You regenerated the icons, but apparently no new icon has been added to the proper folder. Please commit also the new icon or revert your chnages to the iconset.

package.json
```diff
@@ -184,6 +186,7 @@
     "git-revision-webpack-plugin": "5.0.0",
     "history": "4.6.1",
     "html-to-draftjs": "npm:@geosolutions/html-to-draftjs@1.5.1",
+    "html-to-image": "1.11.11",
```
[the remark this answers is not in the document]
offtherailz: Ok. I've seen that effectively html2canvas needs to be removed. So it makes sense to use this instead.

tdipisa:
Here my further feedback on points highlighted above

> * [ ]  Upload file doesn't work.

It was working before, maybe a regression from the previous PR status

> * [ ]  Having 2 ways to edit settings is unuseful, remove one

I can agree. I can provide 2 options for this:
   - simply remove the preferences tab if this doesn't require much time (I think it should be a quick and easy thing)
   - otherwise leave this as it is for now
   
> * [ ]  The info tab uses Icons, but it is not clear what are the values.  Add at least some labels here, what are the values shown? For instance my first impression was that the arrows indicate max and min altitude, but they indicate the active and passive differences in altitude.

I agree. thank you for notifying that @offtherailz 

> * [ ]  "Points" string is not localized (see screen above)

I agree. thank you for notifying that @offtherailz 

> * [ ]  the marker is not aligned in 3d mode

I agree. thank you for notifying that @offtherailz 
It is not a request for now to have a complete support for 3D. Therefore if this requires too much time, simly disable the tool when in 3D mode as I've initially requested 

> * [ ]  Maximize do not fit the window, overlaps on the right
 
I assume you would like to have the same of widgets. In this case I agree with that. fully covering the UI is not so nice.

> * [ ]  When I draw a profile, the geometry is centered without taking into account the panel displacement. Fix this or otherwise is better to not center the geometry at all.

I agree.

> * [ ]  When I draw outside, I see there is a process failure, but it is not notified to the user. ( Maybe I missed some notifications?
 
This is something we have to definitely handle well. In these cases (and this also happens for example if you load a DXF file specifying a wrong projection) the WPS return with an exception like this one:

```
<?xml version="1.0" encoding="UTF-8"?>
<wps:ExecuteResponse xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:xlink="http://www.w3.org/1999/xlink" xml:lang="en" service="WPS" serviceInstance="http://localhost:8080/geoserver/ows?" version="1.0.0">
	<wps:Process wps:processVersion="1.0.0">
		<ows:Identifier>gs:LongitudinalProfile</ows:Identifier>
		<ows:Title>Longitudinal Profile Process</ows:Title>
		<ows:Abstract>The process splits provided linestring to segments, that are no bigger then distance parameter, then evaluates altitude for each point and builds longitudinal profile. Altitude will be adjusted if adjustment layer is provided as parameter. Also supports reprojection to different crs</ows:Abstract>
	</wps:Process>
	<wps:Status creationTime="2023-06-27T12:46:53.927Z">
		<wps:ProcessFailed>
			<ows:ExceptionReport version="1.1.0">
				<ows:Exception exceptionCode="NoApplicableCode">
					<ows:ExceptionText>Process failed during execution
org.opengis.coverage.PointOutsideCoverageException: Coordinate (612.658,086, 4.915.061,527) is outside coverage.
Coordinate (612.658,086, 4.915.061,527) is outside coverage.</ows:ExceptionText>
				</ows:Exception>
			</ows:ExceptionReport>
		</wps:ProcessFailed>
	</wps:Status>
</wps:ExecuteResponse>
```

We have to properly handle the _ows:ExceptionReport_ and notify the user with a popup message in the middle of the screen as we usually do for other plugins.
  
> * [ ]  You regenerated the icons, but apparently no new icon has been added to the proper folder. Please commit also the new icon or revert your chnages to the iconset.

I've nothing to say in this regard.

MV88:
> The info tab uses Icons, but it is not clear what are the values. Add at least some labels here, what are the values shown? For instance my first impression was that the arrows indicate max and min altitude, but they indicate the active and passive differences in altitude.

I'll with third option, but i had prefer something different
image

> Upload file doesn't work.

i was able to replicate, you have to draw and then try to import

> >  Having 2 ways to edit settings is unuseful, remove one
> I can agree. I can provide 2 options for this:
simply remove the preferences tab if this doesn't require much time (I think it should be a quick and easy thing)
otherwise leave this as it is for now

going to remove a Tab

> You regenerated the icons, but apparently no new icon has been added to the proper folder. Please commit also the new icon or revert your changes to the iconset.

yeah i'll try to restore original files


other reports are fixed

web/client/translations/data.it-IT.json
```diff
@@ -3714,6 +3718,70 @@
...
+          "source": "Sorgente",
+          "downloadCSV": "CSV",
+          "downloadPNG": "PNG",
+          "downloadPDF": "PDF",
+          "info": {
+            "points": " punti",
+            "totalPoints": "Numero di punti processati:",
+            "layer":"Livello:",
+            "line":"Distanza:",
+            "up":"Dislivello Positiva:",
```
offtherailz, line 3749:
```suggestion
            "up":"Dislivello Positivo:",
```

web/client/translations/data.it-IT.json
```diff
@@ -3714,6 +3718,70 @@
...
         "sidebarMenu": {
           "showMoreItems": "Mostra più elementi"
+        },
+        "longitudinalProfile": {
+          "open": "Apri profilo longitudinale",
+          "close": "Chiudi profilo longitudinale",
+          "title": "Profilo longitudinale",
+          "draw": "Disegna profilo",
+          "import": "Carica file",
+          "select": "Seleziona il livello",
```
offtherailz, line 3728: Something like "Seleziona da livello" "Linea da livello" or  "Profilo da livello" sounds better in italian, for the menu entry.
```suggestion
          "select": "Profilo da livello",
```

offtherailz (changes requested):
- [x] fix failing tests
- [x] fix inconsistent state of UI (tabs with emtpy chart and empty data in both tabs) when drawing out of the box 
image
- [x] I noticed that selection of lines in 3d do not work
[screencast-localhost_8081-2023.06.28-10_07_57.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/883095b2-d36a-4765-bfed-84f6d9f4281b) maybe is better to disable it at all?
- [x] Inline comments
- [x] icon changes are still present.

Minor things
- [ ] PDF may contain also the info shown in the info tab.
- [ ] PDF on full screen overflows the page size and goes in the next page, like this.  I suggest to reduce a little the vertical size of the chart to make the pdf generate at full screen a 1 page result (or if you need to add the info too, you will need 2 pages, so you can ignore this point.
[Longitudinal profile (2).pdf](https://github.com/geosolutions-it/MapStore2/files/11891971/Longitudinal.profile.2.pdf)
- [ ] Selection of the layer. SelectionThis should be better in recognize the two steps (otherwise Selected label" appended string looks strange.
image
I suggest to show: 
- IF layer is not selected ""Please select layer in TOC and click on the line feature to generate profile",
- IF layer is selected: ".Selected layer: {layerName}<br/>Click on the line feature to generate the profile."

tdipisa:
>- PDF may contain also the info shown in the info tab.
>- PDF on full screen overflows the page size and goes in the next page, like this. I suggest to reduce a little the vertical size of the chart to make the pdf generate at full screen a 1 page result (or if you need to add the info too, you will need 2 pages, so you can ignore this point.

These two points can be excluded for the moment. @offtherailz please can you review again?

offtherailz (approved):

MV88:
@ElenaGallo testable here 

https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/9241_long_profile

PR #9259, opened 2023-06-28
Delete add_to_project.yml

offtherailz:
## Description
We can remove this action in favour of the new automation here https://github.com/orgs/geosolutions-it/projects/10/workflows/10092406

image


tdipisa (commented):
The workflow you have linked doesn't provide what we need. It simply set the status of an item added to a project.

tdipisa (approved):

Issue #9250, opened 2023-06-26
Share plugin - Permalink - Add a permalink functionality


dsuren1:
## Description
Add new permalink functionality to the Share plugin

## Acceptance criteria

- [x] Permalink shares current user session (custom layers in the TOC, custom background layers, all user customization) and not only the initial map of the Context. This means that the current overall map/viewer state will be saved (as it usually happens when you save a map, for example). The saved resource should be public by default
- [x] Create new category named PERMALINK to save all permalink under it --> 
     - [x] Add to XML of initial categories
     - [x] Add the proper migration scripts to [mapstore guidelines ](https://docs.mapstore.geosolutionsgroup.com/en/latest/developer-guide/mapstore-migration-guide/#database-update)
- [x] The permalink saved has its own ID to be used as a query param in the URL to load the same. The same logic will be applied to contexts and maps created from a context.
- [X] Protected layers as part of the permalink that are not visible to a certain user are marked in TOC (which follows existing behaviour of TOC. Hence no change is required in the implementation)
- [x] Permalink should be a separated plugin injected in the share plugin.
- [x] Permalink tab is placed inside the Share panel. The permalink tab is visible only to authenticated users and user with allowed permissions (This could be expanded in the future using some configuration like allowed roles/groups) and can be disabled via configuration.
- [x] On permalink tab open the user will have a form with
    -  "Title" (mandatory)
    - "Description" 
    - "Public" a checkbox, checked by default. If checked, it adds the group everyone can view. Otherwise it is private for the user (that is the default if you do not specify any permission for the links).
- [x] After the sumbit of the form the user will see the following screen
   <img width="407" alt="image" src="https://github.com/geosolutions-it/MapStore2/assets/26929983/e941a008-aa40-4aa8-a587-ce9e9a7c86ec">
- [x] Add a checkbox to permalink panel to make it public
- [x] **Not mandatory** (_nice to have_): It should be possible to include permalinks also for geostories and dashboards with no particular additional effort. To be checked during the implementation 
- [x] Take in consideration also the Other useful information about the suggested implementation and the things about the title.

## Other useful information

https://github.com/georchestra/mapstore2-georchestra/issues/583

### Suggested implementation

The permalink is a map (but can be a context, geostory ...) resource in PERMALINK category that has to be saved on MS DB, with 

- name: (name of the resource) an hash generated on the fly

attributes:

- `title`: the title of the item (mandatory for permalinks, inserted by user) 
[...]

PR #9288, opened 2023-07-18
#9250: Share plugin - Permalink functionality

dsuren1:
## Description
This PR introduces permalink functionality to Share

 - [x] Feature

## Issue

- #9250 

**What is the new behavior?**
User can create permalink of the current state of the map, dashboard, geostory, context map. The permalink resources are saved as PERMALINK category and has unique resource id. The title attribute is shown in Save/SaveAs modal and TOC

image
image

## Breaking change
 - [x] No

### DIFF
--- a/docs/developer-guide/mapstore-migration-guide.md
+++ b/docs/developer-guide/mapstore-migration-guide.md
@@ -95,11 +95,15 @@ The old spring maven repositories that do not exist anymore have been removed fr
 
 ### New Permalink plugin
 
-As part this release, permalink plugin is added to MS.
+As part this release, permalink plugin is added to MapStore. The new plugin is already configured in standard MapStore application, but if you are working on a project or if you customized the configuration files, you may need to update them to introduce the new plugin.
 
-#### Add permalink to localConfig.json
+In any case on an existing installation you **must** update the database adding the category to make the plugin work.
 
-- Update `localConfig.json` and add "Permalink" plugin to the "desktop", "dashboard" and "geostory" section
+#### Add Permalink plugin to localConfig.json
+
+In the case you customized your `configs/localConfig.json` file in your project/installation, to add the permalink plugin you will have to update it as following:
+
+- Add the "Permalink" plugin to the pages you want to use this plugin. Pages plugins are in the `plugins` section in the root of `localConfig.json`,  so you have to add "Permalink" entry to `desktop`, `dashboard` and `geostory` arrays, like this:
 
 ```json
 {
@@ -121,7 +125,7 @@ As part this release, permalink plugin is added to MS.
 }
 ```
 
-- Add new "permalink" section as shown below
+- To activate the functionality, you must add a new `permalink` section to `plugins` root object of `localConfig.json`, as shown below
 
 ```json
 {
@@ -134,10 +138,10 @@ As part this release, permalink plugin is added to MS.
 
 #### Using Permalink in new contexts
 
-Contents of your `pluginsConfig.json` need to be reviewed to allow usage of new "Peramlink" in new contexts.
-Existing contexts need to be updated separately.
+The plugins available for contexts are listed in the file `configs/pluginsConfig.json`. In your project/installation, you may need to edit this configuration to make the plugin selectable for your context.
+Existing contexts need to be updated separately, after applying these changes
 
-- Find "Share" plugin configuration in `pluginsConfig.json` and modify as shown below:
+- Find the "Share" plugin configuration in the `plugins` array in the root of  `pluginsConfig.json` configuration file and modify it as shown below (adding `children` and `autoEnableChildren` sections:
 
 ```json
     {
@@ -157,15 +161,14 @@ Existing contexts need to be updated separately.
     }
 ```
 
-- Add "Permalink" plugin configuration to `pluginsConfig.json`
+- Add "Permalink" plugin configuration to the `plugins` array in the root of `pluginsConfig.json`
 
 ```json
     {
       "name": "Permalink",
       "glyph": "link",
       "title": "plugins.Permalink.title",
-      "description": "plugins.Permalink.description",
-      "denyUserSelection": true
+      "description": "plugins.Permalink.description"
     },
 ```
 
--- a/docs/developer-guide/mapstore-migration-guide.md
+++ b/docs/developer-guide/mapstore-migration-guide.md
@@ -93,6 +93,107 @@ The old spring maven repositories that do not exist anymore have been removed fr
 -        </repository>
 ```
 
+### New Permalink plugin
+
+As part this release, permalink plugin is added to MS.
+
+#### Add permalink to localConfig.json
+
+- Update `localConfig.json` and add "Permalink" plugin to the "desktop", "dashboard" and "geostory" section
+
+```json
+{
+    "desktop": [
+        ...
+        "Permalink",
+        ...
+    ],
+    "dashboard": [
+        ...
+        "Permalink",
+        ...
+    ],
+    "geostory": [
+        ...
+        "Permalink",
+        ...
+    ]
+}
+```
+
+- Add new "permalink" section as shown below
+
+```json
+{
+    "permalink": [
+        "Permalink",
+        "FeedbackMask"
+    ]
+}
+```
+
+#### Using Permalink in new contexts
+
+Contents of your `pluginsConfig.json` need to be reviewed to allow usage of new "Peramlink" in new contexts.
+Existing contexts need to be updated separately.
+
+- Find "Share" plugin configuration in `pluginsConfig.json` and modify as shown below:
+
+```json
+    {
+      "name": "Share",
+      "glyph": "share",
+      "title": "plugins.Share.title",
+      "description": "plugins.Share.description",
+      "dependencies": [
+        "SidebarMenu"
+      ],
+      "children": [
+        "Permalink"
+      ],
+      "autoEnableChildren": [
+        "Permalink"
+      ]
+    }
+```
+
+- Add "Permalink" plugin configuration to `pluginsConfig.json`
+
+```json
+    {
+      "name": "Permalink",
+      "glyph": "link",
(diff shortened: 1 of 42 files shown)


offtherailz:
can you update this pr with latest master, so It can pass the checks, thank you.

docs/developer-guide/mapstore-migration-guide.md
```diff
@@ -93,6 +93,107 @@ The old spring maven repositories that do not exist anymore have been removed fr
...
+
+- Add "Permalink" plugin configuration to `pluginsConfig.json`
+
+```json
+    {
+      "name": "Permalink",
+      "glyph": "link",
+      "title": "plugins.Permalink.title",
+      "description": "plugins.Permalink.description",
+      "denyUserSelection": true
```
offtherailz, line 168: why this option?

web/client/configs/pluginsConfig.json
```diff
@@ -152,8 +152,21 @@
...
+      "autoEnableChildren": [
+        "Permalink"
       ]
     },
+    {
+      "name": "Permalink",
+      "glyph": "link",
+      "title": "plugins.Permalink.title",
+      "description": "plugins.Permalink.description",
+      "denyUserSelection": true
```
offtherailz, line 168: why this option?

web/client/components/share/SharePanel.jsx
```diff
@@ -241,7 +242,9 @@ class SharePanel extends React.Component {
         if (this.props.shareUrlRegex && this.props.shareUrlReplaceString) {
             shareEmbeddedUrl = this.generateUrl(shareEmbeddedUrl, this.props.shareUrlRegex, this.props.shareUrlReplaceString);
         }
-        const currentTab = !this.props.embedPanel && this.state.eventKey === 3 ? 1 : this.state.eventKey; // fallback to tab link if embed is disabled and selected at the same time
+        const {component: Permalink} = this.props.items?.find(({name} = {}) => name === "Permalink") ?? {};
+        const permalink = Permalink ? <Permalink shareUrl={shareUrl}/> : null;
+        const currentTab = ((!this.props.embedPanel && this.state.eventKey === 4) || (!permalink && this.state.eventKey === 3)) ? 1 : this.state.eventKey; // fallback to tab link if embed|permalink is disabled and selected at the same time
```
offtherailz, line 247: Here I'd use `items` with `target` `tabs` to inject the item, without any specific reference to permalinks. 

In general Share plugin should not know anything about the injected tabs
. So the Item should pass the properties (e.g. title) and component, and the eventKey should be incremental (or passed). 
If embed is the last panel `eventKey` of embedPanel should be `2 + items.filter(({target})=>target === 'tabs').length` (or simply something like 10000).

web/client/epics/context.js
```diff
@@ -137,14 +138,16 @@ const createSessionFlow = (mapId, contextName, action$, getState) => {
  */
 export const loadContextAndMap = (action$, { getState = () => { } } = {}) =>
     action$.ofType(LOAD_CONTEXT).switchMap(({ mapId, contextName }) => {
+        const isPermalink = isPermalinkSelector(getState());
+        const resourceCategory = isPermalink ? 'PERMALINK' : 'CONTEXT';
```
offtherailz, line 142: Looking at the code I'm wandering what happens after I open the permalink if I navigate to the context. Permalink id stays there and the session may be calculated wrongly. Please make a double check on this.

by the way seeing `session` in epics of context shows that the functionalities are not properly decoupled, maybe this needs to be improved in the future.

offtherailz (changes requested):
- [x] see my comments inline
- [ ] Title can not be saved or searched. I see you make it work on cards too, and this is great, but for the moment the title functionality is not supported sufficiently for resources to make this title visible everywhere. So for the moment, as I asked, show the title input **only** if the title already exists as attribute (so **only** for the case of permalinks). After merging this PR, let's open a new issue to improve the title functionality.
[title_not_saved.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/3111cae8-f382-4a95-b402-109cb4fdc5c4)
Here the point of the issue telling this 
> - [ ] It could be a nice to have to have the title editor in save form too. IT should be shown for the moment only if it is present as attribute (also an empty string is ok for showing the editor, but if it is an empty string in the layer tree or where it is shown as a string,you should show the name isntead.

- [ ] I notice the controlLabel is not correct. "Via a direct link" . Please fix (use something like "Permalink generated", maybe a subtitle "share the resource using the generated address"

-[ ]  The permalinks URL should contain the hash in the URL (name), instead here I see the ID. 
image

Because of this inconsistency above, any futher test will not be valid because after refactoring I'd need to test all again. Generally the rest seems to work well (please double check the part of the session I commented inline).

- I noticed the share plugin to create permalink is not shown if the map is new ... It may be acceptable @tdipisa ?

web/client/epics/context.js
dsuren1: Upon load finishes, we clear permalink hash. So it loads fine when another context is loaded.

But I have another question, now that permalink is redirected to load `context/hash`, upon refreshing the page, the call is fired as context for permalink hash which will result in `Resource not found` is this okay? Or should we instead save the permalink hash in a localstorage and compare hash to determine the category?

dsuren1:
> * [ ]  Title can not be saved or searched. I see you make it work on cards too, and this is great, but for the moment the title functionality is not supported sufficiently for resources to make this title visible everywhere. So for the moment, as I asked, show the title input **only** if the title already exists as attribute (so **only** for the case of permalinks). After merging this PR, let's open a new issue to improve the title functionality.

The code is already present to show title from the resource on the card. Will make name take precedence on resources and hide input when title is not present

offtherailz:
@tdipisa we should be ready to go. Anyway our installation do not contain the PERMALINK category. Now we have docker and some other items. 
The migration guidelines has been provided, but we have to provide them in our installation. Should we ask to alessandro ?

offtherailz (changes requested):
Everything works. Only some minor issues:

- [ ] Save button has a fixed size as the "create new" button, in certain languages, this is not enough. Expecially in french where  this have to be used mostly.  So please
     - [ ]  Make these buttons growing with the text
     - [ ] I think that strings should be : "Generate permalink" and "Create a new permalink..." (with ... suggesting to do more). 
     - [ ]  Create new permalink should stay below the QR Code, on the right.
image
image

dsuren1:
**Updated:**
Screenshot 2023-07-24 at 11 19 46 AM
Screenshot 2023-07-24 at 11 20 02 AM

dsuren1:
Doc build is complaining about configuration file missing. Strange!

offtherailz:
> Doc build is complaining about missing configuration file. It worked a while back. Strange!

I think is a new requirement. 
https://docs.readthedocs.io/en/stable/config-file/v2.html

Can you open an issue for that? @tdipisa I'll review the PR in the meanwhile, but for merging it requires this to be fixed on the main project
adding a file like `.readthedocs.yml` to the root of the project, with something like this may work, but we have to do some tests with a PR. 

```yml
version: 2

python:
  install:
    - requirements: docs/requirements.txt
  system_packages: true

mkdocs:
  configuration: mkdocs.yml
  fail_on_warning: false

```
@tdipisa can you assign this task to someone?

web/client/epics/permalink.js
```diff
@@ -122,30 +163,20 @@ export const savePermalinkEpic = (action$, { getState = () => {} }) =>
...
-                        )
-                    )
+                    ).switchMap(() => Observable.of(
+                        updatePermalinkSettings({name}),
+                        show({
+                            id: "PERMALINK_SAVE_SUCCESS",
+                            title: "notification.success",
+                            message: `permalink.${categoryCreated ? 'createAndSaveSuccess' : 'success'}`
+                        })
+                    ).do(categoryCreated = false))
```
offtherailz: This do not work as you expected. In particular this always sets categoryCreated to false when thecontaining function is invoked, while it should do it after all. 
to do what you want you may have to do something like `.do(() => categoryCreated = false)`.

I suggest to avoid `categoryCreated` flag handling and emit 2 notifications, one inside the error handler that notifies that the permalink category was missing and it has been created, the other is this one, that always notifies the success of the permalink creation. 
This isolates more the error handling from the rest of the workflow, helping future maintainance.

offtherailz (approved):
I applied the fixes in review by myself, please take a look

offtherailz:
here I added a draft PR to fix the issue of readthedocs, you can start from this. I can not dedicate time today on this, hope it works.
https://github.com/geosolutions-it/MapStore2/pull/9297

tdipisa:
> Hi @tdipisa , there was a missing comment I noticed you didn't replied, so just to double check, I ask for confirmation
> 
> > * I noticed the share plugin to create permalink is not shown if the map is new ... It may be acceptable @tdipisa ?
> 
> I mean that the share plugin do not show when the map is new. So the permalink will not be visible too. Is this acceptable?

yes

offtherailz:
closed and reopen to re-run checks

ElenaGallo:
@dsuren1 some notes after testing the permalink:

- [ ] the permalink description is not reported in the save panel
description

- [ ] if the permalink is saved it is not found as a map on the homepage, is that correct?
save permalink

- [ ] the Details in the permalink are not visible
details

- [ ] the permalink generated from a context does not have the save panel filled in correctly (no name, tile and public rule)
context_permalink

offtherailz:
My feedback on this: 

- Description could be saved as an attribute so it can be persisted
- Permalink should not be listed in maps. For the moment are links, if you do not have the reference, you will not be allowed to use it.
- the Details in the permalink are not visible --> good catch. Anyway the detail is part of the saved resource, so even cloning the map with save as makes it lost. So I think not saving it is the correct behavior too. 

Ther rest is up to @dsuren1

dsuren1:
@ElenaGallo 
> the permalink generated from a context does not have the save panel filled in correctly (no name, tile and public rule)

`Save as` panel will not have name and description. This is an expected behavior.

PR #9309, opened 2023-07-28
Adding reducer addTransformer feature to print module

jdugh:
## Description
Adding addTransformer feature to the print module reducer/action for use this function outside of the MapStore2 framework code (like an MapStore2-geOrchestra extension).

 - [X] Feature

## Issue

When you develop a new extension for Mapstore2-Georchestra, if you [import addTransformer](https://github.com/geosolutions-it/MapStore2/blob/a00d669c5149d8546bb8fcd6a853929abdd672a4/web/client/plugins/Print.jsx#L194-L201) in your MapStore2-geOrchestra extension code and use it, the adding tranformer is in a local copy of the final userTransformerChain variable.
[For now, components retrieved from MapStore (using the import) will be a copy of the existing ones](https://github.com/geosolutions-it/MapStoreExtension#limitations)

So, when you open the print module and click on "Print", the print module call [getSpecTransformerChain](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/PrintUtils.js#L310) and use the variable userTransformerChain  which is not the same as your local copy userTransformerChain  modify by your local copy of addtransformer.
Your transform is not apply.

#<issue>

**What is the new behavior?**
For fix it and use addTransform from outside (MapStore2-geOrchestra extension), I propose to add an action and reducers to the print module for calling the addTransform methode through an action ; Print module in catch this action and apply the addTransformer method in the framework.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/reducers/print.js
+++ b/web/client/reducers/print.js
@@ -9,7 +9,6 @@
 import {
     SET_PRINT_PARAMETER,
     ADD_PRINT_PARAMETER,
-    ADD_PRINT_TRANSFORMER,
     PRINT_CAPABILITIES_LOADED,
     PRINT_CAPABILITIES_ERROR,
     CONFIGURE_PRINT_MAP,
@@ -21,7 +20,6 @@ import {
     PRINT_CANCEL
 } from '../actions/print';
 
-import { addTransformer } from '../utils/PrintUtils';
 import { TOGGLE_CONTROL } from '../actions/controls';
 import { isObject, get } from 'lodash';
 import assign from 'object-assign';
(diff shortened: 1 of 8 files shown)


web/client/reducers/print.js
```diff
@@ -77,6 +79,10 @@ function print(state = {spec: initialSpec, capabilities: null, map: null, isLoad
         }
         return state;
     }
+    case ADD_PRINT_TRANSFORMER: {
+        addTransformer(action.name, action.transformer, action.position);
```
offtherailz, line 83: Theoretically this is a side effect, and so it should go in an epic. reducers should be pure functions.

offtherailz (changes requested):
Very good feature. Thank you for contributing. 
For the implementation, it should be better to place the `addTransformer` in an epic, because it is a side effect, so it makes more sense there. Reducers should be pure functions, and not  cause side effects.  
Could you do this small change.

web/client/reducers/print.js
jdugh: I don't find any epics for print module in the [epcis folder](https://github.com/geosolutions-it/MapStore2/tree/master/web/client/epics).
Should I create a new epics file specially for this one ?
(I prefer asking you before pushing some modification).

thank you for your feedback !

jdugh:
Hi @offtherailz,
Move side effect to epics : Done

Issue #9275, opened 2023-07-11
Invalid date on Firefox


jusabatier:
## Description

I deployed the last "2023.02.00-RC1-geOrchestra" version of mapstore which use the "2023.01.xx" version.

When opening the attribute table of a layer with date attributes on Firefox, it display "Invalid date" instead of the date.

## How to reproduce

-  Use firefox web browser (115.0 here)
-  Add a layer with date attribute
- Open the attribute table
- All date appear as "Invalid date"

![image](https://github.com/geosolutions-it/MapStore2/assets/8310116/1ab1fdde-85c3-457d-ad9a-bacd43ad68a1)

*Expected Result*

The date should appear correctly.

*Current Result*

Date are display as "Invalid date"

- [ ] Not browser related

<details><summary> <b>Browser info</b> </summary>

| Browser Affected | Version |
|---|---|
|Firefox|Tested on 115.0 |
</details>

## Other useful information

In console we should have a related warning : 

```
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: true, _useUTC: true, _l: undefined, _i: 2021-03-09Z, _f: undefined, _strict: undefined, _locale: [object Object]
@https://geoportail.lepuyenvelay.fr/mapstore/dist/geOrchestra.js?5f38512cef2803e9bc4a:2:2761081
30381/At/<@https://geoportail.lepuyenvelay.fr/mapstore/dist/geOrchestra.js?5f38512cef2803e9bc4a:2:2780722
@https://geoportail.lepuyenvelay.fr/mapstore/dist/geOrchestra.js?5f38512cef2803e9bc4a:2:2780773
At@https://geoportail.lepuyenvelay.fr/mapstore/dist/geOrchestra.js?5f38512cef2803e9bc4a:2:2781067
It@https://geoportail.lepuyenvelay.fr/mapstore/dist/geOrchestra.js?5f38512cef2803e9bc4a…

```

landryb:
@tdipisa encountered by @jusabatier with https://github.com/georchestra/mapstore2-georchestra/releases/tag/2023.02.00-RC1-geOrchestra

dsuren1:
[...]

PR #9393, opened 2023-09-05
#9275 Fix issue Invalid date on Firefox 

mahmoudadel54:
## Description
The PR implements fixing the issue when opening the attribute table of a layer with date attributes on Firefox, it display "Invalid date" instead of the date.

 - [x] Bugfix
 
## Issue

#9275 

**What is the new behavior?**
 When opening the attribute table of a layer with date attributes on Firefox, it display the date in this format "YYYY-MM-DD[Z]" for example 2022-02-10Z.

## Breaking change
 - [x] No


web/client/components/data/featuregrid/formatters/index.js
```diff
@@ -26,7 +26,7 @@ const dateTimeFormatter = ({value, format, type}) => {
             ? moment.utc(value).format(format)
             : type === 'time'
                 ? moment(`${DEFAULT_DATE_PART}T${value}`).utc().format(format) // time format append default date part
-                : moment(value).format(format) // date or date-time formats
+                : moment(value, format).format(format) // date or date-time formats
```
allyoucanmap, line 29: The input **format** should not be the same as the output `format`. This function is taking a raw date/date-time/time _value_ from feature and convert it to configured format for visualization so we expect:

moment(_value_, **'YYYY-MM-DD[Z]'**).format(`format`)

We should apply the input **format** when the type is only date and the _value_ ends with **Z**, in general is a good practice to apply the input format when known as suggested [here](https://github.com/geosolutions-it/MapStore2/issues/9275#issuecomment-1655403718).
The current implementation is not working if you try to replace the [localConfig](https://github.com/geosolutions-it/MapStore2/blob/1f30c78dea9212cdcc9ba6890cc45c3ff643e0d9/web/client/configs/localConfig.json#L438) configuration of [FeatureEditor](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.FeatureEditor) with custom formats:

```diff
- "FeatureEditor",
+ {
+  "name": "FeatureEditor",
+  "cfg": {
+    "dateFormats": {
+      "date-time": "MM DD YYYY - HH:mm:ss",
+      "date": "MM DD YYYY",
+      "time": "HH:mm:ss"
+    }
+  }
+ },
```

allyoucanmap (changes requested):

web/client/components/data/featuregrid/formatters/index.js
mahmoudadel54: I check momentjs documentation for this point and I saw something I think it may be helpful. 
If I don't know the exact input format, I can pass all the available date formats in an array (list) and it will match the correct one from the list.
check this image: 
image

ref: https://momentjs.com/docs/#/parsing/string-formats/

allyoucanmap (approved):

ElenaGallo:
Test passed, @mahmoudadel54 please backport this PR to 2023.02.xx, thanks

Issue #9461, opened 2023-09-19
UI issue in Catalog tool using custom metadata template


tdipisa:
## Description

Using a custom metadata template for CSW sources introduce a probem in the UI for which buttons for metadata items in the result list are not well positioned.

## How to reproduce

- Open [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/10358)
- Open the CSW catalog source advanced settings to configure [this metadata template](https://github.com/geosolutions-it/MapStore2/files/12650561/show_metadata_template.txt)
- Load the metadata list and expand items 

*Expected Result*

![image](https://github.com/geosolutions-it/MapStore2/assets/1280027/a4dd4a6a-7864-47a2-95f8-fbfb23a25cac)

*Current Result*

![image](https://github.com/geosolutions-it/MapStore2/assets/1280027/43ac59b0-70d4-4ed1-99f0-995eab288424)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

The problem is affecting also QA and it is an identified regression from the previous state in old MS versions.

PR #9467, opened 2023-09-20
#9461: Fix UI issue in catalog tool using custom metadata template

mahmoudadel54:
## Description
Fix UI issue in catalog tool using a custom metadata template for CSW sources which now the buttons for metadata items in the result list are well positioned.

 - [x] Bugfix

## Issue
#9461 
Using a custom metadata template for CSW sources introduce a problem in the UI for which buttons for metadata items in the result list are not well positioned.
image

#<issue>
#9461 

**What is the new behavior?**
image

## Breaking change
 - [x] No

### DIFF
--- a/web/client/themes/default/less/sidegrid.less
+++ b/web/client/themes/default/less/sidegrid.less
@@ -290,7 +290,6 @@
                     text-overflow: ellipsis;
                     white-space: unset;
                     overflow: visible;
-                    // width: 100%;
                 }
             }
         }
--- a/web/client/themes/default/less/sidegrid.less
+++ b/web/client/themes/default/less/sidegrid.less
@@ -290,7 +290,7 @@
                     text-overflow: ellipsis;
                     white-space: unset;
                     overflow: visible;
-                    width: 100%;
+                    // width: 100%;
                 }
             }
         }


web/client/themes/default/less/sidegrid.less
```diff
@@ -290,7 +290,7 @@
                     text-overflow: ellipsis;
                     white-space: unset;
                     overflow: visible;
-                    width: 100%;
+                    // width: 100%;
```
MV88, line 293: do not comment styles. simply remove it

mahmoudadel54:
> * [ ]  SideCard is used in different part of the application, did you try to check if this change affects them somehow?
>   it could work for the catalog, but you may break it for other components and UI parts.
> * [ ]  Take some time to list here in the PR description where this is used and the result of a proper check by running the application
> * [ ]  commented code or styles should be avoided, there is git history to recover that info if needed

I have checked the code, I found that there is a prop called 'fullText' passed to SideCard Component and based on this prop a class 'full-text' is added. This prop is only passed from RecordItem Component which is in sidebarMenu at the UI issue was. 

The edited style is for:  [ .full-text .msapstore-side-card-info > div span ], which means class 'full-text' must be there. 
So based on that, I think there is no other effect on the other components.

MV88 (approved):

MV88:
@ElenaGallo Please test it in dev

ElenaGallo:
Test passed, @mahmoudadel54 please backport this to 2023.02.xx. Thanks

Issue #9466, opened 2023-09-20
GetFeatureInfo exceptions parameter format configuration


ale-cristofori:
Hi

In MapStore we want to include webservices from "Digitaal Vlaanderen" (eg. https://geo.api.vlaanderen.be/dhmv/wms)

When MapStore sends a GetFeatureInfo request to those services, no result shows up in the identify panel. When taking a look at the response in the network tab of the browser, it is clear that an error occurred:

```xml
   <?xml version="1.0" encoding="UTF-8" standalone="no" ?>
  <!DOCTYPE ServiceExceptionReport SYSTEM "http://schemas.opengis.net/wms/1.1.1/exception_1_1_1.dtd">
  <ServiceExceptionReport version="1.1.1">
  <ServiceException code="InvalidFormat">
  Parameter 'exceptions' contains unacceptable value.
  </ServiceException>
  </ServiceExceptionReport>
```

I see that MapStore adds to every GetFeatureInfo request the following parameter: "exceptions=application%2Fjson"
This parameter seems to be hardcoded in the file MapStore2\web\client\utils\mapinfo\wms.js

When taking a look at the service https://geo.api.vlaanderen.be/dhmv/wms?service=WMS&version=1.1.1&request=GetCapabilities, it is clear the exceptions parameter value that is being send by MapStore is not supported by the service:

```xml
<Exception>
  <Format>application/vnd.ogc.se_xml</Format>
  <Format>application/vnd.ogc.se_inimage</Format>
  <Format>application/vnd.ogc.se_blank</Format>
</Exception>
```

For our own MapStore GeoServers, the format application/json is supported though, so we didn't experienced this problem in the past.

We had contact with the owner of the failing services and they are not planning to add support other exceptions parameters at short term.

In order to fix the issue of malfunctioning Identify, we think it is necessary to make some changes to MapStore to support other exceptions parameter and allowing to configure them (not hardcoded), or to be able to disable the exceptions parameter. 
Or is there already a way to achieve this that I missed? 

What is your option about this? Can you give an estimation on the work needed to be done for this feature request?

MV88:
Hi  @tdipisa @offtherailz 

me and @mahmoudadel54 had a discussion over this task and we noticed that we can simply remove the exceptions parameter but then this has to be handled here differently 
https://github.com/geosolutions-it/MapStore2/blob/600b3bc1968affba2e7859f2451ccca8ab6e9d2c/web/client/epics/identify.js#L133-L137

[...]

PR #9471, opened 2023-09-21
#9466: handling GetFeatureInfo exceptions parameter format configuration

mahmoudadel54:
## Description
- Handling GFI exceptions parameter format by removing "application/json" and putting the geoserver default format for exceptions 'application/vnd.ogc.se_xml' that will work with all WMS servcies including arcgis services.
- Add getIdentifyWorkflow to catch the GFI errors and exceptions. 

 - [x] Bugfix

## Issue
#9466 

Some services has no "application/json" into their exception format list. So an exception of "Parameter 'exceptions' contains unacceptable value" shown for the GFI and prevent showing the identify results.

#<issue>
#9466 

**What is the new behavior?**
There is no exception related to wrong exception format in GFI and identify works well with all WMS servcies including arcgis services and the identify results shown to user normally.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/mapinfo/__tests__/wms-test.js
+++ b/web/client/utils/mapinfo/__tests__/wms-test.js
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2022, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import expect from 'expect';
+
+import axios from '../../../libs/ajax';
+import MockAdapter from "axios-mock-adapter";
+import {INFO_FORMATS} from "../../FeatureInfoUtils";
+import {getFeatureInfo} from "../../../api/identify";
+
+describe('mapinfo wms utils', () => {
+    let mockAxios;
+    beforeEach((done) => {
+        mockAxios = new MockAdapter(axios);
+        setTimeout(done);
+    });
+    afterEach((done) => {
+        if (mockAxios) {
+            mockAxios.restore();
+        }
+        mockAxios = null;
+        setTimeout(done);
+    });
+    it('should return the response object from getIdentifyFlow in case of 200 with empty features,', (done) => {
+        const SAMPLE_LAYER = {
+            type: "wms",
+            name: "test_layer"
+        };
+        mockAxios.onGet().reply(() => {
+            return [200, {
+                "type": "FeatureCollection",
+                "features": [],
+                "totalFeatures": "unknown",
+                "numberReturned": 0,
+                "timeStamp": "2023-09-22T08:50:30.808Z",
+                "crs": null
+            }];
+        });
+        getFeatureInfo(
+            "TEST_URL", {
+                info_format: INFO_FORMATS.PROPERTIES
+            }, SAMPLE_LAYER
+        ).subscribe(
+            n => {
+                expect(n.data.features.length).toEqual(0);
+                expect(n.features).toEqual([]);
+                done();
+            },
+            error => {
+                return done(error);
+            }
+        );
+    });
+    it('should return the response object from getIdentifyFlow in case of 200 with features,', (done) => {
+        const SAMPLE_LAYER = {
+            type: "wms",
+            name: "test_layer"
+        };
+        mockAxios.onGet().reply(() => {
+            return [200, {
+                "type": "FeatureCollection",
+                "features": [{}, {}],
+                "totalFeatures": "unknown",
+                "numberReturned": 2,
+                "timeStamp": "2023-09-22T08:50:30.808Z",
+                "crs": null
+            }];
+        });
+        getFeatureInfo(
+            "TEST_URL", {
+                info_format: INFO_FORMATS.PROPERTIES
+            }, SAMPLE_LAYER
+        ).subscribe(
+            n => {
+                expect(n.data.features.length).toEqual(2);
+                expect(n.features.length).toEqual(2);
+                done();
+            },
+            error => {
+                return done(error);
+            }
+        );
+    });
+    it('should return the response object from getIdentifyFlow in case no data features as a text,', (done) => {
+        const SAMPLE_LAYER = {
+            type: "wms",
+            name: "test_layer"
+        };
+        mockAxios.onGet().reply(() => {
+            return [200, 'no features were found'];
+        });
+        getFeatureInfo(
+            "TEST_URL", {
+                info_format: INFO_FORMATS.TEXT
+            }, SAMPLE_LAYER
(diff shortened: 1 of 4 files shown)


tdipisa (commented):
@mahmoudadel54 it seems the unit test check is failing. Can you please check?

web/client/utils/mapinfo/wms.js
```diff
@@ -59,7 +62,7 @@ export default {
                 service: 'WMS',
                 version: '1.1.1',
                 request: 'GetFeatureInfo',
-                exceptions: 'application/json',
+                exceptions: 'application/vnd.ogc.se_xml',           // the default exception format
```
offtherailz, line 65: I think we can remove this parameter at all... not sure to force it.

offtherailz (changes requested):
Please:
-  `getIdentifyFlow` should have a unit test that checks the effective output for exception.
- the failing unit tests must be fixed

offtherailz (approved):
I effectively see that no event is triggered by openlayers for these load events.
So there is not an easy solution.
Even if we can trigger the load start on add, we can not catch load end... so I think it could be better to open a dedicated issue for that.
@tdipisa let me know what do you think. The rest is ready to be merged.

offtherailz:
sorry I sent my review to this PR in error, but the comment was for another pull request. 

:facepalm:

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright 2022, GeoSolutions Sas.
```
MV88, line 2: this is a minor thing, but for new files the copyright has to be aligned to current year
```suggestion
 * Copyright 2023, GeoSolutions Sas.
```

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -0,0 +1,111 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import expect from 'expect';
+
+import axios from '../../../libs/ajax';
+import MockAdapter from "axios-mock-adapter";
```
MV88, line 12: when you can try to keep external libraries separated by project modules and to be listed as the first thing
```suggestion
import expect from 'expect';
import MockAdapter from "axios-mock-adapter";

import axios from '../../../libs/ajax';
```

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -0,0 +1,111 @@
...
+import {INFO_FORMATS} from "../../FeatureInfoUtils";
+import {getFeatureInfo} from "../../../api/identify";
+
+describe('mapinfo wms utils', () => {
```
MV88, line 16: can you add some test that make intercept ogc error to be triggered?

MV88 (changes requested):
a few things
the rest lgtm

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -108,4 +110,38 @@ describe('mapinfo wms utils', () => {
...
+    it('test intercept ogc error in wms if success', (done)=>{
+        mockAxios.onGet().reply(() => {
+            return [200, {
+                "type": "FeatureCollection",
+                "features": [],
+                "totalFeatures": "unknown",
+                "numberReturned": 0,
+                "timeStamp": "2023-09-22T08:50:30.808Z",
+                "crs": null
+            }];
```
MV88, line 122: I think here you can return the body of an exception in json

try to get it by requesting  a get feature info with exception format in application/json  for a layer that does not exist maybe

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -108,4 +110,38 @@ describe('mapinfo wms utils', () => {
...
+        });
+        wms
+            .getIdentifyFlow(undefined, "/", { features: [] })
+            .subscribe((response) => {
+                expect(response?.data?.features).toEqual([]);
+                done();
+            }, error => {
+                done(error);
```
MV88, line 130: just call done()

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -108,4 +110,38 @@ describe('mapinfo wms utils', () => {
...
+                "numberReturned": 0,
+                "timeStamp": "2023-09-22T08:50:30.808Z",
+                "crs": null
+            }];
+        });
+        wms
+            .getIdentifyFlow(undefined, "/", { features: [] })
+            .subscribe((response) => {
+                expect(response?.data?.features).toEqual([]);
+                done();
```
MV88, line 128: remove this call so that we are sure it works because it triggers the error call back

MV88 (changes requested):

web/client/utils/mapinfo/__tests__/wms-test.js
mahmoudadel54: It gave eslint error here due to not using error

MV88: then remove the error param

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -0,0 +1,163 @@
...
+                        "text": "Could not find layer misc:Bonnebladen1900_index2121 Details:org.geoserver.platform.ServiceException: Could not find layer misc:Bonnebladen1900_index2121"
+                    }]
+            }];
+        });
+        wms
+            .getIdentifyFlow(undefined, "/", { features: [] })
+            .subscribe((response) => {
+                expect(response?.data?.exceptions).toExist();
+                done();
+            });
```
MV88, line 148: are you sure this is actually testing the intercept ogc error call?

shouldn't throw an error and therefore the error callback should be called. so done have to placed there

web/client/utils/mapinfo/__tests__/wms-test.js
```diff
@@ -0,0 +1,163 @@
...
+        mockAxios.onGet().reply(() => {
+            return [404, {}];
+        });
+        wms
+            .getIdentifyFlow(undefined, "/", { features: [] })
+            .subscribe((response) => {
+                expect(response?.data?.features).toEqual([]);
+            }, error => {
+                expect(error.status).toEqual(404);
+                done();
```
MV88, line 160: see previous comment, it should be similar to this one

ElenaGallo:
Test passe, @mahmoudadel54 please backport this to 2023.02.xx. Thanks

PR #9519, opened 2023-09-29
Increase java and javascript packages versions

offtherailz:
## Description
Increase java packages on master for 2024.01.xx 
 to 1.7-SNAPSHOT
Increas `package.json` to 0.8.0


tdipisa (changes requested):
@offtherailz we should probably change the package.json [here](https://github.com/offtherailz/MapStore2/blob/master/package.json#L3)

tdipisa (approved):

Issue #9463, opened 2023-09-19
3D map reloaded when opening or closing the Annotations and Identification panels


tdipisa:
## Description

When the Annotation panel or the Identify one are opened, the Cesium map reload

## How to reproduce

- Open a 3D map and use the identify or open the annotation panel 

*Expected Result*

The 3D map should not reload.

*Current Result*

![Image](https://github.com/geosolutions-it/MapStore2/assets/1280027/00979e67-9fbd-46c6-93ca-91a03bf78995)

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

PR #9530, opened 2023-10-02
#9463 3D map reloaded when opening or closing the Annotations and Identification panels

allyoucanmap:
## Description

Layer added as additionalLayers inside the Map plugin should contain name or id to properly manage the render state of the layer, if one of these information is missing the layer will be re-render causing the problem described in the issue.
This PR improves the rendering of additional layers by adding an id if this is missing inside the configuration.

 - [x] Bugfix

## Issue

#9463

**What is the new behavior?**

The terrain layer is not re-rendered every time the identify panel has been opened

## Breaking change

 - [x] No

### DIFF
--- a/web/client/plugins/Map.jsx
+++ b/web/client/plugins/Map.jsx
@@ -226,7 +226,8 @@ class MapPlugin extends React.Component {
         currentLocaleLanguage: PropTypes.string,
         items: PropTypes.array,
         onLoadingMapPlugins: PropTypes.func,
-        onMapTypeLoaded: PropTypes.func
+        onMapTypeLoaded: PropTypes.func,
+        pluginsCreator: PropTypes.func
     };
 
     static defaultProps = {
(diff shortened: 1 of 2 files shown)


MV88 (changes requested):
seems good, but issue refers to 3d maps and your code targets openlayers only.

also what is the motivation to have pluginsCreator as a prop?

allyoucanmap:
> seems good, but issue refers to 3d maps and your code targets openlayers only.

the use of 2d or 3d maps is irrelevant now because as per description the problem is related to component re-render because of missing key prop generated by id and name that could be missing in the additionalLayers prop configuration

> also what is the motivation to have pluginsCreator as a prop?

The motivation is testing, it was not possible to read the generated key of the component via react. Allowing to override that prop in testing give the  possibility to inject component to visualize the generated key as an id, please take a look to this test

https://github.com/geosolutions-it/MapStore2/pull/9530/files#diff-8beccdbc11a885b74ab7a7c7aec50bfcd1dd7fea6234d9a643cbd8a4c93726f1R93-R107

MV88 (approved):

MV88:

Issue #9539, opened 2023-10-03
Add support for style with properties for Vector and WFS layers


allyoucanmap:
## Description

Allow to choose a property of a feature as value for a symbolizer option. This work includes the update of all libraries style parsers used by the 2D/3D styler for WFS and Vector layers.

Relevant styling properties for which this can be enabled as an option could be the following:

- [x] Color properties
- [x] Width properties
- [x] Radius
- [x] Size
- [x] Rotation properties
- [x] Scale
- [x] Image/3D model URL

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

ElenaGallo:
@allyoucanmap I noticed that the *property value* of a 3D option can be changed even when in 2D mode.

https://github.com/geosolutions-it/MapStore2/assets/56537133/a0d22f2d-998e-4ef6-87ae-10f56499eb36

ElenaGallo:
New issue for https://github.com/geosolutions-it/MapStore2/issues/9539#issuecomment-1758062353 is #9609

PR #9541, opened 2023-10-03
#9539 Add support for style with properties for Vector and WFS layers

allyoucanmap:
## Description

This PR includes the possibility to assign a feature property as value for a style property

The UI includes an inline dropdown to select between constant or property value, it will possible to access the available supported feature attributes if property value is selected. Only property that support string or number are available, boolean and array are not currently supported

image

image

 - [x]  Minor changes to existing features

## Issue

#9539

**What is the new behavior?**

It is possible to select a feature property as value of a style property for WFS and Vector layers

## Breaking change

 - [x] No

## Other useful information

Sample map for testing [here ](https://github.com/geosolutions-it/MapStore2/files/12791415/style-with-properties.zip)


web/client/components/styleeditor/PropertyField.jsx
```diff
@@ -20,9 +21,13 @@ function PropertyField({ children, label, tools, divider, invalid, warning, disa
     return (
         <div
             className={'ms-symbolizer-field' + disabledClassName}>
-            <div className="ms-symbolizer-label"><Message msgId={label} /></div>
+            <div className="ms-symbolizer-label">
+                <Message msgId={label} />
+                {infoMessageId && <>&nbsp;<InfoPopover text={<Message msgId={infoMessageId} />}/></>}
```
MV88, line 26: better to use a ternary with null

web/client/components/styleeditor/PropertySelector.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+import React from 'react';
+import Select from 'react-select';
+import PropertyField from './PropertyField';
+import localizedProps from '../misc/enhancers/localizedProps';
+import DebouncedFormControl from '../misc/DebouncedFormControl';
+
+const LocalizedSelect = localizedProps('placeholder')(Select);
+
+const PropertySelector = ({
```
MV88, line 18: missing proptypes definition

web/client/components/styleeditor/__tests__/PropertySelector-test.jsx
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2020, GeoSolutions Sas.
```
MV88, line 2:
```suggestion
 * Copyright 2023, GeoSolutions Sas.
```

web/client/translations/data.it-IT.json
```diff
@@ -2527,7 +2527,16 @@
...
             "marker": "Marker",
-            "filterByName": "Filtra per nome..."
+            "filterByName": "Filtra per nome...",
+            "fillOpacity": "Opacità di riempimento",
+            "outlineColorOpacity": "Opacità del contorno",
+            "strokeColorOpacity": "Opacità del tratto",
+            "selectProperty": "Seleziona proprietà",
+            "leaderLineColorOpacity": "Opacità linea guida",
+            "colorOpacity": "Opacità",
+            "offset": "Offset",
```
MV88, line 2537: don't we want to translate it ?

MV88 (changes requested):
overall looks great, but reviewing this was not easy, please check inline comments and what listed below


-[ ] when you open the combobox here it does this
image
maybe you can use [this](https://github.com/geosolutions-it/MapStore2/blob/768f282272902ba81e0d6af713b0cc5882b7f49a/web/client/utils/DOMUtil.js#L44) function to change to drop up if needed 

-[ ] idk how,. but playing with ui i was ending up in genrating this, and shape is not recognized (in 3d)
image
[code editor text .txt](https://github.com/geosolutions-it/MapStore2/files/12803321/bug.or.not.txt)

web/client/components/TOC/fragments/StyleBasedLegend.jsx
```diff
@@ -26,7 +27,7 @@ function StyleBasedLegend({ style }) {
             fillOpacity,
             image,
             rotate
-        } = parseSymbolizerExpressions(symbolizer, { properties: {} });
+        } = symbolizer;
```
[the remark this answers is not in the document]
allyoucanmap: no, because symbolizer prop is used also after this in the same file and I wanted to avoid to parse it two times

web/client/translations/data.it-IT.json
allyoucanmap: how should we translate offset?

allyoucanmap:
> - [ ] idk how,. but playing with ui i was ending up in genrating this, and shape is not recognized (in 3d)

This is not a bug, it seems a circle and not a triangle because stroke is using the rotation as value that generate big width, this in combination with a radius equal 1 creates the circle effect.

In the video you can see that the shape is actually a tringle:

https://github.com/geosolutions-it/MapStore2/assets/19175505/f21d7170-2c0f-4139-a41c-fa12beca0334

allyoucanmap:
> - [ ] when you open the combobox here it does this

About this I would prefer to work on this on a separate issue, I just did a quick investigation and it's possible to apply a style to the react-select to get a drop up effect but it needs also a wrapper to manage the state based on the container, similar to the function proposed by @MV88. This problem was not introduced by this PR because the geometry transformation was already there so could we work on this problem inside a separated issue? (cc @tdipisa )

https://github.com/JedWatson/react-select/issues/403#issuecomment-267396731

web/client/translations/data.it-IT.json
MV88: deviazione
distanza
scostamento 
?

MV88 (approved):
I think that at the end we can leave offset translation for ita locale, not a big issue tbh

PR #9542, opened 2023-10-03
Release 2023.02.00

offtherailz:
Contains changes for the release 2023.02.00

+ a fix to `generateChangelog.js` script that will be ported on master too. 

See #9364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## [2023.02.00](https://github.com/geosolutions-it/MapStore2/tree/v2023.02.00) (2023-10-3)
+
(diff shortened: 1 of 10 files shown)


tdipisa (commented):
Only one question @offtherailz 
Should we probably update [this line](https://github.com/geosolutions-it/MapStore2/blob/2023.02.xx/utility/projects/projectLib.js#L71)? 

Having 2023.01.xx as a default hard coded in the code doesn't seem to be correct. Is this probably a typo?

offtherailz:
@tdipisa This function is always called with a parameter so the default will never be used (master by default, or inserted by the user). 
That value was present to support migration to this system, and now can be removed. Anyway it doesn't influece the current creation workflow, so I can change it, to master for instance, on the master branch,

tdipisa (approved):

PR #9569, opened 2023-10-06
Fix #9072 Added advanced options to pie chart

### LINKED ISSUE #9072: Pie Chart
## Description

The support for [advanced options](https://docs.mapstore.geosolutionsgroup.com/en/latest/user-guide/widgets/#advanced-options) should be included also for [Pie](https://docs.mapstore.geosolutionsgroup.com/en/latest/user-guide/widgets/#chart) widgets.

Relevant options could be:

- _Format_ and _Formula_

image

- _Labeling_ options

Image

- Other specific supported by PlotlyJS? To be checked

- [ ] Build related changes
- [ ] CI related changes

MV88:
## Description

Added advanced options to pie chart including:
* prefix
* suffix
* format
* formula
* hide labels

rotation label and never skip labels does not make much sense here I'm afraid, because they are already managed by the chart itself with the space available, 
but it's available rotation of the entire pie chart, do we want it ?

https://github.com/geosolutions-it/MapStore2/assets/11991428/aa67f2c4-7562-41a5-a53e-3a120105318b

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Fix #9072 

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,6 @@
     "concurrently": "6.4.0",
     "connected-react-router": "6.3.2",
     "create-react-class": "15.6.3",
-    "d3-format": "^3.1.0",
     "draft-js": "0.11.0",
     "draft-js-inline-toolbar-plugin": "3.0.1",
     "draft-js-plugins-editor": "2.1.1",
--- a/package.json
+++ b/package.json
@@ -173,6 +173,7 @@
     "concurrently": "6.4.0",
     "connected-react-router": "6.3.2",
     "create-react-class": "15.6.3",
+    "d3-format": "^3.1.0",
     "draft-js": "0.11.0",
     "draft-js-inline-toolbar-plugin": "3.0.1",
     "draft-js-plugins-editor": "2.1.1",
--- a/web/client/components/charts/WidgetChart.jsx
+++ b/web/client/components/charts/WidgetChart.jsx
@@ -1,5 +1,5 @@
 import React, { Suspense } from 'react';
-import {every, includes, isNumber, isString, union, orderBy, flatten} from 'lodash';
+import {round, every, includes, isNumber, isString, union, orderBy, flatten} from 'lodash';
 import {format} from 'd3-format';
 
 import LoadingView from '../misc/LoadingView';
@@ -300,18 +300,18 @@ function getData({
         if (yAxisOpts?.format) {
             try {
                 const formatter = format(yAxisOpts.format);
-                text = y.map(val => formatter(val));
+                text = y.map(val => `${yAxisOpts?.tickPrefix || ""} ${formatter(val)} ${yAxisOpts?.tickSuffix || ""}`);
             } catch (e) {
                 console.error(e);
             }
         }
         let pieChartTrace = {
             name: yAxisLabel || yDataKey,
-            hovertemplate: `%{label}<br>${yDataKey}<br>%{value}<br>%{percent}<extra></extra>`,
+            hovertemplate: `%{label}<br>${yDataKey}<br>%{text}<br>%{percent}<extra></extra>`,
             type,
-            textinfo: textinfo,
+            textinfo: textinfo || "%{percent}",
             // hide labels with textinfo = "none", in this case we have to omit texttemplate which would win over this.
-            texttemplate: textinfo === "none" ? null : `${yAxisOpts?.tickPrefix || ""}${yAxisOpts?.format ? "%{text}" : "%{percent}"}${yAxisOpts?.tickSuffix || ""}`,
+            texttemplate: textinfo === "none" ? null : "%{percent}",
             textposition: 'inside', // this avoids text to overflow the chart div when rendered outside
             values: y,
             text,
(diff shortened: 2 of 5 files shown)


offtherailz (changes requested):
`format`, `prefix` and `suffix` should be applied to the real value (processed by the formula), that for the pie chart is only in the tooltip

image

Using them on the % value on the pie chart makes no sense to me, or at least I can not figure out a use case where this is useful.

Optionally we may give an option to show the real value on the chart instead of the % value, but this doesn't look to be specified by the issue.

MV88:
> `format`, `prefix` and `suffix` should be applied to the real value (processed by the formula), that for the pie chart is only in the tooltip 

mixed values composed by numbers and strings do not work. 
it is expected that the sums of values is 100% so labels like 10M 3.5K  are not ok

In this pr the suffix is applied to the label not the tooltip, tooltip is using a template called hovertemplate

https://github.com/geosolutions-it/MapStore2/blob/9eab5e3155187d7e7e4e8bb9e1ba4ccd208b0953/web/client/components/charts/WidgetChart.jsx#L298

> Using them on the % value on the pie chart makes no sense to me, or at least I can not figure out a use case where this is useful.

in fact they are applied to the labels, it was part of an investigation on how those adv option can be applied to the pie chart

> Optionally we may give an option to show the real value on the chart instead of the % value, but this doesn't look to be specified by the issue.

This can be achieved with textinfo property by setting it to value, but yeah this has to be implemented: max 2h

@tdipisa can you provide a feedback as well ?

tdipisa:
> > `format`, `prefix` and `suffix` should be applied to the real value (processed by the formula), that for the pie chart is only in the tooltip
> 
> mixed values composed by numbers and strings do not work. it is expected that the sums of values is 100% so labels like 10M 3.5K are not ok
> 
> In this pr the suffix is applied to the label not the tooltip, tooltip is using a template called hovertemplate
> 
> https://github.com/geosolutions-it/MapStore2/blob/9eab5e3155187d7e7e4e8bb9e1ba4ccd208b0953/web/client/components/charts/WidgetChart.jsx#L298
> 
> > Using them on the % value on the pie chart makes no sense to me, or at least I can not figure out a use case where this is useful.
> 
> in fact they are applied to the labels, it was part of an investigation on how those adv option can be applied to the pie chart

I agree with @offtherailz. We can discuss it together if you need.

> > Optionally we may give an option to show the real value on the chart instead of the % value, but this doesn't look to be specified by the issue.
> 
> This can be achieved with textinfo property by setting it to value, but yeah this has to be implemented: max 2h

@MV88 if it takes only max 2h please include it.

tdipisa:
@MV88 
> > Optionally we may give an option to show the real value on the chart instead of the % value, but this doesn't look to be specified by the issue.
> 
> This can be achieved with textinfo property by setting it to value, but yeah this has to be implemented: max 2h

> @MV88 if it takes only max 2h please include it.

Considering the current status, after a call I had with @offtherailz I kindly ask you to open a dedicated issue for this to be developer with a separated work.

MV88:
> @MV88
> 
> > > Optionally we may give an option to show the real value on the chart instead of the % value, but this doesn't look to be specified by the issue.
> > 
> > 
> > This can be achieved with textinfo property by setting it to value, but yeah this has to be implemented: max 2h
> 
> > @MV88 if it takes only max 2h please include it.
> 
> Considering the current status, after a call I had with @offtherailz, I kindly ask you to open a dedicated issue for this to be developed with a separated work.

ok doing it now,
anyway, can we merge this right?

MV88:
@tdipisa  
issue opened here https://github.com/geosolutions-it/MapStore2/issues/9587

tdipisa:
> > @MV88
> > > > Optionally we may give an option to show the real value on the chart instead of the % value, but this doesn't look to be specified by the issue.
> > > 
> > > 
> > > This can be achieved with textinfo property by setting it to value, but yeah this has to be implemented: max 2h
> > 
> > 
> > > @MV88 if it takes only max 2h please include it.
> > 
> > 
> > Considering the current status, after a call I had with @offtherailz, I kindly ask you to open a dedicated issue for this to be developed with a separated work.
> 
> ok doing it now, anyway, can we merge this right?

@MV88 I don't think so please better check https://github.com/geosolutions-it/MapStore2/pull/9569#issuecomment-1753227727

MV88:
ready for review 
image

offtherailz (approved):
@MV88 I 
- removed the dependency d3 using internal formatting system, it seems to work, so we do not have to manually do
- Moved textinfo option inside yAxisOpts, so we do not have to drill the new prop and make all the containers know about this new option

allyoucanmap:
@offtherailz we are already using the concept of x, y for pie chart even if following the plotly configuration they should be called values and labels. You can see [here](https://github.com/geosolutions-it/MapStore2/pull/9569/files#diff-0912719c42496ea3d28493535260bf227051ef6fceabb41e828d719758b62819R300) that we are using `yDataKey` key probably to make it easier to switch in between chart types without remapping the configuration. I would prefer that this PR touches less files as possible so it would be easer to include in the ongoing work. Anyway if you have proposal on structure change we could discuss this

MV88:

Issue #9627, opened 2023-10-19
Replace Quill editor with Draftjs editor in Dashboard's Text widget 


dsuren1:
## Description
As part of enhancement to MS in replacing quill editor across the application. We would like to replace Quill editor used in Dashboard Text widget with Draftjs in this issue

### Acceptance criteria

- Replace quill editor with [Draftjs editor](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/resources/modals/fragments/editors/DraftJSEditor.jsx) in Dashboard [Text widget](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/widgets/builder/wizard/text/TextOptions.jsx#L30) (_Note: Refer also Mapview implementation of [text editor](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/mapviews/settings/CompactRichTextEditor.jsx) with Draftjs_)
- With `embedded` toolbar options enabled additionally
- Allow `fontFamilies` to be configurable just like in Geostory plugin

### Screenshot (Expected)
<img width="503" alt="Screenshot 2023-10-19 at 2 57 06 PM" src="https://github.com/geosolutions-it/MapStore2/assets/26929983/d6bbfaec-ec1f-41ea-a15d-5e0fa515bc3f">

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

mumairr:
@dsuren1 a quick question regarding the update, I can see draft-js has been archived this February and won't receive any future updates. 
On the other hand, react-quill is updated around the same time but still active and most probably recieve updates as well.
Any specific reason to upgrade to draftjs from quill? for the pdf thing, I tested it out on already integrated quill and works fine.

Please guide if i am missing something important. Thanks

dsuren1:
@mumairr 
The decision mainly comes from maintaining a single editor across the application, as most of the parts have been replaced already.

mumairr:
@dsuren1 makes sense. Thanks for the explanation

PR #9673, opened 2023-10-29
React Quill replaced with Draftjs in Dashboard's Txt Widget

mumairr:
## Description

React Quill editor has been replaced with Draftjs in Dashboard Text Wigets

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Fix https://github.com/geosolutions-it/MapStore2/issues/9627

**What is the new behavior?**

Draftjs added, allows embedding of PDFs or any link, plus allow upload of image rather than a link via props option

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/mapviews/settings/CompactRichTextEditor.jsx
+++ b/web/client/components/mapviews/settings/CompactRichTextEditor.jsx
@@ -58,7 +58,8 @@ function CompactRichTextEditor({
                     urlEnabled: true,
                     // disable the upload at the moment
                     // it will increase the size of the map too much
-                    uploadEnabled: false,
+                    // allows to set enable imge upload on/off via props
+                    uploadEnabled: props.uploadEnabled || false,
                     alignmentEnabled: false,
                     uploadCallback: (file) => new Promise((resolve, reject) => {
                         const reader = new FileReader();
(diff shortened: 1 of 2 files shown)


allyoucanmap (changes requested):
The implementation seems good, just few notes:

- [ ] we will need to backport this to the version 2022.02.xx supported by geonode-mapstore-client so we should make sure that it will work also on that branch. I think  CompactRichTextEditor.jsx does not exist on 2022.02.xx so just remember to include it in the backport if needed
- [ ] please add unit test to cover at least the requirement of the issue. We should ensure the embedded option is available
- [ ] one of the requirement is also to allow configurable font families but I don't see it implemented here. We could decide to split this task if so a separated issue needs to be created

web/client/components/mapviews/settings/CompactRichTextEditor.jsx
```diff
@@ -58,7 +58,8 @@ function CompactRichTextEditor({
                     urlEnabled: true,
                     // disable the upload at the moment
                     // it will increase the size of the map too much
-                    uploadEnabled: false,
+                    // allows to set enable imge upload on/off via props
+                    uploadEnabled: props.uploadEnabled || false,
```
allyoucanmap, line 62: Why are we allowing to upload images here? I think this is creating embedded base 64 images and I don't think it's a requirement of the task

mumairr: As per comments above, I guess image upload was disallowed for Geo-Stories only due to size issue (map + images), so decided to introduce a prop (false by default) if image upload needs to be allowed or just an embedded link

giohappy:
>@giohappy one of the requirement is also to allow configurable font families but I don't see it implemented here. We could decide to split this task if so a separated issue needs to be created

Custom fonts wasn't a requirement. I guess it was added as a bonus, but we can leave it out if it doesn't come for free.

PR #9676, opened 2023-10-30
OpenStreetMap : use HTTPS and endpoint fixes

RomainBourgATOL:
OSM now use HTTPS
Corrections on OSM URLs (removes old a/b/c aliases)
See https://github.com/openstreetmap/operations/issues/737 for informations about OSM URLs


offtherailz (approved):
Thank you very much for your contribution

PR #9682, opened 2023-11-03
Fix #9623 widgets font customizations

### LINKED ISSUE #9623: Widgets - From the wizard we would like to configure the properties of the text
## Description

From the wizard we would like to configure the properties of the text such as: font size, font family, font style etc

this can be applied to:
- counter widget
- legend in general
- chart widgets

Evaluate also other possibilities within the plotly docs

## Acceptance criteria

- [x] include in the advanced options some options for the [font](https://plotly.com/javascript/reference/layout/#layout-font)
  - [x] font size
  - [x] text color
  - [x] font family

MV88:
## Description

Now you can customize some properties of the font that will be stored inside layout prop or coutnerOpts

- Counter widget fontSize had been hidden because that is automatic
- Pie chart can customize the font inside the hover, but not inside the pie, because  you would lose some auto corrections on  the fontSize
- A new **Layout** tab has been added that will contain general chart layout info, but you can then override it with the values inside the x,y axes
- removed pie options chart comp since it's zombie code

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Fix #9623 

**What is the new behavior?**

## Breaking change
 - [x] No


web/client/components/widgets/builder/wizard/ChartWizard.jsx
```diff
@@ -139,6 +139,11 @@ const ChartWizard = ({
             data
         };
     });
+    useEffect(() => {
+        if (selectedTrace?.type === "pie" && tab === "axis") {
+            setTab("traces");
+        }
+    }, [selectedTrace?.type]);
```
allyoucanmap, line 146: I don't think a useEffect is needed in this case because the current selected tab could be computed directly in the render:

```js
const selectedTab = selectedTrace?.type === "pie" && tab === "axis" ? "traces" : (selectedTrace?.type || "traces");
```

then use:

```js
{tabContents[selectedTab]}
```

the selectedTab can be also passed to other components as tab prop

web/client/utils/GeoStoryUtils.js
```diff
@@ -560,6 +560,7 @@ export const extractFontNames = (fontFamilies) => {
 
 export const DEFAULT_FONT_FAMILIES = [
     'inherit',
+    'Noto Sans',
```
allyoucanmap, line 563: As per main comment I think this should not included in the default

web/client/utils/WidgetsUtils.js
```diff
@@ -28,6 +28,13 @@ import { sameToneRangeColors } from './ColorUtils';
 import uuidv1 from "uuid/v1";
 import { arrayUpsert } from "../utils/ImmutableUtils";
 
+
+export const FONT = {
+    FAMILY: "Noto Sans",
```
allyoucanmap, line 33: As per main comment I think font family should be inherit by default

allyoucanmap (changes requested):
Overall the code is good, just few changes:

- [ ] we should remove font color on pie chart because the default behaviour is better. Bu default it uses the contrast color for each slice of the pie
- [ ] line and bar chart have the layout section for font called **Popup style** but it has only effect on the axis. I think the title should be different like **Font**. It seems there is none font options applied globally to the plolty configuration, is this intentional?
- [ ] the font family select is clearable but it we click on the x an error in console shows up
- [ ] I think the Noto font family should not be included in the list because it's not considered a web safe font and it related to the default theme of mapstore that could be different in a custom project. My proposal is to remove the Noto Sans from the list and display the `Inherit` label so plolty will use the page font by default

web/client/components/widgets/builder/wizard/ChartWizard.jsx
MV88: are you sure about this line, especially in the false part selectedTab could be = to selectedTrace?.type which is not a tab

image

maybe it was `tab || "traces"` or simply tab?

MV88:
> Overall the code is good, just few changes:
> 
> * [ ]  we should remove font color on pie chart because the default behaviour is better. Bu default it uses the contrast color for each slice of the pie

makes sense

> * [ ]  line and bar chart have the layout section for font called **Popup style** but it has only effect on the axis. I think the title should be different like **Font**.

yes I've missed it during refactor

> It seems there is none font options applied globally to the plotly configuration, is this intentional?

what do you mean exactly by "globally"? 
In the Layout tab you can configure generic font attributes that are applied to all traces in that specific chart

> * [ ]  the font family select is clearable but it we click on the x an error in console shows up

I'll remove clearable

> * [ ]  I think the Noto font family should not be included in the list because it's not considered a web safe font and it related to the default theme of mapstore that could be different in a custom project. My proposal is to remove the Noto Sans from the list and display the `Inherit` label so plotly will use the page font by default

sounds good

MV88:
@allyoucanmap 
PR is ready again for review

web/client/components/widgets/builder/wizard/ChartWizard.jsx
allyoucanmap: yes, it was like that, I wrote the code in the comment not tested

web/client/components/widgets/builder/wizard/ChartWizard.jsx
```diff
@@ -143,6 +143,8 @@ const ChartWizard = ({
     if (!selectedTrace) {
         return null;
     }
+    const selectedTab = selectedTrace?.type === "pie" && tab === "axis" ? "traces" : (tab);
```
allyoucanmap, line 146: I think we need to pass [tab={selectedTab}](https://github.com/geosolutions-it/MapStore2/blob/b18c868bc6dc83821db512af83843f1c83e3f7c1/web/client/components/widgets/builder/wizard/ChartWizard.jsx#L160) also to ChartTraceEditSelector

allyoucanmap (changes requested):

web/client/components/widgets/builder/wizard/ChartWizard.jsx
MV88: done

allyoucanmap:

Issue #9807, opened 2023-12-11
Extruded features in Cesium 3D


allyoucanmap:
## Description

Giving one attribute of type Number, it is requested the possibility to extrude layer features (Annotation, Vector and WFS) in the 3D map.

## Acceptance criteria

- [ ] Extrusion of polygon features extending fill symbolizer
- [ ] Extrusion of polyline features extending line symbolizer

ElenaGallo:
@allyoucanmap The UI of *Pattern mark style* and *Pattern icon style* are not correct.

**Pattern mark style**
![1](https://github.com/geosolutions-it/MapStore2/assets/56537133/cb92fdd3-0de2-4128-966a-c993cac47a6b)


**Pattern icon style** 
![2](https://github.com/geosolutions-it/MapStore2/assets/56537133/59cc03e6-c549-4a9a-bd6f-43f7ac34c41f)

**How to Reproduce**
- Open the demo map
- Open the style tab for 'States of US'
- Click on Add fill rule
- Click on Pattern icon style

PR #9808, opened 2023-12-11
#9807 Extruded features in Cesium 3D

allyoucanmap:
## Description

This PR introduce following changes for the style editor:

- possibility to collapse rules. Introduction of new properties is extending the height of rule card and making difficult to sort them 
- legend symbols visible on rules to recognize them when collapsed
- refactor of the cesium parser. The previous parser was using a default Cesium data source class with workaround to support all the customization the we need. The new parsers create a list of styled features that can be managed inside a layer with the GeoJSONStyledFeatures class added in this PR.
- new rules for extrusion on fill symbolizer:
  - msHeight
  - msHeightReference
  - msExtrudedHeight
  - msExtrusionRelativeToGeometry
- new rules for extrusion on line symbolizer
  - msExtrudedShapeName
  - msExtrudedShapeRadius
  - msExtrusionColor
  - msExtrusionOpacity
- The boolean property UI of the visual style editor changed into a checkbox

 - [x] Feature

## Issue

#9807

**What is the new behavior?**

Refactor of the cesium style parser and introduction of extrusion property for line and polygon features

## Breaking change

 - [x] No


offtherailz (changes requested):
### Issues to check /Fix

- [ ] Importing this I have a crash in `GeoJSONStyle...`  [Camb3D_Bldg.zip](https://github.com/geosolutions-it/MapStore2/files/13716040/Camb3D_Bldg.zip) (it was not visible on dev, anyway.
- [ ] The extrusion outline doesn't look to be defined. Looking at the tool UI, I understand it should use the same style for all the edges. 
- [ ] If I select from "property value" from the "..." menu aside the color selector (italian, maybe, caused it), I have a misalignment in tools. Not sure is caused by this PR. The same happens for fill and outline color option.

image

- [ ] This broken image icon is a little ugly on startup. Can we use the icon of the rule type, when not speified? I see it is caused by `href="[object Object]"` in html
image


### Suggestions

- [ ] I'd prefer the drag and drop handle on the left. On the right is not so common, is more natural on the left to me, before the collapse/expand. In that place is too near to the delete button too :) . 

image

- [ ] Label preview looks different from the text icon. I'd do the 2 icons consistent, as well as the 3d model. 

image

- [ ] A tooltip on expand/collapse button should clarify more the role of this tool. 

- [ ]  with this refactor the style is applied all together, instead in the previous it was applied feature by feature. This causes a slow loading that may need a loading spinner. Maybe I used too complex data or features, but it took something like 8-10 seconds to render all and I thought was not working. 

- [ ] I suggest a collapse all / expand all toggle button (near undo-redo maybe with a separator).

- [ ] The legend here looks quite ugly. It should be possible to exclude some elements from the legend. Also indicating the preview of "La" before label is a little misleading.
image


### Questions
I understand that from line the extrusion transform it in a sort of pipe of the shape indicated. This creates a 3D volume for sure. I was expecting to have a surface instead, orthogonal to the ground (defining the height, as well as in the case of polygon). I don't think is it possible isn't it? (this is more a question). 


image

I know the extrusion depends on the software, but it looks a little strange to me that lines have radius and polygons have height. I'd understand more several extrusion modes. But maybe I missed some detail.

allyoucanmap:
@offtherailz 

> Importing this I have a crash in GeoJSONStyle... [Camb3D_Bldg.zip](https://github.com/geosolutions-it/MapStore2/files/13716040/Camb3D_Bldg.zip) (it was not visible on dev, anyway.

I improved the GeoJSONStyledFeatures class to validate the coordinates and if they are not valid they are excluded from the rendering as previous behaviour. It seems this data is not imported correctly in mapstore because it should contain PolygonZ features but instead it returns points with invalid height coordinates. The height coordinate is an array instead of a number.

> The extrusion outline doesn't look to be defined. Looking at the tool UI, I understand it should use the same style for all the edges.

The outline will be disabled if the extrusion height is set. The default outline property does not allow [custom width](https://cesium.com/learn/cesiumjs/ref-doc/PolygonGraphics.html#outlineWidth) in some scenarios neither dash options and for this reason we are creating a polyline as outline when the polygon is flat. Applying the same workflow with entities for extruded polygon will increase the number of geometries slowing down the rendering. It would be nice to fine an alternative approach using primitives but for the moment it's better to hide it.

> If I select from "property value" from the "..." menu aside the color selector (italian, maybe, caused it), I have a misalignment in tools. Not sure is caused by this PR. The same happens for fill and outline color option.

The color field splits in two color and opacity properties so this result is expected. 

> A tooltip on expand/collapse button should clarify more the role of this tool.

discussed with @tdipisa this will not be implemented at the moment

>  with this refactor the style is applied all together, instead in the previous it was applied feature by feature. This causes a slow loading that may need a loading spinner. Maybe I used too complex data or features, but it took something like 8-10 seconds to render all and I thought was not working.

I improved the GeoJSONStyledFeatures class by including a flag to decide to use merged primitives or entities for polygon geometry. By default it will use entity as it was doing before in the Cesium GeoJSON class

>  I suggest a collapse all / expand all toggle button (near undo-redo maybe with a separator).

discussed with @tdipisa this will not be implemented at the moment

> The legend here looks quite ugly. It should be possible to exclude some elements from the legend. Also indicating the preview of "La" before label is a little misleading.

discussed with @tdipisa this will not be implemented at the moment. There is already a related issue https://github.com/geosolutions-it/MapStore2/issues/9419

> I understand that from line the extrusion transform it in a sort of pipe of the shape indicated. This creates a 3D volume for sure. I was expecting to have a surface instead, orthogonal to the ground (defining the height, as well as in the case of polygon). I don't think is it possible isn't it? (this is more a question).

I improved the line symbolizer UI aligning it with the fill one. The line symbolizer have additional properties for extrusion color and type. The default extrusion type will be a vertical wall without with and additional types are only Circle and Square

Line rule

image

Polygon rule

image

offtherailz (approved):
Looks good and working.

offtherailz:
@ElenaGallo, could you please test this on  DEV, when deployed ? Thank you

PR #9809, opened 2023-12-12
Fix #9685 measurements issues and added some enhancements to it

MV88:
## Description

- [ ] Fix vector measurement layer issue
- [ ] Filter out point text label as annotations 
- [ ] Rename features id to use measureType numbered with a counter
- [ ] Transform geodesic lines before calling the process and for highlight features

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**

Fix #9685

**What is the new behavior?**

arc length as a buffer and blue line highlight
image

intersection of a buffer with states, (check lower parts should be the dark one 
image

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #9835, opened 2023-12-19
backport 2023.02.xx - Fix #9685 measurements issues and added some enhancements to it (#9809) 

### LINKED ISSUE #9685: Extend the GeoProcessing tool to work with client side Vector layers as WPS inputs
## Description

Our client is asking for a new additional feature based on the recently developed GeoProcessing tool for supported **Intersection** and **Buffer** processes. 
The existing tool functionalities are expecting to be used only for WMS and/or WFS layers in TOC as inputs for aforementioned processes.
As part of this work it is requested to extend the current tool to also support vector layers defined or imported on the client side.

Possible use case scenarios to supports are:

- Use the output of the GeoProcessing tool (vector layers generated by a buffer or intersection geoprocess) for performing another operation using it as a new input
- Use an imported vector layer (SHP, GeoJson, KML etc) as an input for the GeoProcessing tool
- Use existing annotations as an input for the GeoProcessing tool (to check if possible, it could be excluded initially since we have a new implementation in master)
- Use vector layers TOC coming from measurements

In all cases the UX should be the same and well consistent.

Partial evaluation already done in https://github.com/geosolutions-it/MapStore-C265/issues/29#issuecomment-1752913734

## Acceptance criteria

- [ ] Allow local vector layers listed above to be used as input in the GeoProcessing plugin
  - [ ] measurements
  - [ ] imported GeoJSON
  - [ ] imported Shapefiles
  - [ ] other imports
  - [ ] annotations
- [ ] make sure that the CollectGeometries request sent to the backend is adapted for vector layers
- [ ] handle cases where Geometry collection is generated from a vector layer, because buffer and intersect processes fails with it, we should prevent the user to click run and notifying it via user feedback
- [ ] for intersection it should maintain the previous checks we do, first layer/ feature can't be a point
- [ ] Add possibility to configure a predefined geoserver url to be used as a fallback within process requests when the source layer is a local vector layer

MV88:
## Description

backport 2023.02.xx - Fix #9685 measurements issues and added some enhancements to it (#9809)

### DIFF
--- a/web/client/actions/__tests__/geoProcessing-test.js
+++ b/web/client/actions/__tests__/geoProcessing-test.js
@@ -15,8 +15,6 @@ import {
     errorLoadingDFT, ERROR_LOADING_DFT,
     getFeatures, GET_FEATURES,
     initPlugin, INIT_PLUGIN,
-    increaseBufferedCounter, INCREASE_BUFFERED_COUNTER,
-    increaseIntersectedCounter, INCREASE_INTERSECT_COUNTER,
     runningProcess, RUNNING_PROCESS,
     runProcess, RUN_PROCESS,
     setBufferDistance, SET_BUFFER_DISTANCE,
@@ -97,18 +95,6 @@ describe('Test Geo Processing Tools related actions', () => {
         expect(action).toEqual({
             type: INIT_PLUGIN,
             cfg
-        });
-    });
-    it('increaseBufferedCounter', () => {
-        const action = increaseBufferedCounter();
-        expect(action).toEqual({
-            type: INCREASE_BUFFERED_COUNTER
-        });
-    });
-    it('increaseIntersectedCounter', () => {
-        const action = increaseIntersectedCounter();
-        expect(action).toEqual({
-            type: INCREASE_INTERSECT_COUNTER
         });
     });
     it('runningProcess', () => {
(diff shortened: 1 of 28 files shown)


offtherailz (changes requested):
- [ ] If I export a measure as annotation, the buffer operation includes also the points and do not follow the geodesic lines. Istead the export as a layer works well. @tdipisa should we fix this?
image

 
image

- WFS Intersection works well :+1: 
image

- [ ] Circle annotation buffer generation looks wrong, but at the end the annoation seems to have a strange behavior in any case. In the video below you can see the annotation (yellow) and the buffer generated (green). The buffer looks wrong, but when I go in editing the circle changes its shape, and the buffer seems to be generated on that shape...  maybe this is a bug of the old annotations ? 

[screencast-localhost_8081-2023.12.21-16_01_44.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/8a9819f6-aeed-40e2-b64c-2d570b4c1a3b)

- [ ] Intersection between annotations and remote layer gives the error "It was not possible to collect the geometry from the layer Annotations - undefined"

[screencast-localhost_8081-2023.12.21-16_07_38.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/7a1afea6-ee96-4558-8c8e-45a320777be9)

MV88:
@offtherailz this is ready for review, but i have to specify this:

the third point the one related to the intersection issue is resolved, but it remains problematic since there is this issue in geoserver that occurs
image

@tdipisa i think we should focus on fixing it on master (geoserver side probably) and not in the stable branch

tdipisa:
> @tdipisa i think we should focus on fixing it on master (geoserver side probably) and not in the stable branch

@offtherailz please proceed with the review and then we will discuss this.

offtherailz (approved):
I can not replicate anymore the error on the backend, probably fixed by the NG argorithm configuration.
And circle issue looks to be proper of annotations (re-written), not of this PR. 

So for me this PR is approved.

offtherailz:
Please @ElenaGallo test as well as it is deployed

Issue #9588, opened 2023-10-12
Add support to multi-band color mapping for COG layers


offtherailz:
## Description

As visible [here](https://openlayers.org/en/latest/examples/multiple-cogs.html), we can associate the colors to bands. Actually mapstore supports only standard 3 bands RGB or 1 band gray (because it is the openlayers default)

We have to implement the support to associate colors to bands.
This means: 
- Retrieve bands information from metadata (in advance from catalog, or when needed)
- to persist the association of bands with color. It should be implemented in a 
    - styling model and format open to future updates as color scales or other known raster styling formats.
    - Compatible to be read to cesium implementation for future implementation
- Provide a UI to edit this style (ideally both in catalog and layer style tab)

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

dsuren1:
As part of fetching metadata, we read only one image to get metadata info of the tif source. Reading raster data of the entire image can puts too much load on CPU and can severly impact performance, hence we need to specify a window/bbox, width, height and pooling to read raster data. We can specify some defaults to fetch from but we can't assure that it always reads the area of interest.

So my thoughts, we can get the banding information from the user in catalog and add it to COG source for user defined rendering of tif source and skip any reading of banding info from the image in catalog. 
And In layer -> style, we can allow user to pick a aoi by allowing user to click a point on map and using layer's `getData` to get rgba value to populate the bands fields (slider)?

**Mocks**

- This way we can also modify the band on the fly by changing values via sliders

  <kbd>
  <img width="501" alt="Screenshot 2023-12-11 at 5 37 51 PM" src="https://github.com/geosolutions-it/MapStore2/assets/26929983/7569c087-da21-4153-8828-c517929f7bcd">
  </kbd>

- Or should we also allow user to modify the channels to certain colors as shown here?
  <kbd>
   <img width="252" alt="Screenshot 2023-12-11 at 5 38 38 PM" src="https://github.com/geosolutions-it/MapStore2/assets/26929983/12306e9a-f36d-466f-9acf-96f79ace1b9d">
  </kbd>

We can discuss on this @offtherailz to establish an acceptance criteria

dsuren1:
**Updated acceptance criteria**

[...]

PR #9857, opened 2024-01-09
#9588: Add support to multi-band color mapping for COG layers

dsuren1:
## Description
This PR allows multi band color mapping of COG layers along modification of min and max source data value

 - [x] Enhancement

## Issue

- #9588 

**What is the new behavior?**
image

## Breaking change
 - [x] No


dsuren1:
Looks like GitHub experiencing outage causing actions to fails
image

web/client/api/catalog/COG.js
```diff
@@ -76,6 +76,47 @@ const fromUrl = (url, signal) => {
...
     });
 };
+
+const getLayerConfig = (layer, image) => {
+    const crs = getProjectionFromGeoKeys(image);
+    const extent = image.getBoundingBox();
+    const isProjectionDefined = isProjectionAvailable(crs);
+    const sample = image.getSamplesPerPixel();
+    const isRGB = image.getSampleByteSize(0) === 1 && typeof get(image, 'fileDirectory.PhotometricInterpretation') !== 'undefined';
+    const { STATISTICS_MINIMUM, STATISTICS_MAXIMUM } = image.getGDALMetadata() ?? {};
```
offtherailz, line 86: I suggest to store these source metadata and desume `isRGB` on demand, because conceptually this is not the indication of rgb. Also keep in mind that layer have can potentially have multiple source images.
So something like `fileDirectory.PhotometricInterpretation`  and `{samples: [ byteSize: 1 ]}`. Something like this. I'll ask @dromagnoli some info about geotiffs to clarify this.

offtherailz (changes requested):
- [ ] information for bands and styling should be documented, or is not easy to understand if they are properly made to be compatible for future evolution. Note that style information should be compatible with a future expansion with multiple source (so store style and source information with this in mind). I'll ask to @dromagnoli some help understanding these parameters

For the moment here some bugs I found: 

- [x] once changed, there is no way to restore max and min
    
    [screencast-mail.google.com-2024.01.16-16_55_16.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/25563043-3375-429b-9cdf-ddede45c016d)

- [ ] Sometimes it switches to one band randomly
        
    [screencast-mail.google.com-2024.01.16-17_03_21.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/ad590b82-56ec-4c7a-b7f1-95a87b2c5956)

offtherailz:
Hi, from a brief discussion about it with @dromagnoli I understood that 
Samples count should be the samples count + sope possible  [extra samples ](https://www.awaresystems.be/imaging/tiff/tifftags/extrasamples.html) 

Ideally I'd transfer to the layer the information about the tiff as plain as possible to the layer metadata, so if there is some error in interpretation, we can fix the code and we don't have to fix the layer.

Notice that [photometric interpretation ](https://www.awaresystems.be/imaging/tiff/tifftags/photometricinterpretation.html) can be WhiteIsZero or BlackIsZero that are grays too. 

So I suggest to : 
1. Isolate the metadata extraction function utility so we can allow also refresh of this data in the layer properties in the future. 
2. Save the important metadata as they are, not saving aggregated properties like `isRGB` that was wrongly calculated. In particular you may need to save: 
    - `samples`. I don't know if you need to check if they are 1 byte long. The count should provide the number of "bands" (extra-samples excluded I think). 
    - the `fileDirectory` including `PhotometricInterpretation` value
    - probably the extra-samples (look at [this PR](https://github.com/geotiffjs/geotiff.js/pull/108/files) in `geotiff.js` to see if it is included or not in samples)
3. From these information, you can desume the "isRGB" information and provide the proper styler.

web/client/api/catalog/COG.js
offtherailz: Replied here: 
https://github.com/geosolutions-it/MapStore2/pull/9857#issuecomment-1895513693

dsuren1:
@offtherailz 
> * [ ]  Sometimes it switches to one band randomly

This is because, on `Search` COG layer from the catalog when the cache is expired, the metadata is not fetched as the operation is **restricted** to adding the COG layer as a service. This could also result in a projection warning in the catalog as the metadata information holding projection information is missing.

In the future enhancement, the fetch metadata operation can also be performed from layer settings ex: style editor, should help us handle this case. For now can I just add a warning message in Style Editor letting the metadata information is missing and hence default view is displayed?

web/client/api/catalog/COG.js
```diff
@@ -130,19 +44,19 @@ export const getRecords = (_url, startPosition, maxRecords, text, info = {}) =>
                 title: record.title,
                 type: COG_LAYER_TYPE,
                 sources: [{url}],
+                // sources: [{"url": "https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/31/T/GJ/2022/7/S2A_31TGJ_20220703_0_L2A/B04.tif", "max": 3000}, {"url": "https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/31/T/GJ/2022/7/S2A_31TGJ_20220703_0_L2A/B03.tif", "max": 3000}, {"url": "https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/31/T/GJ/2022/7/S2A_31TGJ_20220703_0_L2A/B02.tif", "max": 3000}, {"url": "https://sentinel-cogs.s3.us-west-2.amazonaws.com/sentinel-s2-l2a-cogs/31/T/GJ/2022/7/S2A_31TGJ_20220703_0_L2A/B08.tif", "max": 3000}],
```
offtherailz, line 47:
```suggestion
```

offtherailz (changes requested):
I understood that you want to allow min/max source data but editing it in the styler looks uncorrect. 

Other stylers allow to edit min/max per band, for instance QGIS

image

I noticed you provided a tooltip to clarify it but it could not be enough. let me take a look at this together with @tdipisa  before to see how to proceed. 

- [ ] In particular I noticed that these values are anyway applied, even if I deactivate the layer.

[screencast-mail.google.com-2024.01.22-11_58_19.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/326022c0-c6b9-4c1c-ba70-0ea3cfdb1b65)

I understand the reason but from the user experience this is very confusing. 

- [ ] saving whole `fileDirectory` looks to be potentially an issue. It contains a lot of data that may potentially make the JSON explode (see `TileOffsets` , `TileByteCounts`, `GeoKeyDirectory`). Please fix and save only the data you effectively need, not relying on them.
- [ ] Save the style in a format compatible with other stylers. Here you are storing it as a string, but it should be ajson. Moreover it should be useful to make it consistent with the other style formats. 
```
{ "style": { "body": {
                    "color": ["array",["band",1],["band",2],["band",3],["band",4]]"
                }, "format": "openlayers"
```
- [ ] please document the raster style in md doc too.

offtherailz:
confirmed the format with @allyoucanmap , let me talk about min and max with tobia.

dsuren1:
@offtherailz 
> * [ ]  In particular I noticed that these values are anyway applied, even if I deactivate the layer.

From the video attached, I see the band styling is only disabled, meaning the bands selected from the channels will be ignored but min and max as these are independent. Min/max can be cleared if the user chooses to apply only source data values. Additionally, there are cases where only either is needed, for this reason, I have provided an option to disable band styling.

> * [ ]  saving whole `fileDirectory` looks to be potentially an issue. It contains a lot of data that may potentially make the JSON explode (see `TileOffsets` , `TileByteCounts`, `GeoKeyDirectory`). Please fix and save only the data you effectively need, not relying on them.

My thoughts exactly and I wanted to discuss when PR is reviewed. Glad you noticed. I will add only the `PhotometricIntepretation` prop for now and add the props as we need in future enhancements.

tdipisa:
@offtherailz some test maps you can use: 
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/46403
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/COG/47132

tdipisa:
@dsuren1 can you please fix conflicts?

offtherailz:
I performed the tests on this map:

http://localhost:8081/#/context/COG/46403

Here my findings:

# Bugs

- The stile toggle is enabled, but if I turn off and on, the style do not work. 

[screencast-mail.google.com-2024.03.14-18_01_02.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/4ba9acc5-31b8-4190-8493-f32b40d04b6d)

# Things to discuss

It is not easy to review the PR because of this regression not related to the PR (verified also on master):

If I add layers to the map, now I don't have the information for zoom etc.. anymore. `Bolivia_LC08_L1TP_001069_20190719_MSf` layer do not seems to work on the map, and I have some errors of incompatibility too, when adding layers to the map. 
@tdipisa we should scheduele a fix for this.

tdipisa:
@offtherailz I think it is better if you open a dedicated issue to report the regression you have identified with proper steps to reproduce it, mentioning also if it is affecting only master or also the release.

offtherailz:
Here the issue @tdipisa https://github.com/geosolutions-it/MapStore2/issues/10072

dsuren1:
@offtherailz @tdipisa 
> Bugs
> * The stile toggle is enabled, but if I turn off and on, the style do not work.
> 
>  screencast-mail.google.com-2024.03.14-18_01_02.webm 

The problem is caused by missing metadata as explained [here](https://github.com/geosolutions-it/MapStore2/issues/10072#issuecomment-2006060930)

web/client/plugins/styleeditor/MultiBandEditor.jsx
```diff
@@ -0,0 +1,206 @@
...
+     * i.e Multi extra samples are not supported
+     * Sample >=3 is generally considered 'RGB' as the GeoTIFF source is set with convertToRGB:'auto'
+     */
+    const samples = get(layer, "sourceMetadata.samples");
+
+    /**
+     * There are instances where the sample is 3 or above with PhotometricInterpretation as 0 or 1 [gray scale indicator],
+     * this could be because the band channels are not properly defined.
+     * Hence we consider if the sample is >=3 or PhotometricInterpretation is not a gray scale image,
+     * to determin the RGB image
```
offtherailz, line 70:
```suggestion
     * to determine the RGB image
```

offtherailz (changes requested):
We are almost in a stable version I think. We have to fix the following: 

- [x]  "enableBandStyling" property should not be used, but the style should depend only on the style object. I think the presence of the body.color section is enough to determine if it is enabled or not, but if you need more, add a flag and proper document it.
- [ ] For certain cogs is still unclear to me how to style. For instance this https://cogeo.craig.fr/opendata/ortho/2018_25cm_ain+isere.tif the styler do not look to be working at all. Maybe because of photometryic interpretation : 6 ?  

a final note: 
Having to save catalogs again and again is counter-intuitive at the end. If we need more metadata, we need to think about a way to update/auto-update layers source metadata.

dsuren1:
@offtherailz 
> * [ ]   "enableBandStyling" property should not be used, but the style should depend only on the style object. I think the presence of the body.color section is enough to determine if it is enabled or not, but if you need more, add a flag and proper document it.

You are right. This prop is not needed. Might have skipped the coffee :wink:

> * [ ]  For certain cogs is still unclear to me how to style. For instance this https://cogeo.craig.fr/opendata/ortho/2018_25cm_ain+isere.tif the styler do not look to be working at all. Maybe because of photometryic interpretation : 6 ?

At times, the alpha channel may prevent the visualization of certain TIFF sources, and the composition of some TIFF sources may disregard the alpha channel. To mitigate this issue and to eliminate no-data tiles surrounding the image, I have incorporated the nodata attribute into the source (**TODO**: should made configurable in the future enhancement), which introduces an alpha channel and facilitates the better visualization of the TIFF image.

**2018_25cm_ain+isere.tif**
image


> a final note: Having to save catalogs again and again is counter-intuitive at the end. If we need more metadata, we need to think about a way to update/auto-update layers source metadata.

At least after this change, all the required metadatas are fetched and persisted. But yes, maybe a force fetch metadata option somewhere in the layer settings is needed in the future.

tdipisa (commented):
@dsuren1 

>At least after this change, all the required metadatas are fetched and persisted. But yes, maybe a force fetch metadata option somewhere in the layer settings is needed in the future.

I think it might be nice to have it soon, maybe adding a layer refresh button in the toolbar valid for all levels which for COGs also updates the level metadata.

offtherailz (approved):

offtherailz:
@ElenaGallo, could you please test this on  DEV  ? Thank you

ElenaGallo:
Test passed, @dsuren1 please backport to 2024.01.xx. Thanks

PR #9967, opened 2024-02-09
C039 - Fixing issues of Math.random()

MV88:
## Description

fixing some vulnerabilities related to Math.Random in a dedicated branch **C039-webmapper-28-10-2022**

see here for more details https://kemilbeltre.medium.com/why-do-not-use-math-random-a6f8b0ad38dd
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#<issue>

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/TOC/fragments/legend/Legend.jsx
+++ b/web/client/components/TOC/fragments/legend/Legend.jsx
@@ -18,6 +18,9 @@ import {
     addAuthenticationToSLD,
     clearNilValuesForParams
 } from '../../../../utils/SecurityUtils';
+import {
+    randomInt
+} from '../../../../utils/cesium/MathUtils';
 import Message from '../../../I18N/Message';
 
 class Legend extends React.Component {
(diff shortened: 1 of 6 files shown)


offtherailz:
Sorry, but the usage here is not cryptographic, so using strong `Math.random()` looks lecit to me.

MV88:
> Sorry, but the usage here is not cryptographic, so using strong `Math.random()` looks lecit to me. Did I missed something? also the article you quote says:
> 
> > You can use [Math.random()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) when the expected value does not compromise the app.

i would agree with you that if Math.random is used in non critical parts is fine to use, altough one of our client has in place a fortify scansion which spotted this as high vulnerability. the thing is that they already marked them as reliable vulnerabilities so i decided to solve them by using another method.

image

Now we could ask the client to unmark them, or 
we can add a rule in our eslint to avoid Math.random method or
we can fix it in another way

for the future we may decide for that client to mark these issues as not vulnerabilities

> 
> > Use strong generation methods as [Crypto.getRandomValues()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) when required.
> 
> That I can understand as "not always"

idk what you mean here tbh

PR #10006, opened 2024-03-01
#10005 update babel configuration

### LINKED ISSUE #10005: Update babel and eslint
## Description

A recent PR sent to a project asked to update babel (and so dependent plugins/eslint) to fix a build problem on mac. 
We should port this Version update on MapStore core. 

Trying to do it, we have several errors in eslint (due to old bugs of eslint not revealing some items) alrady notified in the past by @MV88 (see [this](https://github.com/geosolutions-it/MapStore2/issues/9190) and [this](https://github.com/geosolutions-it/MapStore2/pull/9188)  

This issue want to propose an update strategy for it, as easy as possible.

offtherailz:
## Description

Note: in order to make this PR build correctly, we need to publish eslint node lib. I published it

As explained in 10005 updating babel needs to update eslint too, and this causes several lint error because of old bugs of eslint not revealing the failing parts.

Here I attach the full set of errors.

[eslint-errors.txt](https://github.com/geosolutions-it/MapStore2/files/14459341/eslint-errors.txt)

The errors can be divided in 2 macro-blocks:

- PropTypes used but not defined (>300 errors)
- Other errors (mostly save-state in `componentDidMount` )

For this reason I propose the following: 
- Modify the eslint rules to : 
   - remove the rule that fixes prop-types
   - Remove error of order in class components methods (we are not creating new class components anymore)
   - Disable errors for the remaining lines via comment

We can evaluate to re-introduce the rule of proptypes later and fix the pending eslint errors disabled, but this way we guarantee:
- Update and future updatability
- functionalities to be preserved
- future check of rules that was previously fixes.
   
 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [x] Other... Please describe: Dependency update  

## Issue

Fix #10005

**What is the new behavior?**
Babel updated. Eslint rule catch now finds errors that was not visible before.

## Breaking change
 - [x] No

## Other useful information

I also fixes some require/exports to avoid some lint errors too. 

I noticed babel plugin `plugin-proposal-class-properties` is deprecated and we should use `plugin-transform-class-properties` instead. 
I didn't do this update in this PR, concentrating the effort to a working solution.


utility/eslint/package.json
```diff
@@ -24,7 +24,7 @@
     "@babel/eslint-plugin": "7.11.5"
   },
   "peerDependencies": {
-    "@babel/plugin-proposal-class-properties": ">= 7.8.3",
+    "@babel/plugin-transform-class-properties": ">= 7.8.3",
```
offtherailz, line 27:
```suggestion
    "@babel/plugin-proposal-class-properties": ">= 7.8.3",
```

tdipisa:
@offtherailz we are too close to 2024.01.00 so the plan would be to include this to the next major 2024.02. Let's update this PR and merge on master early after 2024.01 is delivered

allyoucanmap (approved):

PR #10039, opened 2024-03-07
Refactor and simplify the format validator utility logic

DavGoeck:
## Description

When working with MapStore2 in a client project I discovered that the format validator logic is more complicated than necessary. I simplified the code in various tiny-sliced commits for clarity that can be squashed when pulled.

This is Code Style Update and Refactoring (except that the API of the validators was changed, but it was only used in the tests).

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Breaking change
 - [x] No

## Other useful information
I am submitting this Pull Request as an external developer on behalf of DB Systel GmbH.

I found various other code segments where the build-in JS Array method "reduce(method, aggregator)" was used in a way that .filter(predicate) would have sufficed, making it more difficult than necessary to read. This was the starting point for this PR.
Is there a specific reason why this is used? Do you wish more refactorings of this sort?

Kind regards!

### DIFF
--- a/web/client/utils/FeatureInfoUtils.js
+++ b/web/client/utils/FeatureInfoUtils.js
@@ -57,15 +57,16 @@ export function parseXMLResponse(res) {
     }
     return false;
 }
-
-const textNotEmptyAndValid = (res) => res.response !== "" && (typeof res.response === "string" && res.response.indexOf("no features were found") !== 0) && (typeof res.response === "string" && res.response.indexOf("<?xml") !== 0);
-const featuresPresent = (res) => res.response?.features?.length;
+const htmlResponseValid = (res) => !!parseHTMLResponse(res);
+const xmlResponseValid = (res) => !!parseXMLResponse(res);
+const textNotEmptyAndValid = (res) => !!(res.response !== "" && (typeof res.response === "string" && res.response.indexOf("no features were found") !== 0) && (typeof res.response === "string" && res.response.indexOf("<?xml") !== 0));
+const featuresPresent = (res) => !!(res.response?.features?.length);
 const truePredicate = () => true;
 
 const predicate = (format) => {
     switch (format) {
-    case "HTML": return parseHTMLResponse;
-    case "GML3": return parseXMLResponse;
+    case "HTML": return htmlResponseValid;
+    case "GML3": return xmlResponseValid;
     case "TEXT": return textNotEmptyAndValid;
     case "JSON": return featuresPresent;
     case "GEOJSON": return featuresPresent;
--- a/web/client/utils/FeatureInfoUtils.js
+++ b/web/client/utils/FeatureInfoUtils.js
@@ -58,78 +58,25 @@ export function parseXMLResponse(res) {
     return false;
 }
 
-export const Validator = {
-    HTML: {
-        /**
-         *Parse the HTML to get only the valid html responses
-         */
-        getValidResponses(responses) {
-            return responses.filter(parseHTMLResponse);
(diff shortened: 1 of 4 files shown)


DavGoeck:
I'd like to refresh this PR. As commended, I am not sure whether something went wrong with the pipeline or not. It would be great to get some feedback on the issue.

tdipisa:
@DavGoeck thank you so much for your contribution. We will review as soon as possible.

offtherailz (approved):
Thank you very much for your contribution. The division in commits made the PR a lot easier to be reviewed.
Thank you very much. This simplifies a lot the logic of feature info and makes it more maintainable and readable for the future, keeping the functionality untouched.

offtherailz (changes requested):
Could you please check and fix the failing tests in order to make this PR mergiable. 
I reviewed this PR looking at the code and testing the functionality, and in the meanwhile I allowed CI, so I didn't noticed the test failure in my previous review.

Thank you.

offtherailz:
> I found various other code segments where the build-in JS Array method "reduce(method, aggregator)" was used in a way that .filter(predicate) would have sufficed, making it more difficult than necessary to read. This was the starting point for this PR.
Is there a specific reason why this is used? Do you wish more refactorings of this sort?

About this question @DavGoeck There is no particular reason in MapStore to use a reduce where a filter can be used, if there are no corner-cases when the two results may be different. So if you want to provide simplification PRs of this kind, you are welcome.

DavGoeck:
Hey of course, I thought I noticed all failing tests. I will fix it next chance I get.

offtherailz:
Hi, I noticed that the PR is still failing. 
I ping you just in case you didn't noticed, of course if you didn't have time to look at it ignore the message :). 

Btw for the future (we will add something to contributing guidelines soon) force-pushing on a PR that have already a review makes hard the following reviews (in fact destroying the point reached to the review and making necessary to re-review all the files, not only the new changes). 
Internally we avoid to do it, but it is not yet in contribution guidelines, so it's perfectly ok you didn't know it.

DavGoeck:
Hey @offtherailz, I realized that there is in fact a slight API change introduced in my Code. I was aware of one such example and have written it in one of the commit messages, but it seems that there is a general change introduced.

I think the change is good, but I am unsure whether I truly know all the details. I will explain.

I wrote the code so that for each type of response there is a predicate checking whether a response of the given type is valid or not. If so, it can remain in the array. If not, it would be included in the invalid array. However, in the code so far there have been examples of invalid responses that would not be included in the getNoValidResponses Array.

For example. If I have a String validator, then a response that is not of type string is included neither in getValidResponses nor getNoValidResponses. Similarly, a JSON response without a "features" member defined would be neither in getValidResponses nor getNoValidResponses. This changes with my implementation. I think the change is good, but I understand that it is now no longer 100% a refactoring.

The question is, whether you think that there is any bad side-effect from having these extra responses in the getNoValidResponses Array. The output of getValidResponses will stay the same for any call.

Kind regards,
David

offtherailz:
Hi @DavGoeck 
I agree with you in all the parts. I think your changes are fine, but I have to check and test the new implementation in every format and with all possible results, to see the effective result.

offtherailz (approved):
I tested and it looks to work well in all the use cases and corner cases I found. The solution looks correct and clear to me.

offtherailz:
I Did it see #10224
@DavGoeck Thank you very much for your contribution.

Issue #10034, opened 2024-03-06
CSW Catalog - Include `SortyBy` parameter in the CSW Catalog `<csw:Query>`


ale-cristofori:
When composing a CSW query, to be sent to the CSW service, MapStore offers the possibility to include a dynamic filter. The filter can be composed and edited in the MapStore UI as XML text. Example below

![Image](https://github.com/geosolutions-it/MapStore2/assets/14953970/88b4e1f6-be27-4fd3-8e0d-bafbf1f31ad8)

The new feature will allow the user to include the `SortBy` predicate in the CSW query so that results can be retrieved in sorted order.
How the predicate is included could either be through text or adding a UI element to include the attributes the user wants to filter and the sorting order (ASC or DESC). An example of a resulting query containing the `SortBy` predicate could look as follows.

```xml
<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" xmlns:ogc="http://www.opengis.net/ogc" xmlns:gml="http://www.opengis.net/gml" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:dct="http://purl.org/dc/terms/" xmlns:gmd="http://www.isotc211.org/2005/gmd" xmlns:gco="http://www.isotc211.org/2005/gco" xmlns:gmi="http://www.isotc211.org/2005/gmi" xmlns:ows="http://www.opengis.net/ows" service="CSW" version="2.0.2" resultType="results" startPosition="1" maxRecords="4">
	< typeNames="csw:Record">
		<csw:ElementSetName>full</csw:ElementSetName>
		<csw:Constraint version="1.1.0">
			<ogc:Filter>
				<ogc:And>
					<ogc:PropertyIsLike wildCard='*' singleChar='_' escapeChar='\\'>
						<ogc:PropertyName>csw:AnyText</ogc:PropertyName>
						<ogc:Literal>alberi*</ogc:Literal>
					</ogc:PropertyIsLike>
					<ogc:Or>
						<ogc:PropertyIsEqualTo>
							<ogc:PropertyName>dc:type</ogc:PropertyName>
							<ogc:Literal>dataset</ogc:Literal>
						</ogc:PropertyIsEqualTo>
						<ogc:PropertyIsEqualTo>
							<ogc:PropertyName>dc:type</ogc:PropertyName>
							<ogc:Literal>service</ogc:Literal>
						</ogc:PropertyIsEqualTo>
[...]

PR #10077, opened 2024-03-18
#10034: Shown exception text from Catalog in the notification

dsuren1:
## Description
This PR enhances the error notifications in Catalog. The exception texts of the response are shown in the notification and network error are shown with generic error

 - [x] Enhancement

## Issue

- https://github.com/geosolutions-it/MapStore2/issues/10034#issuecomment-1997759268

**What is the new behavior?**
### Exception text
Screenshot 2024-03-18 at 7 33 05 PM
Screenshot 2024-03-18 at 7 32 42 PM

### Generic error
image

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/catalog.js
+++ b/web/client/epics/catalog.js
@@ -72,10 +72,16 @@ import { getResolutions, METERS_PER_UNIT } from "../utils/MapUtils";
 import { describeFeatureType } from '../api/WFS';
 import { extractGeometryType } from '../utils/WFSLayerUtils';
 import { createDefaultStyle } from '../utils/StyleUtils';
+import { removeDuplicateLines } from '../utils/StringUtils';
+
 const onErrorRecordSearch = (isNewService, errObj) => {
     // Exception text is shown as is while the network errors are shown
     // with generic error message in the notification
-    const [errorMsg] = castArray(errObj?.error);
+    let [errorMsg] = castArray(errObj?.error);
+    if (errorMsg) {
+        // Remove any instance of duplicated line string from the exception text
+        errorMsg = removeDuplicateLines(errorMsg);
+    }
     console.warn("Catalog error", errorMsg);
 
     if (isNewService) {
--- a/web/client/epics/catalog.js
+++ b/web/client/epics/catalog.js
@@ -76,6 +76,8 @@ import { removeDuplicateLines } from '../utils/StringUtils';
 import { logError } from '../utils/DebugUtils';
 
 const onErrorRecordSearch = (isNewService, errObj) => {
+    logError({message: errObj});
+
     // Exception text is shown as is while the network errors are shown
     // with generic error message in the notification
     let [errorMsg] = castArray(errObj?.error);
@@ -83,8 +85,6 @@ const onErrorRecordSearch = (isNewService, errObj) => {
         // Remove any instance of duplicated line string from the exception text
         errorMsg = removeDuplicateLines(errorMsg);
     }
-    logError({message: errorMsg});
-
     if (isNewService) {
         const message = errorMsg
             ? truncate(errorMsg, { length: 400 })
--- a/web/client/epics/catalog.js
+++ b/web/client/epics/catalog.js
@@ -9,7 +9,7 @@
 import * as Rx from 'rxjs';
 import axios from 'axios';
 import xpathlib from 'xpath';
-import {head, get, find, isArray, isString, isObject, keys, toPairs, merge, castArray} from 'lodash';
+import {head, get, find, isArray, isString, isObject, keys, toPairs, merge, castArray, truncate} from 'lodash';
 
 import {
     ADD_SERVICE,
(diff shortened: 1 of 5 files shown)


tdipisa (commented):
@dsuren1 from the exception XML output reported in https://github.com/geosolutions-it/MapStore2/issues/10034#issuecomment-1997478994 it should be possible to include the logic to parse the _ows:ExceptionText_ to report a clean and clear notification without duplications. Right?

dsuren1:
> @dsuren1 from the exception XML output reported in [#10034 (comment)](https://github.com/geosolutions-it/MapStore2/issues/10034#issuecomment-1997478994) it should be possible to include the logic to parse the _ows:ExceptionText_ to report a clean and clear notification without duplications. Right?

Yes. Updated screenshot accordingly

tdipisa:
@dsuren1 thank you. @MV88 please proceed with the review.

web/client/epics/catalog.js
```diff
@@ -72,19 +72,33 @@ import { getResolutions, METERS_PER_UNIT } from "../utils/MapUtils";
...
+    let [errorMsg] = castArray(errObj?.error);
+    if (errorMsg) {
+        // Remove any instance of duplicated line string from the exception text
+        errorMsg = removeDuplicateLines(errorMsg);
+    }
+    console.warn("Catalog error", errorMsg);
+
     if (isNewService) {
+        const message = errorMsg
+            ? truncate(errorMsg, { length: 200 })
```
MV88, line 89: i think it would be better to try to recognize the error message and use a short generic message, and if not recognized it will use the previous one.

web/client/epics/catalog.js
```diff
@@ -72,19 +72,33 @@ import { getResolutions, METERS_PER_UNIT } from "../utils/MapUtils";
...
+
 const onErrorRecordSearch = (isNewService, errObj) => {
+    // Exception text is shown as is while the network errors are shown
+    // with generic error message in the notification
+    let [errorMsg] = castArray(errObj?.error);
+    if (errorMsg) {
+        // Remove any instance of duplicated line string from the exception text
+        errorMsg = removeDuplicateLines(errorMsg);
+    }
+    console.warn("Catalog error", errorMsg);
```
MV88, line 85: please use logError from debug utils instead so this will appear only in debug mode

MV88 (changes requested):

web/client/epics/catalog.js
dsuren1: I added it because we faced a similar case where notification occupied a whole app height when exception is lengthier.

@tdipisa @MV88 
Anyway, if the error message is of exception text, then we may show a generic message as **"Failed to save service, please check the attributes of the service"**. Your thoughts?

web/client/epics/catalog.js
dsuren1: I intended for this error to be shown in the console, so the user can easily identify the cause, if any. Kindly let me know if you still prefer to show it only in debug mode

MV88: long time ago (when that function have been created), I was asked from @offtherailz to use console log for showing errors only in debug mode. this can be extended to warnings even if in this case is more appropriate using console.error

if we want to change this convention we have to discuss internally within the team. 
so yes, please from now one consider using it

web/client/epics/catalog.js
tdipisa: @dsuren1 it would be good to have a precise message in this case otherwise it is unuseful. We could consider to leave it as it is but maybe increasing the length a bit more still to an acceptable value (eg. 400?). Let's see together @dsuren1. A solution could also be to improve the component to be expandable when needed but it should be a separated issue.

web/client/epics/catalog.js
offtherailz: I didn't see this PR in detail. 
I can also add some guidelines for general policy, so you can decide what to do:

When an error occours, the application need to handle it, in order to make the user aware of how to proceed, introducing less details as possible, all on UI. The user should never look to the console. 

If the exception is one of the errors that we predicted to happen and maange, this means that it is involved in the normal application workflow, so logging it is  an option only for developers, so in this case a debug-only log makes sense. 

If the error is **not expected** (so it is not one of the errors we predicted to handle), than logging in the console is correct (not masking to the user), so in future we can decide to handle also it. 
The ui in this case should provide a generic error message, with some details if possible.

web/client/epics/catalog.js
```diff
@@ -82,11 +83,11 @@ const onErrorRecordSearch = (isNewService, errObj) => {
         // Remove any instance of duplicated line string from the exception text
         errorMsg = removeDuplicateLines(errorMsg);
     }
-    console.warn("Catalog error", errorMsg);
+    logError({message: errorMsg});
```
MV88, line 86: you should use here the original error, not the parsed one

MV88 (changes requested):

MV88 (approved):

MV88:

ElenaGallo:
Test passed, @dsuren1 please backport to 2024.01.xx. Thanks

Issue #10081, opened 2024-03-19
.geojson file not supported with mac os


congchen1101:
## Description
With MacOS .geojson files will be identified as "application/geo+json" type. So the current code will not look for it in MIME_LOOKUP list. 
 
## How to reproduce

- use a mac os system
- go into one map
- click the import button(import a map file or vector data)  from sidebar menu in the right side of the page
- select one .geojson file to import

*Expected Result*
a modal which can configure the style for this geojson data should appear.

*Current Result*
file not supported message showed

## Other useful information
![Screen Shot 2024-03-19 at 09 31 57 AM](https://github.com/geosolutions-it/MapStore2/assets/161452326/a2efce1e-3774-459b-8327-6fa82b608cc4)

PR #10082, opened 2024-03-19
fixing .geojson file not supported with mac os

congchen1101:
FixBug .geojson file not supported in macos.
- In macOS .geojson file will be identified as type of "application/geo+json". So in checkFileType and readFile functions, they will not go into MIME_LOOKUPS list to find the corresponding type.
- Check first in MIME_LOOKUPS list and then the file.type will solve this problem. On behalf of DB Systel GmbH

## Description

- This is a minimal change in code, so a test and doc was not added

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#10081 

**What is the new behavior?**
now import geojson is also supported with macOS

## Breaking change
 - [x] No

## Other useful information
“On behalf of DB Systel GmbH”

### DIFF
--- a/web/client/components/import/dragZone/enhancers/__tests__/testData.js
+++ b/web/client/components/import/dragZone/enhancers/__tests__/testData.js
@@ -16,7 +16,8 @@ export const getFile = (url, fileName = "file") =>
         responseType: 'arraybuffer'
     }))
         .map( res => {
-            return new File([new Blob([res.data], {type: res.headers['response-type']})], fileName);
+            const contentType = res.headers['content-type'] === 'null' ? undefined : res.headers['content-type'];
+            return new File([res.data], fileName, {type: contentType});
         });
 
 // PDF_FILE: new File(b64toBlob('UEsDBAoAAAAAACGPaktDvrfoAQAAAAEAAAAKAAAAc2FtcGxlLnR4dGFQSwECPwAKAAAAAAAhj2pLQ7636AEAAAABAAAACgAkAAAAAAAAACAAAAAAAAAAc2FtcGxlLnR4dAoAIAAAAAAAAQAYAGILh+1EWtMBy3f86URa0wHLd/zpRFrTAVBLBQYAAAAAAQABAFwAAAApAAAAAAA=', 'application/pdf'), "file.pdf"),
--- a/web/client/components/import/dragZone/enhancers/processFiles.jsx
+++ b/web/client/components/import/dragZone/enhancers/processFiles.jsx
@@ -24,7 +24,8 @@ import {
     readWMC,
     readZip,
     recognizeExt,
-    shpToGeoJSON
+    shpToGeoJSON,
+    readGeoJson
 } from '../../../../utils/FileUtils';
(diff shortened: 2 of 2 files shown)


congchen1101:
Thank you for your suggestions. While attempting to add unit tests, I discovered that unit tests for reading geojson files already existed. This led me to wonder why those tests were passing despite the issues at hand. Upon further investigation into the existing tests in `processFiles-test.jsx`, I noticed that most of them utilize the `getFile` method from `testData.js`. This method employs an axios.get call to read files from local test resources, attempting to extract the `response-type` value from the headers as the type to create a new **Blob** object, which is then used to create a **File** object.
```
export const getFile = (url, fileName = "file") =>
    Rx.Observable.defer( () => axios.get(url, {
        responseType: 'arraybuffer'
    }))
        .map( res => {
            return new File([new Blob([res.data], {type: res.headers['response-type']})], fileName);
        });
```
However, I encountered two main issues. 

1. The `res.headers` does not contain a `response-type`, meaning that all newly created **Blob** objects lack a type attribute. This flaw explains why all current tests could pass: the absence of a type attribute in the checks leads the code to searching file extend in MIME_LOOKUPS table for the corresponding type. 
2. Even after adjusting `response-type` to `content-type`, which allowed the new **Blob** object to successfully have a type property, the new **File** object still failed to inherit the **Blob**'s type attribute.

Given these observations, I opted to directly assign the `content-type` as the type value when creating the new File object, rather than doing so during the **Blob** object's creation. This approach appears to address the core issue more directly and ensures that the intended file type is recognized correctly in tests and code functionality.
```
export const getFile = (url, fileName = "file") =>
    Rx.Observable.defer( () => axios.get(url, {
        responseType: 'arraybuffer'
    }))
        .map( res => {
            const contentType = res.headers['content-type'] === 'null' ? undefined : res.headers['content-type'];
            return new File([res.data], fileName, {type: contentType});
        });
```

I've now also added code to support geojson. However, I noticed that the `readGeoJson` method returns an object includes an `errors` list. Currently, my code does not perform any actions based on the presence of errors in this list, such as throwing a new Error. Implementing such behavior would lead to the failure of existing tests, as the **_caput-mundi.geojson_** file in test resources include the `crs` attribute. This triggers the error `'old-style crs member is not recommended, this object is equivalent to the default and should be removed'` when the `geojsonhint` method is called within `readGeoJson`.

I would greatly appreciate any further advice you could provide. Thank you.

dsuren1 (approved):

dsuren1:
@tdipisa @offtherailz Since this scenario cannot be tested effectively on a window's machine. Can I proceed with the backport?

tdipisa:
You can test with MAC but let's also wait for functional tests as usual made by @ElenaGallo. No? Did I miss something?

dsuren1:
@tdipisa @ElenaGallo Works fine in DEV and tested on a Mac with geosjon [file](https://github.com/user-attachments/files/17474477/india-osm.geojson.zip)

ElenaGallo:
testpassed, @dsuren1 please backport to 2024.02.xx. Thanks

Issue #6913, opened 2021-05-26
Illogical tab order


magnusnil:
## Description
When using tab to shift focus using the keyboard, the order in which the tools are activated doesn´t make sense. It starts top left with Layers, then moves to the lower right corner, then to the tools (such as the zoom button) and then back to home page. It should be in a more straight forward order, such as left to right.

Tested in Chrome and Microsoft Edge, using MapStore 2021.01.03.

fkellner:
PR https://github.com/geosolutions-it/MapStore2/pull/9983 has fixed tab navigation within the layers. However, even if the TOC is hidden, the controls within it are still tabbable.

On the home page:
- Cards are not accessible via keyboard (will do a PR for grid cards fixing this this afternoon)
- Login-Button cannot be tabbed to

fkellner:
Edit: Login button can be reached with `Arrow Down` once the menu is active

PR #10150, opened 2024-04-03
geosolutions-it#6913: grid cards keyboard navigation

fkellner:
## Description

Makes grid cards keyboard navigable and accessible again (with tab + enter,
you can reach a map now). 

Does not change how anything looks for non-keyboard users.

 - [x] Bugfix

## Issue

#6913 Reaching a map only using the keyboard is impossible. 
Power users and people with motoric impairments often prefer
to navigate a website using the keyboard rather than the mouse.

**What is the new behavior?**

The grid cards, which represent buttons to maps, now can be focused
via keyboard and pressed via Enter button.

## Breaking change

 - [x] No

## Other useful information

Tests only include the "clicking with the Enter button" behaviour, since the testing framework does not seem to
support side effects like focusing elements with the tab button.

### DIFF
--- a/web/client/components/misc/GridCard.jsx
+++ b/web/client/components/misc/GridCard.jsx
@@ -47,7 +47,11 @@ class GridCard extends React.Component {
         return (<div
             style={this.props.style}
             className={"gridcard" + (this.props.className ? " " + this.props.className : "")}
-            onClick={this.props.onClick}>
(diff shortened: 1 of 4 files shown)


tdipisa:

allyoucanmap:
@ElenaGallo please test this improvement on dev, thanks

Issue #10158, opened 2024-04-04
WCAG Accessibility Improvements - Low Hanging Fruits


fkellner:
## Description

![image](https://github.com/geosolutions-it/MapStore2/assets/28141677/3617e53e-f964-449e-8dc4-207ac0b121bb)

Running the MapStore dashboard through Lighthouse currently yields a bad accessibility score of 62, and the
map view does not score much better (66). As a german company using MapStore, we are required by law to 
put some effort into the accessibility of public facing websites.

Some changes that we did to improve the score are relatively painless and trivial, others, especially relating
to the labelling of buttons, introduce some small overhead. So, I will provide two PRs with our changes:

- One with the tiny changes that should not be an issue with the rest of the code (like setting document language, alt tag for logo), but already improve the score considerably
- One with the more complicated changes (mainly button labels with i18n). 

I hope the first one should not be a problem for you to merge, but the second one I am interested in your feedback, but it is not a high priority enhancement. 

A small accessibility issue in the navbar is currently caused by a `react-bootstrap`-Component, once you move on to the next major version, it should be fixed. 

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x] Minor changes to existing features

PR #10159, opened 2024-04-04
#10158 painless accessibility improvements

fkellner:
## Description

Tiny changes that improve accessibility. Should hopefully not take longer than a minute to review.

 - [x] Bugfix

## Issue

#10158 

Bad Lighthouse Accessibility Score

**What is the new behavior?**

Better Lighthouse Accessibility Score through:
- Setting Document language
- Fixing a heading that should be a div
- providing a logo with an alt tag
- linking scale label with its form element
- enabling browser zoom

## Breaking change
 - [x] No

## Other useful information

Too much Zoom does seem to break the layout, but handing that responsibility to the end user is more accessible for people who need larger controls than
disabling it. Map zoom is unaffected, and since using browser zoom enlarges map controls, users who tried zooming
into the map using browser zoom should be able to notice the actual controls and recover from their mistake.

### DIFF
--- a/web/client/components/I18N/Localized.jsx
+++ b/web/client/components/I18N/Localized.jsx
@@ -37,6 +37,7 @@ class Localized extends React.Component {
             if (typeof children === 'function') {
                 children = children();
             }
+            document.documentElement.setAttribute("lang", this.props.locale);
 
             return (<IntlProvider key={this.props.locale} locale={this.props.locale}
                 messages={this.flattenMessages(this.props.messages)}
--- a/web/client/components/I18N/Localized.jsx
+++ b/web/client/components/I18N/Localized.jsx
@@ -30,6 +30,16 @@ class Localized extends React.Component {
         };
     }
 
+    componentDidMount() {
+        this.updateDocumentLangAttribute();
+    }
+
+    componentDidUpdate(prevProps) {
+        if (this.props.locale !== prevProps.locale) {
+            this.updateDocumentLangAttribute();
+        }
+    }
+
     render() {
         let { children } = this.props;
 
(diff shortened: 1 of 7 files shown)


tdipisa:

web/client/components/I18N/Localized.jsx
```diff
@@ -37,6 +37,7 @@ class Localized extends React.Component {
             if (typeof children === 'function') {
                 children = children();
             }
+            document.documentElement.setAttribute("lang", this.props.locale);
```
allyoucanmap, line 40: It would be better to move this in `componentDidMount` and `componentDidUpdate` and remove it from the `render` method, it could be something like:

```
componentDidMount() {
    this.updateDocumentLangAttribute();
}

componentDidUpdate(prevProps) {
    if (this.props.locale !== prevProps.locale) {
        this.updateDocumentLangAttribute();
    }
}

updateDocumentLangAttribute() {
    if (document?.documentElement) {
        document.documentElement.setAttribute("lang", this.props.locale);
    }
}
```

allyoucanmap (changes requested):
@fkellner thanks for the contribution, most of the changes are good, there is just to review how we set the lang in the Localized component

tdipisa:
@fkellner is this PR ready for a new review then? Thank you so much.

fkellner:
Yes, I did the requested changes and it is ready from my side

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test on dev and let me know if we can backport on 2024.02.xx, thanks

ElenaGallo:
Test passed, @allyoucanmap please backport on 2024.02.xx. Thanks

PR #10176, opened 2024-04-08
Fix #10175 : fix 2 services search test nondeterminism

### LINKED ISSUE #10175: Search Epic Test with two services is flaky
## Description

The search epic test `"produces the search epic with two services with more results than default limit"` loads two result files:
Wyoming (containing ids `states.x`) and Arizona (containing ids `states-ari.x`). 

It then tests that the combined limited search always produces 5 results from the Wyoming file, even after both services have loaded. However, which service source produces results first seems to be non-deterministic and differ between MacOS and Windows, with the test running on MacOS failing because the Arizona file loads first sometimes. 

## How to reproduce

Run the `search-test` on MacOS.

*Expected Result*

The test to run through.

*Current Result*

The test `"produces the search epic with two services with more results than default limit"` fails because it receives `states-ari.1` instead of `states.1`.

- [x] Not browser related

## Other useful information

We already fixed this in our version by 
- setting a different priority for the services
- accepting results from either file in the first emitted result action
- only accepting results from the file with higher priority in the second result action

but forgot to play it back. Will open a PR shortly.

fkellner:
## Description

Apparently, which search service observable emits results first does not depend on the order in which they are passed into the search. The test however assumed there to be an order, causing it to break on MacOS. 

New version:
- sets a different priority for the services
- accepts results from either file in the first emitted result action
- only accepts results from the file with higher priority in the second result action

Relevant source lines in the search epic:
```js
// merge all results from the streams
.mergeAll()
.scan((oldRes, newRes) => sortBy([...oldRes, ...newRes], ["__PRIORITY__"]))
```

 - [x] Bugfix

## Issue

#10175 

**What is the new behavior?**

Tests run through regardless of OS.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/__tests__/search-test.js
+++ b/web/client/epics/__tests__/search-test.js
@@ -532,7 +532,8 @@ describe('search Epics', () => {
                     typeName: 'topp:states',
                     queriableAttributes: [STATE_NAME],
                     returnFullData: false
-                }
+                },
+                priority: 2
             },
             {
                 type: 'wfs',
@@ -541,7 +542,8 @@ describe('search Epics', () => {
                     typeName: 'topp:states',
                     queriableAttributes: [STATE_NAME],
                     returnFullData: false
-                }
+                },
+                priority: 1
             }],
             maxResults
         };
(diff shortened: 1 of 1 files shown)


tdipisa:
@fkellner Thank you so much for your contribution

@MV88, let's evaluate a possible inclusion in 2024.01.01 (waiting for the review).

MV88 (approved):

MV88:
> @fkellner Thank you so much for your contribution
> 
> @MV88, let's evaluate a possible inclusion in 2024.01.01 (waiting for the review).

@tdipisa 
it's an improvement for a test, i think it is safe to include it in the next release 2024.02.00

@fkellner thanks for spotting this and for your contribution

MV88:
@ElenaGallo no need to test this

Issue #10026, opened 2024-03-06
Interactive legend for TOC layers


mahmoudadel54:
## Description

For WMS layers in TOC it would be interesting to enhance the legend to be interactive so that it is possible for example to switch off layer features belonging to certain classes. For this purpose the legend should be retrieved in JSON format and the UI component managed accordingly.

This is an investigation issue where it is requested to details the way to go with involved limitations if any.

To consider:

- JSON format is supported only by GS (to confirm)
- There should be an option to activate this new legend format in MS (the default is the current one). In Catalog tool for example
- Evaluate the possibility to uniform this functionality also for WFS/Vector layers client side.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [ ] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: Improvement to legend component 

## Other useful information
Sample of request:
https://gs-stable.geo-solutions.it/geoserver/wms?service=WMS&request=GetLegendGraphic&format=application/json&height=50&width=100&layer=gs%3Aus_states&style=pophade&version=1.3.0&SLD_VERSION=1.1.0&LEGEND_OPTIONS=forceLabels%3Aon&LANGUAGE=en&SCALE=36978669

PR #10180, opened 2024-04-09
#10026: Interactive legend for TOC layers [WMS]

mahmoudadel54:
## Description
In this PR, interactive legend functionality for WMS layers is implemented. User can configure enabling the interactive legend from advanced settings in catalog or from display tab in layer settings.
CQL filter format is used to apply this kind of layer filter. User can click on the interactive legend if the legend contain 'filter' expressions and it will reflect on the map. Any type of filter rather than the legend filter will keep the legend filter beside it.
Legend filter will be reset if user change the style and user can reset it manually by clicking on the selected filter legend again to remove the legend filter.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#10026 

#10026 

**What is the new behavior?**
User now can configure and enable interactive legend for WMS and apply filter based on legend filter values from TOC

## Breaking change
 - [x] No

## Other useful information


web/client/plugins/TOC/components/DefaultLayer.jsx
```diff
@@ -300,7 +305,8 @@ const DefaultLayer = ({
                 handleOnChange({ visibility });
             }}
         />),
-        nodeIcon: <Glyphicon className="ms-node-icon" glyph={icon} />
+        nodeIcon: <Glyphicon className="ms-node-icon" glyph={icon} />,
+        onLayerFilterByLegend: onLayerFilterByLegend
```
allyoucanmap, line 309: please remove onLayerFilterByLegend

web/client/plugins/TOC/components/LayersTree.jsx
```diff
@@ -77,6 +77,7 @@ const LayersTree = ({
     layerNodeComponent = DefaultLayer,
     onContextMenu = () => {},
     onSelect = () => {},
+    onLayerFilterByLegend = () => {},
```
allyoucanmap, line 80: please remove onLayerFilterByLegend

web/client/plugins/TOC/components/LayersTree.jsx
```diff
@@ -204,6 +205,7 @@ const LayersTree = ({
                             onSelect={onSelect}
                             nodeItems={nodeItems}
                             nodeToolItems={nodeToolItems}
+                            onLayerFilterByLegend={onLayerFilterByLegend}
```
allyoucanmap, line 208: please remove onLayerFilterByLegend

web/client/plugins/TOC/components/StyleBasedWMSJsonLegend.jsx
```diff
@@ -0,0 +1,153 @@
...
+            });
+        }
+        return '';
+    }
+    renderRules = (rules) => {
+        const isLegendFilterIncluded = this.props.layer?.layerFilter?.filters?.find(f=>f.id === 'interactiveLegend');
+        const prevFilter = isLegendFilterIncluded ? isLegendFilterIncluded?.filters?.[0]?.id : '';
+        const filterWMSLayerHandler = (filter) => {
+            if (!filter) return;
+            this.props.onLayerFilterByLegend(this.props.layer.id, 'layers', filter === prevFilter ? '' : filter);
```
allyoucanmap, line 132: Here we should remove onLayerFilterByLegend and use the passed onChange, I would expect something like this:

```
const newLayerFilter = updateLayerLegendFilter(layer?.layerFilter, filter);
this.props.onChange({ layerFilter: newLayerFilter });

```

Inside the `updateLayerLegendFilter` utils we will place all the logic to update the layerFilter. Note queryform expect an empty filter with specific property available to make it work propertly, from my investigation in review this default properties should be available:

```js
const layerFilter = {
    groupFields: [
        {
            id: 1,
            logic: 'OR',
            index: 0
        }
    ],
    filterFields: [],
    attributePanelExpanded: true,
    spatialPanelExpanded: true,
    crossLayerExpanded: true,
    crossLayerFilter: {
        attribute: 'the_geom'
    },
    spatialField: {
        method: null,
        operation: 'INTERSECTS',
        geometry: null,
        attribute: 'the_geom'
    },
```

- groupFields is needed to show the first row in the attribute filter
- most of the other are needed to make the save button enabled after apply a filter

web/client/plugins/TOC/components/WMSLegend.jsx
```diff
@@ -37,13 +38,15 @@ class WMSLegend extends React.Component {
         scaleDependent: PropTypes.bool,
         language: PropTypes.string,
         legendWidth: PropTypes.number,
-        legendHeight: PropTypes.number
+        legendHeight: PropTypes.number,
+        onLayerFilterByLegend: PropTypes.func
```
allyoucanmap, line 42: please remove onLayerFilterByLegend and pass onChange

web/client/plugins/TOC/components/WMSLegend.jsx
```diff
@@ -92,6 +96,35 @@ class WMSLegend extends React.Component {
                     />
                 </div>
             );
+        } else if (showLegend) {
```
allyoucanmap, line 99: else if it's not needed if you are using an early return, you can use directly if (showLegend)

web/client/plugins/TOC/components/VectorLegend.jsx
```diff
@@ -8,15 +8,15 @@
...
 /**
  * VectorLegend renders the legend given a valid vector style
  * @prop {object} style a layer style object in geostyler format
  */
 function VectorLegend({ style }) {
-
     const renderRules = (rules) => {
         return (rules || []).map((rule) => {
-            return (<div className="ms-vector-legend-rule" key={rule.ruleId}>
+            return (<div className="ms-vector-legend-rule" key={rule.ruleId || rule.name}>
```
allyoucanmap, line 19: This changes seems related to vector layer, please remove them

web/client/plugins/TOC/components/TOC.jsx
```diff
@@ -63,6 +63,7 @@ export function ControlledTOC({
     onChange = () => {},
     onSelectNode = () => {},
     onContextMenu = () => {},
+    onLayerFilterByLegend = () => {},
```
allyoucanmap, line 66: please remove onLayerFilterByLegend

web/client/epics/layerfilter.js
```diff
@@ -114,8 +118,139 @@ export const onApplyFilter = (action$, {getState}) =>
...
+                            "filters": [{
+                                "format": "cql",
+                                "version": "1.0.0",
+                                "body": cqlStatement,
+                                "id": `[${cqlStatement}]`
+                            }]
+                        }
+                    ]
+                ]
+            };
```
allyoucanmap, line 178: This is the logic that should be refactored and ported inside the `updateLayerLegendFilter` utils. Now it seems supported only a selection at times, is it possible instead to select multiple legend entries with a OR logic?

web/client/epics/layerfilter.js
```diff
@@ -114,8 +118,139 @@ export const onApplyFilter = (action$, {getState}) =>
...
+            return Rx.Observable.of(search(searchUrl, filter), addFilterToLayer(layer.id, filter), storeAppliedFilter(filter));
```
allyoucanmap, line 179: After we estracted the filter logic we can remove this epic because the main action we needed addFilterToLayer, it's already covered by the onChange action

web/client/epics/layerfilter.js
```diff
@@ -114,8 +118,139 @@ export const onApplyFilter = (action$, {getState}) =>
...
+            const layer = getLayerByIdFromDashboardMapBuilder(state, layerId);
+            const isDashboard = isDashboardAvailable(state);
+
+            return layer?.type  === 'wms' && isDashboard;
+        })
+        .switchMap(() => {
+            // todo: put the logic of filter wms layer in dashbpard viewer
+            return Rx.Observable.empty();
+        });
+};
```
allyoucanmap, line 202: please remove this epic, it's not used

web/client/themes/default/less/toc.less
```diff
@@ -88,6 +88,13 @@
             .background-color-var(@theme-vars[selected-bg]);
             .outline-color-var(@theme-vars[focus-color]);
         }
+        .wms-json-legend-rule:hover {
+            .color-var(@theme-vars[primary]);
+        }
+        .wms-json-legend-rule.active {
+            .background-color-var(@theme-vars[primary]);
+            .color-var(@theme-vars[primary-contrast]);
+        }
```
allyoucanmap, line 97: we could use the selected colors introduced in the TOC updates

```suggestion
        .wms-json-legend-rule:hover {
            .background-color-var(@theme-vars[selected-hover-bg]);
        }
        .wms-json-legend-rule.active {
            .background-color-var(@theme-vars[selected-bg]);
        }
        .wms-json-legend-rule.active {
            .background-color-var(@theme-vars[primary]);
            .color-var(@theme-vars[primary-contrast]);
        }
```

web/client/utils/FilterUtils.js
```diff
@@ -1143,11 +1143,14 @@ export const getWFSFilterData = (filterObj, options) => {
     return data;
 };
 export const isLikeOrIlike = (operator) => operator === "ilike" || operator === "like";
-export const isFilterEmpty = ({ filterFields = [], spatialField = {}, crossLayerFilter = {}, filters = [] } = {}) =>
-    !(filterFields.filter((field) => field.value || field.value === 0 || field.operator === "isNull").length > 0)
+export const isFilterEmpty = ({ filterFields = [], spatialField = {}, crossLayerFilter = {}, filters = [] } = {}) => {
```
allyoucanmap, line 1146: this seems not related to WMS, please remove it

web/client/utils/StyleEditorUtils.js
```diff
@@ -752,6 +765,12 @@ export const styleValidation = {
     }
 };
 
+export const parseGeoStylerFilterToCql = (geostylerFilter) => {
```
allyoucanmap, line 768: this seems not related to WMS, please remove it

web/client/utils/LayersUtils.js
```diff
@@ -707,7 +707,8 @@ export const saveLayer = (layer) => {
         legendOptions: layer.legendOptions,
         tileSize: layer.tileSize,
         version: layer.version,
-        expanded: layer.expanded || false
+        expanded: layer.expanded || false,
+        enableInteractiveLegend: layer.enableInteractiveLegend || false
```
allyoucanmap, line 711: let's make it undefined by default so it's not included in the map config

web/client/components/data/query/QueryToolbar.jsx
```diff
@@ -263,14 +263,15 @@ class QueryToolbar extends React.Component {
...
+            const isLgendFilterExist = this.props.appliedFilter?.filters?.find(f=>f.id === 'interactiveLegend');
             let filterObj = {
                 featureTypeName: this.props.featureTypeName,
                 groupFields: [],
                 filterFields: [],
                 spatialField: {
                     attribute: this.props.spatialField && this.props.spatialField.attribute
                 },
-
+                filters: isLgendFilterExist ? [isLgendFilterExist] : [],
```
allyoucanmap, line 274: I think if we reset the QueryForm all the filters should be removed, I don't know if it's possible to communicate to the user that other external filter are currently applied and they will be removed on reset (cc @tdipisa )
Anyway I would prefer to remove this custom logic at the moment and reset all the filter

web/client/actions/layerFilter.js
```diff
@@ -62,3 +65,19 @@ export function initLayerFilter(filter) {
...
+    };
+}
+
+export function resetLegendFilter(reason, value) {
+    return {
+        type: RESET_LAYER_FILTER_BY_LEGEND,
+        reason,     // here the reason for reset is change 'style' or change the enable/disable interactive legend config 'disableEnableInteractiveLegend'
+        value
+    };
+}
```
allyoucanmap, line 83: Probably also this function could be removed and create a utils that update the layerFilter removing the filter part related to legend. We need to reset the legend in following places:

- when we disable the interactive from settings. There the onChange should be available
- when we change the style. Also in this case we should use the the updateNode action passed as onUpdateNode

web/client/components/TOC/fragments/settings/Display.jsx
```diff
@@ -261,7 +264,22 @@ export default class extends React.Component {
...
-                        <Col xs={12} sm={6} className="first-selectize">
+                        { this.props.element?.serverType !== ServerTypes.NO_VENDOR &&
+                            <Col xs={12}>
+                                <Checkbox
+                                    data-qa="display-interactive-legend-option"
+                                    value="enableInteractiveLegend"
+                                    key="enableInteractiveLegend"
+                                    onChange={(e) => {
+                                        if (!e.target.checked) this.props.resetLegendFilter('disableEnableInteractiveLegend');
+                                        this.props.onChange("enableInteractiveLegend", e.target.checked);
```
allyoucanmap, line 275: you can use something like this:

```
if (!e.target.checked) {
  const newLayerFilter = resetLayerLegendFilter(this.props.element.layerFilter);
  this.props.onChange("layerFilter", newLayerFilter );
}
```

where resetLayerLegendFilter is a utils that recieve a layerFilter object and return the updated layerFilter

web/client/components/TOC/fragments/settings/Display.jsx
```diff
@@ -261,7 +264,22 @@ export default class extends React.Component {
                         <Col xs={12} className={"legend-label"}>
                             <label key="legend-options-title" className="control-label"><Message msgId="layerProperties.legendOptions.title" /></label>
                         </Col>
-                        <Col xs={12} sm={6} className="first-selectize">
+                        { this.props.element?.serverType !== ServerTypes.NO_VENDOR &&
+                            <Col xs={12}>
```
allyoucanmap, line 268: the checkbox is not aligned to the other

image

allyoucanmap (changes requested):
Some other part to review:

- [ ] if the filter is disabled (filter icon black) it should not be possible to interact with the legend, so the cursor should not be pointer and the hover state should not be visible.
image
- [ ] for some layer the json legend is not available even if the server supports it (eg. Meteorite_Landings_from_NASA_Open_Data_Portal). How should we handle this? (cc @tdipisa ). This happens also with raster layer
 ```
<ServiceException>
      java.lang.IllegalArgumentException: Trying to evaluate the function as Object but the method parameter is set as COLOR
Trying to evaluate the function as Object but the method parameter is set as COLOR
</ServiceException>
```

File: web/client/components/data/query/QueryToolbar.jsx (hunk 16 above)
Comment (tdipisa) [reply web/client/components/data/query/QueryToolbar.jsx:274]: > I don't know if it's possible to communicate to the user that other external filter are currently applied and they will be removed on reset (cc @tdipisa )

I think this one should be handled somehow for consistency and a better UX since, as far as I understand, all filters lends in the same filter object. What about something like this?
image

or this?

image

File: web/client/components/TOC/fragments/settings/Display.jsx (hunk 19 above)
Comment (tdipisa) [reply web/client/components/TOC/fragments/settings/Display.jsx:268]: @allyoucanmap @mahmoudadel54 let's put a proper info message in the helper to well inform what the option does and what it requires. For layers not coming from GeoServer this option should be disabled or simply not available.

File: web/client/themes/default/less/toc.less [hunk 20]
```diff
@@ -88,6 +88,16 @@
...
+        .wms-json-legend-rule.filter-enabled:hover {
+            .background-color-var(@theme-vars[selected-hover-bg]);
+        }
+        .wms-json-legend-rule.filter-enabled.active {
+            .background-color-var(@theme-vars[selected-bg]);
+        }
+        .wms-json-legend-rule.filter-enabled.active {
+            .background-color-var(@theme-vars[primary]);
+            .color-var(@theme-vars[primary-contrast]);
+        }
```
allyoucanmap, line 100: should be removed

allyoucanmap (changes requested):
image

please merge the two tooltip message in a single one:

`
If this option is enabled, filter by legend can be applied by clicking on legend items of this layer from the TOC. Note: This parameter requires specific configurations on GeoServer
`

allyoucanmap (changes requested):
Some finel changes:

- when a item of the legend has no filter it should not show the pointer cursor and the hover highlight
- please add some unit tests for the new two functions in FIlterUtils

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test this new feature on dev, thanks

ElenaGallo:
@mahmoudadel54 
1_ When the legend filter is set and then the map saved. The interactive legend is no longer enabled in the settings layer. Is this the correct behavior?

https://github.com/geosolutions-it/MapStore2/assets/56537133/07684468-c148-49c7-9e78-9f49c4d8919a



2_  When there is an attribute filter for the layer and then interactive legend is enabled, the interactive legend does not work as expected.

https://github.com/geosolutions-it/MapStore2/assets/56537133/5bab59a9-8f96-44c1-aafa-4096fb205e6b



3_ On Dashboard the interactive legend  is available on settings layer but it does not work on layer legend

https://github.com/geosolutions-it/MapStore2/assets/56537133/934138c3-6e95-45bc-ae0a-c60ab4e978f5



4_ On Geostory the interactive legend  is available on settings layer but it works only the first time

https://github.com/geosolutions-it/MapStore2/assets/56537133/d79716ac-d3d6-410c-bb35-c96fdbac526f

tdipisa:
@mahmoudadel54  here what I found after my test: 

>1_ When the legend filter is set and then the map saved. The interactive legend is no longer enabled in the settings layer. Is this the correct behavior?

@mahmoudadel54 this need to be fixed asap. It should be a quick thing to fix.

>2_  When there is an attribute filter for the layer and then interactive legend is enabled, the interactive legend does not work as expected.

We know that @ElenaGallo. This is an expected behavior for the moment. Since it is not possible yet in GS to filter the JSON legend according to the general filter defined in 'Layer filter', if you click on a style class not matching that filter the result if an empty layer in map. I suggest anyway to open a dedicated issue for this right now. @ElenaGallo please create it.

>3_ On Dashboard the interactive legend  is available on settings layer but it does not work on layer legend

The current implementation is partial and provides support only for the map viewer. We have already an issue defined for this purpose in https://github.com/geosolutions-it/MapStore2/issues/10235. That means that until we don't have a working implementation for Dashboards, the _Enable interactive legend_ option should not be available there @mahmoudadel54 

>4_ On Geostory the interactive legend  is available on settings layer but it works only the first time

@mahmoudadel54 is it possible to provide a quick fix for this? Otherwise the same as above for Dashboard is valid and a dedicated issue should be opened similar to #10235

mahmoudadel54:
> > 4_ On Geostory the interactive legend  is available on settings layer but it works only the first time


For the point 4 of the geostory, I have tried it by adding new map and adding layers on this map and it works well with me. 

https://github.com/geosolutions-it/MapStore2/assets/58145645/246e7b18-9dd6-4485-9070-79757f222ffe


After debugging the issue I found you used an existing saved map from mapstore. The existing layer in this map that you tried to apply legend filter on which is 'State' layer in the used map has a disabled flag in its filter object. This flag is set by clicking on this icon 
image

I think it needs to be fixed.

There is an issue which make me to skip showing the interactive legend from geostory is there in an issue related to map consistency. If I make changes into filter object of map in mapEditor it will not reflect to the map viewer. 
How I produce that:
- create geostory
- add new map with layers like 'States of US' the classified one
- enable from map editor the interactive legend of this layer
- clicks on legend to apply multi legend filter
- save it and close 
- open map editor again, disable the interactive legend  --> you will notice the map is updated and layer filter is reset
- save it and close map editor.
- the map viewer will display layer without reset the layer filter [consistency issue --> map of map viewer should be consistent with the changes in map editor]

  

https://github.com/geosolutions-it/MapStore2/assets/58145645/84a0469b-8ace-45b7-be6e-ac7c4552da65

This consistency needs to be checked and fixed as well in the dedicated geostory/dashboard issue

Issue #10136, opened 2024-03-27
Search for Map CRS coordinates


tdipisa:
## Description

It is requested with this issue to improve the CoordinatesSearch tool to be able to use coordinates expressed in the selected Map CRS, now it is possible a search for coordinates in the defined on the map, now it's only possible by Lat/Long.

![Image](https://github.com/geosolutions-it/MapStore2/assets/1280027/de6819a1-abb2-4491-b403-608747255d90)

A proper entry in the current coordinate type selector (decimal/aeronautical) should allow to select the current Map CRS to setup the tool accordingly (e.g. use current Map CRS)

![Image](https://github.com/geosolutions-it/MapStore2/assets/1280027/9c0dea9c-3770-43b4-ab41-407589fdd45f)

Thank into account that the map CRS can vary according to the CRS selected din the [CRS Selector](https://docs.mapstore.geosolutionsgroup.com/en/latest/user-guide/footer/#crs-selector).

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## Design

It would be good to implement the following UI/UX:

- As already done for "Search by location name" in #9751, include a sub menu item for "Search by coordinates"
- Selecting "Search by coordinates" the tool maintain the current behavior with the usual decimal and aeronautical options

![image](https://github.com/geosolutions-it/MapStore2/assets/1280027/cf666e88-9bdb-4121-b3bd-7f3a30159ebc)

- Selecting "Current map CRS" the possibility to switch between decimal/aeronautical disappear from the UI and the tool setup change by switching to map CRS coordinates

![image](https://github.com/geosolutions-it/MapStore2/assets/1280027/59993cf7-7190-4abb-9cf1-7c4e05d44223)

- The "Current map CRS" is available only if the the current Map CRS is not the one normally used by decimal and aeronautical coordinates, which means EPSG:4326

![image](https://github.com/geosolutions-it/MapStore2/assets/1280027/d72ecb0b-6eaa-4074-976c-35052adf08f8)

- As coordinate fields validator when "Current map CRS" is selected, the current one should be preserved which means:
   - The entered value is a number
   - The entered coordinates are within the CRS bounds
   - Same decimal separators should be used according locale as is also for other plugins where geographic coordinates are exposed 

[...]

PR #10220, opened 2024-04-19
#10136: Search for Map CRS coordinates

mahmoudadel54:
## Description
In this PR, the Map CRS coordinate search functionality is implemented. If the curent map CRS is different than `EPSG:4326` a sub menu will be added to search menu and it will include the current map CRS option. Validation of CRS extent, zoom to point are implemented.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#10136 

#10136 

**What is the new behavior?**
User now can apply map CRS coordinate search based on the CRS.

## Breaking change
 - [x] No

## Other useful information
These are the changes added to this PR:
- Add new component for current map CRS coordinates search
- create a util function for getting extent based on extent to validate the mapCRS extent in case of search by mapCRS coordinates
- write some unit tests according the new added code + changes


web/client/components/mapcontrols/search/SearchBar.jsx
```diff
@@ -112,6 +114,13 @@ export default ({
         }
     }, [searchOptions?.services]);
 
+    useEffect(() => {
+        // if user changes the map crs and the search was in  mapCRSCoordinatesSearch tab in search --> switch to default coordniateSearch
```
dsuren1, line 118:
```suggestion
        // Switch back to coordinate search when map CRS is EPSG:4326 and active tool is Map CRS coordinate search
```

web/client/utils/CoordinatesUtils.js
```diff
@@ -207,6 +207,22 @@ export const  reprojectExtent = (extent, projection, isIDL) => {
         ].reduce((a, b) => [...a, b.x, b.y], []));
 };
 
+/**
+ * Get extent by EPSG projection
+ * @param projection {string} the projection of the needed extent
+ * @return {array} extent [minx, miny, maxy, maxy]
+ */
+export const  getExtentByCRSCode = (projection) => {
```
dsuren1, line 215: The requirement ask for entered coordinates are within the CRS bounds. i.e for `EPSG:2154` the bounds are approximately [-9.86, 41.15, 10.38, 51.56]. So the value should be limited in these ranges

web/client/components/mapcontrols/searchcoordinates/CurrentMapCRSCoordSearch.js
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }
+        };
+    };
+    const changeCoordinates = (coord, value) => {
+        const numValue = parseFloat(value);
+        onChangeCoord(coord, numValue);
+        if (coord === 'yCoord') {
+            const latValue = numValue ? reproject([0, numValue], currentMapCRS, 'EPSG:4326', true)?.y : undefined;
+            onChangeCoord('lat', latValue ? Number((latValue).toFixed(4)) : latValue);
```
dsuren1, line 73: I don't think the value should use fixed decimal places. Kindly refer to coordinate search if there is any similar limits applied

dsuren1 (changes requested):
@mahmoudadel54 
I haven't tested the functionality yet as I'm facing gateway timeout often. Will check it once the DEV is back up.
Meanwhile I think you can take a look into the ones requested below. Thanks

web/client/utils/CoordinatesUtils.js
tdipisa: @mahmoudadel54 my question is: why such requirement is so ambiguous for you? 
If one of the coordinates entered by the user is outside the CRS bounds the validator should notify the error. I don't know how to explain it better.

dsuren1 (changes requested):
@mahmoudadel54 I don't see any new commit since the last review but requested for a re-review. Kindly check if you have missed to commit your new changes

web/client/components/misc/coordinateeditors/editors/DecimalCoordinateEditorSearch.jsx
```diff
@@ -0,0 +1,82 @@
...
+    constructor(props, context) {
+        super(props, context);
+        this.state = {
+            value: this.props.value
+        };
+    }
+    componentDidUpdate(prevProps) {
+        const isClearInputs = (prevProps.value && !this.props.value);
+        // (!prevProps.value && this.props.value) in case intiate the component with a prev defined values
+        // like switch from default coordinate search to current map crs coordinate search
```
dsuren1, line 55: Kindly rephrase this comment as it is not clear

dsuren1 (changes requested):
@mahmoudadel54 
Kindly look into these issues. Thanks!
- UI agreed in the issue shows x and y. Interchange the fields
image

- 0,0 is not considered as valid value even when it falls within crs limits
image

- with CRS search performed, logout and loggin back to map, the app crashes

https://github.com/geosolutions-it/MapStore2/assets/26929983/cd02ca1d-3497-41be-b083-75a8421b7b06

- Mouse pointer showing x value doesn't match with user input on the marker placed by the crs search. Projection used `EPSG:2154`. Kindly check why the difference
image

- Unable to edit X or Y value when the value is near limits of the CRS

https://github.com/geosolutions-it/MapStore2/assets/26929983/34e29fe8-0c10-47d9-9912-305e4b32b1dd

web/client/components/mapcontrols/search/SearchBar.jsx
```diff
@@ -115,6 +115,8 @@ export default ({
     }, [searchOptions?.services]);
 
     useEffect(() => {
+        // clear coordinate marker in case of change map CRS
+        onClearCoordinatesSearch({owner: "search"});
```
dsuren1, line 119: Why is it needed to clear the marker, since the value is reprojected to CRS so the marker should be at the same position, right?

dsuren1 (changes requested):
@mahmoudadel54 
Kindly look into these issues
- [ ] Looks like the limit validation is broken, as I can enter value beyond crs limits now. I can see error shown but this behavior is not similar to how coordinate search field validation works. It limits user from entering the value beyond the bound
  image
- [ ] Able to enter 0,0 but CRS search is not performed

  https://github.com/geosolutions-it/MapStore2/assets/26929983/f5d87457-2396-47ed-908f-61bc040dfb93

mahmoudadel54:
> @mahmoudadel54 Kindly look into these,
> 
> * In some case, on changing search to CRS, Y field is cleared and unable to enter value in the field
>   
>     
>       
>       
>   
>       field_issue.mp4
Hi @tdipisa @dsuren1 
For the 1st point, I am testing with 3003, here the extent is [218994.5, 3846433.22, 1415881.28, 5264909.36]

- now if user wants to enter **`x coordinate`** with 218999 which is a valid value. He will start writing with **`2`** in the input
- the 1st change value is **`2`**. For the current requirements, **`2`** is not within the crs extent so **`2`** will not be allowed to be shown in the input and finally he couldn't enter this value one by one 
- only one case will be applied, if user copies the value and makes pasting in the input as it is

What I think is allow out of range, I will highlight the input with red border and showing an error message with that to differentiate bet. out of range case and empty input case

dsuren1:
@tdipisa @mahmoudadel54 
This scenario poses a limitation in simply limiting the user from entering the value beyond bounds as stated above, the 1st search value falls outside limits and hence user is unable to enter any value further. So maybe for CRS search alone, we can instead show field error when value is beyond limits and disable search button or simply skip any search activity.
image
Your thoughts? @tdipisa

tdipisa:
@mahmoudadel54 @dsuren1 

The tool should't prevent the user from entering a number in any case, of course, and it isn't also acceptable that the user needs to copy and paste an entire value because that would make the tool unusable @mahmoudadel54
The user should be facilitated in working with a tool also for changing wrong values and not the contrary. That's quite a UI/UX basis.
What the tool should do is to simply mark in red a field when the value is wrong by preventing the search to be triggered, that's exactly what the tool is doing now in DEV @mahmoudadel54: if coordinate values are wrong, clicking on the search button doesn't produce any result.
So, we should at least to maintain the same behavior, it seems quite obvious to me.

dsuren1:
@mahmoudadel54 
Like I said and aligning with @tdipisa's feedback, please proceed with updating the PR. Thanks

dsuren1 (changes requested):
@mahmoudadel54 
Rest looks good except for the one below. Kindly amend this. Thanks!
- When the coordinate is valid and within range, upon changing either to `Current Map CRS` search or changing CRS from selector, clears the marker. This makes user perform `search` again for the same coordinates

dsuren1 (changes requested):
@mahmoudadel54 
CRS search works fine but `Decimal` coordinate search's Lat and Lon field is affected allowing values beyond limits. Kindly retain these fields behavior as is

web/client/components/misc/coordinateeditors/editors/DecimalCoordinateEditorSearch.jsx
```diff
@@ -0,0 +1,111 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+import {FormGroup} from 'react-bootstrap';
+import DecimalCoordinateEditor from './DecimalCoordinateEditor';
+import IntlNumberFormControl from '../../../I18N/IntlNumberFormControl';
+
+/**
+ This component renders a custom coordiante inpout for decimal degrees for default coordinate CRS and current map CRS as well
+*/
+class DecimalCoordinateEditorSearch extends DecimalCoordinateEditor {
```
dsuren1, line 19: The component name can be distinctive. Modify it in other places too.
Also check if extending DecimalCoordinateEditor is still needed.

```suggestion
class CRSCoordinateEditor extends DecimalCoordinateEditor {
```

dsuren1 (changes requested):
@mahmoudadel54 
I think we are almost there, I only observe this issue with editing
- Unable to edit in-between number digits as the cursor jumps to the end. Kindly fix this 

  https://github.com/geosolutions-it/MapStore2/assets/26929983/6ecd74ef-4097-42fe-91e4-476ec14a0b53

offtherailz:
Fixing milestone from 2024.01.00 to 2024.01.01, accordingly to the issue.

dsuren1:
@mahmoudadel54 
The requirements of the current issue seem to be functioning properly as of this [commit](https://github.com/geosolutions-it/MapStore2/pull/10220/commits/933f2086a1853a6aacef56a5bbb16eb0f6e96892). However, the cursor position issue originates from the Localized number field [component](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/I18N/IntlNumberFormControl.jsx#L23) and addressing it within this pull request is proving to be time-consuming. Additionally, a regression-free fix at this time is not guaranteed. Therefore, I have created a separate [issue](https://github.com/geosolutions-it/MapStore2/issues/10289) to handle it.

Kindly revert your changes pertaining to the number field, formatting etc i.e up until this [commit](https://github.com/geosolutions-it/MapStore2/pull/10220/commits/933f2086a1853a6aacef56a5bbb16eb0f6e96892) and keep any changes from this [comment](https://github.com/geosolutions-it/MapStore2/pull/10220#discussion_r1591856895)

dsuren1 (changes requested):
@mahmoudadel54 
The requirements of the current issue seem to be functioning properly as of this [commit](https://github.com/geosolutions-it/MapStore2/pull/10220/commits/933f2086a1853a6aacef56a5bbb16eb0f6e96892). However, the cursor position issue originates from the Localized number field [component](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/I18N/IntlNumberFormControl.jsx#L23) and addressing it within this pull request is proving to be time-consuming. Additionally, a regression-free fix at this point is not guaranteed. Therefore, I have created a separate [issue](https://github.com/geosolutions-it/MapStore2/issues/10289) to handle it.

Kindly revert your changes pertaining to the number field, formatting etc i.e up until this [commit](https://github.com/geosolutions-it/MapStore2/pull/10220/commits/933f2086a1853a6aacef56a5bbb16eb0f6e96892) and keep any changes from this [comment](https://github.com/geosolutions-it/MapStore2/pull/10220#discussion_r1591856895)

tdipisa:
> @mahmoudadel54 The requirements of the current issue seem to be functioning properly as of this [commit](https://github.com/geosolutions-it/MapStore2/pull/10220/commits/933f2086a1853a6aacef56a5bbb16eb0f6e96892). However, the cursor position issue originates from the Localized number field [component](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/components/I18N/IntlNumberFormControl.jsx#L23) and addressing it within this pull request is proving to be time-consuming. Additionally, a regression-free fix at this point is not guaranteed. Therefore, I have created a separate [issue](https://github.com/geosolutions-it/MapStore2/issues/10289) to handle it.
> 
> Kindly revert your changes pertaining to the number field, formatting etc i.e up until this [commit](https://github.com/geosolutions-it/MapStore2/pull/10220/commits/933f2086a1853a6aacef56a5bbb16eb0f6e96892) and keep any changes from this [comment](https://github.com/geosolutions-it/MapStore2/pull/10220#discussion_r1591856895)

Thank you @dsuren1. @mahmoudadel54 once you have done with this you can keep https://github.com/geosolutions-it/MapStore2/issues/10289 as part of your backlog

dsuren1:
@ElenaGallo Kindly test it in DEV and please factor in this comment while testing. Thanks!

ElenaGallo:
1_ Switching between CRS CURRENT MAP and SEARCH BY COORDINATES and then again CRS CURRENT MAP, the Y value changes

https://github.com/geosolutions-it/MapStore2/assets/56537133/f7c23125-2341-45cd-98e2-e3d07e620151

2_ Switching between CRS CURRENT MAP and SEARCH BY BOOKMARK the mark on the map does not disappear

https://github.com/geosolutions-it/MapStore2/assets/56537133/f2e21f53-ee47-4fb3-819a-019defa9dedd

mahmoudadel54:
Hi @ElenaGallo 
I have checked the first point on DEV, and it works well with me
Shown here:
https://github.com/geosolutions-it/MapStore2/assets/58145645/7639363d-d3d6-4a33-b3e6-28476396d7e1

For the 2nd point, I think it has the same behavior if you switch from lon/lat coordinates to bookmark which means it is preexisting. Should I handle this scenario on both coordinate search [lon/lat and X/Y] ?

ElenaGallo:
@mahmoudadel54 

**Point 1**

Here's how to reproduce step 1:

- Select CURRENT MAP CRS
- Enter x, y coordinate
- Click on Search
- Select SEARCH BY COORDINATES
- Select again CURRENT MAP CRS

**Point 2**

You're right it also happens for LAT/LOG but not when you change from LAT/LOG or X/Y to SEARCH BY LOCATION NAME. @tdipisa do I have to open a new issue for this?

mahmoudadel54:
@ElenaGallo 
For the point 1: It is because I reproject [conversion] the lat/lon coordinates to X/Y coordinates . The implementation of this part is every time user opens the current map CRS search, I retrieve lon/lat coordinates and convert it to X/Y displaying this values using `reproject` method.

Previously, I made round for the X/Y numbers with 4 digits and it was reviewed to remove this rounding : https://github.com/geosolutions-it/MapStore2/pull/10220#discussion_r1576254919 

@tdipisa @dsuren1 What do you think ?

tdipisa:
@ElenaGallo 

>You're right it also happens for LAT/LOG but not when you change from LAT/LOG or X/Y to SEARCH BY LOCATION NAME. @tdipisa do I have to open a new issue for this?

yes please, since it is a preexisting bug unrelated to this PR. Linke me the issue once you created it. Thank you.

@mahmoudadel54 

>Point 1
Here's how to reproduce step 1:

It is happening on DEV but it is a matter of how the number is rounded. In theory we should restore it to the exact value the client entered but, in any case, a minimum variation within an acceptable threshold is good. Any thoughts @dsuren1 @mahmoudadel54?  Can we restore the exact value provided by the user with a really short effort?

dsuren1:
@tdipisa
It is doable but however adds another state management and validation to check if the value is modified as there is a possibility that the value maybe changed by the user between the switches to other search option before coming to CRS search. 

@mahmoudadel54 Do you think you can do it quickly? If you think it would take more effort, then I'm inclined to agree with @tdipisa's point of a minimum variation within an acceptable threshold is good i.e in the fractional 0.00001 or lesser

ElenaGallo:
@tdipisa new issue here: https://github.com/geosolutions-it/MapStore2/issues/10308

ElenaGallo:
Test passed, @mahmoudadel54 please backport to 2024.01.xx. Thanks

mahmoudadel54:
> Test passed, @mahmoudadel54 please backport to 2024.01.xx. Thanks
@ElenaGallo 
Backport is done ---> https://github.com/geosolutions-it/MapStore2/pull/10317

PR #10293, opened 2024-05-08
update dev

Igi-ID:
Fixed search and pagination,
Fixed point GFI,
wip layer metadata

### DIFF
--- a/web/client/actions/layers.js
+++ b/web/client/actions/layers.js
@@ -19,6 +19,7 @@ export const LAYER_LOADING = 'LAYER_LOADING';
 export const LAYER_LOAD = 'LAYER_LOAD';
 export const LAYER_ERROR = 'LAYER_ERROR';
 export const ADD_LAYER = 'ADD_LAYER';
+export const UPDATE_LAYER = 'UPDATE_LAYER';
 export const ADD_GROUP = 'ADD_GROUP';
 export const REMOVE_LAYER = 'REMOVE_LAYER';
 export const SHOW_SETTINGS = 'SHOW_SETTINGS';
@@ -183,6 +184,15 @@ export function addLayer(layer, foreground = true) {
         foreground
     };
 }
+
+export function updateLayer(layerId, layer) {
+    return {
+        type: UPDATE_LAYER,
+        id: layerId,
+        layer
+    };
+}
+
 /**
  * Add a new group
  * @memberof actions.layers
--- a/web/client/epics/catalog.js
+++ b/web/client/epics/catalog.js
@@ -35,7 +35,7 @@ import {
     addLayer,
     setNewServiceStatus
 } from '../actions/catalog';
-import {showLayerMetadata, SELECT_NODE, changeLayerProperties, addLayer as addNewLayer} from '../actions/layers';
+import {showLayerMetadata, SELECT_NODE, changeLayerProperties, addLayer as addNewLayer, updateLayer} from '../actions/layers';
 import { error, success, warning } from '../actions/notifications';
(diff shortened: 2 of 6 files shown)


web/client/reducers/layers.js
```diff
@@ -248,6 +249,24 @@ function layers(state = { flat: [] }, action) {
             groups: newGroups
         });
     }
+    case UPDATE_LAYER: {
```
allyoucanmap, line 252: Why do we need this action? there is already updateNode or addNewLayer, could we use that instead?

allyoucanmap (changes requested):

web/client/actions/layers.js
```diff
@@ -183,6 +183,7 @@ export function addLayer(layer, foreground = true) {
         foreground
     };
 }
+
```
allyoucanmap, line 186:
```suggestion
```

web/client/epics/catalog.js
```diff
@@ -346,6 +349,31 @@ export default (API) => ({
...
+                                        maxy: maxP?.y
+                                    },
+                                    crs: 'EPSG:4326'
+                                },
+                                // include metadata for future improvements
+                                properties
+                            };
+                            return Rx.Observable.from([addNewLayer({...newLayer, id})]);
+                        })
+                        .merge(Rx.Observable.from(actions))
```
allyoucanmap, line 374:
```suggestion
```
you don't need to merge the actions if you need a different workflow

web/client/epics/catalog.js
```diff
@@ -346,6 +349,31 @@ export default (API) => ({
...
+                                        miny: minP?.y,
+                                        maxx: maxP?.x,
+                                        maxy: maxP?.y
+                                    },
+                                    crs: 'EPSG:4326'
+                                },
+                                // include metadata for future improvements
+                                properties
+                            };
+                            return Rx.Observable.from([addNewLayer({...newLayer, id})]);
```
allyoucanmap, line 372: we could instead trigger the sequence of action we need like
```suggestion
                            return Rx.Observable.from([addNewLayer({...newLayer, id}), zoomToExtent(newLayer.bbox.bounds, newLayer.bbox.crs)]);
```

web/client/epics/catalog.js
```diff
@@ -265,7 +266,9 @@ export default (API) => ({
                 const actions = [];
                 const state = store.getState();
                 const id = getLayerId(layer);
-                actions.push(addNewLayer({...layer, id}));
+                if (layer.type !== 'arcgis') {
+                    actions.push(addNewLayer({...layer, id}));
+                }
```
allyoucanmap, line 271: finally we can restore this because it's not used by arcgis type anymore
```suggestion
                actions.push(addNewLayer({...layer, id}));
```

allyoucanmap (changes requested):

web/client/epics/catalog.js
Igi-ID: Added here https://github.com/geosolutions-it/MapStore2/pull/10293/commits/3da19898fbd8ddfbd249392a27b25306277b3350

web/client/epics/catalog.js
Igi-ID: Removed here https://github.com/geosolutions-it/MapStore2/pull/10293/commits/fc115a23255e03cef0cda6992ab3ff195b6b3d03

web/client/reducers/layers.js
Igi-ID: Removed action and reverted to the already implemented addNewLayer here https://github.com/geosolutions-it/MapStore2/pull/10293/commits/98f88d219a9dd148be8abdc52d713616a5820935

allyoucanmap (approved):

Issue #10322, opened 2024-05-14
Fix scale selector to make it compatible with different projections


offtherailz:
## Description

When switching between projections, the scale selector becomes buggy and is not responding correctly. 

## How to reproduce

I tested this locally and I verified that the issue you find happens even without the projection, so also switching in EPSG:4326 from Web mercator
Here I recorded the evolution of the state locally that can be used for the investigation .

https://github.com/geosolutions-it/support/assets/1279510/d331e009-e06c-4ab6-adc8-1c865e9ddfd3

*Expected Result*
The selector should be consistent with the map and the final print.

*Current Result*
When switching SRS, the selector do not work. 

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

Review to be assigned to @offtherailz 
Connected support issue https://github.com/geosolutions-it/support/issues/3313

PR #10344, opened 2024-05-20
#10322: Fix scale selector to make it compatible with different projections

mahmoudadel54:
## Description
In this PR, the issue of scale selector which was not compatible with different projections is resolved.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#10322

#10322 

**What is the new behavior?**
Now if user opens the print modal, switches the projection to 4326 for example and select a scale from the scale selector list, it will selected successfully without choosing wrong one.

## Breaking change
 - [x] No

## Other useful information
I have noticed some points within this issue: 

1- wrong calculate scale zoom level in case user changes projection in the print modal, this causes the noticed wrong calculated scale in DD scales. 
If you see this line which calculate the scaleZoom if useFixedScales = true 
https://github.com/geosolutions-it/MapStore2/blob/7443a4839e8fed1f1cd981a578b404a7e2ccde55/web/client/plugins/Print.jsx#L593 
it uses getNearestZoom method: https://github.com/geosolutions-it/MapStore2/blob/7443a4839e8fed1f1cd981a578b404a7e2ccde55/web/client/utils/PrintUtils.js#L187-L192 and into this method it uses defaultScales is applied at each projection and this defaultScales is the google mercator scales : https://github.com/geosolutions-it/MapStore2/blob/7443a4839e8fed1f1cd981a578b404a7e2ccde55/web/client/utils/PrintUtils.js#L42

So what I did is passing the scales of the actual correct projection to this method to get the correct calculated scaleZoom to display the correct expected scale in the scales drop down list.

2- Based on the 1st point above, the scale value of the configure print map was calculated wrongly: https://github.com/geosolutions-it/MapStore2/blob/7443a4839e8fed1f1cd981a578b404a7e2ccde55/web/client/plugins/Print.jsx#L594. Another point I need to mention is that mapZoom of the map viewer is passed to **configurePrintMapProp** instead of the calculated zoom value of the print map based on the selected projection here: 
https://github.com/geosolutions-it/MapStore2/blob/7443a4839e8fed1f1cd981a578b404a7e2ccde55/web/client/plugins/Print.jsx#L596 and here as well 
https://github.com/geosolutions-it/MapStore2/blob/7443a4839e8fed1f1cd981a578b404a7e2ccde55/web/client/plugins/Print.jsx#L599

so here I refactored the method to calculate needed values based on the projection value

3- using '**getZoomForExtent**' in '**configurePrintMap**' method is not the proper way to get the mapZoom of the print map --> in some cases the change bet. main map viewer and the print map viewer >= 2 and this makes change in the print result: https://github.com/geosolutions-it/MapStore2/blob/7443a4839e8fed1f1cd981a578b404a7e2ccde55/web/client/plugins/Print.jsx#L583-L592

so I used reprojectZoom method to calculate print map zoom instead using bbox and mapSize for that: https://github.com/geosolutions-it/MapStore2/blob/138ae35d89e9cf124314e41602cb3cf39a731a8f/web/client/plugins/Print.jsx#L586.

4- when user opens the print panel, the sent projection 'prop' should be set initially to be consistent with the UI [selected projection in DD] and to be not misleading if the shown projection into UI not the initial one.

So I used initiated the projection into useEffect in Projection component at initiating the component: 
https://github.com/geosolutions-it/MapStore2/blob/138ae35d89e9cf124314e41602cb3cf39a731a8f/web/client/plugins/print/Projection.jsx#L68-L75

### DIFF
--- a/web/client/plugins/__tests__/Print-test.jsx
+++ b/web/client/plugins/__tests__/Print-test.jsx
@@ -222,9 +222,9 @@ describe('Print Plugin', () => {
                     capabilities: {...initialState.print.capabilities,
                         scales: [1000000, 500000, 100000].map(value => ({name: value, value}))}
                 }}
-        }).then(async({ Plugin }) => {
+        }).then(({ Plugin }) => {
             try {
-                await ReactDOM.render(<Plugin
+                ReactDOM.render(<Plugin
                     projectionOptions={{
                         "projections": [{"name": "UTM32N", "value": "EPSG:23032"}, {"name": "EPSG:3857", "value": "EPSG:3857"}, {"name": "EPSG:4326", "value": "EPSG:4326"}]
                     }}
@@ -233,11 +233,12 @@ describe('Print Plugin', () => {
                         enableScalebox: true
                     }}/>, document.getElementById("container"));
                 const comp = document.getElementById("container");
-                await ReactTestUtils.act(async() => comp);
-                expect(comp).toExist();
-                const scaleBoxComp = document.querySelector("#mappreview-scalebox select");
-                expect(scaleBoxComp).toExist();
-                done();
+                ReactTestUtils.act(() => new Promise((resolve) => resolve(comp))).then(()=>{
+                    expect(comp).toExist();
+                    const scaleBoxComp = document.querySelector("#mappreview-scalebox select");
+                    expect(scaleBoxComp).toExist();
+                    done();
+                });
             } catch (ex) {
                 done(ex);
             }
@@ -263,9 +264,9 @@ describe('Print Plugin', () => {
                     capabilities: {...initialState.print.capabilities,
                         scales: [1000000, 500000, 100000].map(value => ({name: value, value}))}
                 }}
-        }).then(async({ Plugin }) => {
+        }).then(({ Plugin }) => {
             try {
-                await ReactDOM.render(<Plugin
+                ReactDOM.render(<Plugin
                     projectionOptions={{
                         "projections": [{"name": "UTM32N", "value": "EPSG:23032"}, {"name": "EPSG:3857", "value": "EPSG:3857"}, {"name": "EPSG:4326", "value": "EPSG:4326"}]
                     }}
@@ -274,11 +275,12 @@ describe('Print Plugin', () => {
                         enableScalebox: false
                     }}/>, document.getElementById("container"));
                 const comp = document.getElementById("container");
-                await ReactTestUtils.act(async() => comp);
-                expect(comp).toExist();
-                const scaleBoxComp = document.querySelector("#mappreview-scalebox select");
-                expect(scaleBoxComp).toNotExist();
-                done();
+                ReactTestUtils.act(() => new Promise((resolve) => resolve(comp))).then(()=>{
+                    expect(comp).toExist();
+                    const scaleBoxComp = document.querySelector("#mappreview-scalebox select");
+                    expect(scaleBoxComp).toNotExist();
+                    done();
+                });
             } catch (ex) {
                 done(ex);
             }
--- a/web/client/plugins/__tests__/Print-test.jsx
+++ b/web/client/plugins/__tests__/Print-test.jsx
@@ -202,6 +202,88 @@ describe('Print Plugin', () => {
         });
     });
 
+    it('test configuration with useFixedScales and enableScalebox = true', (done) => {
+        const printingService = {
+            getMapConfiguration() {
+                return {
+                    layers: [],
+                    center: {
+                        x: 0,
+                        y: 0,
+                        crs: "EPSG:4326"
+                    }
+                };
+            },
+            validate() { return {};}
+        };
+        getPrintPlugin({
+            state: {...initialState,
+                print: {...initialState.print,
+                    capabilities: {...initialState.print.capabilities,
+                        scales: [1000000, 500000, 100000].map(value => ({name: value, value}))}
+                }}
+        }).then(async({ Plugin }) => {
+            try {
+                await ReactDOM.render(<Plugin
+                    projectionOptions={{
+                        "projections": [{"name": "UTM32N", "value": "EPSG:23032"}, {"name": "EPSG:3857", "value": "EPSG:3857"}, {"name": "EPSG:4326", "value": "EPSG:4326"}]
+                    }}
+                    printingService={printingService}
+                    useFixedScales mapPreviewOptions={{
+                        enableScalebox: true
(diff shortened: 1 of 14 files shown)


tdipisa:
@offtherailz I would like if you can schedule a call with @mahmoudadel54 to review issues requirements before moving on with this review.

offtherailz (changes requested):
- [ ] I  noticed that the printing result is still wrong when switching coordinates. It have to be consistent with the map. 

[screencast-localhost_8081-2024.06.18-18_21_32.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/32493050-c870-4455-85d1-b55db021b146)

Everything have to be consistent. I'll review the code where it is fixed. 

- [ ] Other useful information are not clear, please explain them better.

web/client/plugins/__tests__/Print-test.jsx
```diff
@@ -202,6 +202,88 @@ describe('Print Plugin', () => {
...
+        };
+        getPrintPlugin({
+            state: {...initialState,
+                print: {...initialState.print,
+                    capabilities: {...initialState.print.capabilities,
+                        scales: [1000000, 500000, 100000].map(value => ({name: value, value}))}
+                }}
+        }).then(async({ Plugin }) => {
+            try {
+                await ReactDOM.render(<Plugin
```
offtherailz, line 227: I don't think ReactDOM.render returns a promise

web/client/plugins/__tests__/Print-test.jsx
```diff
@@ -202,6 +202,88 @@ describe('Print Plugin', () => {
...
```
offtherailz, line 268: same here. Await makes sense only on promises.

web/client/plugins/__tests__/print/Projection-test.jsx
```diff
@@ -119,11 +120,12 @@ describe('PrintProjection Plugin', () => {
...
     it('map transformer with user chosen crs', (done) => {
-        getPrintProjectionPlugin().then(({ Plugin, store }) => {
+        getPrintProjectionPlugin().then(async({ Plugin, store }) => {
             try {
-                ReactDOM.render(<Plugin projections={[{"name": "Mercator", "value": "EPSG:3857"}, {"name": "WGS84", "value": "EPSG:4326"}]}/>, document.getElementById("container"));
+                const comp = ReactDOM.render(<Plugin projections={[{"name": "Mercator", "value": "EPSG:3857"}, {"name": "WGS84", "value": "EPSG:4326"}]}/>, document.getElementById("container"));
+                await ReactTestUtils.act(async() => comp);
                 const combo = getByXPath("//select");
-                ReactTestUtils.Simulate.change(combo, {
+                await ReactTestUtils.Simulate.change(combo, {
```
offtherailz, line 128: `.Simulate.change` does not return a promise, so using `await` is wrong.

web/client/plugins/__tests__/print/Projection-test.jsx
```diff
@@ -139,11 +141,12 @@ describe('PrintProjection Plugin', () => {
...
     it('validator without allowPreview', (done) => {
-        getPrintProjectionPlugin().then(({ Plugin, store }) => {
+        getPrintProjectionPlugin().then(async({ Plugin, store }) => {
             try {
-                ReactDOM.render(<Plugin projections={[{"name": "Mercator", "value": "EPSG:3857"}, {"name": "WGS84", "value": "EPSG:4326"}]}/>, document.getElementById("container"));
+                const comp = ReactDOM.render(<Plugin projections={[{"name": "Mercator", "value": "EPSG:3857"}, {"name": "WGS84", "value": "EPSG:4326"}]}/>, document.getElementById("container"));
+                await ReactTestUtils.act(async() => comp);
                 const combo = getByXPath("//select");
-                ReactTestUtils.Simulate.change(combo, {
+                await ReactTestUtils.Simulate.change(combo, {
```
offtherailz, line 149: .Simulate.change does not return a promise, so using await is wrong.

mahmoudadel54:
> I noticed that the printing result is still wrong when switching coordinates. It have to be consistent with the map. [screencast-localhost_8081-2024.06.18-18_21_32.webm](https://github.com/geosolutions-it/MapStore2/assets/1279510/32493050-c870-4455-85d1-b55db021b146)
> 
> Everything have to be consistent. I'll review the code where it is fixed.

Hi @offtherailz 

After checking this issue, I found out, this is the same currently behavior in print [in DEV or QA] ---> if the print projection is one of the geographic coordinate systems [like 4326] and the **'geodetic'** = true  into the "overrideOptions" in print cfg, and when I edited **'geodetic'** to be false, printing map with geographic projection e.g: 4326 printing works for me.
Honestly I don't know the reason.
This is a demo of my debugging: 

https://github.com/geosolutions-it/MapStore2/assets/58145645/b0084bac-1579-4742-b888-db22244b1863

mahmoudadel54:
> * [ ]  Other useful information are not clear, please explain them better.

@offtherailz  I have updated the useful information section. I hope it is more clear now

offtherailz (changes requested):
AS we discussed in a call @mahmoudadel54 the PR has still the inonsistencies I notified in the Preview.

As we checked, the preview looks consistent when `geodetic` is set to `false`, so our development will do it in that direction, assuming that `geodetic` is set to `false`.
Now print plugin allows to send the geodetic parameter via configuration, but it doesn't handle to change the preview to fit the final result.

offtherailz:
Now it looks a lot better.
I'm reporting here in a comment my findings just to keep track of them.
I'll check if these issues of the preview already existed and let you know.

Discrepancy between the preview and the scale sent (as we have seen in our meeting).

image

offtherailz (changes requested):
Please fix the tests and do a sert of manual tests more.

offtherailz (approved):

tdipisa:
@ElenaGallo please test this in DEV once deployed and let us know for the backport.

ElenaGallo:
Test passed, @mahmoudadel54 please backport to 2024.01.xx Thanks

mahmoudadel54:
> Test passed, @mahmoudadel54 please backport to 2024.01.xx Thanks

@ElenaGallo Backport is done ---> https://github.com/geosolutions-it/MapStore2/pull/10452

tdipisa:
@ElenaGallo I've also tested using test map [here](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/demo_context/49601) and all seems to work as expected with this fix.

PR #10365, opened 2024-05-28
Update User Guide - Add support for HTML response for WFS layer

ElenaGallo:
## Description

Add support for HTML response for WFS layer on [Layer Settings](https://mapstore--10365.org.readthedocs.build/en/10365/user-guide/layer-settings/#feature-info-form) section


docs/user-guide/layer-settings.md
```diff
@@ -527,6 +527,9 @@ In particular, the user can choose between:
 !!!note
     Without selecting any format here, the [Identify Tool](navigation-toolbar.md#identify-tool) will return the layers information with the format chosen in Map Settings ( in the [Side Toolbar](mapstore-toolbars.md#side-toolbar)). Once a user specifies the information format in layers settings, instead, that format will take precedence over the map settings only for that specific layer.
 
+!!!warning
```
tdipisa, line 530: Should it be really a _warning_ or a simple _note_ is maybe better? Anyway, in this case there should be a specific mention that the HTML format for WFS layers is available only with GS and if the corresponding extension is used on the GS side. Also a link the GS extension page would be useful as a reference.

tdipisa (changes requested):

Issue #10488, opened 2024-07-24
'ilike' as default operator for text field in attribute table quick filter


tdipisa:
## Description

With https://github.com/geosolutions-it/MapStore2/issues/9706 the Attribute Table quick filter has been enhanced with the operator drop down but the default operator for text fields now is `=` instead of `ilike` as it should be.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

PR #9743, opened 2023-11-23
#9706: Allow time range filtering for Attribute Table quick filter

mahmoudadel54:
## Description
- adding filter operators dropdown list in attribute table that includes range operator for time/date attributes 
- creating a custom picker for range
- write unit tests
- Edit in style 

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [x] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue
#9706 

**What is the current behavior?**
#9706 

**What is the new behavior?**
Different filter operators dropdown is added to the attribute table columns including range time/date

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #10496, opened 2024-07-25
#10488: 'ilike' as default operator for text field in attribute table quick filter

mahmoudadel54:
## Description
In this PR, handling specifying 'ilike' as the default operator in DD operators in attribute table for the text fields is implemented and adding a unit test for that.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
#10488 

#10488 

**What is the new behavior?**
Now the attribute table shows 'ilike' operator for the text fields as a default operator for quick filter.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/featuregrid/filterRenderers/AttributeFilter.jsx
+++ b/web/client/components/data/featuregrid/filterRenderers/AttributeFilter.jsx
@@ -23,7 +23,7 @@ class AttributeFilter extends React.PureComponent {
         column: PropTypes.object,
         placeholderMsgId: PropTypes.string,
         tooltipMsgId: PropTypes.string,
-        operator: PropTypes.string,
+        defaultOperator: PropTypes.string,
         type: PropTypes.string,
         isWithinAttrTbl: PropTypes.bool
     };
@@ -38,7 +38,7 @@ class AttributeFilter extends React.PureComponent {
         onChange: () => {},
         column: {},
         placeholderMsgId: "featuregrid.filter.placeholders.default",
-        operator: "=",
+        defaultOperator: "=",
         isWithinAttrTbl: false
     };
     constructor(props) {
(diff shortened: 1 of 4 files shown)


web/client/components/data/featuregrid/filterRenderers/AttributeFilter.jsx
```diff
@@ -50,7 +50,7 @@ class AttributeFilter extends React.PureComponent {
             booleanOperators: ["="],
             defaultOperators: ["=", ">", "<", ">=", "<=", "<>", "isNull"],
             timeDateOperators: ["=", ">", "<", ">=", "<=", "<>", "><", "isNull"],
-            operator: this.props.isWithinAttrTbl ? "=" : "",
+            operator: this.props.isWithinAttrTbl ? (this.props.operator || '=') : "",
```
allyoucanmap, line 53: @mahmoudadel54 The `this.props.operator` was introduced in a previous PR https://github.com/geosolutions-it/MapStore2/pull/9743 and in this new PR is used to apply a default value. Few questions:
- What was the role of the prop `operator` before this change? It was used?
- If the prop `operator` was existing before why do we are using it to apply a default value?

mahmoudadel54: @allyoucanmap 
this.props.operator  ---> it is used in BaseDateTimeFilter component - which inherits from AttributeFilter class component - in onChange handlers but after double checking I think I can remove it and just use the operator state

for the second point, I think again I can just use initiate the state operator based on this.props.type to implement ilike in case of string field

If Ok, I will push the code for that, thanks

allyoucanmap: @mahmoudadel54 I would suggest to introduce a specific prop for the default value such as defaultOperator, we cannot use type because the current request is a change just to the attribute table while the usage of type could effect also the query builder.
About the removal of `this.props.operetor` we should do it only if we are 100% sure it's not really used

allyoucanmap (changes requested):

allyoucanmap (approved):

allyoucanmap:

Issue #8113, opened 2022-04-19
Allow WMS/WFS request through the MapStore proxy


tdipisa:
## Description

MapStore is preventing you to add sources in HTTP when you are in HTTPs to avoid mixed content issues. We implemented this through #6127. This is due to a browser security constraint (eg. Chrome from v81 tries to upgrade the http request to https and blocks them by default if they fail to load over https, see also here). 

Through this issue it is requested to let certain OGC requests (like GetMap requests) to pass through the MapStore proxy.

## Acceptance criteria

From our preliminary tests the proxy  should already works for the purpose of this task. Anyway there, given the proxy is ok we need to: 

- Allow to add catalog in http-not-secure
- Allow WMS layers to load tiles 
- Do not support other type of protocol for the moment (TMS, WMTS... ) because these require additional work on specific implementations 

In order to support what requested for this task, we need to change several things here and there on MapStore UI: 

- Add a flag in the UI of advanced options of catalog of type WMS and WFS.  Let's call it in the UI "Allow not secure layers". Option in catalog should be "allowUnsecureLayers"
- A WFS layer, with the fix above, should work. To be checked
- Also WMS with Cesium should work, because they already use the proxy. To be checked
- In case a **WMS** layer has been added using a catalog, a flag `forceProxy: true` option in the layer should be added. This option forces the layer to use the proxy and load the data, instead of doing a direct request. Only Openlayers supports this parameter, Leaflet doesn't, therefore mobile maps will not work if that option is set to **true**

This option has to be : 
   - Documented on MapStore
   - Saved in map (there is a filter of savable options) and so restored in a second time when the map is loaded
   - Fixed in implementation (Openlayers has some implementation but is partially overridden by other things, so it is not fully working). 
   - The flag should be added to the display tab of layer setting to allow to enable/disable in a second time at layer level

PR #10526, opened 2024-09-02
fix #10505 Allow to specify use of proxy or cors at layer level

### LINKED ISSUE #10505: Allow to specify use of proxy or cors at layer level
## Description

The way the http-proxy is used in the application need to be changed and improved to guarantee more automation, better flexibility and exempting also users from interacting with certain concepts CORS related.

At the moment, the application proxy is used by default for all Ajax requests cross domain. If the requested base URL is specified in the [useCORS](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/configs/localConfig.json#L4C10-L4C17) configuration, then MS assumes that CORS headers are enabled and allowed for the origin and the proxy is not used automatically.

There are also different behaviors on how layers tile requests are performed based on the mapping library, e.g.:
- OpenLayers by default doesn't use the proxy for loading layer tiles
- Cesium by default configures the proxy based on the global _[proxyUrl](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/configs/localConfig.json#L3)_ property in _localConfig.json_

With this issue it is requested to change the current behavior **to not use the proxy by default** (1) and let the MS application to automatically establish when the use of the proxy is necessary through a proper logic depending on the request.

## Acceptance criteria

- [x]  A preflight request should be performed (e.g simply checking the [useCORS](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/configs/localConfig.json#L4C10-L4C17) first and performing an test request early after if necessary) to check if CORS are enabled and allowed. This could happen for example when a catalog source is created or updated but other valid solutions are also accepted 
- [x]  Basically, where necessary, the proxy should be automatically used for all requests the application needs to perform (e.g. Ajax requests) if CORS are not allowed. The MS logic should be anyway enough robust and smart to understand if under certain special conditions the proxy is necessary (even if CORS are allowed, for example) such as:
   - [x] To prevent mixed contents when HTTP resources are requested under HTTPs (the proxy is needed anyway). A message should be also notified to the user in catalog tool when a source is added/saved by saying something like: 
         `The specified URL is not secure as it uses HTTP. An URL using HTTPs protocol should be provided. Using HTTP, requests will be handled automatically to prevent browser security issues by passing through the proxy.`
   - [x] To load OL layer tiles in map according to the source type (OL doesn't need tile URLs passing through the proxy for WMS GetMap requests, for example, even if CORS are not allowed)
- [x] The `Force proxy` (Layer Settings) and the `Allow not secure layers` (Catalog tool for WMS/WFS sources) are no longer needed and the logic can be removed
- [x] The requested default overall behavior (1) should be coherent for all source types supported (WMS, WFS, WMTS, TMS, ArcGIS, IFC and 3D Tiles) as well as mapping libraries (OL, Leaflet and Cesium)

## Other useful information
As a consequence of this implementation the PR https://github.com/geosolutions-it/MapStore2/pull/10474 and the [connected issue](https://github.com/geosolutions-it/MapStore2/issues/10473) become out of date and can be closed with a comment.

rowheat02:
## Description

- fixes #10505. proxy by default implemented if any endpoint fails due to CORS
- UI removed for: Force proxy (Layer Settings) and the Allow not secure layers
 - Interceptor has been used to use retry with proxy to all Axios endpoints.
- For Cesium, before adding a layer to the map one preflight request is used to check CORS and use the proxy.
- Common proxy cache for both cases.

 - [ ] Bugfix
 - [ ] Feature
 - [x] Refactoring 
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

 #10505

**What is the new behavior?**
AutoProxy for CORS disabled endpoints.

## Breaking change
 - [x] No


rowheat02:
- UI removed for: Force proxy (Layer Settings) and the Allow not secure layers
- If any api fails due to CORS, proxy will be used by default in ajax interceptor.
- In case of cesium, preflight CORS check api will be called before adding layer and forceProxy option will be used to create new layer which handle proxy logic in layers.
- common proxy cache logic implemented.
- Old Failed test cases other than "should create a bil terrain provider with wms config" have been solved. In this last failed test, two async parts are there throwing timeout error.

tdipisa:
@allyoucanmap I think this is ready for a new review now.

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo this PR introduced an automatic way to apply the proxy url (`/proxy/?url=`) to layer and requests, we need to test the following:
- test all the layer types available in MapStore in the 3D viewer (Cesium). The should work as usual, we should also find some layer services with CORS disabled to check if the automatic detection is working
- verify that the allow unsecure layer options from the advanced setting of catalog has been removed and a new warning message is visible when a user add an http service in a https instance (see #8113). When an unsecure layer is added to the map the proxy url (`/proxy/?url=`) must be applied automatically (you can verify in the network tab)
- verify that the force proxy option in layer settings has been removed

PR #10531, opened 2024-09-04
Fix #10530 fix catalog editor for array scenario

### LINKED ISSUE #10530: If a the url of service catalog is defined as array when you open it  the application crash
## Description

it crash because [this](https://github.com/geosolutions-it/MapStore2/blame/master/web/client/components/catalog/editor/MainForm.jsx#L166) gets trigger when editing the catalog and an array is not accepted, so we have to pass the first element of the array in case of array to prevent the crash to happen

in the onChangeUrl of defaultservice the value is an array, but when the onchange is triggered, the output value becomes comma separated string, so there we do not have the problem

## How to reproduce

-  create a service with an array of urls (also same url) in the url property
-  open the map with this service and try to edit it

*Expected Result*

*Current Result*

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MV88:
## Description

fixing case when catalog url service is array, all details in issue

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Fix #10530

**What is the new behavior?**

no crash in case of array 

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/catalog/editor/MainForm.jsx
+++ b/web/client/components/catalog/editor/MainForm.jsx
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 import React, { useState, useEffect } from 'react';
-import {get, find, isEmpty} from 'lodash';
+import {isArray, get, find, isEmpty} from 'lodash';
 import Message from '../../I18N/Message';
 import HTML from '../../I18N/HTML';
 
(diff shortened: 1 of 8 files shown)


tdipisa:
@MV88 there is a failing test which is not depending on this PR. @offtherailz will check it during review.

offtherailz (changes requested):
Instead of using the first URL, An error should be shown because the format with array is not supported, and so it is error prone.

Please show misconfiguration error.

MV88:
> Instead of using the first URL, An error should be shown because the format with array is not supported, and so it is error prone.
> 
> Please show misconfiguration error.

done, please review again

offtherailz (approved):
The catalog still works for some reason in search, while gives error in the editor.
even if not properly consistent, I think that it is enough for a misconfiguration, that by definition can have inconsistent results.

PR #10532, opened 2024-09-04
Fixed test build failures caused by ChromeHeadless height and a test dependent on window height

offtherailz:
## Description
This PR is to fix the issue of failures happening during build with `npm test`. 

image

After some attempts I noticed that the failing test was not failing using ``test:watch` but as well as I resized the window to a smaller height (`523px` was the limit in the test), it started failing with the same error.
So I supposed that some changes on `ChromeHeadless` may set the window to a smaller value, so the test fails. 

I tried to execute the test with smaller height values on `scroll-container` in the given test and I noticed that the size needed to make it fail by resizing the window height decreased. 
So I tried the same approach with tests with ChromeHeadless, where I can not resize the window, by adjustind the height of `scroll-container`. 
I found the limit of scroll container to 430px in scroll-container to make it fail. 429 is successful on ChromeHeadless.

So I decided to put the pixel at a very small value (for security), that make anyway the tests work, and it looks to work also with ChromeHeadless.

Hope this fixes also the failures on the backend. This fix may need to be backported

 - [x] CI related changes

## Issue

No issue at the moment. @tdipisa should we create one.

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/geostory/common/enhancers/__tests__/withMediaVisibilityContainer-test.jsx
+++ b/web/client/components/geostory/common/enhancers/__tests__/withMediaVisibilityContainer-test.jsx
@@ -64,11 +64,12 @@ describe('withMediaVisibilityContainer HOC', () => {
 
     it('scroll in view should render the actual component (lazy loading)', (done) => {
         const DEBOUNCE_TIME = 1;
+        // note: this test fails if the window is too small in height
         ReactDOM.render(
             <div
                 id="scroll-container"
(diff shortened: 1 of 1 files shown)


tdipisa (approved):

tdipisa:
@offtherailz it seems a backport to 2024.01.xx is necessary: https://github.com/geosolutions-it/MapStore2/actions/runs/10573476730/job/29713019887#step:9:9988

I ran the last build job on that branch and it fails for the same reason.

Issue #10247, opened 2024-04-23
Normalize layers/groups management in TOC 


tdipisa:
## Description

To maintain retro-compatibility with old maps, as reported here in https://github.com/geosolutions-it/MapStore2/issues/10246, with the new TOC there are still a few inconsistent but not blocking behaviors due to how the 'default group' is managed. 

Loading a Map in one of the following scenarios:

- Importing a Map
- Opening an existing Map 

and  logic should intervene to automatically update the TOC internal structure to definitely fix retro-compatibility issues.

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [X] Minor changes to existing features
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

mahmoudadel54:
What I understood from the description, for some old maps, it is requested to handle the logic of adding layer without group in case not selecting any existing group on TOC before adding. is it correct ?

The current implemented code is Ok with any new created map based on this PR: https://github.com/geosolutions-it/MapStore2/pull/10278, right ?

PR #10609, opened 2024-10-10
Normalize layers/groups management in TOC  #10247

rowheat02:
## Description

fixes: #10247.
Some old maps and exports have groups in legacy format which shows some bugs like those mentioned in #10247.
Old formats can be converted to new standard formats with UI-wise invisible default group as a parent of all groups.
```
legacyGroups = [
  { id: '1', name: 'Group 1' },
 { id: DEFAULT_GROUP_ID }
 ];

newStandardFormat=[
      {
          id: <DEFAULT_GROUP_ID>,
         name: <DEFAULT_GROUP_ID>,
        nodes: [
             { id: '1', name: 'Group 1' },
             { id: 'DEFAULT_GROUP_ID.<uuid>' } // Unique UUID generated
         ],
        expanded: true
    }
 ]
 
 ```

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#10247
There are some interactivity bugs on TOC in old maps and exports.

**What is the new behavior?**

There are no interactivity bugs on TOC in old maps and exports.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/LayersUtils.js
+++ b/web/client/utils/LayersUtils.js
@@ -23,6 +23,9 @@ import { addAuthenticationParameter } from './SecurityUtils';
 import { getEPSGCode } from './CoordinatesUtils';
 import { ANNOTATIONS, updateAnnotationsLayer, isAnnotationLayer } from '../plugins/Annotations/utils/AnnotationsUtils';
 import { getLocale } from './LocaleUtils';
+import { isSingleDefaultGroup } from '../plugins/TOC/utils/TOCUtils';
+import uuid from 'uuid';
+
 
 let LayersUtils;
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rowheat02** [comment]
@tdipisa 
- This PR solves the bugs mentioned here #10247
- Now the conversion is done considering groups from this map.https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/39079. Are there any other maps with TOC bugs to verify.

**Comment by tdipisa** [comment]
@allyoucanmap it seems @rowheat02 is waiting for your feedback to unblock and finalize this PR. Can you please have a look?

**Review by allyoucanmap — CHANGES_REQUESTED** [review]
This changes in this PR are technically making the presented cases in the issue work but it's not addressing this part:

> As part of this issue is definitely requested to provide also the necessary clean up for completely removing the concept of "default group" to make also old maps consistent with new TOC behaviors.

The solution provided is adding a group prefix to the default group resulting in a new group id (`Default.Default`) that it making it works someway.
We should instead understand the following:

- Why the default group has the sort handler blocked?
- Why the sort/move action used by the TOC is preventing the default to be moved?
- Why the sort/move action used by the TOC is preventing the layers to be placed in the root?
- Is it possible to remove the concept of Default group and place nodes (groups and layers) inside the TOC root?

We need to take this issue to really understand the problem related to groups and layers in TOC and if needed also refactor logic and component.
I think we could accept this PR only once we answered the questions above and there isn't other alternatives.

---

Side note:
Usually group ids in mapstore are following this structure `parentGroupId.groupId` but it seems with the proposed changes we are adding the following case, where `group01` and `group02` don't have a the parentId group:

```js
[
    {
        "id": "Default",
        "name": "Default",
        "nodes": [
            {
                "id": "Default.Default",
                ...
            },
            {
                "id": "group01",
                ...
            },
            {
                "id": "group02",
                ...
            }
        ]
    }
]
```
image

We need also to understand and investigate why is this still working

rowheat02:
This task will be placed on hold for the following reasons.

Currently, there is a single group wrapper named "Default," which is not visible in the UI. This wrapper supports all utility functions related to adding/moving nodes, exporting, and importing configurations from the existing codebase. The legacy system's default group (node) can coexist with other sibling groups. Additionally, there is the concept of a "root group" for nodes that are not assigned to any group. Two potential solutions to make it compatible for both new and legacy format of nodes have been identified, both of which involve breaking changes and require careful updates to legacy utility functions, followed by thorough testing.

**Solution 1**: Include both layers and groups in the group state, as the group state serves as the starting point for the Table of Contents (TOC) UI. However, this approach would disrupt all legacy utility functions that assume the group state should only contain groups as level-one nodes.

the format that every util function supports ( all group nodes as level one child in  group state)
```json
[
  {
    "id": "group node 1",
    "nodes": [
      {
        "id": "layer node 1"
      }
    ]
  },
  {
    "id": "group node 2"
  }
]

``` 
format proposed which breaks every util functions( layer node in group as level one child)
``` json
[
  {
    "id": "group node 1",
    "nodes": [
      {
        "id": "layer node 1"
      }
    ]
  },
  {
    "id": "layer node 1"
  },
  {
    "id": "group node 2"
  }
]
``` 

Implementing this would require a deep understanding of all utility functions and meticulous updates to ensure compatibility.

**Solution 2**: Introduce a root group that can wrap the Default group and other sibling groups while hiding it in the UI. 
```json
{
  "id": "root",
  "node": [
    {
      "id": "group node 1",
      "nodes": [
        {
          "id": "layer node 1"
        }
      ]
    },
    {
      "id": "layer node 1"
    },
    {
      "id": "group node 2"
    }
  ]
}

```
However, this adds an additional layer of hierarchy, which could complicate maintenance. It also risks breaking the format and functionality during the export and import of configurations.
Both solutions involve significant changes and need careful updates to legacy utility functions, along with extensive testing.

tdipisa:
closing as out of date. We will reopen when necessary @allyoucanmap

PR #10677, opened 2024-11-14
Returning wrong pixel xy value on mousemove for leaflet engine

shivadas1980:
## Description

**On mouse move over map using leaflet engine, it returns wrong pixel xy value.  It always return x value for x and y.**

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

**On mouse move on map using leaflet engine, return x pixel value for both x and y pixel value. **
#<issue>

**What is the new behavior?**

**After this fix, on mouse move over map return correct xy pixel value for leaflet engine**
## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/map/leaflet/Map.jsx
+++ b/web/client/components/map/leaflet/Map.jsx
@@ -488,7 +488,7 @@ class LeafletMap extends React.Component {
             crs: "EPSG:4326",
             pixel: {
                 x: event.containerPoint.x,
-                y: event.containerPoint.x
+                y: event.containerPoint.y
             },
             latlng: {
                 lat: event.latlng.lat,


ash03e608:
Hello @tdipisa , @allyoucanmap ,
This MR has been submitted from a team at Halliburton who is currently using Mapstore via the JS API. We are currently evaluating the stability of the Mapstore JS API and hence we are interested to know the history of this bug and if this was a recent regression. If not, is this a bug thats only exposed on the JS API and not via the React components. Since the workflow is fundamental and the error very obvious, i  wanted to hear your thoughts on it.

tdipisa:
@ash03e608 it is not a regression. That incorrect setup was introduced many years ago by this PR https://github.com/geosolutions-it/MapStore2/pull/1211
Probably because Leaflet map engine is less used and supported than OL/Cesium ones, this issue has never been reported before. Thank you very much for using MapStore.

PR #10761, opened 2025-01-16
chore: Fixed the attribute value of tooltipid ("remove-alias-button" instead of "add-alias-button").

youssefallalix:
## Description

This pull request fixes the incorrect attribute value for `tooltipid` in the relevant component. The previous value was set to `"add-alias-button"`, which has been corrected to `"remove-alias-button"` to ensure proper functionality and alignment with the intended behavior.

## Changes

- Updated the `tooltipid` attribute value from `"add-alias-button"` to `"remove-alias-button"`.

### DIFF
--- a/web/client/components/catalog/editor/AdvancedSettings/WMSDomainAliases.js
+++ b/web/client/components/catalog/editor/AdvancedSettings/WMSDomainAliases.js
@@ -50,7 +50,7 @@ export default ({
             {toInteger(k) !== 0 &&
             <TooltipButton
                 tooltip={<Message msgId="catalog.domainAliases.removeAliasTooltip" />}
-                tooltipid="add-alias-button"
+                tooltipid="remove-alias-button"
                 tooltipPosition="left"
                 className="remove-alias"
                 onClick={onRemoveAlias(k)}>


web/client/components/catalog/editor/AdvancedSettings/WMSDomainAliases.js
```diff
@@ -50,7 +50,7 @@ export default ({
             {toInteger(k) !== 0 &&
             <TooltipButton
                 tooltip={<Message msgId="catalog.domainAliases.removeAliasTooltip" />}
-                tooltipid="add-alias-button"
+                tooltipid="remove-alias-button"
```
allyoucanmap, line 53: Dear @youssefallalix , thanks for the contribution. I double check this prop and it seems it is not actually used by the component and it is passed directly to the html [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attributes)

image

so we could simple remove it.

allyoucanmap (changes requested):

tdipisa:
Hi @youssefallalix I see this PR is opened for a long time. Is it finally possible for you to provide the requested change according to the review above? Otherwise, we will have to close the PR. In the meantime, I would like to take this opportunity to ask you to visit the updated [contributing rules](https://github.com/geosolutions-it/MapStore2/wiki/Contributing-to-MapStore) in the project's WIKI where you can see info related to the CLA that is necessary for contributing updates to the MS core. Thank you so much again for your contribution in evolving MS and the MS community.

tdipisa:
> Hi @youssefallalix I see this PR is opened for a long time. Is it finally possible for you to provide the requested change according to the review above? Otherwise, we will have to close the PR. In the meantime, I would like to take this opportunity to ask you to visit the updated [contributing rules](https://github.com/geosolutions-it/MapStore2/wiki/Contributing-to-MapStore) in the project's WIKI where you can see info related to the CLA that is necessary for contributing updates to the MS core. Thank you so much again for your contribution in evolving MS and the MS community.

I'm going to close for now @youssefallalix. If necessary, we can consider reopening the PR according to your feedback.

Issue #10825, opened 2025-02-19
After saving a context requests for home loading are performed three times


tdipisa:
## Description

Requests for loading the MS home are performed more than enough after saving a context

![Image](https://github.com/user-attachments/assets/abbc3fed-d5e1-4016-8fe0-14b087279902)

## How to reproduce

- Search a context from home 
- Open the context in edit mode and save it 

*Expected Result*

The needed requests for loading the home are performed only one time

*Current Result*

As for the screenshot above requests are performed three times

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

allyoucanmap:
@dsuren1 I did an initial investigation, here my findings:

1 - I was not able to understand why only when redirected from context the page rerender 3 times mounting/unmounting component
2 - the cancel token is not passed to the final request https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/ResourcesCatalog/hooks/useQueryResourcesByLocation.js#L107-L109
3 - when this component unmount is missing the clearTimout for requestTimeout.current https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/ResourcesCatalog/hooks/useQueryResourcesByLocation.js#L209-L214

all three should be addressed

dsuren1:
**Other issue to check**
- Previous requests should be cancelled when performing filter by search
- Check if extensions are loaded properly

allyoucanmap:
@dsuren1(cc @ElenaGallo) I'm still seeing multiple requests on development environment:

![Image](https://github.com/user-attachments/assets/ad57c0fd-4f0c-4dda-977d-af8921addb81)

please @dsuren1 check again the problem, I noticed there is a loadExtension action right after the push, and I'm guessing that's causing the problem. I don't think we need that action when the redirect destination is different from `/context/${context.name}` or `/context-manager`

![Image](https://github.com/user-attachments/assets/146b7af5-d183-46c9-89b5-385e574eb3d3)

tdipisa:
@dsuren1 

>@tdipisa Should we handle this is as part of this issue or a separate issue?

[...]

PR #10868, opened 2025-02-26
#10825: Fix - Skip loading extensions when redirected to homepage

dsuren1:
## Description
This PR skips loading of extensions when redirected to homepage

 - [x] Bugfix

## Issue

- #10825 

**What is the new behavior?**
Upon saving the context and navigating to the homepage, no duplicate resource requests are triggered

## Breaking change
 - [x] No


web/client/epics/contextcreator.js
```diff
@@ -75,23 +75,23 @@ export const saveContextResource = (action$, store) => action$
...
+                (destLocation === destLocations.manager ? action$.ofType(CONTEXTS_LIST_LOADED).take(1).switchMap(() => Rx.Observable.of(
                     show({
                         title: "saveDialog.saveSuccessTitle",
                         message: "saveDialog.saveSuccessMessage"
                     }))) : Rx.Observable.empty()),
                 Rx.Observable.of(
                     contextSaved(rid),
-                    push(destLocation || `/context/${context.name}`),
-                    loadExtensions(),
+                    push(destLocation || destLocations.context),
```
allyoucanmap, line 93: If we remove the action we have this problem now:

https://github.com/user-attachments/assets/f867624e-8f64-4f4b-bc2b-3765eeb17dbf

Unfortunately my guess on the `loadExtensions` was wrong and this function is needed because it adds the new uploaded extensions in the page and it triggers a full application reload, so it's not possible to add it to an internal workflow at the moment (eg. while uploading an extension).
Not sure if we can check somewhat that a new extension has been uploaded during the current context creator save session and only in that case trigger the `loadExtensions` function?

allyoucanmap (changes requested):

allyoucanmap:

PR #10950, opened 2025-03-20
Fix #10947 Updated dockerfile to static files for standard templates

### LINKED ISSUE #10947: Builds using dockerfile complains about a binary config that should not be present
## Description

these two lines in the dockerfile are causing a little build issue in jenkins 
```
COPY binary/tomcat/conf/server.xml "${CATALINA_BASE}/conf/"
RUN sed -i -e 's/8082/8080/g' ${CATALINA_BASE}/conf/server.xml
```
so standard template Dockerfile file must be fixed

## How to reproduce

- try to configure a build with jenkins for example based on
- verify build works

*Expected Result*

success 

*Current Result*

fail 

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MV88:
## Description

Fixing the dockerfile for projects

```
# COPY binary/tomcat/conf/server.xml "${CATALINA_BASE}/conf/"
# RUN sed -i -e 's/8082/8080/g' ${CATALINA_BASE}/conf/server.xml
```

this is the only chagned part from original dockerfile

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Fix #10947

**What is the new behavior?**

## Breaking change
 - [x] No


project/standard/static/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM tomcat:9-jdk11 AS mother
+LABEL maintainer="Alessandro Parma<alessandro.parma@geosolutionsgroup.com>"
+ARG MAPSTORE_WEBAPP_SRC="https://github.com/geosolutions-it/MapStore2/releases/latest/download/mapstore.war"
```
offtherailz, line 3: This will copy mapstore from latest release. This is not what we want for project, or it is unuseful and they can take directly the docker container from dockerhub.
By the way, if this is the standard docker file, a future build of the docker will build always against the altest version, and this is not what we want. So this looks to be a bug of the current dockerfile too.

project/standard/templates/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM tomcat:9-jdk11 AS mother
+LABEL maintainer="Alessandro Parma<alessandro.parma@geosolutionsgroup.com>"
+ARG MAPSTORE_WEBAPP_SRC="https://github.com/geosolutions-it/MapStore2/releases/latest/download/mapstore.war"
```
offtherailz, line 3: same here, file downloads the latest mapstore, so dockerfile is unuseful in this case.
(2 copies of same docker file, too).

offtherailz (changes requested):
I put some notes, because if we want to provide docker file for projects, we have several bugs. 
Probably this was a problem existing before your contribution, but we have to take into account to avoid this (unuseful) feature for project, if it not makes sense.

offtherailz:
@tdipisa let's evaluate if makes sense to delay this to next release, and assign it to a devops.
We should : 
- include in the build the current built war
- also in project, same logic. The current project

project/standard/templates/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM tomcat:9-jdk11 AS mother
+LABEL maintainer="Alessandro Parma<alessandro.parma@geosolutionsgroup.com>"
```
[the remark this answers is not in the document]
MV88: since is present in the static folder i think we can safely remove it from here

project/standard/templates/Dockerfile
MV88: will remove this file

MV88:
> I put some notes, because if we want to provide docker file for projects, we have several bugs. Probably this was a problem existing before your contribution, but we have to take into account to avoid this (unuseful) feature for project, if it not makes sense.

@tdipisa @offtherailz 
this happened because in one of the latest projects we had, our devop took dockerfile from main mapstore, so if we need a dedicated version for projects then yes, it might be better assigned to a devop

What should i do here other than removing a duplication?

offtherailz:
It is needed to : 

- set MAPSTORE_WEBAPP_SRC to the default `mapstore.war` file of the project and it should work
- we need to document it properly (E.g. build before to run `docker build` ). 
- the same for mapstore, MAPSTORE_WEBAPP_SRC should be set to `product/target/mapstore.war` or whatever it is, and properly docuemnt it too.

tdipisa:
@offtherailz the backport need to be done anyway, I guess...

offtherailz:
@MV88 I think you can backport

Issue #11021, opened 2025-04-15
Implement an option to have static WMS legends


giohappy:
A [recent enhancement](https://github.com/geosolutions-it/MapStore2/pull/10756) to MapStore2 has introduced dynamic WMS legends, which adapt the legend for a WMS layer to the content inside the map viewport.
This changes the previous behavior where legends were static. 

Such a change introduces a few problems:

 - projects where legends are expected always to show the full symbology
 - increased rate of GetLegendGraphics requests that, in certain cases, can lead to performance issues (https://github.com/GeoNode/geonode-mapstore-client/issues/2022)

We want to implement an option to revert to the previous behavior.

- The option can be set globally (localConfig) to affect the default behavior for legends.
- It can be overridden for a layer through Layer Settings 

In my opinion, the default behavior should be the previous one, since this is what is expected normally, with the option to have dynamic legends.

tdipisa:
Thank you for reporting that @giohappy 
In my opinion it should be simply one of the following:
1. A dedicated option for activating that ("legend filtered by viewport") also from UI as is for the interactive legend one (just after it in UI)
2. Once you activate interactive legend you activate also that legend filter behavior.

tdipisa:
@dsuren1

>Should we also allow configuring this option from Catalog or only in Display tab of TOC ? @tdipisa 

No, I don't think so, TOC is enough. Ok for the rest, I would say. @giohappy any concern that I'm not considering?

giohappy:
I'm not sure about the constraints between interactive legend and filtering by viewport, but I trust you.

The goal is to have the old behavior as the default.

allyoucanmap:
> Thank you for reporting that [@giohappy](https://github.com/giohappy) In my opinion it should be simply one of the following:
> 
> 1. A dedicated option for activating that ("legend filtered by viewport") also from UI as is for the interactive legend one (just after it in UI)
> 2. Once you activate interactive legend, you activate also that legend filter behavior
> 
> What do you think?

@tdipisa it's not just viewport/bbox filter, the work done mentioned in the issue's description was also containing filter to CQL that should be also excluded if we want to have the previous behavior. My suggestion is to exclude also this kind of filter and call the property `dynamicLegend` (`Dynamic legend` checkbox in UI)

dsuren1:
**Updated requirement**

- when dynamic legend is true, add content dependent legend params
- when interactive legend is true, add content dependent legend params
[...]

PR #10756, opened 2025-01-15
[Backport 2024.02.xx] #10684 & #10236 & #10241 - Legend filtering enhancement 

dsuren1:
## Backport includes

- #10684
- #10236
- #10241

PR #10718, opened 2024-12-05
#10684: Legend filtering for GeoServer WMS layers

dsuren1:
## Description
This PR implements legend filtering of the WMS layer including both the formats (image and json). Incorporates filtering factoring map extents and additional layer filter and interactions between them

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Enhancement

## Issue

**What is the current behavior?**
- #10684 
- #10241 

**What is the new behavior?**
- **Layer and legend filtering**

  https://github.com/user-attachments/assets/7144bf51-c7f4-4f51-b38e-460d9027b163

- **Legend filtering in printing**

  Screenshot 2024-12-05 at 7 26 25 PM
  Screenshot 2024-12-05 at 7 27 10 PM

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #10249, opened 2024-04-23
#10026: fix functional tests for Interactive legend for TOC layers [WMS Support]

mahmoudadel54:
## Description
This PR includes fixing functional tests for interactive legend #10026:
- Hide checkbox of enabling interactive legend for WMS layers in dashboard and geostory as they will be implemented in a separate issue.
- fix not saving the flag of enabling interactive legend into wms layers in case save map.

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue
#10026

**What is the current behavior?**
Saving map with a layer that has a flag of enabling interactive legend to keep it in each open this map was not working well.
Dashboard and geostory were containing the checkbox of enabling interactive legend in Display tab in layer settings.

**What is the new behavior?**
Save map is handling storing the enable interactive legend flag to keep it stored if user close the map and open it again.
For dashboard and geostory pages, the enable interactive legend checkbox in layer settings in Display tab is hidden now.

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #11034, opened 2025-04-17
#11021: Implement an option to have static WMS legends

dsuren1:
## Description
This PR adds the ability to opt in or out of legend filtering by viewport. The field is available in Display tab of the layer setting. By default this option is not active. The feature also applies to print tool

image

**Tooltips**
image

image

 - [x] Enhancement

## Issue

- #11021 

**What is the new behavior?**
- User can view legend by filtering based on map viewport or static legend
- Same applies to print tool

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/LegendUtils.js
+++ b/web/client/utils/LegendUtils.js
@@ -56,19 +56,20 @@ export const getWMSLegendConfig = ({
         SLD_VERSION: "1.1.0",
         LEGEND_OPTIONS: legendOptions
     };
-
     if (layer.serverType !== ServerTypes.NO_VENDOR) {
-        const filterViewport = layer.enableLegendFilterByViewport;
+        const addContentDependantParams = layer.enableDynamicLegend || layer.enableInteractiveLegend;
         return {
             ...baseParams,
-            // hideEmptyRules is applied for all layers except background layers
-            LEGEND_OPTIONS: `hideEmptyRules:${layer.group !== "background"};${legendOptions}`,
-            SRCWIDTH: mapSize?.width ?? 512,
-            SRCHEIGHT: mapSize?.height ?? 512,
-            SRS: projection,
-            CRS: projection,
-            ...optionsToVendorParams({ ...layer, layerFilter: getLayerFilterByLegendFormat(layer, format) }),
-            ...(filterViewport && mapBbox?.bounds && {BBOX: getExtentFromViewport(mapBbox, projection)?.join(',')})
+            ...(addContentDependantParams && {
+                // hideEmptyRules is applied for all layers except background layers
+                LEGEND_OPTIONS: `hideEmptyRules:${layer.group !== "background"};${legendOptions}`,
+                SRCWIDTH: mapSize?.width ?? 512,
+                SRCHEIGHT: mapSize?.height ?? 512,
+                SRS: projection,
+                CRS: projection,
+                ...optionsToVendorParams({ ...layer, layerFilter: getLayerFilterByLegendFormat(layer, format) }),
+                ...(mapBbox?.bounds && {BBOX: getExtentFromViewport(mapBbox, projection)?.join(',')})
+            })
         };
     }
 
--- a/web/client/utils/LegendUtils.js
+++ b/web/client/utils/LegendUtils.js
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { isEmpty } from "lodash";
+import isEmpty from "lodash/isEmpty";
 import { getExtentFromViewport } from "./CoordinatesUtils";
 import { ServerTypes } from "./LayersUtils";
 import { optionsToVendorParams } from "./VendorParamsUtils";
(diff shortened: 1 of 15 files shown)


web/client/utils/LegendUtils.js
```diff
@@ -66,8 +67,8 @@ export const getWMSLegendConfig = ({
             SRCHEIGHT: mapSize?.height ?? 512,
             SRS: projection,
             CRS: projection,
-            ...(mapBbox?.bounds && {BBOX: getExtentFromViewport(mapBbox, projection)?.join(',')}),
-            ...optionsToVendorParams({ ...layer, layerFilter: getLayerFilterByLegendFormat(layer, format) })
+            ...optionsToVendorParams({ ...layer, layerFilter: getLayerFilterByLegendFormat(layer, format) }),
+            ...(filterViewport && mapBbox?.bounds && {BBOX: getExtentFromViewport(mapBbox, projection)?.join(',')})
```
allyoucanmap, line 71: A requirement of the issue https://github.com/geosolutions-it/MapStore2/issues/11021#issue-2995902456 is:

> In my opinion, the default behavior should be the previous one, since this is what is expected normally, with the option to have dynamic legends.

The referenced enhancements are listed here https://github.com/geosolutions-it/MapStore2/pull/10756, in particular https://github.com/geosolutions-it/MapStore2/pull/10718 where the `BBOX` parameter was introduced.

In the same issue were introduced other parameters: `SRCWIDTH`, `SRCHEIGHT`, `hideEmptyRules` (in `LEGEND_OPTIONS`), `SRS`, `CRS`, `CQL_FILTER` and `BBOX`  (please double check if other parameters were introduced).

If we want to cover the requirement request we need to exclude all parameters that were not available before. I would like also to propose a change in the property name from enableLegendFilterByViewport to `enableDynamicLegend`.

allyoucanmap (changes requested):

web/client/utils/LegendUtils.js
tdipisa: @allyoucanmap all options you listed have been included together within the same macro task because them all are connected to achieve the same purpose (which is https://github.com/geosolutions-it/MapStore2/pull/10756, indeed). Therefore, it make sense to me what you are saying. I was assuming the same but maybe I was wrong.
I also agreed with _enableDynamicLegend_ which is a terminology that better groups many options aimed at providing a specific functionality all together.

dsuren1: After discussion, the only way to make the legend params meaningful in both the cases it have content dependant legend params available when either `enableDynamicLegend` is true or interactiveLegend is true. Otherwise the params used doesn't fetch meaningful legend data.

1. enableDynamicLegend is true, add content dependent legend params
2. interactiveLegend is true, add content dependent legend params
3. when interactiveLegend is true, enableDynamicLegend is auto enabled and the field interaction is disabled
4. enableDynamicLegend is false, no param is added


I see another flag called `hideInteractiveLegendOption` initally added to hide the interactive legend from geostory and dashboard (however, i don't see it being used. We can see the options from geostory and dashboard in dev, maybe used by downstream project not sure). But this flag purpose is not properly fulfilled, as the interactive legend can still be enabled when adding a layer from catalog. So we should enable interactive legend only when hideInteractiveLegendOption is false. @allyoucanmap @tdipisa Your thoughts?

why this is needed, because it contradicts with new approach suggested above, i.e when hideInteractiveLegendOption is true and layer added from catalog enables interactive legend. This makes the new dynamicLegend field disabled (point 3) and user is left with no way to toggle it

allyoucanmap: > I see another flag called hideInteractiveLegendOption initally added to hide the interactive legend from geostory and dashboard (however, i don't see it being used. We can see the options from geostory and dashboard in dev, maybe used by downstream project not sure). But this flag purpose is not properly fulfilled, as the interactive legend can still be enabled when adding a layer from catalog. So we should enable interactive legend only when hideInteractiveLegendOption is false. @allyoucanmap @tdipisa Your thoughts?

@dsuren1 @tdipisa If we are looking to the PR https://github.com/geosolutions-it/MapStore2/pull/10249 introducing this, we can see the `hideInteractiveLegendOption` used to temporary hide this checkbox inside the TOCItemsSettings of the MapEditor inside the geostory and used directly in the node editor of dashboard. Now this options is exposed has documented [configuration](https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.TOCItemsSettings) for the plugin so we cannot remove it. The documentation is also explaining that the sole purpose of this property is to hide the checkbox so I would not enable interactive legend only when hideInteractiveLegendOption is false.
My proposal:

- add deprecated flag to the property `hideInteractiveLegendOption` in documentation for TOCItemsSettings 
- if both `hideInteractiveLegendOption` and `enableInteractiveLegend` are true also the checkbox for dynamic legend will not be visible

I think the aim should be to remove `hideInteractiveLegendOption`

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test this on dev and let use know when we can backport it to 2025.01.xx and 2024.02.xx, thanks

ElenaGallo:
@dsuren1 @tdipisa The deployment to DEV doesn't seem to have been done. Can someone check?

tdipisa:
@ElenaGallo the update is now in DEV ready to be tested.

PR #11067, opened 2025-05-02
Fix #10966 basic auth for services

### LINKED ISSUE #10966: Insert credentials at runtime when connecting to secured services
## Description

## Acceptance criteria

  - [ ] for openlayers
  - [ ] for cesium
  - [ ] ~not for leaflet~

MV88:
## Description

A new SecurityPopup plugin has been added to handle secured services using basic auth.

main features
- you can defined protected services in the catalog service editor of main viewer through a new button 
- this button is injected from SecurityPlugin to enhance Catalog one
image
- when defined a new username and password for a particular service and you save the service all layers added now will be "secured"
image
- credentials are stored in session storage
- security object of layers are stored, but it only stores the sourceId to be able to fetch from session storage an the type of method  ("basic")
- also services has a protectedId property that detemines if a service requires basic auth
- this is applied to maps, dashboards and geostories
- when resource page is loaded a check is done and if we find one or more services/layers protected we ask for introducing credentials if  they are not present is session storage otherwise
- all layers update functions have been considering a security change (rand property is used to trigger the refresh and therefore render the layer)

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Fix #10966 

**What is the new behavior?**

## Breaking change
 - [x] No

## Other useful information

This is how a secured service look like 
image

### For tester
We created a set of rules using GeoFence in gs-stable for testing. i have created a layer "linea costa basic" that has access using some credentials, can can ask in pvt. so you can try using it
image

the idea is that this will only work if a protected service is configured, and failing if not. This will be valid for many layer types in ol and cesium

### DIFF
--- a/web/client/actions/layers.js
+++ b/web/client/actions/layers.js
@@ -25,7 +25,6 @@ export const SHOW_SETTINGS = 'SHOW_SETTINGS';
 export const HIDE_SETTINGS = 'HIDE_SETTINGS';
 export const UPDATE_SETTINGS = 'UPDATE_SETTINGS';
 export const REFRESH_LAYERS = 'REFRESH_LAYERS';
-export const REFRESH_SECURITY_LAYERS = 'LAYERS:REFRESH_SECURITY_LAYERS';
 export const UPDATE_LAYERS_DIMENSION = 'LAYERS:UPDATE_LAYERS_DIMENSION';
 export const LAYERS_REFRESHED = 'LAYERS_REFRESHED';
 export const LAYERS_REFRESH_ERROR = 'LAYERS_REFRESH_ERROR';
@@ -234,11 +233,6 @@ export function refreshLayers(layers, options) {
         type: REFRESH_LAYERS,
         layers,
         options
-    };
-}
-export function refreshSecurityLayers() {
-    return {
-        type: REFRESH_SECURITY_LAYERS
     };
 }
 
--- a/web/client/actions/layers.js
+++ b/web/client/actions/layers.js
@@ -25,6 +25,7 @@ export const SHOW_SETTINGS = 'SHOW_SETTINGS';
 export const HIDE_SETTINGS = 'HIDE_SETTINGS';
 export const UPDATE_SETTINGS = 'UPDATE_SETTINGS';
 export const REFRESH_LAYERS = 'REFRESH_LAYERS';
+export const REFRESH_SECURITY_LAYERS = 'LAYERS:REFRESH_SECURITY_LAYERS';
 export const UPDATE_LAYERS_DIMENSION = 'LAYERS:UPDATE_LAYERS_DIMENSION';
 export const LAYERS_REFRESHED = 'LAYERS_REFRESHED';
 export const LAYERS_REFRESH_ERROR = 'LAYERS_REFRESH_ERROR';
@@ -233,6 +234,11 @@ export function refreshLayers(layers, options) {
         type: REFRESH_LAYERS,
         layers,
         options
+    };
+}
+export function refreshSecurityLayers() {
+    return {
+        type: REFRESH_SECURITY_LAYERS
     };
 }
 
--- a/web/client/actions/security.js
+++ b/web/client/actions/security.js
@@ -29,6 +29,9 @@ export const LOGOUT = 'LOGOUT';
 export const REFRESH_SUCCESS = 'REFRESH_SUCCESS';
 export const SESSION_VALID = 'SESSION_VALID';
 
+export const SET_SHOW_MODAL_STATUS = 'SECURITY:SET_SHOW_MODAL_STATUS';
+export const SET_CREDENTIALS = 'SECURITY:SET_CREDENTIALS';
+
 export function loginSuccess(userDetails, username, password, authProvider) {
     return {
         type: LOGIN_SUCCESS,
@@ -168,3 +171,24 @@ export function verifySession() {
 }
 
 export const checkLoggedUser = () => ({type: CHECK_LOGGED_USER});
+
+/**
+ * set status to show modal for inserting credentials
+ * @param {boolean} status
+ */
+export const setShowModalStatus = (status) => {
+    return {
+        status,
+        type: SET_SHOW_MODAL_STATUS
+    };
+};
+/**
+ * set credentials for a service ogc
+ * @param {object} protectedService
+ */
+export const setCredentials = (protectedService) => {
+    return {
+        protectedService,
+        type: SET_CREDENTIALS
+    };
+};
--- a/web/client/actions/security.js
+++ b/web/client/actions/security.js
@@ -33,7 +33,7 @@ export const SET_SHOW_MODAL_STATUS = 'SECURITY:SET_SHOW_MODAL_STATUS';
 export const SET_CREDENTIALS = 'SECURITY:SET_CREDENTIALS';
 export const CLEAR_SECURITY = 'SECURITY:CLEAR_SECURITY';
 export const SET_PROTECTED_SERVICES = 'SECURITY:SET_PROTECTED_SERVICES';
-
+export const REFRESH_SECURITY_LAYERS = 'SECURITY:REFRESH_SECURITY_LAYERS';
 export function loginSuccess(userDetails, username, password, authProvider) {
     return {
         type: LOGIN_SUCCESS,
@@ -220,3 +220,12 @@ export const setCredentialsAction = (protectedService, creds) => {
         type: SET_CREDENTIALS
     };
 };
+
+/**
+ * action to use to rerender layers in map when security changed
+ */
+export function refreshSecurityLayers() {
+    return {
+        type: REFRESH_SECURITY_LAYERS
+    };
+}
--- a/web/client/actions/security.js
+++ b/web/client/actions/security.js
@@ -11,7 +11,7 @@
  */
 import AuthenticationAPI from '../api/GeoStoreDAO';
 
-import {getToken, getRefreshToken} from '../utils/SecurityUtils';
+import {setCredentials, getToken, getRefreshToken} from '../utils/SecurityUtils';
 import {encodeUTF8} from '../utils/EncodeUtils';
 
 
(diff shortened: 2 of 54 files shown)


web/client/actions/layers.js
```diff
@@ -235,6 +236,11 @@ export function refreshLayers(layers, options) {
         options
     };
 }
+export function refreshSecurityLayers() {
```
allyoucanmap, line 239: Could you add a comment or jsdoc to explain this action?

web/client/actions/security.js
```diff
@@ -168,3 +173,50 @@ export function verifySession() {
...
+/**
+ * set protected services to request to provide username and password
+ * @param {object[]} protectedServices
+ */
+export const setProtectedServices = (protectedServices) => {
+    return {
+        protectedServices,
+        type: SET_PROTECTED_SERVICES
+    };
+};
```
allyoucanmap, line 196: These two actions seems specific for the `SecurityPopup` plugin but they are inside the security actions.

web/client/actions/security.js
```diff
@@ -168,3 +173,50 @@ export function verifySession() {
...
+/**
+ * clear security of layers
+ * @param {string} id
+ */
+export const clearSecurity = (protectedId) => {
+    return {
+        protectedId,
+        type: CLEAR_SECURITY
+    };
+};
```
allyoucanmap, line 206: This action seems to have a behaviour opposite to refreshSecurityLayers, both acting on layers but they are implemented in different actions files.

web/client/components/map/cesium/plugins/WMTSLayer.js
```diff
@@ -115,11 +116,19 @@ function wmtsToCesiumOptions(_options) {
...
+            "Authorization": `Basic ${btoa(storedProtectedService.username + ":" + storedProtectedService.password)}`
+        };
+    }
     return assign({
         // TODO: multi-domain support, if use {s} switches to RESTFul mode
         url: new Cesium.Resource({
             url: head(getURLs(isArray(options.url) ? options.url : [options.url], queryParametersString)),
-            proxy: proxy && new WMTSProxy(proxy) || new NoProxy()
+            proxy: proxy && new WMTSProxy(proxy) || new NoProxy(),
+            ...(headers)
```
allyoucanmap, line 131: Looking at the doc, [headers](https://cesium.com/learn/ion-sdk/ref-doc/Resource.html?classFilter=Resource#headers) should be an object not spread inside the `Cesium.Resource` class, could you double check if it's working?

web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -18,7 +18,7 @@ import { getProjection } from '../../../../utils/ProjectionUtils';
 import { getConfigProp } from '../../../../utils/ConfigUtils';
 
 import {optionsToVendorParams} from '../../../../utils/VendorParamsUtils';
-import {addAuthenticationToSLD, addAuthenticationParameter, getAuthenticationHeaders} from '../../../../utils/SecurityUtils';
+import {/* getCredentials, */addAuthenticationToSLD, addAuthenticationParameter, getAuthenticationHeaders} from '../../../../utils/SecurityUtils';
```
allyoucanmap, line 21: please remove the commented method

web/client/plugins/MetadataExplorer.jsx
```diff
@@ -259,6 +261,17 @@ class MetadataExplorerComponent extends React.Component {
...
 }
 
+const MetadataExplorerComponentWrapper = (props, context) => {
+    const { loadedPlugins } = context;
+    const addonsItems = usePluginItems({ items: props.items, loadedPlugins }).filter(({ target }) => target === 'url-addon');
+    return <MetadataExplorerComponent {...props} addonsItems={addonsItems}/>;
+};
+
+
+MetadataExplorerComponentWrapper.contextTypes = {
```
allyoucanmap, line 271: If we are going in the direction of injected plugin we need documentation on how to use the `url-addon` items

web/client/actions/security.js
```diff
@@ -168,3 +173,50 @@ export function verifySession() {
...
+        setCredentials(protectedService.protectedId, {
+            ...creds,
+            url: protectedService.url
+        });
+    }
+    return {
+        protectedService,
+        type: SET_CREDENTIALS
+    };
+};
```
allyoucanmap, line 222: Not sure if it would be better to have the setCredentials side effect inside the reducer instead of the action

web/client/utils/LayersUtils.js
```diff
@@ -717,6 +717,7 @@ export const saveLayer = (layer) => {
     layer.localizedLayerStyles ? { localizedLayerStyles: layer.localizedLayerStyles } : {},
     layer.options ? { options: layer.options } : {},
     layer.credits ? { credits: layer.credits } : {},
+    layer.security ? { security: layer.security } : {},
```
allyoucanmap, line 720: Just for confirmation, does this property contain only the sourceId and type?

allyoucanmap (changes requested):

web/client/utils/LayersUtils.js
MV88: afaik yes

web/client/actions/security.js
allyoucanmap: @MV88 please ignore this comment

web/client/plugins/MetadataExplorer.jsx
MV88: will

web/client/actions/security.js
MV88: i did not want to create a securityPopop folder for just these actions but since the plugin is about security it might be good to have theme here

web/client/plugins/MetadataExplorer.jsx
```diff
@@ -341,6 +354,43 @@ const AddLayerButton = connect(() => ({}), {
  * @prop {number} cfg.zoomToLayer enable/disable zoom to layer when added
  * @prop {number} cfg.autoSetVisibilityLimits if true, allows fetching and setting visibility limits of the layer from capabilities on layer add (Note: The default configuration value is applied only on new catalog service (WMS/CSW))
  * @prop {number} [delayAutoSearch] time in ms passed after a search is triggered by filter changes, default 1000
+ * @prop {object[]} addonsItems this property contains the items injected from the other plugins,
```
allyoucanmap, line 357:
```suggestion
 * @prop {object[]} items this property contains the items injected from the other plugins,
```

web/client/plugins/MetadataExplorer.jsx
```diff
@@ -341,6 +354,43 @@ const AddLayerButton = connect(() => ({}), {
...
+ * createPlugin(
+ *  'MyPlugin',
+ *  {
+ *      containers: {
+ *          MetadataExplorer: {
+ *              name: "TOOLNAME", // a name for the current tool.
+ *              target: "url-addon", // the target where to insert the component
+ *              Component: MyAddonComponent,
+ *              // selector is `optional` and it will receive same prop as the Component eg.:
+ *              // selector: ({ status, statusTypes }) => status === statusTypes.DESELECT,
```
allyoucanmap, line 390:
```suggestion
 *              Component: MyAddonComponent
```

web/client/actions/security.js
```diff
@@ -168,3 +173,59 @@ export function verifySession() {
...
+    return {
+        protectedServices,
+        type: SET_PROTECTED_SERVICES
+    };
+};
+/**
+ * clear security of layers
+ * @param {string} id
+ */
+export const clearSecurity = (protectedId) => {
```
allyoucanmap, line 201: Probably this should be called clearSecurityLayers if the only purpose is to clear the layers security

web/client/actions/security.js
```diff
@@ -168,3 +173,59 @@ export function verifySession() {
...
+    return {
+        protectedId,
+        type: CLEAR_SECURITY
```
allyoucanmap, line 204: same as previous comment CLEAR_SECURITY_LAYERS

allyoucanmap (approved):

ElenaGallo:
@MV88 
1_ If the credentials are not entered in the service, it is still possible to add the "basic coastline" level to the map

https://github.com/user-attachments/assets/2b00c235-62a9-454a-aa51-e3bc463e2fe8


2_ when I set the credentials and click on the Confirm button, the credentials popup does not close automatically. You have to click on Cancel

https://github.com/user-attachments/assets/aa78eafc-759e-48e9-b2f2-017433c7ad98


3_ Once the Authentication button turns green, when you open it, the username and password are not visible

https://github.com/user-attachments/assets/6ff96a7c-dbd7-497b-bb52-71b3125a411a

MV88:
> 1_ If the credentials are not entered in the service, it is still possible to add the "basic coastline" level to the map

@tdipisa do we have to check this before adding the layer? from getCapabilities id no not know if a layer is protected via basic auth, the admin should know it, otherwise we have to try to make a request per each record to see if the getMap will fail or not without basic auth, but i',m not sure if we want this now or in the future

> 2_ when I set the credentials and click on the Confirm button, the credentials popup does not close automatically. You have to click on Cancel

checking probably a quick issue 

> 3_ Once the Authentication button turns green, when you open it, the username and password are not visible

this is because credentials are not persisted in the application, but only if a service / layers is protected using ids

tdipisa:
> > 1_ If the credentials are not entered in the service, it is still possible to add the "basic coastline" level to the map
> 
> @tdipisa do we have to check this before adding the layer? from getCapabilities id no not know if a layer is protected via basic auth, the admin should know it, otherwise we have to try to make a request per each record to see if the getMap will fail or not without basic auth, but i',m not sure if we want this now or in the future
> 
> > 2_ when I set the credentials and click on the Confirm button, the credentials popup does not close automatically. You have to click on Cancel
> 
> checking probably a quick issue
> 
> > 3_ Once the Authentication button turns green, when you open it, the username and password are not visible
> 
> this is because credentials are not persisted in the application, but only if a service / layers is protected using ids

Status after last sync reported in https://github.com/geosolutions-it/MapStore2/issues/10966#issuecomment-2855065683

PR #11070, opened 2025-05-05
backport c027-2024.01.xx -  Fix #10966 basic auth for services (#11067)

### LINKED ISSUE #10966: Insert credentials at runtime when connecting to secured services

MV88:
## Description

backport c027-2024.01.xx -  Fix #10966 basic auth for services (#11067)


web/client/plugins/ResourcesCatalog/components/InputControl.jsx
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright 2024, GeoSolutions Sas.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { FormControl as FormControlRB } from 'react-bootstrap';
+import withDebounceOnCallback from '../../../components/misc/enhancers/withDebounceOnCallback';
```
allyoucanmap, line 11: This file can be removed it's not used

web/client/components/layout/ConfirmDialog.jsx
```diff
@@ -0,0 +1,111 @@
...
+                        <Message msgId={confirmId} />
+                        {loading ? <>{' '}<Spinner /></> : null}
+                    </Button>
+                </FlexBox>
+            </FlexBox>
+        </Modal>
+    );
+}
+
+export default ConfirmDialog;
```
allyoucanmap, line 80: This file can be removed it's not used

web/client/components/catalog/editor/MainForm.jsx
```diff
@@ -7,45 +7,74 @@
...
-
 import {getConfigProp} from '../../../utils/ConfigUtils';
-
 import InfoPopover from '../../widgets/widget/InfoPopover';
-import { FormControl as FC, Form, Col, FormGroup, ControlLabel, Alert } from "react-bootstrap";
-
 import localizedProps from '../../misc/enhancers/localizedProps';
 import {defaultPlaceholder, isValidURL} from "./MainFormUtils";
 
+import IconRB from '../../../plugins/ResourcesCatalog/components/Icon';
```
allyoucanmap, line 19: In this case it's better to use directly Glyphicon so we can remove Icon from this PR

web/client/plugins/ResourcesCatalog/components/Icon.jsx
```diff
@@ -0,0 +1,63 @@
...
+import { Glyphicon } from 'react-bootstrap';
+import { loadFontAwesome } from '../../../utils/FontUtils';
+import useIsMounted from '../../../hooks/useIsMounted';
+import PropTypes from 'prop-types';
+
+function FaIcon({
+    name,
+    className,
+    style
+}) {
```
allyoucanmap, line 19: We can remove this component if we use Glyphicon instead in the other files. Then we can remove the ResourcesCatalog folder

web/client/components/layout/FlexBox.jsx
```diff
@@ -0,0 +1,98 @@
...
+ * @prop {string} gap gap in between flex items, one of `xs`, `sm`, `md` or `lg`
+ * @prop {bool} wrap it makes the flex layout wrap on the next line
+ * @prop {bool} centerChildren center children vertically and horizontally
+ * @prop {bool} centerChildrenHorizontally center children horizontally
+ * @prop {bool} centerChildrenVertically center children vertically
+ */
+const FlexBox = forwardRef(({
+    children,
+    className,
+    classNames = [],
```
allyoucanmap, line 30: missing less file with css rules

web/client/components/layout/Spinner.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+/**
+ * Spinner circle component, it inherits color and size from the css font rules
+ * @prop {string} id html identifier
+ * @prop {string} className custom class name
+ * @prop {object} style custom style object
+ */
+function Spinner({
```
allyoucanmap, line 17: missing less file with css rules

web/client/components/layout/Text.jsx
```diff
@@ -0,0 +1,71 @@
...
+ * Text component with utilities classes
+ * @prop {string} className custom class name
+ * @prop {string[]} classNames list of custom class names
+ * @prop {any} component a valid component to replace the default one
+ * @prop {string} fontSize one of `sm`, `md`, `lg`, `xl` or `xxl`
+ * @prop {bool} ellipsis if true it applies ellipsis when text overflows
+ * @prop {string} textAlign one of `left`, `right` or `center`
+ * @prop {bool} strong if true it applies strong style
+ */
+const Text = ({
```
allyoucanmap, line 25: missing less file with css rules

web/client/components/layout/Button.jsx
```diff
@@ -0,0 +1,34 @@
...
+const Button = forwardRef(({
+    children,
+    variant,
+    size,
+    square,
+    className,
+    borderTransparent,
+    ...props
+}, ref) => {
+    return (
```
allyoucanmap, line 21: missing less file with css rules

allyoucanmap (changes requested):

web/client/components/catalog/editor/MainForm.jsx
MV88: this is a custom branch where we backport that feature, i would prefer to avoid to have too many differences from the master branch. i would skip this

allyoucanmap (changes requested):

allyoucanmap (approved):

PR #11077, opened 2025-05-06
Fix #10966 improved basic auth modal and logic

### LINKED ISSUE #10966: Insert credentials at runtime when connecting to secured services

MV88:
## Description

new modal is like this

image

Adding these points to be done
- [x] basic auth must be added also to GetCapabilities requests
- [x] add x in top right, remove cancel button
- [x] position of clear button same height of ok, all fields inline taking flex space
- [x] show credentials when modal opens
- [x] check context edit button --> mandatory and hidden of pluginsconfig
- [x] add show password button

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Fix #10966

**What is the new behavior?**

see description

## Breaking change
 - [x] No

### DIFF
--- a/web/client/api/ArcGIS.js
+++ b/web/client/api/ArcGIS.js
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { getCredentials } from '../utils/SecurityUtils';
+import { getAuthorizationBasic } from '../utils/SecurityUtils';
 import axios from '../libs/ajax';
 import { reprojectBbox } from '../utils/CoordinatesUtils';
 import trimEnd from 'lodash/trimEnd';
@@ -88,14 +88,8 @@ export const searchAndPaginate = (records, params) => {
     };
 };
 const getData = (url, params = {}) => {
-    let headers = {};
     const protectedId = params?.info?.options?.service?.protectedId;
-    const storedProtectedService = getCredentials(protectedId);
-    if (storedProtectedService) {
-        headers = {
-            "Authorization": `Basic ${btoa(storedProtectedService.username + ":" + storedProtectedService.password)}`
-        };
-    }
+    let headers = getAuthorizationBasic(protectedId);
     const request = _cache[url]
         ? () => Promise.resolve(_cache[url])
         : () => axios.get(url, {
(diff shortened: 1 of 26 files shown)


web/client/api/ArcGIS.js
```diff
@@ -87,12 +88,21 @@ export const searchAndPaginate = (records, params) => {
...
 };
 const getData = (url, params = {}) => {
+    let headers = {};
+    const protectedId = params?.info?.options?.service?.protectedId;
+    const storedProtectedService = getCredentials(protectedId);
+    if (storedProtectedService) {
+        headers = {
+            "Authorization": `Basic ${btoa(storedProtectedService.username + ":" + storedProtectedService.password)}`
+        };
+    }
```
allyoucanmap, line 98: This same operation is repeated multiple times in this PR, probably we could create a function in SecurityUtils that given a service it returns headers.

allyoucanmap (commented):
I noticed that here we are using `protectedId` to select the credential while inside WFSLayerUtils is using `sourceId`, is this correct?

https://github.com/geosolutions-it/MapStore2/blob/0ec0d929ae447450e7f4e21859a49ee08c2cb7b3/web/client/utils/WFSLayerUtils.js#L65-L95

web/client/api/ArcGIS.js
MV88: sure

MV88:
> I noticed that here we are using `protectedId` to select the credential while inside WFSLayerUtils is using `sourceId`, is this correct?
> 
> https://github.com/geosolutions-it/MapStore2/blob/0ec0d929ae447450e7f4e21859a49ee08c2cb7b3/web/client/utils/WFSLayerUtils.js#L65-L95

I answer also here in order to explain the reasoning:
**protectedId** is for services data, and **sourceId** was a preexisting name used in **layer.security** object (for WFS mainly)
So in order to avoid confusion I have created two different names, but I can quickly replace them.

allyoucanmap (approved):

ElenaGallo:
@MV88 If the credentials are still "basic", once set in the protected service pop-up, the service can no longer be saved.

https://github.com/user-attachments/assets/0eda4737-dcf7-4337-bfd1-dd2f5824f31d

PR #11082, opened 2025-05-08
fix #10966 some inconsistencies on the security popup modal

### LINKED ISSUE #10966: Insert credentials at runtime when connecting to secured services

MV88:
## Description

fix #10966 [some inconsistencies](https://github.com/geosolutions-it/MapStore2/pull/11077#issuecomment-2858144922) on the security popup modal

The form sometimes were not correctly populated when opening 

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#10966 

**What is the new behavior?**

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/SecurityPopup.jsx
+++ b/web/client/plugins/SecurityPopup.jsx
@@ -65,12 +65,19 @@ function SecurityPopup({
 
     const id = services[currentFormIndex]?.protectedId;
     const show = currentFormIndex + 1 < services.length;
-    const currentCreds = getCredentials(id);
-    const [creds, setCreds] = useState(currentCreds);
+    const [creds, setCreds] = useState({});
 
     useEffect(() => {
-        setCreds(currentCreds);
-    }, [id]);
+        const credentials = getCredentials(id);
+        setCreds(credentials);
+    }, [id, currentFormIndex]);
+    useEffect(() => {
+        const credentials = getCredentials(id);
+        setCreds(credentials);
+        return () => {
+            setCreds({});
+        };
+    }, []);
     function handleCancel() {
         const nextIndex = show ? currentFormIndex + 1 : 0;
         onSetShowModal(show);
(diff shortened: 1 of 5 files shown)


web/client/plugins/SecurityPopup.jsx
```diff
@@ -65,12 +65,19 @@ function SecurityPopup({
...
+        const credentials = getCredentials(id);
+        setCreds(credentials);
+    }, [id, currentFormIndex]);
+    useEffect(() => {
+        const credentials = getCredentials(id);
+        setCreds(credentials);
+        return () => {
+            setCreds({});
+        };
+    }, []);
```
allyoucanmap, line 80: I tested locally and I found this issue, where the form is empty after saving:

https://github.com/user-attachments/assets/5994678e-4607-4e8b-a6ac-c0fa29116dee

Here I opened a PR with a proposal to change the management of credential directly inside the popup to take advantage of mount unmount event triggered by the key prop https://github.com/MV88/MapStore2/pull/7

Please double check if the changes included in the PR works as expected

allyoucanmap (changes requested):

web/client/plugins/SecurityPopup.jsx
MV88: > I tested locally and I found this issue, where the form is empty after saving:
> 
>  security-empty.mp4 
> Here I opened a PR with a proposal to change the management of credential directly inside the popup to take advantage of mount unmount event triggered by the key prop [MV88#7](https://github.com/MV88/MapStore2/pull/7)
> 
> Please double check if the changes included in the PR works as expected

i fixed also tests. you can review again

PR #11287, opened 2025-07-04
Cannot exit the filter configuration panel during the setup of a widget for dashboards #11269

### LINKED ISSUE #11269: No action provided to exit the filter configuration panel during the setup of a widget for dashboards
## Description
If the user opens the filter configuration during the creation of a widget for a dashboard, there's nor clear way to exit it and return to the setup.

## How to reproduce

- Create an empty dashboard  
- Add a table widget (but it can be any widget that supports filtering)
- Click the "Filer" icon 
- The filter configuration panel is shown
- No action is provided to exit the configuration

*Expected Result*
I expect that when the user creates a widget and opens the filter configuration panel, an action is clearly provided to exit the configuration and return to the setup of the widget.

*Current Result*
When the user creates a widget and opens the filter configuration panel, no clear action is provided to exit the configuration and return to the setup.

## Other useful information
It can be reproduced on both demo instances, stable and dev.

Image

rowheat02:
## Description

On  the Dashboard, Query panel and Query Map were overlapping/hiding the brandNavbar. Top position was adjusted for both of them for dashboard

 - [x] Bugfix

fixes #11269 

#11269
**What is the new behavior?**

- On the Dashboard page, while performing filter, query panel and map will not overlap/hide brandnavbar.
- Search icon is changed to tick icon
image

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/query/QueryBuilder.jsx
+++ b/web/client/components/data/query/QueryBuilder.jsx
@@ -204,6 +204,7 @@ class QueryBuilder extends React.Component {
                 storedFilter={this.props.storedFilter}
                 advancedToolbar={this.props.advancedToolbar}
                 loadingError={this.props.loadingError}
+                queryBtnGlyph="ok"
             /></div>);
         const { spatialMethodOptions, toolsOptions, spatialOperations} = this.props;
         return this.props.attributes.length > 0 ?
--- a/web/client/plugins/querypanel/SpatialFilterMap.jsx
+++ b/web/client/plugins/querypanel/SpatialFilterMap.jsx
@@ -6,10 +6,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 import React from 'react';
+import { createPortal } from 'react-dom';
 
 import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
-import Portal from '../../components/misc/Portal';
 
 import withContainer from '../../components/misc/WithContainer';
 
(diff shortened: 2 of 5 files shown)


web/client/components/data/query/QueryBuilder.jsx
```diff
@@ -136,7 +137,8 @@ class QueryBuilder extends React.Component {
         items: [],
         selectedLayer: false,
         standardItems: {},
-        style: {}
+        style: {},
+        dashboardAvailable: false
```
offtherailz: In general `dashboardAvailable` is not a good name for a property. The component should not be aware of the dashboard existence. 
Anyway this is not necessary, because I suggested to remove the queryBtnGlyph to be "search", because there is no case when search is really needed.

web/client/components/data/query/QueryBuilder.jsx
```diff
@@ -204,6 +206,7 @@ class QueryBuilder extends React.Component {
                 storedFilter={this.props.storedFilter}
                 advancedToolbar={this.props.advancedToolbar}
                 loadingError={this.props.loadingError}
+                queryBtnGlyph={this.props.dashboardAvailable ? "ok" : "search"}
```
offtherailz: I'd make this always "ok", there is no search anymore.

web/client/plugins/QueryPanel.jsx
```diff
@@ -286,7 +290,9 @@ class QueryPanel extends React.Component {
                     sidebar: {
                         ...this.props.layout,
                         zIndex: 1044,
-                        width: 600
+                        width: 600,
+                        // Since the query panel in dashboard page is from top, need to set the top position to 38px which is the height of the Brand Navbar
+                        ...(this.props.dashboardAvailable ? { top: 38 } : {})
```
offtherailz: @allyoucanmap there is a way to avoid this in the new layout system?
Making it more abstract and maybe injected by some external property?

offtherailz (changes requested):
I asked to simplify one thing, moreover I asked @allyoucanmap a sugestion for a better management of layout, if any,

web/client/plugins/QueryPanel.jsx
allyoucanmap: @offtherailz I think this could be solved by using CSS and review the way the portal is working for the SpatialFilterMap component.

1 - adding this cssless rule the QueryPanel will be correctly visualized. With the relative position to the body wrapper the position absolute of the QueryPanel will start as 0,0 to the top left corner of the body wrapper
```css
#page-dashboard > .ms2-border-layout-body {
    position: relative;
}
```

2 - The SpatialFilterMap component has a strange setup, is using the withContainer HOC and at the same time the Portal component (note, the container value has no effect to the Portal). My proposal here is just to change the target container using the body wrapper as a target using the react-dom createPortal instead, something like:

```js
export default withContainer((props) => {
    const {
        container,
        useEmbeddedMap,
        hideSpatialFilter,
        queryPanelEnabled
    } = props;
    return useEmbeddedMap && !hideSpatialFilter && queryPanelEnabled ?
        createPortal(
            <div className="mapstore-query-map">
                <MapComponent {...props}/>
            </div>, document.querySelector('#page-dashboard > .ms2-border-layout-body') || container)
        : null;
});
```

Maybe we can make the custom container selector configurable (`#page-dashboard > .ms2-border-layout-body`).
If the SpatialFilterMap component becomes child of the body wrapper it will have same behaviour of Query panel and fit the available space in the body without covering the BrandNavbar, we need just to increase the z-index of .mapstore-query-map class

rowheat02:
@offtherailz , @allyoucanmap 
As suggested.
- Added
``` 
    #page-dashboard > .ms2-border-layout-body {
        position: relative;
    } 
```
- default gylp icon to "Ok"
- The portal is replaced by a create portal  
- targetContainerSelector is configurable in SpatialFilterMap
- In the query builder, the prop for SpatialFilterMap is passed for targetContainerSelector: "#page-dashboard > .ms2-border-layout-body"
    - If queryselector gets the element, the target container is that element; otherwise, the `container` that comes from `withContainer ` is the target container
    - - updated z-index of mapstore-query-map
    
image

allyoucanmap:
@ElenaGallo please test this on dev and let us know when we can backport, thanks

ElenaGallo:
Test passed, @rowheat02 please backport to 2025.01.xx. Thanks

PR #11330, opened 2025-07-18
Support of getfeatureinfo to change by time

### LINKED ISSUE #11327: Support of getfeatureinfo to change by time
We have improve the feature info tool to refresh responses on timeline changes, by performing again the request with the new time value

subashtiwari1010:
## Description

This PR fixes issue #11327. It adds the support of the getFeatureInfo for layers in the Info popup when the time changes from the Timeline Plugin.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

Even if the time in the Timeline plugin is updated, the feature information does not change, as no API calls are made.
Fix #11327

**What is the new behavior?**
When the time changes, the API request for the getFeatureInfo is made, hence updating the information in the Feature info popup.

## Breaking change
 - [x] No

## Other useful information


web/client/epics/identify.js
```diff
@@ -440,6 +440,36 @@ export const setMapTriggerEpic = (action$, store) =>
...
+            // Only process if params is time and there's a click point
+            const state = getState();
+            return includes(Object.keys(params), "time") && clickPointSelector(state);
+        })
+        .switchMap(({ layer = [], params = {} }) => {
+            const state = getState();
+            const groups = rawGroupsSelector(state);
+            const queryableLayers = reverse(getDerivedLayersVisibility(queryableLayersSelector(state), groups));
+            const filterNameList = queryableLayers.filter(({ id, visibility }) => visibility && layer.includes(id)).map(({ name }) => name);
+            const overrideParams = filterNameList.reduce((obj, filterName) => ({ ...obj, [filterName]: params}), {});
```
offtherailz: `OverrideParams` doesn't look to be necessary. I removed it and it works anyway. Did you needed to use it for some reason?
But for other params, I'm not sure they can all be reproduced except this.
I'd suggest to use as sample `onUpdateFeatureInfoClickPoint` epic, that should do more or less what we should do...

Did you tried to do it yet?

offtherailz (commented):
Technically works, but there is something not necessary.
In case of different options etc... I'm not sure if it will work. See my suggestion below and try

web/client/epics/identify.js
subashtiwari1010: Used the `OverrideParams` to use the parameters that have been updated.
As in the layers tree, the time params are updated when there are changes made, the updated layers are used in the `featureInfoClick` function, so even if the `OverrideParams` is not passed, the functionality works.

As for other parameters, I have taken the reference from `onUpdateFeatureInfoClickPoint` and made the necessary changes to the function.

offtherailz (commented):
It works well. 
I noticed that if I have 2 time layers and ( e.g. raster_speed and vector_speed, or meteorilte landings) and I selectr in feature info combobox on top a different layer, when I update the feature info due to coordinate change or to timeline change, the selection resets. 
Maybe we can improve it somehow, but I don't want to break things to have this behavior (e.g. what happens if I keep selection but next version do not have information for the given layer...
Can you try to keep selected layer and see the various use case, to see if it somehow can break?

subashtiwari1010:
> Can you try to keep selected layer and see the various use case, to see if it somehow can break something or if it is a safe thing?

@offtherailz I have updated the functionality, so that it will always keep the last selected layer, and will update to new layer only when there is no information for the selected layer.

PR #11348, opened 2025-07-21
Extend the charts to show the current time

### LINKED ISSUE #11326: Extend the charts  to show the current time
## Description
Extend the charts  to show the current time .: Here we need to receive a depeendency for widget of currentTime. If one axis is of type time, then we can report the value (or the interval) included.
We have to properly insert the dependency as the bounding box for time. Avoid to include it when the timeline is out of data range (see anyway during dev, what's the final effect and what is needed for this. 
We need also to design how to make this properly extensible with alternatives like time filtering that have to be managed.

## Acceptance criteria

- If the user create a widgets on the map from a layer that has time data (e.g. use meteorite landings layer).
- When selecting axis properties, if the user forces the type of the axis to "date", a checkbox appears, to ask to show the current time.
- If the option is checked, the user sees the options to customize the current time line style
- the preview of the chart the user can see the line configured on the chart, in corrispondence of the current time in timeline
- When saved the widget will show the chart and the map, with same line that moves when time on timeline changes.

**Note**: when the widget and timeline are on the same map, they can be switched each other, but with "pin" of the widget, they can be shown at the same time.

## Other useful information

Basically we should: 
- add to the "Axis" section (X or Y) a checkbox to "show current time on the axis" . Thinking if make sense to show it only when the user forces the time value to date... let's see what's better during dev.
here: 

https://github.com/user-attachments/assets/db29675c-8a1e-46ca-b0d6-4f509eadac74

- The current time should be shown on the chart
    - as a dashed line
    - as an area below the chart, similar to the one that shows on the timeline, in case of range.

subashtiwari1010:
## Description
This PR adds the functionality to display the current time in the charts, issue #11327. It adds the checkbox to display the current time in the Axes tab of the charts.
image    image

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

There is no option to add the current time in the chart.
Fixes #11326 

**What is the new behavior?**
````
1. UI to Show the current time on the axis.
2. When checked, the user can view the current time either single or range on the axis.
3. When saved, the same feature will be available in the chart widgets too.
````

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/charts/WidgetChart.jsx
+++ b/web/client/components/charts/WidgetChart.jsx
@@ -579,6 +579,7 @@ export const toPlotly = (_props) => {
             }} : {}),
             hovermode: 'x unified',
             uirevision: true,
+            shapes: [...(layout?.shapes || [])],
             ...gridProperty
         },
         data: traces.filter((trace, idx) => !!dataProp[idx]).reduce((acc, {
--- a/web/client/components/widgets/builder/wizard/ChartWizard.jsx
+++ b/web/client/components/widgets/builder/wizard/ChartWizard.jsx
@@ -36,6 +36,7 @@ import ChartTraceEditSelector from './chart/ChartTraceEditSelector';
 import TraceAxesOptions from './chart/TraceAxesOptions';
 import TraceLegendOptions from './chart/TraceLegendOptions';
 import { isChartOptionsValid } from '../../../../utils/WidgetsUtils';
+import { addCurrentTimeShapes } from '../../../../utils/widgetUtils';
 
 const loadingState = loadingEnhancer(({ loading, data }) => loading || !data, { width: 500, height: 200 });
 const hasNoAttributes = ({ featureTypeProperties = [] }) => featureTypeProperties.filter(({ type = "" } = {}) => type.indexOf("gml:") !== 0).length === 0;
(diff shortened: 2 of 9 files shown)


offtherailz (changes requested):
The functionality works 👍 
Anyway I'd ask you some changes to improve readability and compatibility.

- [ ] Use dependencies. We can not add a parameter on every new dependency. For this reason there is a dependency mapping system and several enhancer that allows to get values for dependencies and update the widget every time it . I'd suggest a "dependenciesToShapes" enhancer that checks timeRange dependency (the flag and it s settings, to draw the shapes on the chart). 
- [x] Name not simply range but timeRange
- [x] Maybe it is the case to allow to select also the style of the line, when checked (size, dash style and color).

offtherailz (changes requested):
- [x] For options and labels, instead of spread generic shapeColor, shapeSize, shapeStyle, use an object `currentTimeOpts: {style, size, color}` or we can confuse with other shapes.
- [x] For labels, shape style etc... is confusing. I suggest to use color, style and size, and group them, something like this shold work 

Image

final effect (replace labels with simply "Color" "Style" and "Size"
Image

offtherailz:

ElenaGallo:
@subashtiwari1010 When the "Show the current time on the axis" option is enabled, the highlighted date on the timeline is January 1st, 2021 instead of the current date.

Used [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/59233) for testing.

bgvf

offtherailz:
Hi @ElenaGallo , current time looks to be  January 1st, 2021 in the map you provided:

image

and the layer looks to contain some data for the given value. 

image

So it looks ok to me

Issue #11358, opened 2025-07-24
Performace Lag when there is a layer with very large geomerty


rowheat02:
## Description
Performance issue observed when there is at least one layer with big geometry
Replicate the issue
- create a map with
- Upload a big GeoJSON as a vector layer.
- Notice the performance lag.
- Check this already prepared [Map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/59249) with big vector layers

**What kind of improvement you want to add?** (check one with "x", remove the others)
- [x]  Enhancement

Issue #11142, opened 2025-05-26
Performance Lag on Map with Large Number of Layers


rowheat02:
## Description
Significant performance lag has been observed when working with maps that contain a large number of layers. This issue primarily impacts core UI interactions such as panning, zooming, accessing the resource catalogue, etc.

https://github.com/user-attachments/assets/593b743c-32d1-45c9-840e-5309fdb6c20a

## How to reproduce

-  Create a map with a huge number of layers. I have prepared [this map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/55231). Open this map. It has 77 (7 backgrounds, 70 others) layers.
- Do basic UI interactions like zooming, panning on the map, and try to open the catalogue
- The system will be lagging noticeably.
- Noticed: Lagging is more on the local development build than the production build

*Expected Result*
The system should not be lagging on basic UI interaction

*Current Result*
The system lags when there are huge number of layers

- [x] Not browser related

## Initial Investigation Output

Steps taken to identify the problem:

1. Created a [test map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/55231) with large number of layers.
2. Disabling the TOC plugin and removing groups did not improve performance.
3. I compared the same map with the production version(see above video), where it worked smoothly.
4. Enabled plugins one at a time:
   - No lag with only the **Map** plugin enabled.
   - Lag started again when enabling plugins like **BrandNavbar**, **Save**, **SaveAs**, **DeleteResource**, and **ResourceDetails**.
5. All these plugins use the same [function](https://github.com/geosolutions-it/MapStore2/blob/94d9822b2f62cc4075ecc8d881b7cdbaed178ca7/web/client/plugins/ResourcesCatalog/selectors/save.js#L91-L97): `getPendingChanges`, which includes a heavy calculation (`computePendingChanges`).
6. After commenting out all uses of this function, the lag disappeared.

---

## Root Cause

- The [function](https://github.com/geosolutions-it/MapStore2/blob/94d9822b2f62cc4075ecc8d881b7cdbaed178ca7/web/client/plugins/ResourcesCatalog/selectors/save.js#L91-L97) `getPendingChanges`, and more specifically `computePendingChanges`, is called very often.
- It runs on every UI interaction and state update, which causes lag when there are many layers.

---
## Suggested Next Steps

- Look into optimizing or reducing the number of times `getPendingChanges ` runs.

PR #11359, opened 2025-07-24
Performace Lag when there is a layer with very large geomerty #11358

rowheat02:
## Description

fixes https://github.com/geosolutions-it/MapStore2/issues/11358
fixes https://github.com/geosolutions-it/MapStore2/issues/11142
A significant performance lag was observed when there is a map with a layer with big geometry and map with huge number of layers. This PR optimized the functionality to check if the config has changed, which was responsible for the lag.

 - [x] Enhancementy

## Issue

https://github.com/geosolutions-it/MapStore2/issues/11358
 https://github.com/geosolutions-it/MapStore2/issues/11142

**What is the new behavior?**

Will not experience lag when there is a layer with big geometry or  a map with huge number of layers.
On dev server two map has been created :
Map with big geometry -> `viewer/59249`
Map with a huge number of layers ->  `viewer/55231`

https://github.com/user-attachments/assets/76fc9843-b6fc-4b39-8f1d-68762f72e60a

On these maps, checks can be performed.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/utils/MapUtils.js
+++ b/web/client/utils/MapUtils.js
@@ -833,71 +833,125 @@ export const getIdFromUri = (uri, regex = /data\/(\d+)/) => {
     return findDataDigit && findDataDigit.length && findDataDigit.length > 1 ? findDataDigit[1] : null;
 };
 
-/**
- * Method for cleanup map object from uneseccary fields which
- * updated map contains and were set on map render
- * @param {object} obj
- */
-
-export const prepareMapObjectToCompare = obj => {
-    const skippedKeys = ['apiKey', 'time', 'args', 'fixed'];
-    const shouldBeSkipped = (key) => skippedKeys.reduce((p, n) => p || key === n, false);
-    Object.keys(obj).forEach(key => {
-        const value = obj[key];
-        const type = typeof value;
-        if (type === "object" && value !== null && !shouldBeSkipped(key)) {
-            prepareMapObjectToCompare(value);
-            if (!Object.keys(value).length) {
-                delete obj[key];
-            }
-        } else if (type === "undefined" || !value || shouldBeSkipped(key)) {
-            delete obj[key];
-        }
-    });
-};
 
 /**
- * Method added for support old key with objects provided for compareMapChanges feature
- * like text_serch_config
- * @param {object} obj
- * @param {string} oldKey
- * @param {string} newKey
+ * Determines if a field should be included in the comparison based on picked fields and exclusion rules.
+ * @param {string} path - The full path to the field (e.g., 'root.obj.key').
+ * @param {string} key - The key of the field being checked.
+ * @param {*} value - The value of the field.
+ * @param {object} rules - The rules object containing pickedFields and excludes.
+ * @param {string[]} rules.pickedFields - Array of field paths to include in the comparison.
+ * @param {object} rules.excludes - Object mapping parent paths to arrays of keys to exclude.
+ * @returns {boolean} True if the field should be included, false otherwise.
  */
-export const updateObjectFieldKey = (obj, oldKey, newKey) => {
-    if (obj[oldKey]) {
-        Object.defineProperty(obj, newKey, Object.getOwnPropertyDescriptor(obj, oldKey));
-        delete obj[oldKey];
+export const filterFieldByRules = (path, key, value, { pickedFields = [], excludes = {} }) => {
+    if (value === undefined) {
+        return false;
     }
+    if (pickedFields.some((field) => field.includes(path) || path.includes(field))) {
+        // Fix: check parent path for excludes
+        const parentPath = path.substring(0, path.lastIndexOf('.'));
+        if (excludes[parentPath] === undefined) {
+            return true;
+        }
+        if (excludes[parentPath] && excludes[parentPath].includes(key)) {
+            return false;
+        }
+        return true;
+    }
+    return false;
 };
 
 /**
- * Feature for map change recognition. Returns value of isEqual method from lodash
- * @param {object} map1 original map before changes
- * @param {object} map2 updated map
- * @returns {boolean}
+ * Prepares object entries for comparison by applying aliasing, filtering, and sorting.
+ * @param {object} obj - The object whose entries are to be prepared.
+ * @param {object} rules - The rules object containing aliases, pickedFields, and excludes.
+ * @param {string} parentKey - The parent key path for the current object.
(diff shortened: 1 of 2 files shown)


web/client/utils/MapUtils.js
```diff
@@ -833,71 +833,125 @@ export const getIdFromUri = (uri, regex = /data\/(\d+)/) => {
...
+        // Fix: check parent path for excludes
+        const parentPath = path.substring(0, path.lastIndexOf('.'));
+        if (excludes[parentPath] === undefined) {
+            return true;
         }
-    });
+        if (excludes[parentPath] && excludes[parentPath].includes(key)) {
+            return false;
+        }
+        return true;
```
allyoucanmap, line 860: I'm expecting to check just `excludes[path]`, why here we are looking also for the parent field? Is there a particular reason for this?

web/client/utils/MapUtils.js
```diff
@@ -833,71 +833,125 @@ export const getIdFromUri = (uri, regex = /data\/(\d+)/) => {
...
+            }
+        }
+        return false;
     }
+
+    // Fallback for primitives
+    if (!isEqual(a, b)) {
+        return true;
+    }
+    return false;
```
allyoucanmap: Why not using `return a !== b` ? Do we really need `isEqual`?

web/client/utils/MapUtils.js
```diff
@@ -833,71 +833,125 @@ export const getIdFromUri = (uri, regex = /data\/(\d+)/) => {
...
+    if (Array.isArray(a)) {
+        if (!Array.isArray(b) || a.length !== b.length) {
+            return true;
+        }
+        // same reference
+        if (a === b) return false;
+        for (let i = 0; i < a.length; i++) {
+            if (recursiveIsChangedWithRules(a[i], b[i], rules, `${parentKey}[]`)) return true;
+        }
+        return false;
```
allyoucanmap, line 902: Why are you using for loop instead of `.some(` ?

web/client/utils/MapUtils.js
```diff
@@ -833,71 +833,125 @@ export const getIdFromUri = (uri, regex = /data\/(\d+)/) => {
...
+        const bEntries = prepareObjectEntries(b || {}, rules, parentKey);
+        if (aEntries.length !== bEntries.length) {
+            return true;
+        }
+        for (let i = 0; i < aEntries.length; i++) {
+            const [key, value] = aEntries[i];
+            if (recursiveIsChangedWithRules(value, bEntries[i]?.[1], rules, `${parentKey}.${key}`)) {
+                return true;
+            }
+        }
```
allyoucanmap, line 918: Why are you using for loop instead of `.some(` ?

web/client/utils/MapUtils.js
```diff
@@ -833,71 +833,125 @@ export const getIdFromUri = (uri, regex = /data\/(\d+)/) => {
...
+    // Handle arrays
+    if (Array.isArray(a)) {
+        if (!Array.isArray(b) || a.length !== b.length) {
+            return true;
+        }
+        // same reference
+        if (a === b) return false;
+        for (let i = 0; i < a.length; i++) {
+            if (recursiveIsChangedWithRules(a[i], b[i], rules, `${parentKey}[]`)) return true;
+        }
```
allyoucanmap, line 901: Why are you using for loop instead of `.some(` ?

web/client/utils/MapUtils.js
allyoucanmap, line 918: Why are you using for loop instead of `.some(` ?

allyoucanmap (changes requested):

web/client/utils/MapUtils.js
rowheat02: The excludes object is defined as { parent: [keys to exclude] }, so we need to check the parent path to match the exclude structure.
**Example**

-  Exclude: { 'root.map.layers[]': ['apiKey', 'time'] }
- Path: 'root.map.layers[].apiKey'
-  Parent path: 'root.map.layers[]' (to match the exclude key)

That's why I'm checking excludes[parentPath] instead of excludes[path] because excludes are defined by parent paths, not full field paths.

web/client/utils/MapUtils.js
rowheat02: I thought of using isEqual just to be safe. But I think it can be replaced by `a !== b`. No reason to use isEqual for primitives.
Updated the PR.

web/client/utils/MapUtils.js
rowheat02: For loop is faster than .some.
Ref: https://www.measurethat.net/Benchmarks/Show/15248/4/arraysome-vs-for-loop
Although the difference is not huge on a small scale. But still, it is good to use a faster technique where performance improvement is needed.

web/client/utils/MapUtils.js
rowheat02: For loop is faster than .some.
Ref: https://www.measurethat.net/Benchmarks/Show/15248/4/arraysome-vs-for-loop
Although the difference is not huge on a small scale. But still, it is good to use a faster technique where performance improvement is needed.

web/client/utils/MapUtils.js
rowheat02: For loop is faster than .some.
Ref: https://www.measurethat.net/Benchmarks/Show/15248/4/arraysome-vs-for-loop
Although the difference is not huge on a small scale. But still, it is good to use a faster technique where performance improvement is needed.

web/client/utils/MapUtils.js
rowheat02: For loop is faster than .some.
Ref: https://www.measurethat.net/Benchmarks/Show/15248/4/arraysome-vs-for-loop
Although the difference is not huge on a small scale. But still, it is good to use a faster technique where performance improvement is needed.

web/client/utils/MapUtils.js
rowheat02: Exclude definitions can be updated to be defined by a full path. 
I will check the performance difference and push the changes if it makes sense.

allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test this improvement on dev and let us know if we can backport, thanks

ElenaGallo:
Test passed, @rowheat02 please backport to 2025.01.xx. Thanks

Issue #11376, opened 2025-08-08
Support  widget creation in map even when layer have no geometry and fix time dependencies for the widget


subashtiwari1010:
## Description
1. When the layer does not have geometry, the creation of widget is blocked in the map even though the widget can be created in the dashboard. 
2. Current time in widget: When the map sync is disabled and enabled again, the dependency of the timeline i.e. currentTime and offsetTime will be removed, hence the time shapes is removed from the widget.

## How to reproduce
 1. For the issue of layer with no geometry: 
- Go to create new map
- Add a layer with no geometry in it
- Click the layer in the TOC, you will not be able to see the tool to create widget
Exit the map page now:
- Go to create new dashboard
- Add the same layer that have no geometry
- Click to add widget, now you can create the widget for that layer

2. For the current time in widget:
- Add a layer
- Click the layer to create a widget
- Add the field in the X and Y axis
- Go to axes, in the x axis, change the type to date and enable the current time
- Disable the connection with map
- Re-enable the connection with map

*Expected Result*
1. Even when the layer fails to load in the map (i.e. due to no geometry in it), the button to create the widget is displayed. So, for such layer the parameter `mapSync: false` is passed so the widget can be created without syncing the data and map.
2. Even when the connection is disabled and enabled, the time shape should not disappear.

*Current Result*
1. When the layer fails to load, on the TOC, the create new widget button will be hidden.
2. When connection is disabled and enabled, the time shape disappears from the widget.

- [x] Not browser related

offtherailz:
@ElenaGallo, could you please test this on  DEV ? Thank you

PR #11377, opened 2025-08-08
Support of widget in map even when layer have no geometry and fix time dependencies for the widget

subashtiwari1010:
## Description
This PR fixes issue https://github.com/geosolutions-it/MapStore2/issues/11376. It adds the support of the widget for the layer with no geometry and fixes the issue of time dependencies when the map connection disabled and re-enabled.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

When the layer is added in the map with no geometry, then there is error for that layer (as it does not have any geometry). So, the creation of the widget is not available for it. But, when the same layer is added to the dashboard, then the widget can be created for that layer using other fields.
image
Also, for the current time in the widget, when the map connection is disabled and re-enabled, then the current time shape disappears from the widget.
image image

Fixes https://github.com/geosolutions-it/MapStore2/issues/11376

**What is the new behavior?**
Even if the layer does not have any geometry, the widget can be created just like in the dashboard. But for such case, the connection with the map is disabled so no geometry filtering applies and the widget can be displayed properly.
image

Also, for the current time, even if the connection is toggled, the current time will be visible.
image image

## Breaking change
 - [x] No

## Other useful information

### DIFF
--- a/web/client/epics/widgets.js
+++ b/web/client/epics/widgets.js
@@ -58,7 +58,7 @@ import { updateDependenciesMapOfMapList, DEFAULT_MAP_SETTINGS } from "../utils/W
 const updateDependencyMap = (active, targetId, { dependenciesMap, mappings}) => {
     const tableDependencies = ["layer", "filter", "quickFilters", "options"];
     const mapDependencies = ["layers", "groups", "viewport", "zoom", "center"];
-    const dimensionDependencies = ["currentTime", "offsetTime"];
+    const dimensionDependencies = ["dimension.currentTime", "dimension.offsetTime"];
     const id = (WIDGETS_REGEX.exec(targetId) || [])[1];
     const cleanDependenciesMap = omitBy(dependenciesMap, i => i.indexOf(id) === -1);
 
@@ -67,7 +67,7 @@ const updateDependencyMap = (active, targetId, { dependenciesMap, mappings}) => 
         if (includes(dimensionDependencies, k)) {
             return {
                 ...ov,
-                [`dimension.${k}`]: `dimension.${mappings[k]}`
+                [k]: `dimension.${mappings[k]}`
             };
         }
         if (!endsWith(targetId, "map") && includes(tableDependencies, k) && !includes(dimensionDependencies, k)) {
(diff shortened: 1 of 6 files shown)


web/client/plugins/WidgetsBuilder.jsx
```diff
@@ -129,13 +128,13 @@ const WidgetsBuilderButton = connect((state) => ({ available: widgetBuilderAvail
...
     const layer = selectedNodes?.[0]?.node;
-    if (available && [statusTypes.LAYER].includes(status) && layer?.search && layer.search !== 'vector' && !layer?.error) {
+    if (available && [statusTypes.LAYER].includes(status) && layer?.search && layer.search !== 'vector') {
         return (
             <ItemComponent
                 {...props}
                 glyph="stats"
                 tooltipId={'toc.createWidget'}
-                onClick={() => onClick()}
+                onClick={() => layer?.error ? onClick({ mapSync: false }) : onClick()}
```
offtherailz:
```suggestion
                onClick={() => layer?.error ? onClick({ mapSync: false }) : onClick()} // allows anyway to create a widget, not connected to map
```

web/client/epics/widgets.js
```diff
@@ -58,18 +58,25 @@ import { updateDependenciesMapOfMapList, DEFAULT_MAP_SETTINGS } from "../utils/W
...
+            };
+        }
+        if (!endsWith(targetId, "map") && includes(tableDependencies, k) && !includes(dimensionDependencies, k)) {
             return {
                 ...ov,
                 [k]: `${targetId}.${mappings[k]}`
             };
         }
-        if (endsWith(targetId, "map")) {
+        if (endsWith(targetId, "map") && !includes(dimensionDependencies, k)) {
```
offtherailz: I think this can be removed, because !includes already returns.

```suggestion
        if (!endsWith(targetId, "map") && includes(tableDependencies, k)) {
            return {
                ...ov,
                [k]: `${targetId}.${mappings[k]}`
            };
        }
        if (endsWith(targetId, "map")) {
```

web/client/epics/widgets.js
```diff
@@ -81,6 +88,7 @@ const updateDependencyMap = (active, targetId, { dependenciesMap, mappings}) =>
                 [k]: `${depToTheWidget}.${mappings[k]}`
             };
         }
+
```
offtherailz:
```suggestion
```

offtherailz (approved):

Issue #11400, opened 2025-08-25
Include quickFilters  when exporting data from TableWidget.


anup39:
## Description

Currently when exported the data from the TableWidget , the filter applied on table headers (quickFilters) are  not included in the downloaded file. 

LayerDownload was enhanced in this issue #11071 , however the quickFilters are not included when exporting the  data

## Other useful information

The preliminary  evaluation was done here https://github.com/geosolutions-it/support/issues/4824#issuecomment-3139534526

PR #11401, opened 2025-08-25
Fix : #11400 Added quickFilters also  when exporting data from TableWiget

anup39:
## Description

When exporting the data from the TableWidget , quickFilters are also included

 - [x] Bugfix
 - [x] Enhancement

## Issue

#11400 

**What is the new behavior?**

When exporting the data now , both the quick filter and filter configured in the widget is included.

Screenshot 2025-08-25 at 14 26 27

## Breaking change
 - [x] No

### DIFF
--- a/web/client/selectors/layerdownload.js
+++ b/web/client/selectors/layerdownload.js
@@ -11,6 +11,7 @@ import { createSelector } from 'reselect';
 import { isFeatureGridOpen } from './featuregrid';
 import { getSelectedLayer } from './layers';
 import { wfsFilter } from './query';
+import { composeFilterObject } from '../components/widgets/enhancers/utils';
 
 import { getTableWidgets } from './widgets';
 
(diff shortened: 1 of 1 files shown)


MV88 (changes requested):
- [ ] functionality wise looks to be working in the dashboard page, but when downloading from the map it is downloading the entire set
image
 and not just the 10 items 

- [ ] Also please add a unit test covering this case for the selector chagned

anup39:
> * [ ]  functionality wise looks to be working in the dashboard page, but when downloading from the map it is downloading the entire set
> 
> image and not just the 10 items
> * [X]  Also please add a unit test covering this case for the selector chagned

@MV88 @tdipisa  Upon checking it seems like in Map Page  quickFilters does not include geometry filtering (only attributes). In map page the data is filtered according to the current extent(geometry) . In this case we are not updating the quick filter with the new geometry every time . This is the reason in map page you are getting full datasets when you export. In the production also it is behaving the same. From my understanding this should be addressed in the new separate issue.

MV88:
> > * [ ]   functionality wise looks to be working in the dashboard page, but when downloading from the map it is downloading the entire set
> > 
> > image and not just the 10 items
> > 
> > * [x]   Also please add a unit test covering this case for the selector chagned
> 
> @MV88 @tdipisa Upon checking it seems like in Map Page quickFilters does not include geometry filtering (only attributes). In map page the data is filtered according to the current extent(geometry) . In this case we are not updating the quick filter with the new geometry every time . This is the reason in map page you are getting full datasets when you export. In the production also it is behaving the same. From my understanding this should be addressed in the new separate issue.

@anup39 
I agree this might be addressed into a separate issue, but let me rephrase a bit. The quickFilters you mention are the input field that are placed on the second row, and these are stored usually in a quickFilters property. Other filters like the spatial one applied to the layer in the map viewer or other attribute filters, are stored into layerFilter (ie. as a global filter on this layer). 
I think that if  a global filter is active and applied to the map it should be used also in other tools like download etc. for consistency. 
So yes please create a new issue specifying these things

@tdipisa FYI

MV88 (approved):
I'm going to merge this then

MV88:

ElenaGallo:
Test passed, @anup39 please backport to 2025.01.xx. Thanks

PR #11421, opened 2025-09-01
Fix As NON-ADMIN user not possible to manage own resource

### LINKED ISSUE #11371: As NON-ADMIN user not possible to manage own resource after operating on the properties panel
## Description
If a NON-ADMIN user operates on the properties panel for a newly created resource (by the same user), for example a map, it can get to a state where the user cannot manage that same resource, for example deleting the resource or changing it since deleting is disabled and when attempting to save the user gets an error message (500 from the server)

## How to reproduce
-  as a NON-ADMIN user access the home page and create a new map and save the new map.
- 

*Expected Result*

*Current Result*

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

stefanocudini:
## Description
fix #11371 
fix lose of user owner on save resource in GeoStore module. The fix tries to keep the user attribute fixed and not lost in case the group permissions are changed, so the resource owner can still modify it.

here is defined the bug replication conditions and other considerations about server side:
https://github.com/geosolutions-it/MapStore2/issues/11371#issuecomment-3242750557

And the result or PR:
https://github.com/user-attachments/assets/bb3e922f-e324-4f09-9c33-b6bfb27661ae

The bug of resource permissions is solved, **but some tests fail that I couldn't fix**:
- computePendingChanges: in `utils/__tests__/GeostoreUtils-js?:349:76`
- getPendingChanges in `ResourcesCatalog/selectors/__tests__/save-test.js?:34:9`

The test payload is probably wrong in new conditions.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

### DIFF
--- a/web/client/plugins/ResourcesCatalog/Save.jsx
+++ b/web/client/plugins/ResourcesCatalog/Save.jsx
@@ -62,7 +62,7 @@ function Save({
             const api = Persistence.getApi();
             api.updateResource(addNameToResource(saveResource))
                 .toPromise()
-                .then((resourceId) => api.getResource(resourceId, { includeAttributes: true, withData: false }).toPromise())
+                .then((resourceId) => api.getResource(resourceId, { includeAttributes: true, withData: false, withPermissions: true }).toPromise())
                 .then(resource => ({ ...resource, category: { name: resourceType } }))
                 .then((resource) => {
                     onSelect(resource);


web/client/utils/GeostoreUtils.js
```diff
@@ -184,6 +185,7 @@ export const computePendingChanges = (initialResource, resource, resourceData) =
         },
         changes: {
             ...pendingChanges,
+            permissions: mergedPermissions,
```
allyoucanmap: The `changes` property includes all the current pending changes and the `pendingChanges` object already includes permissions so we don't need to add this here. The `changes` is used mainly to prevent to leave the page and it has not direct effect on the payload used to save the resource (`saveResource`). Please remove this addition.

web/client/utils/GeostoreUtils.js
```diff
@@ -163,6 +163,7 @@ export const computePendingChanges = (initialResource, resource, resourceData) =
     const unlinkTags = (initialResource?.tags || []).filter(tag => !(resource?.tags || []).find(t => t.id === tag.id)).map(tag => ({ tag, action: 'unlink' }));
     const linkTags = (resource?.tags || []).filter(tag => !(initialResource?.tags || []).find(t => t.id === tag.id)).map(tag => ({ tag, action: 'link' }));
     const mergedTags = [...unlinkTags, ...linkTags];
+    const mergedPermissions = merge(initialResource.permissions, pendingChanges.permissions); // maintain always user
```
allyoucanmap: Merging permissions could cause problem when a user start with a resource with multiple permissions applied and want to remove them, so I don't think it's the correct solution. Anyway there aren't modification in the `saveResource` property for permissions so this change has not real effect on the save workflow.

I tried to replicate the problem locally following the steps and the provided video without success. After investigating the video I noticed the following:

- the root cause of the problem seems the onChange callback of the `Permissions` component where we are trying to recover the userPermissions from the resource prop. Probably at function creation the resource permissions are not ready because of the async request causing an and empty array.
https://github.com/geosolutions-it/MapStore2/blob/5a95d704d3c0c186d1ee8305977ff4ff8f7f9fd1/web/client/plugins/ResourcesCatalog/containers/ResourcePermissions.jsx#L104
- The Permissions component is managing the state internally and only on mount it store the initial entries this is the reason the missing owner is visible in the video only when the user change tab (e.g. from About to Permission again). To note how there aren't changes on the user permissions internally because it is part of the array list
https://github.com/geosolutions-it/MapStore2/blob/5a95d704d3c0c186d1ee8305977ff4ff8f7f9fd1/web/client/plugins/ResourcesCatalog/components/Permissions.jsx#L38
https://github.com/geosolutions-it/MapStore2/blob/5a95d704d3c0c186d1ee8305977ff4ff8f7f9fd1/web/client/plugins/ResourcesCatalog/components/Permissions.jsx#L84-L98

Here my proposal:

- include a new property `data` with the original payload inside the user entry permissions so we can access later inside the onChange callback
https://github.com/geosolutions-it/MapStore2/blob/5a95d704d3c0c186d1ee8305977ff4ff8f7f9fd1/web/client/plugins/ResourcesCatalog/containers/ResourcePermissions.jsx#L51-L66

```js
        return {
            type: 'user',
            id: entry?.user?.id,
            name: entry?.user?.name,
            permissions: 'owner',
            data: entry
        };
    });
```
- extract the user permissions from the entry list instead of the props
https://github.com/geosolutions-it/MapStore2/blob/5a95d704d3c0c186d1ee8305977ff4ff8f7f9fd1/web/client/plugins/ResourcesCatalog/containers/ResourcePermissions.jsx#L104
```js
const userPermissions = (entries || []).filter((entry) => entry.type === 'user').map(entry => entry.data);
```
- finally we could also improve the saveResource to ensure to include the user permission from the initial resource
  - check if the new pending permissions are not undefined
  - if the new pending permissions does not contain the user permission we extract this information from the intialResource and apply to the new pending permissions
https://github.com/geosolutions-it/MapStore2/blob/5a95d704d3c0c186d1ee8305977ff4ff8f7f9fd1/web/client/utils/GeostoreUtils.js#L172

allyoucanmap (changes requested):

tdipisa:
Dear @stefanocudini it seems I can't reproduce the problem either by following steps in issue description [here](https://github.com/geosolutions-it/MapStore2/issues/11371#issue-3283254264) nor the ones in [this](https://github.com/geosolutions-it/MapStore2/issues/11371#issuecomment-3242750557) comment. I can only suppose some strange race condition in managing permissions.
Anyway, the review of @allyoucanmap above seems consistent with the problem investigated by @rowheat02 [here](https://github.com/geosolutions-it/MapStore2/issues/11371#issuecomment-3187255543). It make sense to follow @allyoucanmap indications definitely. I saw you only need to fix for failing checks. Please go ahead and finalize the PR. Thank you.

stefanocudini:
> Dear @stefanocudini it seems I can't reproduce the problem either by following steps in issue description [here](https://github.com/geosolutions-it/MapStore2/issues/11371#issue-3283254264) nor the ones in [this](https://github.com/geosolutions-it/MapStore2/issues/11371#issuecomment-3242750557) comment. I can only suppose some strange race condition in managing permissions. Anyway, the review of @allyoucanmap above seems consistent with the problem investigated by @rowheat02 [here](https://github.com/geosolutions-it/MapStore2/issues/11371#issuecomment-3187255543). It make sense to follow @allyoucanmap indications definitely. I saw you only need to fix for failing checks. Please go ahead and finalize the PR. Thank you.

last commit apply the review suggestions by @allyoucanmap, testing in local this fix the original bug #11371

the CI Fails is caused only by this https://github.com/geosolutions-it/MapStore2/issues/11441 just fixed

allyoucanmap (approved):
@tdipisa reported an additional problem on the save workflow where the permissions were not restored. There is already a commit to fix this here 77f6e8ef260c193a62dfa07cfb9ca373b5d02b7f

- [x] steps to reproduce the problem:
  - login as an user
  - create a map
  -  open the resource details panel and select the permissions panel
  - click on edit properties button
  - edit permissions adding everyone group
  - click again on edit properties button to exit editing
  -  click on save
  - **Current**: No permission available message
  - **Expected**: The permissions are still visible in the panel

stefanocudini:
> @tdipisa reported an additional problem on the save workflow where the permissions were not restored. There is already a commit to fix this here [77f6e8e](https://github.com/geosolutions-it/MapStore2/commit/77f6e8ef260c193a62dfa07cfb9ca373b5d02b7f)

Dear @allyoucanmap I was investigating issue #11436  as noted below to be a related to this issue #11421
Actually, trying this additional attribute `withPermissions` also resolves the issue #11436!

@tdipisa  this commit could solve also #11436 in this same PR

tdipisa:
No worries guys @allyoucanmap @stefanocudini I see we are aligned. @stefanocudini you can simply connect the issue #11436 to this PR then. Thank you.

allyoucanmap:
@ElenaGallo please test this fix on dev and let us know when we can backport, thanks

ElenaGallo:
Test passed, @stefanocudini please backport to 2025.01.xx. Thanks

stefanocudini:
> Test passed, @stefanocudini please backport to 2025.01.xx. Thanks

@ElenaGallo backport merged now

PR #11425, opened 2025-09-02
Update the style of the square buttons and add border radius in all buttons, inputs, selectors

### LINKED ISSUE #10954: Uniform square buttons size to have the same size for all buttons in the whole application

subashtiwari1010:
## Description
This PR updates the style of the square buttons, and adds the border radius in button, inputs and selectors as per the requirement in #10954.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

The buttons were categorized in small, medium and default size. And the components did not have the border radius to them.
Fixes #10954 

**What is the new behavior?**
All the buttons are now categorized under `square-button` only. Other categories of square-button-md and square-button-sm are deprecated. Also, the border radius have been implemented in the button, inputs and select components.

## Breaking change
 - [x] No

## Other useful information

### DIFF
--- a/web/client/plugins/TOC/components/Toolbar.jsx
+++ b/web/client/plugins/TOC/components/Toolbar.jsx
@@ -10,7 +10,7 @@ import React, { useRef, useEffect, useState } from 'react';
 import { Glyphicon, Dropdown } from 'react-bootstrap';
 import TableOfContentItemButton from './TableOfContentItemButton';
 import { NodeTypes, ROOT_GROUP_ID, DEFAULT_GROUP_ID } from '../../../utils/LayersUtils';
-import { StatusTypes } from '../utils/TOCUtils';
+import { markEdgesForToolbar, StatusTypes } from '../utils/TOCUtils';
 
 /**
  * Toolbar component for the TOC
@@ -74,6 +74,7 @@ function Toolbar({
             const width = ref.current.clientWidth;
             const scrollWidth = ref.current.scrollWidth;
             const overflow = width < scrollWidth;
+            markEdgesForToolbar(ref.current);
             if (overflow) {
                 const sumUntilIndex = (arr, index) => arr.filter((val, idx) => idx <= index).reduce((acc, val) => acc + val, 0);
                 const newBreakpoint = [...ref.current.children]
--- a/web/client/plugins/TOC/utils/TOCUtils.js
+++ b/web/client/plugins/TOC/utils/TOCUtils.js
@@ -189,3 +189,35 @@ export const selectedNodesIdsToObject = (selectedNodesIds, layers, tree) => {
         return null;
     }).filter(selectedNode => !!selectedNode);
 };
+
+/**
+ * Determines if a DOM element is visible based on its computed style.
+ *
+ * @param {HTMLElement} btn - The DOM element to check for visibility.
+ * @returns {boolean} True if the element is visible; otherwise, false.
+ */
+const isVisible = (btn) => {
+    const style = window.getComputedStyle(btn);
+    return style.display !== 'none' && style.visibility !== 'hidden';
+};
+
+/**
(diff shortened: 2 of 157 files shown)


web/client/plugins/TOC/utils/TOCUtils.js
```diff
@@ -189,3 +189,35 @@ export const selectedNodesIdsToObject = (selectedNodesIds, layers, tree) => {
...
+    // filter out excluded + invisible
+    const eligible = buttons.filter((btn) => isVisible(btn));
+    // remove old markers
+    buttons.forEach((btn) => btn.classList.remove('is-first', 'is-last'));
+    // mark new first/last
+    if (eligible.length > 0) {
+        eligible[0].classList.add('is-first');
+        eligible[eligible.length - 1].classList.add('is-last');
+    }
+};
```
allyoucanmap: We should not manipulate directly classes in the DOM, in general we should give priority on react or css solutions and as last resource access to the DOM trying to reduce at minimum the manipulation.

Here my suggestion is to only find which are the first and last indices of the visible button and then inside the React component use useState to store and add the classes to the div wrapper

```js
export const getEdgesIndexForToolbar = (el) => {
    if (!el) return [];
    const eligible = [...(el.children || [])].map((child, idx) => [idx, child]).filter((entry) => entry[1].children.length > 0 && entry[1].style.visibility !== 'hidden');
    const [firstIndex] = eligible[0] || [];
    const [lastIndex] = eligible[eligible.length - 1] || [];
    return [firstIndex, lastIndex];
};
```

web/client/plugins/TOC/components/Toolbar.jsx
```diff
@@ -74,6 +74,7 @@ function Toolbar({
             const width = ref.current.clientWidth;
             const scrollWidth = ref.current.scrollWidth;
             const overflow = width < scrollWidth;
+            markEdgesForToolbar(ref.current);
```
allyoucanmap: Here we could add a internal useState:

`setEdgesIndex(getEdgesIndexForToolbar(ref.current));`

and then apply the class to the div wrapper only if the idx match the first and last visible not empty item

allyoucanmap:
@subashtiwari1010 I started to review the PR but not completed it yet. You can start to address the changes requested in this initial review, thanks

web/client/plugins/TOC/utils/TOCUtils.js
subashtiwari1010: Updated the function to use the above implementation instead of using the DOM manipulation.

web/client/plugins/TOC/components/Toolbar.jsx
subashtiwari1010: Used the new function here to set the edges button index and used the index to set the class in the component.

allyoucanmap:

allyoucanmap:
@subashtiwari1010 after a test I noticed following problems:

- [x] button groups in some cases was not rendering correctly
image
image

- [x] Problem in spacing of forms (I don't think is directly connected to this one but it needs to be solved)
image

- [x] Even if square-button-md and square-button-sm are deprecated we don't provide a fallback

I have already reviewed locally and I'll prepare a PR then I'll assign the review to you, thanks

allyoucanmap:
@subashtiwari1010 here the PR to review https://github.com/geosolutions-it/MapStore2/pull/11961

tdipisa:
@subashtiwari1010 
New rounded buttons are very nice. At a very first look I only noticed this

image


This button seems not rounded. Is there some css still to align for this?

As you requested I report also the following which you confirmed is related. Thank you @subashtiwari1010 

image

subashtiwari1010:
Here is some issues in the filter widget:
image

PR #11429, opened 2025-09-04
Introducing broken map context state #11140

### LINKED ISSUE #11140: Introducing broken map context state
## Description

If an user can view a context map without having enough permissions on the context, the map should be marked as broken with a proper message (or tooltip somewhere and right symbology) to properly inform the user that the map can be loaded without the context configuration and more permissions on the context are necessary for a complete loading of the map resource.

- [ ] Build related changes
- [ ] CI related changes

rowheat02:
## Description

 - [x] Enhancement

fixes #11140

#11140
When a user attempts to open a context map without the context permission to the user, they can open the map on the viewer route /viewer/id, which is incorrect.

**What is the new behavior?**

- On the homepage, the warning tooltip will be shown for the context map without context permission (yellow border line). The user can not open this map
    image

- If the user tries to access this type of map, it already shows the context permission message.
   image

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/Share.jsx
+++ b/web/client/plugins/Share.jsx
@@ -143,6 +143,11 @@ const ActionCardShareButton = connect(
         });
         onToggle();
     }
+    // hide button if view url is missing
+    // e.g. when context map have missing permissions on the context resource
+    if (viewerUrl === false) {
+        return null;
+    }
     return (<Component
         iconType="glyphicon"
         glyph="share-alt"
--- a/web/client/themes/default/less/resources-catalog/_resource-card.less
+++ b/web/client/themes/default/less/resources-catalog/_resource-card.less
@@ -16,10 +16,7 @@
     .ms-resource-card {
         &.ms-resource-issue-dependency-missing {
             .border-color-var(@theme-vars[warning]);
-            border-width: 2px;
-            border-style: solid;
-            border-radius: 4px;
-            opacity: 0.85;
+            opacity: 0.75;
         }
     }
 }
--- a/web/client/translations/data.de-DE.json
+++ b/web/client/translations/data.de-DE.json
@@ -4513,7 +4513,7 @@
         "errorDefault": "Netzwerkfehler"
       },
       "resourceIssues": {
-        "dependencyMissing": "Required dependency (e.g: parent context) is not accessible"
+        "dependencyMissing": "Diese Ressource kann nicht geöffnet werden. Fehlende Berechtigungen für eine zugehörige Ressource."
       },
       "deleteError": {
         "error403": "Sie dürfen die Ressource nicht löschen",
--- a/web/client/translations/data.en-US.json
+++ b/web/client/translations/data.en-US.json
@@ -4484,7 +4484,7 @@
         "errorDefault": "Network error"
       },
       "resourceIssues": {
-        "dependencyMissing": "Required dependency (e.g: parent context) is not accessible"
+        "dependencyMissing": "This resource can't be open. Missing permissions on an associated resource."
       },
       "deleteError": {
         "error403": "You are not allowed to delete the resource",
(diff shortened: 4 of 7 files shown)


web/client/plugins/ResourcesCatalog/components/ResourceCard.jsx
```diff
@@ -71,17 +71,27 @@ const ResourceCardWrapper = ({
...
             column
             classNames={[
                 '_relative',
                 '_interactive',
                 ...(active ? ['_active'] : [])
             ]}
+            style={contextMapWithoutContextPermission ? {
+                border: '2px solid #f0ad4e',
+                borderRadius: '4px'
+            } : undefined}
```
allyoucanmap: We should avoid inline hardcoded style and we should think on a different way to represent this not sure yet on the usage of a yellow border for a broken link

web/client/plugins/ResourcesCatalog/components/ResourceCard.jsx
```diff
@@ -71,17 +71,27 @@ const ResourceCardWrapper = ({
...
     query,
     target,
+    contextMapWithoutContextPermission,
     ...props
 }) => {
     const showViewerLink = !!(!readOnly && viewerUrl);
+    const CardWrapper = contextMapWithoutContextPermission ? tooltip(FlexBox) : FlexBox;
+    const tooltipProps = contextMapWithoutContextPermission ? {
+        tooltipId: "resourcesCatalog.warningForContextMapWithoutContextPermssion"
+    } : {};
```
allyoucanmap: The `contextMapWithoutContextPermission` prop and the `warningForContextMapWithoutContextPermssion` message are too specific for the GeoStore backend we should avoid this kind of naming and keep things generic

web/client/plugins/ResourcesCatalog/components/ResourceCard.jsx
```diff
@@ -71,17 +71,27 @@ const ResourceCardWrapper = ({
     layoutCardsStyle,
     query,
     target,
+    contextMapWithoutContextPermission,
     ...props
 }) => {
     const showViewerLink = !!(!readOnly && viewerUrl);
+    const CardWrapper = contextMapWithoutContextPermission ? tooltip(FlexBox) : FlexBox;
```
allyoucanmap: We should not define a component inside the render function of another component. This should be declared outside

web/client/plugins/ResourcesCatalog/components/ResourceCard.jsx
```diff
@@ -402,7 +412,8 @@ const ResourceCard = forwardRef(({
     const {
         icon,
         viewerUrl,
-        thumbnailUrl
+        thumbnailUrl,
+        contextMapWithoutContextPermission
     } = getResourceInfo(resource);
```
allyoucanmap: In addition to the previous comment we should use status object instead of info
https://github.com/geosolutions-it/MapStore2/blob/e6f0797cf03a6268dd6ae4296ce34e53b6ce63ce/web/client/utils/GeostoreUtils.js#L102-L120

If we want we could use the already available items that are icon visible on the grid card but we could also expose new property e.g.:

```js
{
  cardClassNames: ['ms-broken-resource-card'], // based on condition
  cardTooltipId: 'resourcesCatalog.warningForContextMapWithoutContextPermssion' // based on condition
  items: [...]
}
```

if we include something `cardClassNames` and `cardTooltipId` we can avoid to have hardcoded inline style and too specific name for props inside the generic component.

tdipisa:
@rowheat02 there are failing eslint checks. Can you please fix?

rowheat02:
- Now error is handled using getResourceState.
- Error message is generalized: Required dependency (e.g: parent context) is not accessible
- Css is not inline , moved to .less
- cardClassNames and cardTooltipId are taken from resource status and passed to the `CardWrapper`
- The style same yellow border. This style can be changed if you think another style is better for this case.

allyoucanmap:

Issue #11257, opened 2025-06-30
Add support for 3D Tiles imagery layers


allyoucanmap:
## Description

CesiumJS introduces the possibility to drape imagery layers on top of 3D Tiles in latest version (1.130).
We should design a solution to support this functionality, here snippet of code on how the imageryLayers are associated to 3D Tiles:

```js
const tileset = await Cesium.Cesium3DTileset.fromUrl(
  "http://localhost/tilesets/tileset.json"
);
scene.primitives.add(tileset);

tileset.imageryLayers.add(imageryLayer);

```

## Acceptance criteria

- [ ] ~Design UI to associate an imagery layer to 3D Tiles~ removed due to https://github.com/geosolutions-it/MapStore2/issues/11257#issuecomment-3252612469
- [ ] Include and review support of imagery layers inside the cesium layer component

allyoucanmap:
@anup39 please propose a  UI design for this new feature

tdipisa:
@subashtiwari1010 will proceed with the investigation aimed at evaluating this implementation. He will provide a report here with his proposal. As for the design part is concerned @subashtiwari1010, I would like to adopt the usual logic which is expecting to have the TOC driving the rendering logic in term of rendering order of layers.

subashtiwari1010:
@tdipisa
While exploring the implementation for this, it is possible for draping the imagery layers on the 3D tiles. I tried it on the base map and some WMS layer. For the reference, I used the above provided code and the [imagery layers](https://cesium.com/learn/ion-sdk/ref-doc/ImageryLayer.html#:~:text=tileset.imageryLayers.add(imageryLayer)%3B) in Cesium documentation.

Here are the results:
<img width="1877" height="922" alt="Image" src="https://github.com/user-attachments/assets/b1d50117-15f1-4dc1-9502-d212005b03f2" />
<img width="1877" height="922" alt="Image" src="https://github.com/user-attachments/assets/83ad0274-5532-433c-a4a5-e14a7ada07ed" />

Code used:
```bash
function drapeLayer(tileSet, map) {
    if (!tileSet || tileSet.isDestroyed()) return;
    const sceneLayers = map.imageryLayers;
    const tileSetLayers = tileSet.imageryLayers;
    // Remove stale layers
    for (let i = tileSetLayers.length - 1; i >= 0; i--) {
        const layer = tileSetLayers.get(i);
        if (!sceneLayers.contains(layer)) {
[...]

PR #11466, opened 2025-09-12
Add the draping of imagery layers on 3d tiles functionality

subashtiwari1010:
## Description
Introduced the possibility to drape imagery layers on top of 3D Tiles. The drape is only done of those imagery layers that are above the 3d tiles (upon which the drape is done) in the ToC.

 - [ ] Bugfix
 - [x] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

There is no possibility to drape the imagery layers on the 3d tiles.
Fixes https://github.com/geosolutions-it/MapStore2/issues/11257

**What is the new behavior?**
When the imagery layers are above the 3d tiles in the ToC, those layers can be draped on the 3d tiles.

## Breaking change
 - [x] No

## Other useful information

### DIFF
--- a/web/client/translations/data.de-DE.json
+++ b/web/client/translations/data.de-DE.json
@@ -182,8 +182,8 @@
           "eyeDomeLightingStrength": "Lichtstärke",
           "eyeDomeLightingRadius": "Beleuchtungsradius"
         },
-       "enableImageryOverlay": "Überlagerung von Bildebenen aktivieren",
-       "enableImageryOverlayInfo": "Hinweis: Aktiviert Bildebenen wie WMS, TMS oder WMTS, die über der 3D-Tiles-Ebene gerendert und in der festgelegten Reihenfolge angezeigt werden."
+        "enableImageryOverlay": "Überlagerung von Bildebenen aktivieren",
+        "enableImageryOverlayInfo": "Hinweis: Aktiviert Bildebenen wie WMS, TMS oder WMTS, die über der 3D-Tiles-Ebene gerendert und in der festgelegten Reihenfolge angezeigt werden."
       },
       "modelLayer": {
         "modelCenterLat": "Breitengrad des Zentrums (DD)",


allyoucanmap (changes requested):
There are few issue to solve:

- [x] We should move the _msUpdatePrimitivesImageryLayersTimeout function setup inside componentDidMount of the cesium/Map.jsx file so we are sure is created once. We should also set it to undefined on componentWillUnmount
- [ ] The overlay is not working properly when a imagery layer is moved after the 3D Tiles or removed by the TOC, steps:
  - open this [map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/61306)
  - move the test_dataset layer below the 3D Tiles layer (test again by remove it from the TOC)
  - **Expected** the overlay is not applied
  - **Current** the overlay is still applied
- [x] It seems the overlay system is a bit slow to apply so I think it would be better to enable it with an options. My proposal is to include a new property called `enableImageryOverlay` for 3D tiles layer. This property should be undefined by default and accessible as checkbox in the display panel
image

subashtiwari1010:
> * [ ]  The overlay is not working properly when a imagery layer is moved after the 3D Tiles or removed by the TOC, steps:
>   
>   * open this [map](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/61306)
>   * move the test_dataset layer below the 3D Tiles layer (test again by remove it from the TOC)
>   * **Expected** the overlay is not applied
>   * **Current** the overlay is still applied

- The imagery overlay on 3D Tiles is not updating correctly when an imagery layer is moved below the 3D Tiles layer in the TOC, removed, or enable imagery overlay is disabled from the 3d tiles settings.
- Adding the drape overlay over the 3D Tiles works as expected.
- The function to remove draped imagery layers works fine; the layers are removed from the `imageryLayers` field of the 3D Tiles primitive.
- However, the draped imagery remains visible on the 3D Tiles in the map. So, I think the reference to the overlay is not being visually cleared from the scene.

web/client/translations/data.en-US.json
```diff
@@ -181,7 +181,8 @@
           "eyeDomeLighting": "Enable lighting",
           "eyeDomeLightingStrength": "Lighting strength",
           "eyeDomeLightingRadius": "Lighting radius"
-        }
+        },
+        "enableImageryOverlay": "Enable imagery overlay"
```
allyoucanmap: I think we should explain a bit more, e.g.
```suggestion
        "enableImageryOverlay": "Enable imagery layers overlay"
```
I think we should add an inline question mark icon info with additional information explaining that all the imagery layers such as WMS, TMS, WMTS, ... positioned above the 3D Tiles will be rendered in order on top of it.

Please use the same component already used for other checkbox in settings, e.g. this is visible for WMS

image

allyoucanmap (changes requested):
@subashtiwari1010 I pushed a code refactor because the way we implemented it before was not making it possible to remove imagery layers. The refactor moved all the logic to the 3D Tiles again similar to your initial approach but added logic in the generic Layer component to trigger an update count every time a new imagery layer is added removed or sorted in the layer tree. Additionally few fixes have been introduced inside the imagery visibility and opacity management to use as much as possible the native CesiumJS api.

Note: with this approach we are forced to recreate the 3D Tiles when the layer tree is reorder to ensure all the imagery layer are correctly placed

Please complete the PR with following:
- [x] review the overall code and test it locally to ensure everything is working as expected
- [x] review all the messages are requested inline
- [x] fix failing test

allyoucanmap:
@ElenaGallo please test this fix on dev and let use know if we can backport on 2025.02.xx. thanks

ElenaGallo:
Test passed, @subashtiwari1010 please backport to 2025.02.xx. Thanks

Issue #11437, opened 2025-09-09
Isochrone plugin implementation


dsuren1:
### Introduction
The goal of this tool is to enable users to **compute and visualize isochrones and isodistances** using both proprietary and open-source APIs. The solution is intended to provide a user-friendly interface, configuration options, and compatibility with MapStore’s existing geospatial tools.

### Technical Solution
The plugin will be implemented as a **MapStore Core plugin, and a list of child core plugins and extensions implementations for the different APIs**.

- **Generic host**: This component provides the generic functionality and will host the client component that implements the specific APis. 
- **[GrapHopper](https://www.graphhopper.com/)**: This component implements the **[GraphHopper Isochrones API](https://docs.graphhopper.com/openapi/isochrones)**. 

The plugin implements shared functionality and abstractions that specific components, like the one for GrapHopper, implement and extend.  In the future additional components (services APIs) can be added to the plugin.

Specific configurations for child components will be available through the dedicated configuration section or the plugin configuration inside map viewers.

Where requested / available, a configuration key will be provided to add an API Key for service authentication.

### Isochrone/Isodistance Core plugin
This plugin implementation serves as the main container and provides core shared functionalities, including:

- An input field to retrieve a point location, either by searching or selecting a point on the map
- An input field for selecting time or distance values
- Business logic to support the integration of isochrone / isodistance provider components
- Tools to initiate the isochrone / isodistance  request workflow
- Management of generated runs, with the following available actions:
- Edit: Modify run parameters and generate a new result upon saving
- Delete: Remove the selected run from the list of saved runs
- Save as Layer: Convert the current run into a MapStore vector layer embedded in the map configuration
- Export: Export the selected run as a GeoJSON file
- Visibility: Toggle the visibility of the generated isochrone layer on the map
- Opacity: Adjust the opacity of the generated isochrone layer on the map

> [!NOTE]
>- Runs are temporarily saved in the panel. After refreshing or changing the page runs will be reset. Only a run converted to layer will be persisted >inside the map
>- Some theming aspects proposed for the panel in the mockup could be different in the final delivery to follow the overall MapStore theme

#### GraphHopper Provider
This component is an isochrone / isodistance provider that implements the GraphHopper Isochrone API, which is an open-source implementation offering free and commercial cloud services, as well as the option to set up an on-premise instance. 
The provider functions as a child component of the main Isochrone plugin and includes the following elements:
[...]

PR #11478, opened 2025-09-15
#11437: Isochrone plugin implementation

dsuren1:
## Description
This PR implements the Isochrone and Isodistance plugin with Graphhopper as the default provider

 - [x] Feature

## Issue

- #11437

**What is the new behavior?**
image

## Breaking change
 - [x] No


web/client/actions/additionallayers.js
```diff
@@ -105,3 +106,16 @@ export const addAdditionalLayers = (layers = []) => ({
...
+ * @memberof actions.additionallayers
+ * @param {string} id identifier
+ * @param {object} options options to merge
+ * @return {object} of type `MERGE_OPTIONS_BY_ID` with id and options
+ */
+export const mergeOptionsById = (id, options) => ({
+    type: MERGE_OPTIONS_BY_ID,
+    id,
+    options
+});
```
allyoucanmap, line 121: Why do we need a new action when we have `updateAdditionalLayer` ? There is the `override` flag

web/client/plugins/Isochrone/utils/IsochroneUtils.js
```diff
@@ -0,0 +1,131 @@
...
+        id: uuid(),
+        name: ISOCHRONE_ROUTE_LAYER,
+        title: CONTROL_NAME,
+        visibility: true,
+        features: computeIsochroneBands(features),
+        style: {
+            format: 'geostyler',
+            body: {
+                rules: features.map((_, index) => ({
+                    filter: [ '==', 'id', `isochrone-polygon-${index}` ],
```
allyoucanmap: It would be better to use with or so the visual style editor works correctly
```suggestion
                    filter: ['||', [ '==', 'id', `isochrone-polygon-${index}` ] ],
```

web/client/plugins/Isochrone/epics/isochrone.js
```diff
@@ -0,0 +1,290 @@
...
+                type: 'Feature',
+                geometry: { type: 'Point', coordinates: [latlng.lng, latlng.lat]},
+                properties: { id: "point" }
+            }],
+            style: {
+                format: "geostyler",
+                body: {
+                    rules: [
+                        {
+                            filter: [ '==', 'id', 'point' ],
```
allyoucanmap: Same as before for compatibility with the visual style editor
```suggestion
                            filter: ['||', [ '==', 'id', 'point' ]],
```

web/client/plugins/Isochrone/epics/isochrone.js
```diff
@@ -0,0 +1,290 @@
...
+                                    image: {
+                                        name: "msMarkerIcon",
+                                        args: [
+                                            {
+                                                glyph: "chevron-down",
+                                                color: "blue",
+                                                shape: "circle"
+                                            }
+                                        ]
+                                    },
```
allyoucanmap: here we could use similar marker we had in itinerary 
image

web/client/plugins/Isochrone/Isochrone.jsx
```diff
@@ -0,0 +1,221 @@
...
+ *     }
+ * }
+ * ```
+ *
+ * Configure a custom provider (PLUGIN) as follows: (Same pattern applies when creating an extension)
+ * 1. Create a plugin component for the custom provider (Take reference from GraphHopperProvider)
+ * 2. Register the provider api as follows from the custom provider with custom functions.
+ * ```js
+ * registerApi(PROVIDER_NAME, {
+ *     getDirections, // API call function to fetch route details
```
allyoucanmap: In don't think the api promise should be call `getDirections` I'm also seeing reference to route.
Please review the documentation and rename the `getDirections` to `getIsochrones`, I'm open to other suggestions

web/client/plugins/Isochrone/epics/isochrone.js
```diff
@@ -0,0 +1,290 @@
...
+ * @param {external:Observable} action$ manages `TOGGLE_CONTROL`
+ * @return {external:Observable}
+ */
+export const onOpenIsochroneEpic = (action$, {getState}) =>
+    action$.ofType(TOGGLE_CONTROL)
+        .filter(({control}) => control === CONTROL_NAME && enabledSelector(getState()))
+        .switchMap(() =>
+            Observable.of(
+                purgeMapInfoResults(),
+                changeMapInfoState(false)
```
allyoucanmap, line 180: It seems the map info state is never restored when the plugin is closed. Steps:

- open a map with isochrone plugin
- select a point on the map
- run a request
- close the panel

Expected:

- the identify is working

Current:

- the identify is not working

This should be also verified when a user leave the page with the panel open

Probably the identify should be restored after the click on the map is performed and the marker placed?

web/client/plugins/Isochrone/components/RouteDetail.jsx
```diff
@@ -0,0 +1,184 @@
...
+    onDeleteIsochroneData = () => {},
+    onSetCurrentRunParameters = () => {},
+    onUpdateLocation = () => {}
+}) => {
+
+    const getMarkerLayer = (id) => {
+        const markerLayerId = getMarkerLayerIdentifier(id);
+        const {features, style} = get(isochroneLayers.find(l => l.id === markerLayerId), 'options', {});
+        const [markerFeature] = features ?? [];
+        const [markerRule] = get(style, 'body.rules', []);
```
allyoucanmap: I would prefer to use another rule for the center created layer to distinguish current pointer with generated ones. We should also use `center` instead of point as identifier

```
{
      "name": "center",
      "ruleId": "c6a13230-93d7-11f0-a44f-6dc088d58fba",
      "symbolizers": [
        {
          "kind": "Mark",
          "wellKnownName": "Circle",
          "color": "#dddddd",
          "fillOpacity": 0,
          "strokeColor": "#000000",
          "strokeOpacity": 1,
          "strokeWidth": 2,
          "radius": 4,
          "rotate": 0,
          "msBringToFront": false,
          "msHeightReference": "none",
          "symbolizerId": "c6a13233-93d7-11f0-a44f-6dc088d58fba"
        }
      ],
      "filter": [
        "||",
        [
          "==",
          "id",
          "center"
        ]
      ]
    },
```

image

web/client/plugins/Isochrone/components/GraphHopperProvider.jsx
```diff
@@ -0,0 +1,254 @@
...
+                <ColorRamp
+                    items={getColors(providerBody.buckets)}
+                    samples={providerBody.buckets}
+                    rampFunction = {({ colors }) => colors}
+                    value={{ name: get(providerBody, 'ramp.name', DEFAULT_RAMP) }}
+                    onChange={(ramp) => {
+                        handleProviderBodyChange("ramp", ramp);
+                    }}
+                />
+            </FlexBox.Fill>
```
allyoucanmap, line 265: When we have a single interval it would be better to switch to a color picker and show the color ramp only when there are more than one interval

allyoucanmap (changes requested):
- [ ] I noticed that the panel clear the content when closed, I think we should add a confirm dialog to prevent closing if there are generated runs

web/client/actions/additionallayers.js
dsuren1: The 'override' flag applies overrides on active layers (layers state) using options retrieved from additional layers and overlay layer is skipped when type is override. However, I only want to update the overlay layer options. We already have a similar approach that merges options by owner. I find this action makes it more convenient to update the options without affecting other properties of the additional layers

web/client/plugins/Isochrone/epics/isochrone.js
dsuren1: Identify will be restored when Isochrone panel is closed

web/client/plugins/Isochrone/utils/IsochroneUtils.js
```diff
@@ -0,0 +1,161 @@
...
+ * @param {string} rampColors - The ramp colors to generate the isochrone layer from
+ * @returns {object} The isochrone layer
+ */
+export const getIsochroneLayer = (data = [], config = {}, rampColors = BUCKET_COLORS) => {
+    const features = (data ?? []).map((feature, index)=> ({
+        type: 'Feature',
+        geometry: feature.geometry,
+        properties: {
+            id: `isochrone-polygon-${index}`
+        }
```
allyoucanmap: properties should include value of `time` or `distance`, eg:

```js
{
   id: `isochrone-polygon-${index}`,
   time: 2,
   timeUom: 'min',
   label: 'Time 2 km'
}
```

```js
{
   id: `isochrone-polygon-${index}`,
   distance: 2,
   distanceUom: 'km',
   label: 'Distance 2 km'
}
```

If there are multiple buckets the time of each bucket should be computed

allyoucanmap (changes requested):
Still some issues to solve:

- [x] After searching for a location there isn't automatic pan to position. This makes difficult to find the marker when it is out of the current map view

https://github.com/user-attachments/assets/0899d2cb-2fce-48fa-a9ea-bf29c68876e3

- [x] There is not update of the marker while changing the coordinates manually

https://github.com/user-attachments/assets/d36c7d88-5b43-4ac9-859e-6a67076c68e8

- [ ] After the run the marker is removed even if the input location is still filled. The marker should be visible if the input field is filled and valid

https://github.com/user-attachments/assets/c2343e6c-e366-4331-ac14-39efd4eb95be

- [ ] When using Itinerary and Isochrone together there is a `Maximum call stack size exceeded`. Please investigate this and before proceed let's discuss on how to fix this. Step:
   - open isochrone
   - open itinerary without closing isochrone

https://github.com/user-attachments/assets/bb169815-2899-4cad-b286-a36dde509fef

- [ ] The created layer should report:
  - the title similar as we have in the list. I'm seeing now the title in the isochrone list is using message with params, maybe we should translate just distance and time with getMessageById so we can compose the title as a string in js and use it for the title of layer.
  - in the legend should be used the actual distance or time value

image

image

- [x] Rules in visual stile editor cannot be removed. I already pushed some changes for this also replacement for color picker using the same one of the style editor, could you please double check them?

allyoucanmap:
@ElenaGallo please test this new plugin on dev, thanks

allyoucanmap:
@dsuren1 

- [x] after testing the plugin in geonode environment I noticed the Direction option is not working. The request is always sending reverse_flow property at false.
image

dsuren1:
@allyoucanmap @ElenaGallo 
FYI, a fix has been created [here](https://github.com/geosolutions-it/MapStore2/pull/11507)

ElenaGallo:
@dsuren1 se my comments below:

1_ Add a place in the search bar, then select a point on the map using the coordinate editor. When switching back to “Search by location”, the previous location remains visible on the map, but the search is performed using the coordinate point.


https://github.com/user-attachments/assets/5493fba1-24f9-4d09-8d9a-1b5df8d86b7f



2_ Add a place in the search bar and then click Reset — the place remains selected instead of being cleared.


https://github.com/user-attachments/assets/4fa44056-bb62-4b3e-8645-4079c945041b


3_ The Direction options are not set correctly when Use run parameters is clicked.


https://github.com/user-attachments/assets/802c63c7-515e-4d7f-aed1-bea5ea54edfc


4_ If the Catalog panel is opened while the Isochrone panel is active, the Isochrone panel should automatically close and the Catalog panel should open.


https://github.com/user-attachments/assets/c3d7e3ec-2d2c-4b73-9c89-f9911f973154

dsuren1:
@ElenaGallo Thanks, I will check and provide an enhancement PR

PR #11498, opened 2025-09-19
Fix #11465 Problem with feedback mask and user permissions

### LINKED ISSUE #11465: Problem with feedback mask and user permissions
## Description

The feedback mask at the moment is not covering the Brand Navbar as it should be instead. This highlighted an issue with the navbar itself while switching between users with different permissions.

## How to reproduce

- login as admin
- open a resource that only admin can see: https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/60717
- logout
- login as normal user

*Expected Result*

The feedback should cover the Brand Navbar and the application state in term of permissions should be consistent when the user change.

*Current Result*

The feedback doesn't cover the Brand Navbar as it should be. From this situation we can see some defects where we shoudl probably intervene to consolidate the application state on the frontend. A few relevant points reported below:

1. As per steps above, when you login as normal user just after the logout, the Brand Navbar is visible, yes, but also _Save_ and _Save As_ and _Delete_ buttons are available. and it should not be, in theory. Clicking then on _Open properties_ nothing happens and mostly the same for the _About_ button.

   Image

2. If you try to access the same resource directly as user (e.g. without authenticating as admin before), you have only the _Save as_. The same for the rest (when clicking on _Open properties_ or _About_ buttons)

   Image

3. Giving point 2, you can now click on _Save as_ and create a new resource but, doing this, you create an empty map.

   Image

   with an infinite loading if you try to open it then

   Image

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

allyoucanmap:
## Description

This PR includes two fixes:

- ensure the feedback mask uses portal to be render on top of everything
- remove canCopy permissions to map with id but without info to hide SaveAs. This combination refers to inaccessible map

 - [x] Bugfix

## Issue

#11465

**What is the new behavior?**

The feedback mask covers the page to prevent unwanted actions

## Breaking change

 - [x] No

### DIFF
--- a/web/client/components/misc/enhancers/__tests__/withMask-test.jsx
+++ b/web/client/components/misc/enhancers/__tests__/withMask-test.jsx
@@ -66,4 +66,10 @@ describe('withMask enhancer', () => {
         expect(document.querySelector('.ms2-mask')).toNotExist();
         expect(document.querySelector('.custom-class')).toExist();
     });
+    it('withMask with enablePortal', () => {
+        const Sink = withMask(undefined, undefined, { enablePortal: true })(() => <div id="test">test</div>);
+        ReactDOM.render(<div id="content"><Sink /></div>, document.getElementById("container"));
+        expect(document.querySelector('#content').children.length).toBe(0);
+        expect(document.body.children.length).toBe(2);
+    });
 });
--- a/web/client/components/misc/enhancers/withMask.js
+++ b/web/client/components/misc/enhancers/withMask.js
@@ -8,18 +8,24 @@
 import React from 'react';
 
 import { branch, nest } from 'recompose';
+import Portal from '../Portal';
 /**
  * Add a grayed out mask to a component.
  * @param {function} showMask gets props as argument and returns true if the enhancer should be applied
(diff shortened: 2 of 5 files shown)


offtherailz (changes requested):
From my tests the page do not always work: 

- [x] Trying to open in local do not prompt the login, showing empty page (on dev it works better)

https://github.com/user-attachments/assets/d8b277c5-aa0e-4d17-8974-32b3f2d86d2e

- [ ] Same issue on logout

https://github.com/user-attachments/assets/23e18c6f-0eaa-4cbc-8b12-c401db8cdfea

allyoucanmap:
@tdipisa this fix is ready for testing then we will backport to 2025.01.xx

allyoucanmap:
@ElenaGallo please test this fix on dev and let me know when I can backport to 2025.01.xx, thanks

allyoucanmap:
test passed, @allyoucanmap backport

allyoucanmap:
test passed

PR #11484, opened 2025-09-16
Performance Optimization: Refactor Pending Changes Logic and Save Resource Selector #11476

rowheat02:
## Description

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Optimization
 - [x] Refactoring 

fixes #11476
fixes #11435
fixes #11458

**What is the current behavior?**

 #11476
 #11435
 #11458

**What is the new behavior?**

- More optimized than before, while doing UI/Map interaction.
- No pending change Prompt shown when the map is deleted when there are some pending changes on Map page.

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

## Other useful information 

https://github.com/user-attachments/assets/a067e462-f3ff-4932-a7e1-e74dc3c14e03

- `compareMapChanges` is out of selector and is called using debounce logic. Managed in `useComputedPendingChanges`
  - By default, 
     - it waits for 500 ms before calling
     - if for 2000 ms, not called due to debouncing, it will be forced to call
 
- Another expensive function, ' saveMapConfiguration`, is only called on demand. Will not be called infinitely while doing UI/Map Interaction
- for Optimization, createSelector has been used on the main selector
- Also, now, while saving a new or existing map, the mapSave payload is calculated, unlike before, taking from the selector.

Issue #11435, opened 2025-09-08
Changes dialog show after resource is deleted


stefanocudini:
## Description
Changes dialog show immediately after a resource is deleted

## How to reproduce
- Create a new map resource
- Save the new resource with a name (Save As button)
- Click the Delete button in the top left of the navbar
- A dialog will open asking "Are you sure you want to delete this resource?" to confirm deletion.
- Immediately afterward, a second dialog will automatically appear (which is the bug):
"Are you sure to leave without applying your changes?"

*Expected Result*
Redirect to homepage

*Current Result*
Show changes dialog

https://github.com/user-attachments/assets/2e376c50-b8e3-4ba2-8318-4949f110c7e3

ElenaGallo:
Test passed. @rowheat02, please backport to version 2025.01.xx. Thanks.

PR #11503, opened 2025-09-22
[Backport 2025.01.xx] - Fixes #11476 , #11435 : Performance Optimization: Refactor Pending Changes Logic and Save Resource Selector #11476 (#11484)

### LINKED ISSUE #11476: Performance Optimization: Refactor Pending Changes Logic and Save Resource Selector
## Description

It has been observed that performance degrades when multiple there are huge number of layers, particularly affecting map/UI interaction. This slowness is caused by the logic for computing pending changes being placed inside a selector, which is called multiple times under heavy load.

- [x] Optimization

### Proposed Enhancements

1. **Separate Logic**
   - Move the logic for:
     - Computing **pending changes**.
     - Preparing the **resource to save**.
   - The resource preparation should be handled via a **selector** (without debounce) to avoid inconsistencies.

2. **Component-Based Optimization**
   - Shift the pending changes computation into a **connected component**.
   - Apply **debounce** to reduce frequent executions of heavy computations.
   - Optimize more if possible

3. **Store in State**
   - Save the computed **pending changes** in the **application state** for access by other components.

Ultimately, when performing map interaction on a map with a large number of layers (or has layer with heavy geometry), it should not block the JS main thread, causing the entire application to hang.

rowheat02:
[Backport 2025.01.xx] - Fixes #11476 , #11435 : Performance Optimization: Refactor Pending Changes Logic and Save Resource Selector #11476 (#11484)

### DIFF
--- a/web/client/plugins/ResourcesCatalog/DeleteResource.jsx
+++ b/web/client/plugins/ResourcesCatalog/DeleteResource.jsx
@@ -13,7 +13,8 @@ import { connect } from 'react-redux';
 import { createStructuredSelector } from 'reselect';
 import Persistence from '../../api/persistence';
 import { searchResources } from './actions/resources';
-import { getPendingChanges } from './selectors/save';
+import {setPendingChanges as setPendingChangesAction} from './actions/save';
+import { getResourceInfoByType } from './selectors/save';
 import { push } from 'connected-react-router';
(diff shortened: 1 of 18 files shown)


allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test this enhancement on QA, thanks

Issue #11614, opened 2025-10-21
MapViews - Radio Button controls enabling on nested layer groups won't exclude each other


ale-cristofori:
Take this map as example (data is public)
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/61376

## Description
In Map View plugins where there are nested groups (a group that contains more than another map group) and one of those nested layer groups have radio buttons as visualization command, radio buttons, (that should be mutually exclusive) are both enabled 

## How to reproduce

Using the map above as example

- Open the map Views plugin and act on the TEST view editing the map view (pencil button).  
- Open the 'Radio button groups' parent group.
- Navigate to the 'Checked radio button group' subgroup
- Enable one of the disabled radio buttons 

*Expected Result*
If one radio button of a layer belonging to the group is enabled the other layer radio button of the same group should automatically disable itself

*Current Result*
If one radio button of a layer belonging to the group is enabled the other layer radio button of the same group will not automatically disable itself, both radio buttons on the relative layers of the layer groups are enabled. see below

<img width="570" height="267" alt="Image" src="https://github.com/user-attachments/assets/6d376cb6-d470-431e-843f-aa81a43fcf71" />

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information
No evident error on the DevOps browser console

Issue #11613, opened 2025-10-21
MapViews - Nested Layer groups visualization won't respond consistently to TOC controls


ale-cristofori:
Take this map as example (data is public)
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/61376

## Description
In Map View plugins where there are nested groups (a group that contains more than another map group) operating visualization on the parent group of the nested groups will have no effect on the map

## How to reproduce
Using the map above as example

-  Open the map Views plugin and act on the TEST view editing the map view (pencil button).
-  Open the 'Checkbox groups' parent group.
-  Operate on the 'Checkbox groups'  visualization (check/uncheck the checkbox)
- Notice how the layers on the map won't respond to the general parent group visualization setting, the layers on the map will still be visible even though the paren layer visualization checkbox is unchecked, responding to the sub-group setting but not to the parent group setting

*Expected Result*
All the layers contained in the parent group 'Checkbox groups' should respond to the checked command on the parent groups checkbox and be not visible if the checkbox is unchecked 

*Current Result*
All the layers contained in the parent group 'Checkbox groups' do not respond to the checked command on the parent groups checkbox and remain visible despite the parent group is unchecked 

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information
No errors in the dev tool console

PR #11656, opened 2025-11-06
Fix issues on mapview's TOC

rowheat02:
## Description

This PR fixes issues with the TOC of Mapviews

 - [x] Bugfix

1. #11614
2. #11613

**What is the new behavior?**

1. Inside mutually exclusive groups' layers, only one layer will be selected.
   image

2. The visibility setting of a grandparent group (without direct layer children) now properly controls the visibility of all its descendant layers on the map.

https://github.com/user-attachments/assets/ceda6d77-a16e-4f09-8481-4efa4f0bc420

## Breaking change
 - [x] No

## Other useful information
1. The first issue was due to a race condition. So, a batched callback is used to update the state.
2.  In the getMapViews selector, the group filter was only accounting for groups that were direct parents of a layer. As a result, any group containing only subgroups (without direct layer children) was not being recognized as a valid group. So this PR fixes to consider all of the groups.

### DIFF
--- a/web/client/selectors/mapviews.js
+++ b/web/client/selectors/mapviews.js
@@ -6,8 +6,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { DEFAULT_GROUP_ID } from "../utils/LayersUtils";
-import { layersSelector } from "./layers";
+import {  flattenArrayOfObjects } from "../utils/LayersUtils";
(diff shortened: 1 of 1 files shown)


allyoucanmap:
@ElenaGallo please test these fixes on dev and let use know when we can backport, thanks

ElenaGallo:
Test passed, @rowheat02 please backport to 2025.02.xx. Thanks

PR #11658, opened 2025-11-07
Update maximum PostgreSQL version to 18

offtherailz:
## Description

According with https://groups.google.com/g/mapstore-developers/c/9OoIzbKQiQo we could increase this version to 18


offtherailz:
@tdipisa I increased only the max version.

Issue #11660, opened 2025-11-07
Map Templates list is not retained when replacing the current map


tdipisa:
## Description

As per issue title the list of available map templates disappear when replacing the current map. This is a regression.

## How to reproduce

- Open [this](https://dev-mapstore.geosolutionsgroup.com/mapstore/#/context/ISPRAContext) context  
- Open the Map Templates tool 
- Choose one and replace the current map

*Expected Result*

The map is replaced and the list of available map templates is retained as expected

*Current Result*

The list disappear and it is necessary to close the tool and reopen it again

https://github.com/user-attachments/assets/29955688-0a9f-4f02-90e1-54f4ec0d663a

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

## Other useful information

Also QA affected, but this is anyway a regression.

rowheat02:
-  [MAP_CONFIG_LOADED](https://github.com/geosolutions-it/MapStore2/blob/05f21649e556c6026ccf7e3d3bc50827e630d438/web/client/reducers/maptemplates.js#L20) overwrite templates when a user session restores them, because those configs always come with the template list that must be reapplied.
- And the template replace flow emits a config with mapTemplates missing or empty, which currently wipes the list and leaves the panel blank.
- Fix: We will skip the overwrite whenever mapTemplates is undefined or empty, so the replace flow stops clearing the list while the user session flow keeps working(because in the user session flow, it only makes sense to replace the  template when it is not empty)

PR #11661, opened 2025-11-07
Map Templates list is not retained when replacing the current map #11660

rowheat02:
## Description

 - [x] Bugfix

#11660 

#11660 

**What is the new behavior?**

Now map templates will not wipe out after replacing

## Breaking change
 - [x] No

## Other useful information
Skiped the overwrite whenever mapTemplates is undefined or empty, so the map-template replace flow stops clearing the list while the user session flow keeps working(because in the user session flow, it only makes sense to replace the template when it is not empty)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ MapStore's architecture is designed for modularity and extensibility, allowing d
 
 For more information check the <a href="https://docs.mapstore.geosolutionsgroup.com/en/latest/" target="_blank">MapStore documentation!</a>
 
+Also check out the MapStore project entry page available online at [mapstore.io](https://mapstore.io/)
+
 ## Documentation
 
 You can find more documentation about how to build, install or develop with MapStore on the <a href="https://docs.mapstore.geosolutionsgroup.com/en/latest/" target="_blank">documentation site</a>.


web/client/reducers/maptemplates.js
```diff
@@ -18,7 +18,13 @@ export default (state = {}, action) => {
...
     }
     case MAP_CONFIG_LOADED: {
-        return set('templates', action.config?.mapTemplates, state) ?? [];
+        const mapTemplates = action.config?.mapTemplates;
+        // when the map config originated from a template replace flow: the template list may
+        // come missing or empty, so preserve the existing list rather than wiping it
+        if (mapTemplates === undefined || (Array.isArray(mapTemplates) && mapTemplates.length === 0)) {
+            return state;
+        }
+        return set('templates', mapTemplates, state);
```
offtherailz, line 36: if this happens on every MAP_CONFIG_LOADED, what happesn if we open a new map immediately after this?
I'm afraid the map will preserve last mapTemplates anyway. Can you check?

rowheat02: @offtherailz 
After some digging around on both map templates and user sessions:
I have added a clarification comment. As mapTemplates is a dynamic plugin (which initializes during run time, the initial MAP_CONFIG_LOADED is not triggered). It only handles three cases:
1. While removing the user session.
2. After replacing map templates
3. After adding templates to the map

These three cases are explained in the comment on the reducer function.

So the previous fix is fine with the newly added comment.

offtherailz (approved):

rowheat02:
Hi @ElenaGallo, can you test it on the Dev and let us know if it can be backported?

ElenaGallo:
Test passed, @rowheat02 please backport to 2025.02.xx. Thanks

PR #11733, opened 2025-11-26
Fix #11721 Add support for line traces classification

### LINKED ISSUE #11721: Add support for line traces classification
## Description

We should introduce the possibility to classify line traces in a similar way as we are doing with bar chart. The main use case we are focusing are time series. As an example we could have a dataset including at least following attributes:

- an identifier or category e.g. `id`
- a date/time attribute e.g. `year`
- a value to represent on the y axis e.g. `population`

In this type of dataset structure we would like to use the `year` on the x axis, `population` on the y axis and use the classification style to group by the `id` to get multiple traces for each identifier.

Requirements:

- [x] A new style classification type for the line trace

- [x] Minor changes to existing features

## Other useful information

Datasets for testing:

- [NYPD Complaint Data Historic](https://data.cityofnewyork.us/Public-Safety/NYPD-Complaint-Data-Historic/qgea-i56i/data_preview) data see main chart in https://github.com/geosolutions-it/events-tracker

subashtiwari1010:
## Description

This PR adds the classification mode for the line chart so that the line traces can be categorized for the visualization.

 - [x] Feature

## Issue

Currently, the line trace only supports line, line+marker, and scatter type charts. 
#11721 

**What is the new behavior?**

The line traces type will have classification style in dropdown and after it is selected the line traces will be categorized according to the classification attribute field selected.
image image
image image

## Breaking change
 - [x] No

## Other useful information

### DIFF
--- a/web/client/components/widgets/builder/wizard/chart/__tests__/ChartStyleEditor-test.jsx
+++ b/web/client/components/widgets/builder/wizard/chart/__tests__/ChartStyleEditor-test.jsx
@@ -32,6 +32,7 @@ describe('ChartStyleEditor', () => {
             data={{
                 type: 'line',
                 style: {
+                    msMode: 'simple',
                     mode: 'lines'
                 }
             }}
@@ -39,6 +40,7 @@ describe('ChartStyleEditor', () => {
         const controlLabelsNodes = document.querySelectorAll('.control-label');
         expect([...controlLabelsNodes].map(node => node.innerText)).toEqual([
             'widgets.advanced.mode',
+            'widgets.advanced.type',
             'widgets.advanced.lineColor',
             'widgets.advanced.lineWidth'
         ]);
@@ -48,6 +50,7 @@ describe('ChartStyleEditor', () => {
             data={{
                 type: 'line',
                 style: {
+                    msMode: 'simple',
                     mode: 'markers'
                 }
             }}
@@ -55,6 +58,7 @@ describe('ChartStyleEditor', () => {
         const controlLabelsNodes = document.querySelectorAll('.control-label');
         expect([...controlLabelsNodes].map(node => node.innerText)).toEqual([
             'widgets.advanced.mode',
+            'widgets.advanced.type',
             'widgets.advanced.markerColor',
             'widgets.advanced.markerSize'
         ]);
@@ -64,6 +68,7 @@ describe('ChartStyleEditor', () => {
             data={{
                 type: 'line',
                 style: {
+                    msMode: 'simple',
                     mode: 'lines+markers'
                 }
             }}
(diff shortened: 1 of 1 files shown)


web/client/components/charts/WidgetChart.jsx
```diff
@@ -18,7 +18,8 @@ import {
     parseNumber,
     parsePieNoAggregationFunctionData,
     enableBarChartStack,
-    FONT
+    FONT,
+    DEFAULT_CLASSIFICATION
```
allyoucanmap: why do we need DEFAULT_CLASSIFICATION? Are we doing a similar thing in bar chart?

web/client/components/widgets/builder/wizard/chart/ChartStyleEditor.jsx
```diff
@@ -29,13 +37,22 @@ const chartStyleEditors = {
                             options={[
                                 { value: 'lines+markers', label: 'Line with markers' },
                                 { value: 'lines', label: 'Line' },
-                                { value: 'markers', label: 'Scatter' }
+                                { value: 'markers', label: 'Scatter' },
+                                { value: 'classification', label: <Message msgId={'styleeditor.classificationStyle'} /> }
```
allyoucanmap, line 61: From UI I think we should split the `mode` from `msMode` property and have two select. This means we should have:

- `msMode` property select with two options (label `Mode`):
    - Simple style
    - Classification style
-  `mode` property select with three options (label `Type`):
    - Line with markers
    - Line
    - Scatter

Note: when we switch from Simple to Classification style we may need to manage the `Line with markers` options because it needs two different colors one for line and one for markers. I think we could hide Line with markers when classification style is selected.

In this way we will be able to use the scatter chart in classification mode

allyoucanmap (changes requested):
There is an error by following these steps:

- create a new chart with `mapstore:states` layer on gs-stable
- select Line Chart trace
- select `state_name` as X Attribute
- select `land_km` as Y Attribute
- select AVG as Operation
- select Classification style
- select `sub_region` as Classification Attribute
- now switch the chart trace to Bar Chart
- switch again the trace to Line Chart
- select Classification style

Expected:
- chart builder is visible wit the correct preview

Current:
- error in console and grey page

Note: this seems to happen with the specific `mapstore:states` layer

allyoucanmap:

ElenaGallo:
@allyoucanmap here are some notes from my test:


1_ Once the line chart widget is added to the dashboard, it is not possible to select individual items in the legend and view them on the chart.
Use this dah sample: https://dev-mapstore.geosolutionsgroup.com/mapstore/#/dashboard/63584

*Expected Behavior*
(same behavior as the bar chart)

https://github.com/user-attachments/assets/7cdd240b-793b-4315-abd4-7696a76e2a8d

*Current Behavior*

https://github.com/user-attachments/assets/137a8da5-12ec-48fb-ad19-ac8d1a32c072

2_ When switching from a bar chart to a line chart, the classification style option causes an error.


https://github.com/user-attachments/assets/c5a2b0a9-590c-41cb-97ba-9880b7a27018

PR #11789, opened 2025-12-10
Improve layout in the map viewer

subashtiwari1010:
## Description
This PR updates the layout of the map viewer with the possibility of adding the map as the background and letting all the components act as the overlay over the map. The feature grid is added in the bottom container of the map layout.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue
The feature grid is in the dock container of the map viewer.
https://github.com/geosolutions-it/MapStore2/issues/11794

**What is the new behavior?**
The feature grid is displayed in the bottom container of the map layout.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/configs/localConfig.json
+++ b/web/client/configs/localConfig.json
@@ -977,7 +977,10 @@
         "cfg": {
           "titleMsgId": "widgets.mapWidget.mapEditorTitle",
           "overrideDefaultPlugins": [
-            { "name": "FeatureEditor", "cfg": { "containerPosition": "bottom" } }
+            {
+              "name": "FeatureEditor",
+              "cfg": { "containerPosition": "bottom" }
+            }
           ]
         }
       },


web/client/components/layout/MapViewerLayout.jsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import FlexBox, { FlexFill } from "./FlexBox";
+
+export default ({
```
allyoucanmap: please avoid direct default export, write instead

```js
const MapViewerLayout = () => ({
...

export default MapViewerLayout;
```

web/client/components/layout/MapViewerLayout.jsx
```diff
@@ -0,0 +1,37 @@
...
+                    <FlexFill classNames={['_relative', 'ms2-layout-content']}>
+                        {children}
+                    </FlexFill>
+                    <div className="_relative ms2-layout-right-column">{rightColumn}</div>
+                    <div className="ms2-layout-columns">{columns}</div>
+                </FlexFill>
+                <div className="_relative">{bottom}</div>
+            </FlexFill>
+            {footer}
+        </FlexBox>
```
allyoucanmap, line 41: Let's change name of the classes to ensure it is not the same of BorderLayout. We could use `ms-map-viewer-layout`, `ms-map-viewer-layout-...`

web/client/product/assets/css/viewer.css
```diff
@@ -154,3 +154,14 @@ html, body, #container, .fill {
...
+/* Disable pointer events on the main content container but enable them for all its children */
+.ms2-layout-main-content {
+    pointer-events: none;
+}
+.ms2-layout-main-content .ms2-layout-content > *,
+.ms2-layout-main-content .ms2-layout-left-column > *,
+.ms2-layout-main-content .ms2-layout-right-column > *,
+.ms2-layout-main-content .ms2-layout-columns > * {
+    pointer-events: auto;
+}
```
allyoucanmap: review classes based on previous comment. In addition we should ensure these classes are included in the main theme, the viewer.css may be included in component not used inside other project, I think we should move this inside the  .less files, maybe in common.less

web/client/reducers/featuregrid.js
```diff
@@ -77,7 +77,7 @@ const emptyResultsState = {
     drawing: false,
     newFeatures: [],
     features: [],
-    dockSize: 0.35,
+    dockSize: 0,
```
allyoucanmap: We need to review how dockSize has been used and in case remove completely the logic from MapStore including related actions. Could you check if we had previously logic to set the initial size of the Attribute table panel?

We need also to clean up all the logic related to the feature editor dock bottom size in [updateMapLayoutEpic](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/epics/maplayout.js#L48-L159), it's not needed anymore

web/client/plugins/featuregrid/hoc/withResize.jsx
```diff
@@ -0,0 +1,115 @@
...
+                        transition: 'border-color 0.2s'
+                    }}
+                    onMouseEnter={(e) => {
+                        e.currentTarget.style.borderTopColor = '#ccc';
+                    }}
+                    onMouseLeave={(e) => {
+                        if (!isResizing) {
+                            e.currentTarget.style.borderTopColor = 'transparent';
+                        }
+                    }}
```
allyoucanmap: Could we move all of this in a less or css? We could also use a direct import but I think it would be better to use the the less/css option for the following reasons:

- use of css variable for styling
- use of pseudoselector to apply border color (e.g.: :hover instead of onMouseEnter/onMouseLeave)

web/client/plugins/featuregrid/hoc/withResize.jsx
```diff
@@ -0,0 +1,115 @@
...
+                document.body.style.cursor = 'row-resize';
+                document.body.style.userSelect = 'none';
+            }
+
+            return () => {
+                document.removeEventListener('mousemove', handleMouseMove);
+                document.removeEventListener('mouseup', handleMouseUp);
+                document.body.style.cursor = '';
+                document.body.style.userSelect = '';
+            };
```
allyoucanmap, line 51: Maybe it's better to use pointer events instead? pointermove pointerup

web/client/components/layout/MapViewerLayout.jsx
```diff
@@ -0,0 +1,37 @@
...
+        <FlexBox id={id} className={className} column classNames={['_fill', '_absolute']}>
+            {header}
+            <FlexFill flexBox column className={bodyClassName} classNames={['_relative', 'ms2-layout-body']}>
+                <div className="_fill _absolute">{background}</div>
+                <div className="_relative">{top}</div>
+                <FlexFill flexBox classNames={['_relative', 'ms2-layout-main-content']}>
+                    <div className="_relative ms2-layout-left-column">{leftColumn}</div>
+                    <FlexFill classNames={['_relative', 'ms2-layout-content']}>
+                        {children}
+                    </FlexFill>
```
allyoucanmap: This component container represent the visible content on top of the map, when this component resize we should update also the state of the `boundingMapRect` of the maplayout reducers. We need to:

- include a callback inside this component that return information about this component changes (we could use a resize observer). This component should not be directly connected to the state
- later we will connect this component when we import it in the MapViewer container
- at the moment we need just to update information for the bottom side, left and right are still managed by the updateMapLayout epic

allyoucanmap:

tdipisa:
@subashtiwari1010 
it seems there are a few inconsistencies I've noticed.

1. Open both toc and catalog, then close catalog

image

2. Open a map the first time then open catalog

image

Reproduced with: https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/60692

subashtiwari1010:
Few inconsistencies that I've noticed:
- Export panel
There is some style issue with the header, footer for the Export panel
image

- Import panel
There is some style issue with the header, footer for the import panel
image

- Resizable / movable popup -> Share modal
When such popup is moved to the header or footer, it goes under them.
image
image

allyoucanmap:
@subashtiwari1010 just noticed the second step of context creation is not showing map

image

PR #11865, opened 2026-01-13
Fix #11864 : In the Permission tab when username is very long text then it is not clipped

### LINKED ISSUE #11864: In the Permission tab when username is very long text, then it is  not clipped.
## Description

In the Permissions tab, usernames with long text are not clipped (ellipsis is not applied) which causes overflow.

This happens because the text is currently rendered inside a React Fragment, which prevents text overflow handling.

https://github.com/geosolutions-it/MapStore2/blob/master/web/client/plugins/ResourcesCatalog/components/ALink.jsx#L13

To fix the default behavior, we can wrap the username with the Text component and apply the ellipsis class along with a title attribute.

Image

anup39:
## Description
In the Permissions tab, usernames with long text are now clipped (ellipsis)

 - [X] Bugfix

## Issue

#11864 

**What is the new behavior?**

fix_long_username

## Breaking change
 - [x] No

### DIFF
--- a/web/client/themes/default/less/resources-catalog/_permissions.less
+++ b/web/client/themes/default/less/resources-catalog/_permissions.less
@@ -27,6 +27,7 @@
     }
 }
 
+
 .ms-permission-entryname {
     max-width: 145px;
 }


web/client/plugins/ResourcesCatalog/components/ALink.jsx
```diff
@@ -8,9 +8,20 @@
...
+const EllipsisText = ({ children, maxWidth = 145, ...props }) => (
+    <Text
+        {...props}
+        title={children}
+        ellipsis
+        style={{ maxWidth }}
+    >
+        {children}
+    </Text>
+);
```
allyoucanmap: I think this will change all the usage of the ALink and may cause different problem, could we just solve this for the specific username component

anup39: Reverted and fixed

web/client/plugins/ResourcesCatalog/components/Permissions.jsx
```diff
@@ -156,10 +155,13 @@ function Permissions({
                                                         ? <img src={item.avatar}/>
                                                         : <Glyphicon glyph={item.type} />}
                                                 </Text>
-                                                <ALink
-                                                    href={item.link}>
+                                                <Text
+                                                    title={item.name}
+                                                    ellipsis
+                                                    style={{ maxWidth: 145 }}
```
subashtiwari1010: I would suggest to use a className instead of using the inline css.

anup39:
Close due to github config problem. And opened link another PR for this.

Issue #11633, opened 2025-10-27
Dynamic filter widget


allyoucanmap:
## Description

This epic collects all related issues to complete this task

## Acceptance criteria

- [ ] A new widget to dynamically filter connected content

ElenaGallo:
@offtherailz  I found this small layout issue. Should I create a new issue?

When the title is added to the filter set as a data source, the title box can be moved as shown in the video

https://github.com/user-attachments/assets/13e68564-c294-4925-b3a8-5f222ebfac80

tdipisa:
>@offtherailz I found this small layout issue. Should I create a new issue?

@ElenaGallo thank you. I would say yes. Please, include it in the epic's issues list

PR #11867, opened 2026-01-13
Dynamic filter widget

rowheat02:
## Description

 - [x] Feature

## Issue

 Dynamic filter widget
https://github.com/geosolutions-it/MapStore2/issues/11633

**What is the new behavior?**

Implement the Dynamic filter widget

## Breaking change
 - [x] No

## Other useful information
This is the draft PR More discussions and reviews will be in the comment section.

### DIFF
--- a/web/client/plugins/widgetbuilder/utils/filterBuilderDefaults.js
+++ b/web/client/plugins/widgetbuilder/utils/filterBuilderDefaults.js
@@ -31,7 +31,7 @@ export const createNewFilter = (filtersCount = 0) => {
             icon: 'filter',
             selectionMode: 'multiple',
             direction: 'vertical',
-            maxHeight: 240,
+            maxHeight: undefined,
             label: getFilterName(filtersCount),
             titleStyle: {
                 fontSize: 14,


rowheat02:
Current Status for initial dynamic widget development: 
- The user can create a filter widget. Similar to other widget creation.
- The user can define a filter of different data sources, "User defined" and "Features."
  - On user-defined: The Filter can be defined as shown in the video
  - On Features: Fetched features will be used to generate the filter
- Can be connected to interactionTree (where filter should be applied, like trace of chart, layer of map, layer of table)
- Applied filter refreshed after filter widget update and deleted
- Applied filter persisted
- Basic demo is still there

https://github.com/user-attachments/assets/24dfc09e-b18c-4871-b685-6a36670d9034

Remaining issues:
- Unused code is still there, and translation work is also remaining. Will be covered along with upcoming issues
- Unit tests need to be covered
- Known bugs:
    - Trace item is not persisting the applied filter after page refresh
    - After a long change, sometimes changing the Value and Sort attributes does not fetch new data 
- Still under development, bugs may appear
- Code clean up and enhancement will be done along with other issues

web/client/reducers/widgets.js
```diff
@@ -499,4 +499,4 @@ function widgetsReducer(state = emptyState, action) {
     }
 }
 
-export default widgetsReducer;
+export default widgetsReducer;
```
offtherailz:
```suggestion
export default widgetsReducer;

```

web/client/reducers/interactions.js
offtherailz, line 1: This is not needed at all I think

offtherailz (changes requested):
- [ ] no selection and clear must be consistent each other. If no selection is denied, you should never go in a condition to have something unseleted, neither the clear should be visible. You have also the clear "x" in dropdown. If says "no selection" and turning it on, it means that the label must be "force selection". Image
- [ ] Seeing the things, looks like the no-selection button is not working at all at the moment, probably need to remove to make the PR mergiable.
- [ ] We should indicate that the user has to select at least one attribute to proceed (required fields are not specified)  Image
- [ ] Strings are not localized. I'd suggest to localize things from the beginning, or you loose pieces.
- [ ] Maybe we need to redesign select all / clear button
- [ ] "Connect all" interactions button should be consistent (if it can not be, let's remove it for the moment, better one functionality less that a thing that do not work well, maybe let's hide it for the moment).
- [ ] Apply filter (anyway the first item) should be expanded.
- [ ] The change style button still breaks everything. Remove it before merging.

rowheat02:
The dynamic widget is ready for review.
- The user can define a filter widget with a collection of filters
- Based on DataSource: "User Defined and Features"
   - User-defined: User can define a filter on the filter builder 
   - Features: Features will be fetched to make the filter options
- The filter can be connected to possible connectable nodes (chartWidget's traces, Map layers)
- On a saved filter widget, if the user clicks on filter items to select
   - Filters will be applied to the connected node(traces, layers, counter)
- Deleting defined filters or disconnecting the interaction will remove the applied filter

- Quick demo can be checked in the video below


https://github.com/user-attachments/assets/85132aab-df9f-427a-8977-e1dc96347fac

PR #11884, opened 2026-01-16
Fix #11646 baselayer selector doesn't adjust to the map after Save

### LINKED ISSUE #11646: Baselayer selector doesn't adjust to the map after Save
## Description
After editing annotations that cause the panel to open on the left, the position of the baselayer selector on the map not shifts accordingly. See video below.
This seems to only happen when saving an existing resource and not manually closing the panel. Closing or opening the panel directly always seems to position the baselayer selector correctly. It seems that the coordinates or position offset aren't automatically restored when the save state changes.
The problem seems to be present on DEV and QA

## How to reproduce

The steps to reproduce the problem are simple:
1. Create a new map resource and save it with a name (with or without layers, it seems to behave the same way). It's important that the map viewer is already open on an existing resource, because the problem seems to only affect the "Save" button.
2. Open the left panel and click "Add annotations." At this point, the left panel will become wider than before, and the baselayer selector will move to the right.
3. Edit the annotation text or geometry and click the `Save` button in the navbar to save the map resource, this will make the panel disappear automatically.

*Expected Result*
The baselayer selector should reposition itself in bottom-left with a small margin.

*Current Result*
The baselayer selector remains in the same position without adapting to the map's resizing.

- [x] Not browser related
the same behavior exists on chrome and firefox in the same way

## Other useful information
Even resizing the browser window doesn't move it; this should indicate that it's not just a CSS float bug.

https://github.com/user-attachments/assets/4686ead5-c554-4efa-8214-88726dd1d902

subashtiwari1010:
## Description
This PR fixes the issue of position of the baselayer selector on the map after editing the annotation and saving.

 - [x] Bugfix

## Issue

The baselayer selector remains in the same position without adapting to the map's resizing.
https://github.com/geosolutions-it/MapStore2/issues/11646

**What is the new behavior?**
The baselayer selector reposition itself in bottom-left with a small margin.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/Annotations/epics/annotations.js
+++ b/web/client/plugins/Annotations/epics/annotations.js
@@ -18,7 +18,7 @@ import {
     closeIdentify,
     purgeMapInfoResults
 } from '../../../actions/mapInfo';
-import { MAP_INFO_LOADED } from '../../../actions/config';
+import { MAP_SAVED } from '../../../actions/config';
 import {
     EDIT_ANNOTATION,
     DOWNLOAD,
@@ -198,7 +198,7 @@ export const mergeAnnotationsFeaturesEpic = (action$, { getState }) =>
  * This ensures annotations.editing is reset to fix layout calculation issues
  */
 export const closeAnnotationsOnMapInfoLoadedEpic = (action$, { getState }) =>
-    action$.ofType(MAP_INFO_LOADED)
+    action$.ofType(MAP_SAVED)
         .switchMap(() => {
             const selectedAnnotationLayer = getAnnotationsSession(getState());
             if (selectedAnnotationLayer) {
--- a/web/client/plugins/Annotations/epics/__tests__/annotations-test.js
+++ b/web/client/plugins/Annotations/epics/__tests__/annotations-test.js
@@ -44,7 +44,7 @@ import {
     mergeAnnotationsFeaturesEpic,
     closeAnnotationsOnMapInfoLoadedEpic
 } from '../annotations';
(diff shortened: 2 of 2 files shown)


web/client/plugins/Annotations/epics/annotations.js
```diff
@@ -191,12 +193,27 @@ export const mergeAnnotationsFeaturesEpic = (action$, { getState }) =>
             );
         });
 
+/**
+ * Handles closing annotations when map info is loaded (e.g., after saving)
+ * This ensures annotations.editing is reset to fix layout calculation issues
+ */
+export const closeAnnotationsOnMapInfoLoadedEpic = (action$, { getState }) =>
+    action$.ofType(MAP_INFO_LOADED)
```
anup39: @subashtiwari1010  I think `MAP_SAVED`  is the correct action to listen. However `MAP_INFO_LOADED` also works. Just wanted to know which is better in this case.

subashtiwari1010: Here, as the refresh of the map data is occuring due to the `MAP_INFO_LOADED`, so we need to listen to that action so no further issues arises. As, all the closing of the annotation panel and calculation of the map layout needs to be done during that time.

allyoucanmap:
@ElenaGallo please test this fix on dev and let us know if we can backport, thanks

ElenaGallo:
Test passed, @anup39 please backport. Thanks

offtherailz:
Successfully created backport PR for `2025.02.xx`:
- #11933

Issue #11973, opened 2026-02-12
MetadataExplorer plugin config - hideThumbnails value setting has no effect


mahmoudadel54:
## Description

In the MetadataExplorer plugin, called also catalog, it is possible to configure it so that when layers are searched from the service they show as a list, by default the list items contain a thumbnail on the left that is possible to hide using a configuration item in localConfig.json cfg.hideThumbnail see https://mapstore.geosolutionsgroup.com/mapstore/docs/api/plugins#plugins.MetadataExplorer.
The setting does not seem to have any effect, both when set on localConfig.json and when set as plugin configuration for contexts having this plugin

## How to reproduce

-  Open a localConfig.json of MapStore and include the cfg.hideThumbnail in the MetadataExplorer plugin configuration, or alternatively create a new context adding the Catalog plugin and change the configuration to include such setting
- Save the localConfig.json or the context and open a map or a map created from the context just created

*Expected Result*

When opening the Catalog and searched in a service for a layer, the retrieved list of items should have no thumbnails on the left part.

*Current Result*

When opening the Catalog and searched in a service for a layer, the retrieved list of items still has thumbnails showing on the left part.

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MV88:
i don't think backport needed label is needed if the milestone is next major

MV88:
@ElenaGallo please test it in DEV

MV88:
tested in prod env of the client and is working

PR #11975, opened 2026-02-12
#11973: fix issue of MetadataExplorer plugin config - hideThumbnails value setting has no effect

mahmoudadel54:
## Description

This PR includes:
- fix issue of not reflect cfg value for **hideThumbnail** in **MetadataExplorer** plugin
- handle use priority for hideThumbnail of service if configred/set then global cfg hideThumbnail from plugin

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#11973 

**What is the new behavior?**

If user configured **hideThumbnail** into **MetadataExplorer ** plugin with value, this value will be used and will reflect into catalog cards and advanced settings as well.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/plugins/MetadataExplorer.jsx
+++ b/web/client/plugins/MetadataExplorer.jsx
@@ -370,7 +370,9 @@ export const BackgroundSelectorAdd = connect(
  * @class
  * @name MetadataExplorer
  * @memberof plugins
- * @prop {string} cfg.hideThumbnail shows/hides thumbnail
+ * @prop {string} cfg.hideThumbnail - Global configuration for thumbnail visibility.
+ *   - **Value**: `true` to hide thumbnails globally, `false` to show thumbnails in catalog.
+ *   - **Overrides**: Service-specific configurations take precedence.
  * @prop {object[]} cfg.serviceTypes Service types available to add a new catalog. default: `[{ name: "csw", label: "CSW" }, { name: "wms", label: "WMS" }, { name: "wmts", label: "WMTS" }, { name: "tms", label: "TMS", allowedProviders },{ name: "wfs", label: "WFS" },{ name: "flatgeobuf", label: "FlatGeobuf" }]`.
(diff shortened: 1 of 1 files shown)


web/client/reducers/catalog.js
```diff
@@ -209,7 +209,7 @@ function catalog(state = {
         });
     }
     case TOGGLE_THUMBNAIL: {
-        return set("newService.hideThumbnail", !state.newService.hideThumbnail, state);
+        return set("newService.hideThumbnail", action.toggleValue, state);
```
MV88, line 212: a toggle action usually changes negate the previous value, in this case you are just setting it

it would be better to rename the function to SET_THUMBNAIL_FLAG or SET_THUMBNAIL_VISIBILITY

otherwise don't negate the value in the onChange method.
better this way then

web/client/components/catalog/Catalog.jsx
```diff
@@ -227,7 +227,7 @@ class Catalog extends React.Component {
         // defaults for recordItem elements
         let metadataTemplate = "";
         let showTemplate = false;
-        let hideThumbnail = false;
+        let hideThumbnail = this.props.hideThumbnail;
```
MV88, line 230: is this needed?

like we define it anyway here in line 241

if (!isNil(selectedService.hideThumbnail)) {
                hideThumbnail = selectedService.hideThumbnail;
            }

mahmoudadel54: Yes it is needed as it belongs to global cfg of the plugin not the service one

MV88: can you better document the behaviour of [this](https://github.com/mahmoudadel54/MapStore2/blob/662d8342a40775b28a9ca31e1435798a860ca69a/web/client/plugins/MetadataExplorer.jsx#L373C23-L373C36) via jsdoc?

like the new logic is global can be defined via cfg, but a service config will override this?

PR #12019, opened 2026-02-25
#11987: Rules Manager - opening old rules using standalone geofence does not work

mahmoudadel54:
## Description

This PR handles fix issue in edit existing rule for geofence standalone if layer or workspace missing in rule object. It includes:

- refactor(EditorEnhancer): ensure fetching instances list for dropdown even if layer/workspace are not yet selected
- Extract reusable `fetchStylesAndAttributesStream` helper to reduce duplication

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**

#11987 

**What is the new behavior?**

The GS Instances are loaded to edit rule screen showing all form fields of edit

https://github.com/user-attachments/assets/ab605f1f-a2de-4632-8974-0e68ff885c93

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #12034, opened 2026-03-02
[Backport 2026.01.xx] #11987: Rules Manager - opening old rules using standalone geofence does not work

### LINKED ISSUE #11987: Rules Manager - opening old rules using standalone geofence does not work
## Description

the rule panel is empty when opening an existing rule using a standalone instance of geofence AND when a rule has no layer defined. so technically this is reproducible outside client insfrastructure

the EditorEnhancer has a filter that prevents its execution, therefore we have to remove that filter and adapt the styles and attribute fetch adding a if to check if there re layer and workspace. so handle that case.

we can also create a function for the styles and attributes to be reused (3 times)

### How to reproduce

- open rule manager
- edit existing rule

*Expected Result*

to be able to 
*Current Result*

not able to blank page becasue gsInstances array is empty 
- [x] Not browser related

|---|---|
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |

</details>

offtherailz:
# Description
Backport of #12019 to `2026.01.xx`.

Fixes #11987

### DIFF
--- a/web/client/plugins/manager/EditorEnhancer.js
+++ b/web/client/plugins/manager/EditorEnhancer.js
@@ -13,9 +13,33 @@ const emitStop = stream$ => stream$.filter(() => false).startWith({});
 import { getStylesAndAttributes } from '../../observables/rulesmanager';
 import { gsInstancesDDListSelector } from '../../selectors/rulesmanager';
 import { storeGSInstancesDDList } from '../../actions/rulesmanager';
+
+/**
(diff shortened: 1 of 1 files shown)


tdipisa (changes requested):
Waiting for this before merging the PR:
https://github.com/geosolutions-it/MapStore2/issues/11987#issuecomment-3984916593

PR #12176, opened 2026-03-27
Fix #12170: Attribute table  toolbar buttons  border-radius issue.

### LINKED ISSUE #12170: Attribute Table toolbar buttons in Edit mode are not rounded
## Description

with recent layour/css updates we have missed the attribute table toolbar in edit mode. Furthermore, also in view mode the size of the first two buttons in the toolbar is a bit bigger.

Image

## How to reproduce

- Open a map 
- Go in edit mode

*Expected Result*

Button should be rounded as in view mode

Image

*Current Result*

Image

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

anup39:
## Description

In the Attribute table toolbar, some buttons were hidden by applying inline styles (width: 0, padding: 0, border-width: 0) instead of removing them from the DOM.

Because the toolbar uses a ButtonGroup, border-radius is applied to first and last child elements. When hidden buttons still exist in the DOM, they represents first or last children, so the visible edge buttons do not receive the expected rounded corners.

Additionally, the split button component wrapped content with `ContainerDimensions` without consuming width or height values, which added unnecessary resize-detector DOM nodes (erd_scroll_detection_container). This element was also considered as last element which is unnecessary in this case

Also, style related to class `square-button no-border` was not applied to the final button DOM in some of the tools because the class was passed via `buttonClassName` in places where the effective styling path expected `className`.
wrong_attribute_buttonClassname

This PR:
1. Updated toolbar button visibility behavior to unmount hidden buttons instead of shrinking them.
2. Updated split toolbar button visibility behavior to unmount hidden buttons as well.
3. Removed unused ContainerDimensions wrapper from split toolbar button, since measured dimensions (width and height) were not used.
4. Updated tools class  for snap and viewport filter so `square-button no-border` is passed through `className` (instead of relying on `buttonClassName`), ensuring expected classes are applied to rendered toolbar elements.

 - [X] Bugfix

## Issue

#12170 

**What is the new behavior?**

- First and last visible toolbar buttons consistently get correct rounded corners.
- Toolbar DOM no longer contains zero-size hidden buttons for visibility control.
- Split button no longer introduces unnecessary and unused resize-detector helper nodes from ContainerDimensions.
- Snap and viewport filter tools correctly carry `square-button no-border` in the effective rendered DOM/classes.

https://github.com/user-attachments/assets/511ad2a5-5863-42ee-a046-c4312613bcf8

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/data/featuregrid/toolbars/Toolbar.jsx
+++ b/web/client/components/data/featuregrid/toolbars/Toolbar.jsx
@@ -30,8 +30,18 @@ const getSaveMessageId = ({ saving, saved, error }) => {
     }
     return "featuregrid.toolbar.saveChanges";
 };
+const VisibleTSplitButton = ({visible, ...props})=>{
+    if (!visible) return null;
+    return <TSplitButton visible={visible} {...props}/>;
+};
+
+const VisibleTButton = ({visible, ...props})=>{
+    if (!visible) return null;
+    return <TButton visible={visible} {...props}/>;
+};
+
 const standardButtons = {
-    editMode: ({disabled, mode, isEditingAllowed, layer, events = {}}) => (<TButton
+    editMode: ({disabled, mode, isEditingAllowed, layer, events = {}}) => (<VisibleTButton
         id="edit-mode"
         keyProp="edit-mode"
         tooltipId="featuregrid.toolbar.editMode"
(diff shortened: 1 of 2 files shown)


web/client/components/data/featuregrid/toolbars/Toolbar.jsx
```diff
@@ -186,8 +186,7 @@ const standardButtons = {
         onMount={() => !snappingConfig && events.setSnappingConfig && events.setSnappingConfig(null, null, pluginCfg)}
         title={isSnappingLoading ? <Spinner spinnerName="ball-beat" overrideSpinnerClassName="spinner" key="loadingSpinner" noFadeIn /> : <Glyphicon glyph="magnet" />}
         tooltipPosition="top"
-        className="snap-tool"
-        buttonClassName="square-button no-border"
+        className="snap-tool square-button no-border"
```
subashtiwari1010: Please revert this changes as this is a TSplitButton so the buttonClassName is valid props.

anup39: Done

offtherailz:
Successfully created backport PR for `2026.01.xx`:
- #12182

subashtiwari1010:
@ElenaGallo  please test this fix on dev, thanks

PR #12179, opened 2026-03-27
Fix COG and FlatGeobuf layers support params for protected storage

### LINKED ISSUE #12162: COG and FlatGeobuf layers support params for protected storage
## Description
When adding a COG or FlatGeobuf source files The parameters defined in the localConfig section: requestsConfigurationRules are not passed in remote requests.
This prevents the opening of protected COG or FlatGeobuf files on remote storage.

This occurs at time to opening the catalog and also when adding layers to the map.

## How to reproduce
add some rules in section requestsConfigurationRules
(This is just an example. You need to use keys defined on the server that distributes the resources.)
```json
 {
      "urlPattern": "https://remote.storage/cog/.*",
      "params": {
          "sig": "574XQctJduYVZLpxS5I3yHBaMHFIIuCbYQZ1sXpr7iA="
      }
  },
  {
      "urlPattern": "https://remote.storage/flatgeobuf/.*",
      "params": {
          "sig": "m7ESsEudYbnzjGW3cD7fZNwNolPKzBbEDc2Y3tXrjEI="
      }
  }
```

*Expected Result*
loading remote resources using correct url parameters by config

*Current Result*
parameters defined in requestsConfigurationRules is ignored

- [x] Not browser related

stefanocudini:
Fix #12162 

- fix cog catalog sugin security params
- new function setSecurityParams
- added unit test for setSecurityParams
- secure params in 3D layers
- fix tests SecurityUtils

## Description
In short, in cases of datasets using custom rules in the requestsConfigurationRules section of localConfig.json, the URL parameters for resource addresses corresponding to COG and FlatGeobuf files were not being passed. See the issue for an example of rules to use for testing.

 - [x] Bugfix

## Issue
Fix #12162 

#<issue>

**What is the new behavior?**
A new method `setSecurityParams` has been added to avoid code replication that automatically appends parameters to the URL when necessary. In cases where no rules exist, the original URL is used

## Breaking change
 - [x] No

## Other useful information
See the linked issue for bug reproducing details.
Geonode-client porting is required

### DIFF
--- a/web/client/api/FlatGeobuf.js
+++ b/web/client/api/FlatGeobuf.js
@@ -7,7 +7,7 @@
  */
 import axios from '../libs/ajax';
 import isEmpty from 'lodash/isEmpty';
-import { setSecurityParams } from '../utils/SecurityUtils';
+import { updateUrlParams } from '../utils/URLUtils';
 
 export const FGB = 'fgb';
 export const FGB_LAYER_TYPE = 'flatgeobuf';
@@ -46,8 +46,8 @@ export const getCapabilities = (url) => {
     return getFlatGeobufGeneric().then(flatgeobuf => {
         return axios.get(url, {
             adapter: config => {
-                const axiosUrl = setSecurityParams(config.url);
-                return flatgeobuf.readMetadata(axiosUrl);
+                const secureUrl = updateUrlParams(config.url, config.params);
+                return flatgeobuf.readMetadata(secureUrl, true, config.headers);
             }
         }).then((metadata) => {
 
--- a/web/client/components/map/cesium/plugins/COGLayer.js
+++ b/web/client/components/map/cesium/plugins/COGLayer.js
@@ -7,7 +7,8 @@
  */
 
 import Layers from '../../../../utils/cesium/Layers';
-import { setSecurityParams } from '../../../../utils/SecurityUtils';
+import { getRequestConfigurationByUrl } from '../../../../utils/SecurityUtils';
+import { updateUrlParams } from '../../../../utils/URLUtils';
 
 import isEqual from 'lodash/isEqual';
 import proj4 from 'proj4';
@@ -90,11 +91,17 @@ const createLayer = (options) => {
     if (!options.visibility) {
         return null;
     }
-    const url = setSecurityParams(options.url || options?.sources[0]?.url);
 
+    const url = options.url || options?.sources[0]?.url;
+    const {headers, params} = getRequestConfigurationByUrl(url, options?.security?.sourceId);
+    const secureUrl = updateUrlParams(url, params);
     const renderOptions = buildRenderOptions(options);
 
-    return TIFFImageryProvider.fromUrl(url, {
+    // https://github.com/hongfaqiu/TIFFImageryProvider?tab=readme-ov-file#api
+    return TIFFImageryProvider.fromUrl(secureUrl, {
+        requestOptions: {
+            headers
+        },
         projFunc: (code) => {
             const epsgCode = `EPSG:${code}`;
             if (isProjectionAvailable(epsgCode)) {
--- a/web/client/components/map/cesium/plugins/FlatGeobufLayer.js
+++ b/web/client/components/map/cesium/plugins/FlatGeobufLayer.js
@@ -16,7 +16,8 @@ import {
 import { applyDefaultStyleToVectorLayer } from '../../../../utils/StyleUtils';
 import GeoJSONStyledFeatures from  '../../../../utils/cesium/GeoJSONStyledFeatures';
 import { createVectorFeatureFilter } from '../../../../utils/FilterUtils';
-import { setSecurityParams } from '../../../../utils/SecurityUtils';
+import { updateUrlParams } from '../../../../utils/URLUtils';
+import { getRequestConfigurationByUrl } from '../../../../utils/SecurityUtils';
 
 import {
     FGB_LAYER_TYPE,
@@ -70,9 +71,9 @@ const createLayer = (options, map) => {
         };
 
         let rect = mapBbox();
-
-        const secureUrl = setSecurityParams(options.url);
-        for await (const feature of flatgeobuf.deserialize(secureUrl, rect)) {
+        const {headers} = getRequestConfigurationByUrl(options.url, options?.security?.sourceId);
+        const secureUrl = updateUrlParams(options.url, options.params);
+        for await (const feature of flatgeobuf.deserialize(secureUrl, rect, headers)) {
             geojson.features.push(feature);
         }
 
--- a/web/client/components/map/openlayers/plugins/COGLayer.js
+++ b/web/client/components/map/openlayers/plugins/COGLayer.js
@@ -13,21 +13,22 @@ import get from 'lodash/get';
 import GeoTIFF from 'ol/source/GeoTIFF.js';
 import TileLayer from 'ol/layer/WebGLTile.js';
 import { isProjectionAvailable } from '../../../../utils/ProjectionUtils';
-import { getRequestConfigurationByUrl, setSecurityParams } from '../../../../utils/SecurityUtils';
+import { getRequestConfigurationByUrl } from '../../../../utils/SecurityUtils';
+import { updateUrlParams } from '../../../../utils/URLUtils';
 
 function create(options) {
     let sources = [];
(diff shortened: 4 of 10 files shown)


web/client/api/FlatGeobuf.js
```diff
@@ -45,7 +46,8 @@ export const getCapabilities = (url) => {
     return getFlatGeobufGeneric().then(flatgeobuf => {
         return axios.get(url, {
             adapter: config => {
-                return flatgeobuf.readMetadata(config.url);
+                const axiosUrl = setSecurityParams(config.url);
+                return flatgeobuf.readMetadata(axiosUrl);
```
allyoucanmap: It seems possible to include also headers https://github.com/flatgeobuf/flatgeobuf/commit/ec78df6aa1c4e7df7e45fc25650c5c53e4f621ea#diff-b03111f7d0b3f8b9e9642ebf0ea8d6fbaff46d335a83257901ff5acb357f3a4cR187-R188.

Note the axios library is already including the params and headers based on security rules with the [addAuthenticationToAxios](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/libs/ajax.js#L49-L89) interceptor, could you verify if the config headers and params are already including needed information and use here directly url parse format without need of setSecurityParams?

web/client/components/map/cesium/plugins/COGLayer.js
```diff
@@ -89,7 +90,7 @@ const createLayer = (options) => {
     if (!options.visibility) {
         return null;
     }
-    const url = options.url || options?.sources[0]?.url;
+    const url = setSecurityParams(options.url || options?.sources[0]?.url);
```
allyoucanmap: In this case I would prefer we use getRequestConfigurationByUrl and extract both headers and params and pass them to a cesium Resource class https://cesium.com/learn/cesiumjs/ref-doc/Resource.html. In the .fromUrl(url we should be able to pass the resource instead of the url.

web/client/components/map/cesium/plugins/FlatGeobufLayer.js
```diff
@@ -70,7 +71,8 @@ const createLayer = (options, map) => {
 
         let rect = mapBbox();
 
-        for await (const feature of flatgeobuf.deserialize(options.url, rect)) {
+        const secureUrl = setSecurityParams(options.url);
+        for await (const feature of flatgeobuf.deserialize(secureUrl, rect)) {
```
allyoucanmap: also in this case the headers are supported, let's include them https://github.com/flatgeobuf/flatgeobuf/blob/master/src/ts/geojson/featurecollection.ts#L55, please let's try to include them. I think the params and header can be computed outside the loadingBbox function.

web/client/components/map/openlayers/plugins/COGLayer.js
```diff
@@ -13,16 +13,23 @@ import get from 'lodash/get';
...
 function create(options) {
+    let sources = [];
     let sourceOptions = {};
-    if (options.security && options.sources && options.sources.length > 0) {
+    if (options.sources && options.sources.length > 0) {
         const firstSource = options.sources[0];
-        const requestConfig = getRequestConfigurationByUrl(firstSource.url, null, options.security?.sourceId);
-        if (requestConfig.headers) {
-            sourceOptions.headers = requestConfig.headers;
+        const {headers} = getRequestConfigurationByUrl(firstSource.url, null, options.security?.sourceId);
```
allyoucanmap: The params are already computed here. Could we extract the params from here and maybe use [updateUrlParams](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/URLUtils.js#L164-L169) from URLUtils ?

web/client/components/map/openlayers/plugins/FlatGeobufLayer.js
```diff
@@ -47,14 +49,15 @@ const getWFSStyle = (layer, options, map) => {
 
 const createLoader = (source, options, strategy) => (extent, resolution, projection) => {
     getFlatGeobufOl().then(flatgeobuf => {
-        const loader = flatgeobuf.createLoader(source, options.url, 'EPSG:4326', strategy, true);
+        const secureUrl = setSecurityParams(options.url);
+        const loader = flatgeobuf.createLoader(source, secureUrl, 'EPSG:4326', strategy, true);
```
allyoucanmap: could we include also header in this case, they seems supported https://github.com/flatgeobuf/flatgeobuf/blob/master/src/ts/ol.ts#L87

web/client/utils/cog/LayerUtils.js
```diff
@@ -47,9 +48,10 @@ export const fromUrl = (url, signal) => {
     if (signal?.aborted) {
         return abortError(Promise.reject);
     }
+    const secureUrl = setSecurityParams(url);
     return new Promise((resolve, reject) => {
         signal?.addEventListener("abort", () => abortError(reject));
-        return fromGeotiffUrl(url)
+        return fromGeotiffUrl(secureUrl)
```
allyoucanmap: also in this case the headers seem supported, could you include them? https://geotiffjs.github.io/geotiff.js/functions/geotiff.fromUrl.html https://geotiffjs.github.io/geotiff.js/interfaces/geotiff.RemoteSourceOptions.html

web/client/utils/SecurityUtils.js
```diff
@@ -356,6 +356,30 @@ export const  getRequestConfigurationByUrl = (url, securityToken, sourceId) => {
...
+            });
+            // we need to remove this to force the use of query
+            delete url.search;
+            return URL.format(url);
+        } catch (e) {
+            return sourceUrl;
+        }
+    }
+    return sourceUrl;
+}
```
allyoucanmap: In most of the cases reviewed we need to extract both headers and params. My proposal in this case is to use:

- const { headers, params } = getRequestConfigurationByUrl(url, ...); to extract all at once
- then use [updateUrlParams](https://github.com/geosolutions-it/MapStore2/blob/master/web/client/utils/URLUtils.js#L164-L169)  to apply params when not supported natively

web/client/components/map/cesium/plugins/COGLayer.js
allyoucanmap: Note all layers may recieve the options.security?.sourceId property that could be used with getRequestConfigurationByUrl

web/client/components/map/cesium/plugins/FlatGeobufLayer.js
allyoucanmap: Note all layers may recieve the options.security?.sourceId property that could be used with getRequestConfigurationByUrl

web/client/components/map/openlayers/plugins/FlatGeobufLayer.js
allyoucanmap: Note all layers may recieve the options.security?.sourceId property that could be used with getRequestConfigurationByUrl

web/client/api/FlatGeobuf.js
stefanocudini: yes check axios juect return headers anche params

web/client/components/map/openlayers/plugins/COGLayer.js
stefanocudini: replaced all `setSecurityParams` with `updateUrlParams`

web/client/utils/cog/LayerUtils.js
stefanocudini: solved and linked 3d third party docs

web/client/utils/SecurityUtils.js
stefanocudini: removed setSecurityParams and it's tests

web/client/components/map/cesium/plugins/FlatGeobufLayer.js
stefanocudini: passed options.security

web/client/components/map/openlayers/plugins/FlatGeobufLayer.js
stefanocudini: passed options.security and headers

allyoucanmap:
@ElenaGallo please test this fix on dev and let us know when we can backport, thanks

offtherailz:
Successfully created backport PR for `2026.01.xx`:
- #12251

PR #12217, opened 2026-04-03
backport C040_2025.02.xx -  #11987: Rules Manager - opening old rules using standalone geofence does not work (#12019)

MV88:
## Description

backport C040_2025.02.xx -  #11987: Rules Manager - opening old rules using standalone geofence does not work (#12019)


mahmoudadel54 (approved):
LGTM

PR #12347, opened 2026-05-07
Dynamic Filter Widget: "APPLY DIMENSION" to update Time and Elevation #12267

### LINKED ISSUE #12267: Dynamic Filter Widget: "APPLY DIMENSION" to update Time and Elevation
## Description

* Time (datetime)
* Elevation (number)

Image

On apply_dimension targets:

 

* [x] Enhancement

---

### Notes

rowheat02:
## Description

This PR adds Apply Dimension as an optional target for interactions. Can connect to update the layer time and elevation. If a timeline exists, a different global node "MapTime" to control currentTime of timeline

Here are the conditions applied for Apply Dimension
- Maptime(target) can only be connected alone; other targets can not be connected on the same filter. This is done to make consistent reerse sync. 
- When range is enabled in timeline, mapTime connected widget will be disabled
- An apply dimension node can only be connected when selection mode is "Single"
- For elevation, the value field must be numeric, and for time, it must be datetime/date

Demo for Map:

https://github.com/user-attachments/assets/e1ad10a2-c2ca-4782-a2a9-a008dcb4c6ad

Demo for Dashboard

https://github.com/user-attachments/assets/9a793e69-0991-438a-8cf0-6b7370ac54c3

 - [x] Enhancement

## Issue

#12267

**What is the new behavior?**

The user can connect the dimension update node of the layer(time and elevation)

## Breaking change
 - [x] No


web/client/components/widgets/builder/wizard/common/interactions/InteractionsRow.jsx
offtherailz, line 1: Logic for dimension should be isolated somehow. Interaction row component can not contain the logic for every possible target.
Or we have different components  for different actions to do (that maybe share a basic graphic component, that allows to place the row icon, text, tooltops, components etc..., if makes sense), or a logic should be in the container using component. This depends of course on the items.

offtherailz (changes requested):
I still need to check the code. Make sure to design the application with best practices. I see a lot of code and ifs that looks to be hard to maintain

- [x] Conditional Visibility for "+" Button: it should show when at least one additional event can be added. In the stile case, when userDefined style is selected. In case of apply dimension it only appears when a dimension can actually be applied. This prevents users from clicking an action that has no effect.
- [x] Option vetical/horizonal makes no sense on slider (or still not supported) disable it. 
- [x] Implement backward compatibility for missing attribute types.
- [x] See comments inline

rowheat02:
Fixed points mentioned [here](https://github.com/geosolutions-it/MapStore2/pull/12347#pullrequestreview-4264637006) and tick issue to support empty space(which helps in labelling empty string) and alway show tick even label is hidden.
image

offtherailz (commented):
We are near, but there are still some things that are not in line with specification.


- [ ] in dashboard I see the items, but i can not connect to the a layer of the map. Same for generic map, I can connect it only to generic time. Not for single layer. Did we removed this specification?

https://github.com/user-attachments/assets/9a7bb547-e31a-4b44-b6a1-68ba8fbf462d

Image

- [ ] As far as I remember filter should be allowed anyway. The only thing that should be denied is to mutually exclude map time or layer specific time. Is there a specific reason for denying it? 

https://github.com/user-attachments/assets/9a7bb547-e31a-4b44-b6a1-68ba8fbf462d

- [ ] Elevation do not seems to be in sync

https://github.com/user-attachments/assets/436c688c-24c4-4743-be8f-91b5aed4cea7

- [x] "apply regardless the data source" is shown here ? It doesn't make sense to me. 

Image

rowheat02:
@offtherailz 
> **1.  in dashboard I see the items, but i can not connect to the a layer of the map. Same for generic map, I can connect it only to generic time. Not for single layer. Did we removed this specification?**

  => For the first point of the dashboard, there can be two reasons:
       - Selection mode is multiple
       - That node is already connected to another filter source. For consistency, for one layer dimension, only one filter can be connected like done in Apply Style
  
   =>  For the second point 
       - From one filter instance, we can not connect other nodes if mapTime is connected. This is because global time(currenTime) should be synced to the maptime node only. It does not make sense if it is connected to other nodes also.  Example, if we connect to mapTime from applyDimension, layer 1 apply filter. or any other type of future target  that can be changed from outside also(like dimension, style now are changing). Should sync from all sources?. For this complexity, restriction have been applied to mapTime to be connected alone for reverse sync.
       -  we need discussion if we need to sync from all the targets  like target(1) updateds value -> widget sync -> update other targets (2, 3). Currently, we are just syncing the selection of the widget selection only if filter is connected to Maptime


>  2. As far as I remember filter should be allowed anyway. The only thing that should be denied is to mutually exclude map time or layer specific time. Is there a specific reason for denying it?

=> the same explanation from number 1 (second point)



> 3. Elevation do not seems to be in sync

 => This comes around similar reason: Sync was done only for maptime( single connected ), for other they can be connected from multiple targets. So from all targets should we sync? like elevation is connected to multiple layers. from all targets we should sync? Then a bit complex syncing system should be introduced. As enhancement, we can show warning we some target is not synced like for applyStyle.
image

> 4 "apply regardless the data source" is shown here ? It doesn't make sense to me.

=> this is fixed

offtherailz:
I updated but "apply regardless data source"  appears on map time, at least on map, and in this case also blocks be to disable the time dimension on map

image

Please remove this apply regardless data source for dimensions, even with tooltip.

image


Regarding your point, there is no reason to block the interactions for filters that do not receive any value back. I can agree on style. For this mechanism, I agree with you that 2 way sync complicates a lot the interactions, if multiple targets are present. We can think about to relax this binding in the future. 

So please remove the constraint on filters, but I agree with you to keep it on style.

offtherailz (approved):

offtherailz:
@ElenaGallo, could you please test this on  DEV  ? Thank you

offtherailz:
 - Pull request Pull request is a draft

PR #12372, opened 2026-05-12
fix WMS getMap unhandled exception

### LINKED ISSUE #12371: WMS GetMap unhandled error for custom requestsConfigurationRules
## Description
In the case where a new requestsConfigurationRules rule defines custom headers, for example, to authenticate a WMS layer with API tokens.
The case where geoserver returns an exception in the body, causing an attempt to reload the WMS layer, is not handled.

This could cause an infinite loop of WMS layer reload attempts.

## How to reproduce
If you need to call a WMS layer, causing a response exception, this can be simulated by calling a non-existent layer.
The layer must match the custom rules, adding custom headers like the following:

```json
"requestsConfigurationRules": [
    {
      "urlPattern": "http://localhost/geoserver/wms.*",
      "headers": {
        "X-Custom-Header": "CustomTOKEN"
      }
    }
]
```

*Expected Result*
layer loading must be blocked

- [x] Not browser related

stefanocudini:
Fix issue #12371 

## Description
Improvement of rendering block condition for the Openlayers WMS layer. If an exception is returned in the XML response by geoserver body, the individual image tile is blocked and removed from the queue.
the modified logic is here:
`/web/client/components/map/openlayers/plugins/WMSLayer.js` in custom loader: `loadFunction()`

A generic WMS utils has been added `parseOGCException` for exception parsing to display in the console in the event of specific ogc errors.

 - [x] Bugfix

## Issue

#12371 

**What is the new behavior?**
In the case of exceptions from the geoserver, the attempt to reload the layer is blocked in any case and the specific exception message is displayed in the console

## Breaking change
 - [x] No

## Other useful information
For a complete reproduction of the fix, the production environment where the bug was identified is required, which requires a distribution of the layers on the geoserver via authentication and consistent configuration of the requestsConfigurationRules. See on related issues.

### DIFF
--- a/web/client/components/map/openlayers/plugins/WMSLayer.js
+++ b/web/client/components/map/openlayers/plugins/WMSLayer.js
@@ -74,11 +74,11 @@ const loadFunction = (options, headers) => function(image, src) {
             axios.get(newSrc, {
                 headers,
                 responseType: 'blob'
-            }).then(response => {
+            }).then(async response => {
                 if (isValidResponse(response)) { // not contains OGC exception
                     image.getImage().src = URL.createObjectURL(response.data);
                 } else {
-                    const exception = parseOGCException(response);
+                    const exception = await parseOGCException(response);
                     throw new Error('response exception: ' + exception);
                 }
             }).catch(e => {
--- a/web/client/components/map/openlayers/plugins/WMSLayer.js
+++ b/web/client/components/map/openlayers/plugins/WMSLayer.js
@@ -29,7 +29,7 @@ import VectorTileSource from 'ol/source/VectorTile';
 import VectorTileLayer from 'ol/layer/VectorTile';
 
 import { isVectorFormat } from '../../../../utils/VectorTileUtils';
-import { isValidResponse, parseOGCException } from '../../../../utils/WMSUtils';
+import { isValidResponse } from '../../../../utils/WMSUtils';
 import { OL_VECTOR_FORMATS, applyStyle } from '../../../../utils/openlayers/VectorTileUtils';
 
 import { proxySource, getWMSURLs, wmsToOpenlayersOptions, toOLAttributions, generateTileGrid } from '../../../../utils/openlayers/WMSUtils';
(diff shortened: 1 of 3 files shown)


web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -66,26 +66,25 @@ const loadFunction = (options, headers) => function(image, src) {
...
         });
     } else {
-        if (headers) {
+        if (headers) { // case of custom headers is setted in localConfig, example requestsConfigurationRules
             axios.get(newSrc, {
                 headers,
                 responseType: 'blob'
-            }).then(response => {
-                if (isValidResponse(response)) {
+            }).then(async response => {
```
allyoucanmap: Why this is async?

web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -66,26 +66,25 @@ const loadFunction = (options, headers) => function(image, src) {
...
+                if (isValidResponse(response)) { // not contains OGC exception
                     image.getImage().src = URL.createObjectURL(response.data);
                 } else {
-                    // #10701 this is needed to trigger the imageloaderror event
-                    // in ol otherwise this event is not triggered if you assign
-                    // the xml content of the exception to the src attribute
-                    image.getImage().src = null;
-                    console.error("error: " + response.data);
+                    const exception = await parseOGCException(response);
```
allyoucanmap: could we pass the response data error as is if it's a string, without parsing it? So we have the full error and remove the introduction of TextDecoder and DOMParser for each error.

allyoucanmap (changes requested):

web/client/components/map/openlayers/plugins/WMSLayer.js
stefanocudini: The problem is that the previous axios request use responseType: 'blob'.
`Blob` inevitably requires an asynchronous operation to be converted to text.

web/client/components/map/openlayers/plugins/WMSLayer.js
stefanocudini: I did some tests and this seems to be the cleanest solution which only converts the exceptions into text and removed parseOGCException method.
Edited in the ladt commit

```js
.then((response) => {
                    return response.data.type === "text/xml"
                        ? response.data.text().then(dataText => ({...response, dataText}))
                        : response;
                })
                .then(response => {
                    if (isValidResponse(response)) { // not contains OGC exception
                        image.getImage().src = URL.createObjectURL(response.data);
                    } else {
                        throw new Error(response.dataText);
                    }
                }).catch(errorMessage => {
                    image.getImage().src = null; // needed to trigger the MS imageloaderror event in Map.onLayerError
                    image.setState(3); // set error state for tile and removed from the queue and will block, prevent reloading loops
                    console.error(errorMessage);  // show ogc exception in console for debugging
                });
```
Then show xml eception in console:
image

allyoucanmap (approved):

offtherailz:
 - Pull request Pull request is in unstable status

PR #12386, opened 2026-05-14
[Backport geonode-5.0.x] fix WMS getMap unhandled exception

### LINKED ISSUE #12371: WMS GetMap unhandled error for custom requestsConfigurationRules

offtherailz:
# Description
Backport of #12372 to `geonode-5.0.x`.

Fixes #12371

### DIFF
--- a/web/client/components/map/openlayers/plugins/WMSLayer.js
+++ b/web/client/components/map/openlayers/plugins/WMSLayer.js
@@ -66,27 +66,31 @@ const loadFunction = (options, headers) => function(image, src) {
                 }
             }
         }).catch(e => {
+            image.setState(3);
             console.error(e);
         });
     } else {
-        if (headers) {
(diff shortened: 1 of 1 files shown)


allyoucanmap:
not needed

PR #12392, opened 2026-05-15
[Backport 2026.01.xx] Dynamic Filter Widget: "APPLY DIMENSION" to update Time and Elevation #12267

### LINKED ISSUE #12267: Dynamic Filter Widget: "APPLY DIMENSION" to update Time and Elevation

offtherailz:
# Description
Backport of #12347 to `2026.01.xx`.

Fixes #12267


tdipisa (changes requested):
@rowheat02 there are conflicts to be resolved here.

rowheat02 (approved):

tdipisa (approved):

ElenaGallo:
@offtherailz  I would like to add a note because the behavior on QA is currently different from DEV.
In QA, it is not possible to view the time on the timeline and the filter widget slider simultaneously, as you cannot open the widgets and the timeline at the same time.

https://github.com/user-attachments/assets/9c9d9dab-8999-4cd8-992d-984fe68ab373

offtherailz:
@ElenaGallo It looks to be the new timeline for 2026.02.xx that allows to show together widgets and timeline.

PR #12413, opened 2026-05-20
Extend Apply dimension to connect with Chart Axis current Time. #12408

rowheat02:
## Description

This PR extends the apply Dimension to be able to connect with the axis of the chart(if context has no timeline)

https://github.com/user-attachments/assets/10d7df1b-1cb1-4b6c-824e-41b8ebfd785e

Now the user can directly connect to the axis's current Time. If a timeline exists, timeline currentTime is used. If not, then the current value saved in the axis from filterWidget will be used. 

Other improvements / fixes from [report1](https://github.com/geosolutions-it/MapStore2/issues/12268#issuecomment-4488581005) and [here](https://github.com/geosolutions-it/MapStore2/issues/12267#issuecomment-4497921625) also included

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [x] Enhancement

## Issue

**What is the current behavior?**

#12408
Fixes Reported here  https://github.com/geosolutions-it/MapStore2/issues/12268#issuecomment-4488581005
Improvement reported here: https://github.com/geosolutions-it/MapStore2/issues/12267#issuecomment-4497921625

**What is the new behavior?**

- Apply dimension on axis
- Fixes for Slider-related bugs/glitches and fixes for reported issues/improvements mentioned above in "What is the current behavior?"
- User can configure the on-axis to enable a tooltip only on tickValues match.

    https://github.com/user-attachments/assets/c225f889-80a1-4d1d-b032-165ad2c0d766

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #12423, opened 2026-05-22
[Backport 2026.01.xx] Extend Apply dimension to connect with Chart Axis current Time. #12408

### LINKED ISSUE #12267: Dynamic Filter Widget: "APPLY DIMENSION" to update Time and Elevation

### LINKED ISSUE #12408: Extend Apply dimension to connect with Chart Axis current Time.
## Description

Right now, from the dynamic filter widget, only we can connect a layer's dimension or Timeline.
So this issue's intention is to extend the Apply dimension(time), to be able to connect with the axis of the chart trace as current Time(currently which is only coming from Timeline). It can come from the widget also after this issue. (This feature useful when there is no timeline on the context)

Image

- [x] Enhancement

rowheat02:
# Description
Backport of #12413 to `2026.01.xx`.

Fixes #12267,12268,12408


offtherailz (approved):
LGTM.

offtherailz:
@ElenaGallo please test this on QA as soon as deployed

Issue #11779, opened 2025-12-05
Setting the map projection crashes MapStore if resolutions configured in new.json file with a different projection


mahmoudadel54:
## Description

Setting the map projection as admin causes MapStore to crash with error "Out of Memory" if there are some resolutions preset in new.json file with a different projection rather than the selected one.

## How to reproduce

-  in new.json file add resolutions array in map.mapOptions.view for a projection like: 3857
- add new map resource
- select a different CRS from CRS selector like: 4326
- the app will crash

*Expected Result*

Selection CRS should be done smoothly without any crash

*Current Result*

The mapstore crash

- [x] Not browser related

<details><summary> <b>Browser info</b> </summary>

(use this site: <a href="https://www.whatsmybrowser.org/">https://www.whatsmybrowser.org/</a> for non expert users)

| Browser Affected | Version |
|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

MV88:
@mahmoudadel54 we have analyzed the scenario where:
- then the admin opens the map and
- changes crs from the original one to another one
- and saves the map

currently when saving the map projection might not be aligned to the crs projection creating a misalignement between custom resolutions  and projection of the map

What we should do is:

- Add the ability to define custom resolutions for a defined CRS to the CRSSelector plugin.
- When the CRS is changed, the map's resolutions must also be modified/updated. set_map_resolutions
(if custom resolutions are present in the CRSSelector plugin then use them, otherwise re-calculate them from CRS).
- When saving the map, the resolutions must be persisted (saved) along with the currently selected CRS 
- if scales are defined... 
- Remove the management of the projection parameter within mapOptions.view in the new.json file 
(we are reverting the pr anyway, we have reverted pr because it was incomplete and it was not considering saving of new resolutions, but the pr can be resumed as the logic was ok, needs some adjustment in fetching custom resolutions from CRSSelector instead of mapOptions.view.resolutions of map config)
- update documentation, including migration guidelines mentioning the fact that custom resolutions from now on has to be defined in CRSSelector instead of new.json
- at the end of the day we will ignore what present in the new.json
- for old maps that has a CRS and there is customResolutions defined in CRSSelector plugin, we have to use them, updating the map state


Scenario: 

In CRSSelector cfg we have 
```
  *     "customResolutions": {
  *       "EPSG:3003": [RES_3003],
  *       "EPSG:4326": [RES_4326]
[...]

PR #12424, opened 2026-05-22
Revert "[Backport 2026.01.xx] #11779: FIX Setting the map projection freezes MapStore causing OOM if resolutions configured in new.json file with a different projection"

tdipisa:
Reverts geosolutions-it/MapStore2#12394


MV88 (approved):
this was reverted because it was an incomplete work, for more details  see here

https://github.com/geosolutions-it/MapStore2/issues/11779#issuecomment-3654978544

PR #12451, opened 2026-06-03
About/details panel configured to show at startup does not appear consistently when map is loaded in a context  #12441

### LINKED ISSUE #12441: About/details panel configured to show at startup does not appear consistently when map is loaded in a context
## Description
When a map includes an "About" section configured to be shown at startup, this behavior does not work consistently when the map is loaded within a context.
Specifically, when opening the map for the first time in a user session from the homepage, the About panel fails to appear at startup in most cases.

## How to reproduce

TBD
- See https://qa-mapstore.geosolutionsgroup.com/mapstore/#/context/no-timeline/49226
- It is configured with "Show on startup" 

Image

*Expected Result*
- The about / details panel should appear

*Current Result*
- Observe that the About/details panel does not appear on map open

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

rowheat02:
## Description

The About panel sometimes did not open on startup, even when “Show at startup” was enabled, because the open action could run before the Details plugin was ready during slow context loading.

This PR now retry the startup open after the Details plugin is registered, and only open it if the map has details, “Show at startup” is enabled, and the panel is not already open.

To test the throttle to 3G and test the context map with “Show at startup” enabled.

https://github.com/user-attachments/assets/b9758876-b613-4b09-8a1b-3ece3814a5cd

 - [x] Bugfix

## Issue

#12441

**What is the new behavior?**

So the About panel will be opened regarless of the network speed when “Show at startup”  is enabled

## Breaking change
 - [x] No

### DIFF
--- a/web/client/epics/__tests__/details-test.js
+++ b/web/client/epics/__tests__/details-test.js
@@ -12,11 +12,13 @@ import { createEpicMiddleware, combineEpics } from 'redux-observable';
 
 import {
     closeDetailsPanelEpic,
-    fetchDataForDetailsPanel
+    fetchDataForDetailsPanel,
+    openDetailsPanelOnStartup
 } from '../details';
 
 import {
     CLOSE_DETAILS_PANEL,
+    OPEN_DETAILS_PANEL,
     closeDetailsPanel,
     openDetailsPanel,
     UPDATE_DETAILS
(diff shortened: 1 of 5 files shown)


offtherailz (changes requested):
This is a workaround that duplicates what happens in `config.js` . 
This looks to trigger an "openDetailsPanel" from a "REDUCERS_LOADED" to re-trigger the event in case the reducer was not set...

I think we should extract instead details management from `config.js` in order to handle all or part of the `openDetailsPanel` triggering (and conflict resolution with tutorial ) inside the main epic dedicated to this. 
Let's separate the concerns

rowheat02:
@offtherailz 
I updated it as suggested.  Moved the About auto open logic out of config/tutorial and centralized it in the Details epic.

offtherailz (approved):

PR #12454, opened 2026-06-03
MIP - AutoRefresh Plugin - Automatic Layer Refresh for MapStore #11887

### LINKED ISSUE #11887: AutoRefresh Plugin
# MIP <num> - AutoRefresh Plugin – Automatic Layer Refresh for MapStore

## Overview

This proposal introduces a new MapStore plugin named AutoRefresh, designed to automatically refresh selected map layers at a defined interval.
The plugin allows map publishers to configure which layers are refreshed and at which frequency, while end users can easily enable or disable the auto-refresh functionality through a persistent UI control.

### Proposed By

Julien Longchamp - arx iT

### Assigned to Release

This proposal is for MapStore <version>.

### State

* [x] Under Discussion
* [ ] In Progress
* [ ] Completed
* [ ] Rejected
* [ ] Deferred

### Motivation

Many MapStore use cases rely on near real-time or frequently updated data sources (e.g. monitoring dashboards, operational maps).
Currently, MapStore does not provide a native, map-level mechanism to automatically refresh layers without manual interaction.

The AutoRefresh plugin addresses this limitation by providing:
- Automatic refresh of selected layers
- Non-intrusive refresh behavior
- Clear and always-visible UI feedback

## Proposal

The AutoRefresh plugin introduces a UI element always visible in the map viewer, positioned in the bottom-right corner (similar to CRS selector or scale selector plugins).

The plugin provides:
- A toggle button to activate or deactivate auto-refresh
- A settings button (visible to publishers only)
- A last refresh timestamp indicator when active
- a tooltip indicating current autorefresh map settings

**Viewer Mode**
- Plugin is always visible (injected in the MapFooter using the plugin containers approach)
- Users can just enable or disable auto-refresh
- Users can see the last automatic refresh date/time + current configuration
- The toggle itself acts as the visual indicator of activation

Image

Image

**Publisher Mode**
- Additional Settings button available
- Settings panel allows selection of layers to refresh  and configuration of a refresh interval per layer (in seconds)

Image

**Layer selection**
- Layer list retrieved from the TOC
- Only layers already supporting refresh capabilities are displayed (i.e those providing a specific refresh URL)
- Layers are listed in a flat structure
- Layers list is updated when adding/removing a layer to/from the map
- Some maps could have hundreds of layers so the panel should be able to limit its own height

**Data Model Changes**
AutoRefresh settings are stored per map, not per context.
The MapStore map > layers object will be extended to include auto-refresh configuration:

Image

AutoRefreshInterval is set in ms in config.

**Refresh Behavior**
- Only layers currently visible + selected in the settings are refreshed
- Smooth refresh with no visual flickering
- Most user interactions remain uninterrupted
- Autorefresh of a specific layer will also update potential other widgets showing layer data : Counter, Chart, Table, Dashboard Map, Dashboard Legend

For this last point, it is proposed to introduce a proper action to set the refresh option to widgets. So, the implementation will require:
- An action to set the refresh property inside the widgets based on the layer name they are connected with
- Needed logic to initialize a refresh interval inside the widgets component for each type
- The list of layers inside the Autorefresh plugin should list unique layers available in the map or dashboard

**Plugin configuration (when adding to context)**
- Default refresh interval
- Minimum refresh interval

### Backwards Compatibility

No impact on existing maps, the new AutoRefreshInterval property is optional
Maps without this property remain fully compatible even id AutoRefresh plugin is added.

Note : potential conflict with https://github.com/geosolutions-it/MapStore2/issues/7289 

## Future evolution

NA

## Feedback

Update this section with relevant feedbacks, if any.

## Voting

Project Steering Committee:

* @tdipisa:
* @offtherailz: 
* @allyoucanmap:
* @MV88:
* @giohappy:

## Links

Remove unused links below.

* [Email Discussion]()
* [Pull Request]()
* [Linked Issue]()

endorria:
## Description 
This is an early stage version of the plugin to validate the approach.
Mainly works with Openlayers on few layer types (the ones who implements the `refresh`'s method).

Footer Plugin next to the CRS Plugin.

  - [x] Feature
 

## Issue

Feature not available

**What is the new behavior?**

#11887

## Breaking change
 - [x] No


tdipisa:
Thank you very much for your contribution to the MapStore codebase. We will review it as soon as possible. Many thanks.

tdipisa:
Thank you so much for this! 
As expected by [Community Bylaws](https://github.com/geosolutions-it/MapStore2/wiki#community-bylaws), we need the [CLA signed](https://github.com/geosolutions-it/MapStore2/wiki/Contributing-to-MapStore#contributor-license-agreement) to accept your contribution. Can you please provide it? Thank you again.

endorria:
> Thank you so much for this! As expected by [Community Bylaws](https://github.com/geosolutions-it/MapStore2/wiki#community-bylaws), we need the [CLA signed](https://github.com/geosolutions-it/MapStore2/wiki/Contributing-to-MapStore#contributor-license-agreement) to accept your contribution. Can you please provide it? Thank you again.

I work for the company (my GH's profile is up to date), arx iT, we already send the CLA. We can update it if necessary.

jlo-arxit:
Hi @tdipisa 
I confirm Pull Request author (Marc M.) is listed in the CLA signed by arx iT
Is it ok for you ?

tdipisa:
> Hi @tdipisa I confirm Pull Request author (Marc M.) is listed in the CLA signed by arx iT Is it ok for you ?

Yes sure. I've seen it. I'm sorry, it seems we missed to include it. Pending PR [here](https://github.com/geosolutions-it/MapStore2/pull/12576).

web/client/components/widgets/enhancers/dependenciesToAutorefresh.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { compose, withProps } from 'recompose';
+import Rx from 'rxjs';
+import { isNumber } from 'lodash';
+
+import propsStreamFactory from '../../misc/enhancers/propsStreamFactory';
```
MV88, line 13: avoid using recompose for the new development as it is going to be removed at all in the future

web/client/plugins/AutoRefresh/components/AutoRefreshForm.jsx
```diff
@@ -0,0 +1,96 @@
...
+import tooltip from '../../../components/misc/enhancers/tooltip';
+import ButtonRB from '../../../components/misc/Button';
+import Message from '../../../components/I18N/Message';
+import { AUTOREFRESH_STEP_INTERVAL_IN_SECONDS, formatDate } from '../constants';
+
+const Button = tooltip(ButtonRB);
+
+const AutoRefreshForm = ({
+    defaultRefreshInterval,
+    minRefreshInterval,
```
MV88, line 21: this is not named minimumRefreshInterval

MV88 (changes requested):
First of all thanks for this contribution Here a few feedbacks about this PR:

### Handling the auto-refresh

We noticed that you are introducing a lot of logic in order to handle the refresh inside the core components like:

- web/client/components/map/openlayers/* and other mapping library
- web/client/utils/openlayers/*  and other mapping library

we think that this might be too much and we propose a different direction you can take:

- there is already for WFS and WMS a mechanism to trigger an update which is called refreshLayerVersion

<https://github.com/geosolutions-it/MapStore2/blob/1de66c1ae5e2a3ba00fe4f658fb0bd1f2f5fea46/web/client/actions/layers.js#L222-L230>

that changes a param `_v_` dedicated to "refresh" recreate the layer instance hence updating it. So it is enough to trigger this action instead of the AUTOREFRESH_TICK  one.  This is also a more abstracted way to handle refresh since layers doesn't need to be aware of ticks and if another plugin needs to trigger a refresh we can reuse `refreshLayerVersion` logic, so instead of having a refresh method inside layers you can reuse the logic and add the missing logic on `_v_` parameter in the `update` method, therefore a cleanup here is needed, implementing this logic in the remaining layers.

All the ticks data and logic probably can be removed using derived data from layers object, triggering the action in the epic itself.

### Manage changes and different intervals

In order to manage visibility or param changes and activate/deactivate the epic you need to recollect the possible trigger actions that require a reset,
We suggest a set of `TRIGGERS` actions, and a set of timers that triggers the refresh action on each timeout.
When one Trigger changes one of the intervals, or one of the layers arrives/goes out, the timer is added/recreated/destroyed.

Here a sample code:

```js
const TRIGGERS = [
    CHANGE_LAYER_PROPERTIES,   // visibility toggle
    CHANGE_LAYER_PARAMS,
    UPDATE_NODE,               // autoRefreshInterval update
    ADD_LAYER,
    REMOVE_NODE,
    MAP_CONFIG_LOADED          // init
    AUTOREFRESH_START, // initiate the flow
    AUTOREFRESH_STOP, // stops the flow
    AUTOREFRESH_PAUSE // ADD PAUSE ACTION THAT blocks execution if when feature editor is in edit mode, this action will be triggered by the refresh plugin itself that will listen for changes to the featuregrid.mode property and trigger it when that the mode value changes like a toggle action, the pause value can be stored inside autorefresh reducer  
];

const layersAutoRefreshEpic = (action$, store) => {
    const getLayersMap = () =>
        isPaused(store.getState()) || !isActiveRefresh(store.getState()) ? {} : refreshableLayers(store.getState())
            .reduce((acc, l) => ({ ...acc, [l.id]: l.interval }), {});
    // this stream should emit refreshable layers map on every possible change, but not when paused or refesh is not active (toggled on)
    const active$ = action$.ofType(...TRIGGERS)
        .map(getLayersMap)
        .startWith(getLayersMap())
        .distinctUntilChanged(isEqual) // if some of the layer interval changed (or list change) triggers
        .publishReplay(1).refCount(); // activate multicast

    return active$
        .mergeMap(m => Object.keys(m).map(id => ({ id, interval: m[id] })))
        .groupBy(d => d.id)
        // this merges N stream actions, 1 for each layer that needs a refresh, each one with its own timer
        .mergeMap(group$ => {
            const id = group$.key;
            const removed$ = active$.filter(m => !m[id]);   // this intercepts when a layer is not in the map anymore
            return group$
                .map(d => d.interval)
                .distinctUntilChanged()                     // if change interval, rebuild the timer
                .switchMap(interval =>
                    Rx.Observable.timer(interval, interval)
                        .map(() => refreshLayerVersion(id, Date().getTime()))
                )
                .takeUntil(removed$);                        // teardown of the single timer
        });
};

```

`refreshableLayersSelector` will exclude layers being edited from attribute table or from widget wizard

### State handling

Considering that `autorefreshInterval` is stored within layers object, you don't need `availableLayers: {}`, or `activeLayers: {}` in reducer. with a dedicated selector you can fetch layers with `autorefreshInterval` defined from the `layers` state and then filter the array of layers with only the one with visibility set to true, so they can trigger the refreshLayerVersion action. Try to reuse the layers selectors

### User Interface

It is not clear from how the settings panel should look like, for example how to handle multiple layers inside a single widget? Probably not all layers inside a widget must refresh? When a layer is selected to be refreshed, (in layers state) it will receive the `autorefreshInterval` property. Considering that widgets have their own state, then also this state should receive the `autorefreshInterval` param for all the layers instances. If unselected in all places (widgets & layers state) it will remove the `autorefreshInterval` param from layer object. it would be nice to visually know if layer is coming from the map or from which widget

# Map Widgets

The **refreshLayerVersion** action trigger an update of `_v_` and by keeping track of this change we can retrigger the update on widgets. By monitoring changes to `_v_` param in [this part](https://github.com/geosolutions-it/MapStore2/pull/12454/changes#diff-f18c61e94e546ec7b93c8ac833d728705729b755fb8e2d138f16fd50d0f1d298R55) we can trigger a refresh on the widget

# Dashboards

Regarding dashboards we have some concerns on how to handle connections between widgets(dependencies managemente) even after introducing the logic to listen for changes on `_v_` param it might not be enough. Therefore we suggest to create a dedicated issue to better investigate this part after the others are done.

PR #12471, opened 2026-06-08
#12466: Enhancements to Dynamic filter functionality

### LINKED ISSUE #12466: Enhancements to Dynamic filter functionality
## Description
Enhance dynamic filter to support the following

## Acceptance criteria
- [x] Implement Map-level widget holder and configuration, similar to what we have for widgets in Dashboards. This will contain the Map-level Filter widget.
- [x] Adapt the current Filter widget to allow selecting from all layers in the map
- [x] Allow collapsing Filters (compact view) on widget
- [x] Option to add a short description for a filter, which will be visible to end users. The description will be visible as a tooltip
- [x] Option to configure if the filter should be expanded by default

### Clarification Needed
~- [ ] Quick action to allow zooming to the filtered features for the layer.~
~- [ ] Quick action to open the layer's filter panel
~  -  If the user wants to filter on another attribute not yet preconfigured as a quick filter, he can switch to the layer's filter panel of the corresponding layer~
~- [ ] Export filtered data from a filter
~  - The user has the capacity to export Quick filters results as CSV and/or shapefile, similar to the export we have inside the table widget|~

dsuren1:
## Description
This PR enhances the dynamic filter functionality by allowing users to create a global widget that enables preselection of a layer from the list of layers available on the map.

- Allows users to preselect a layer before creating the filter and reselect a different layer from the list of layers available on the map.
- Adds support for configuring a description from the Dynamic Filter layout panel.
- Allows filters to be enabled or disabled directly from the widget.
- Allows the filter to be expanded or collapsed directly from the widget.

 - [x] Enhancement

## Issue

- #12466

**What is the new behavior?**

- Users can create a global Dynamic Filter widget with a preselected layer from the list of layers available on the map.
- The selected layer can be changed at any time by choosing a different layer from the available map layers.
- A description can be configured through the Dynamic Filter Layout panel and displayed in the widget.
- Filters can be enabled or disabled directly from the widget.
- The filter panel can be expanded or collapsed directly from the widget.

  https://github.com/user-attachments/assets/b83b434e-347e-40b6-82fb-ae5d5eb78362

### Pending
- Zoom to filtered features
- Open layer filter panel for attributes modification
- Export of filtered data from the widget

## Breaking change
 - [x] No


web/client/plugins/WidgetsBuilder.jsx
```diff
@@ -146,13 +148,19 @@ export default createPlugin('WidgetsBuilder', {
...
             Component: WidgetsBuilderButton,
             position: 10
-        },
+        }, {
+            doNotHide: true,
+            name: "FilterWidgetTOC",
+            target: 'toolbar',
+            Component: FilterWidgetTOCButton,
+            position: 99
+        }],
```
allyoucanmap: Why do we have two different buttons? Could we have a single button that based on selection is changing how the widget workflow start?

web/client/utils/WidgetsUtils.js
```diff
@@ -258,8 +258,15 @@ export const getDefaultAggregationOperations = () => {
...
-export const FILTER_PROPS = ["selectedFilterId", "id", "widgetType", "filters", "selections", "dataGrid", "title", "description", "interactions"];
+export const CHART_PROPS = ["selectedChartId", "selectedTraceId", "id", "mapSync", "widgetType", "charts", "dependenciesMap", "dataGrid", "title", "description", "globalWidgetMode"];
+export const FILTER_PROPS = ["selectedFilterId", "id", "widgetType", "filters", "selections", "dataGrid", "title", "description", "interactions", "globalWidgetMode"];
+
+/**
+ * Check if layer can support a widget
+ * @param {object} layer the map layer
+ * @return {boolean} true if the layer can support a widget
+ */
+export const isWidgetLayerSupported = (layer) => !!layer?.search && layer.search !== 'vector';
```
allyoucanmap, line 269: Are the WFS layers supported for the widgets?

web/client/plugins/widgetbuilder/MapCatalogLayerSelector.jsx
```diff
@@ -0,0 +1,199 @@
...
+        });
+    },
+    getCatalogRecords: (data) => {
+        if (data && data.records) {
+            return data.records.map((layer) => ({
+                ...layer,
+                serviceType: 'maplayers',
+                title: MAP_LAYERS_API.getLayerTitle(layer),
+                description: layer.description,
+                identifier: layer.name || layer.id,
```
allyoucanmap: Probably the identifier should be the layer id, because a map may contain multiple layer with the same name

web/client/plugins/widgetbuilder/MapCatalogLayerSelector.jsx
```diff
@@ -0,0 +1,199 @@
...
+                API={{ [MAP_LAYERS_SERVICE_KEY]: MAP_LAYERS_API }}
+                title={<Message msgId="widgets.builder.wizard.selectALayer" />}
+                services={services}
+                selectedService={MAP_LAYERS_SERVICE_KEY}
+                showCatalogSelector={false}
+                multiSelect={false}
+                includeAddToMap={false}
+                locales={currentLocale}
+                selected={selected}
+                onSelect={onItemClick}
```
allyoucanmap, line 186: If we are using the id we could use the prop showIdentifier to false so it's not exposed in the UI

allyoucanmap (changes requested):
I noticed some additional inconsistencies that we should probably address now:

- creating a widget from a layer does not allow to edit and change the layer. I think this should be allowed. Creation from a layer should skip the catalog as is doing now and applying the selected layer in the filter but it should allow the modification
- creating a widget from global is working but it would be nice if we get a feedback if the layer is supported before moving to the next step (similar to dashboard)
  - open demo map #/viewer/60692
  - create a widget from global
  - select verdepubblico
  - go next step
  - not supported page without possibility to go back (It would be nice a check before going on next step as in dashboard)

web/client/utils/WidgetsUtils.js
dsuren1: Yes

dsuren1:
@allyoucanmap 
> * not supported page without possibility to go back (It would be nice a check before going on next step as in dashboard)

In the Map Viewer, unsupported layers currently redirect users to the error panel. For global widget creation, we can perform a pre-check on the layer selection page. However, this is not possible for preselected-layer widget creation because there is no layer selection step.

To keep both flows consistent, we could reuse the existing error panel logic for unsupported layers and provide a Back button so users can return to the previous step and continue the widget creation process.

Your thoughts?

image

web/client/plugins/WidgetsBuilder.jsx
dsuren1: We could, but each component serves a different purpose and is conditionally rendered. Apart from the button icon, the implementations are quite different, so I initially thought it would be better to keep them separate. That said, I've merged them now.

web/client/plugins/widgetbuilder/MapCatalogLayerSelector.jsx
dsuren1: Yeah I agree. Updated

web/client/plugins/widgetbuilder/MapCatalogLayerSelector.jsx
dsuren1: `hideIdentifier` was not properly propagated through the component hierarchy. I have fixed that and updated this as well

web/client/plugins/WidgetsBuilder.jsx
allyoucanmap: I included a commit to merge the buttons e4e277ec100053fd3643361a0c29e24c7bf92c4e, Now there was still two different components in the container configuration

allyoucanmap (approved):

ElenaGallo:
@dsuren1 When the *Default expanded* option is disabled, and after the filter widget has been created, editing the widget causes the filter to no longer be visible in the preview.


https://github.com/user-attachments/assets/86e3e52c-de79-4bfb-9003-6976c96c9424

PR #12356, opened 2026-05-08
Fix #12353 Click events not working outside the projection extent

allyoucanmap:
## Description

This PR introduces following changes:

- align behaviours of click and mouse move events
- limit the click events only when the reprojected point is not finite
- allow to trigger CHANGE_MAP_VIEW when the bbox is valid
- validate bounding box before triggering fit action

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix

## Issue

**What is the current behavior?**

#12353

**What is the new behavior?**

Allow to click outside of the projection maximum extent

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)

 - [x] No

PR #12590, opened 2026-07-07
Fix #12589 Elevation in MousePosition not shown in 2D

### LINKED ISSUE #12589: Elevation 2D regression
## Description

The altitude value shown by the MousePosition plugin (`showElevation: true`) next to the coordinates is not shown anymore in 2D when an `elevation` layer is configured.

Regression of #12356: since that change the mousemove/click handlers pass lon/lat degrees to the elevation lookup, while the elevation layer resolves the cached tile using its tile grid (map projection units) and the current map zoom as tile level: the computed key never matches the cached tiles, so the altitude is never shown. It works on releases that do not include #12356.

## How to reproduce

- Configure an `elevation` layer and MousePosition with `showElevation: true` (e.g. import the reproduction context attached to #12583, based on `sf:sfdem` on gs-stable).
- Open the map in 2D and turn on the mouse position tool (mouse button at the bottom right).
- Move the mouse over the DEM area.

*Expected Result*

`Alt: <value> m` is shown next to the coordinates while moving the mouse (e.g. `Alt: 1372 m`).

*Current Result*

No altitude is shown, even if the `application/bil16` tiles are correctly requested and loaded.

- [x] Not browser related

## Other useful information

Only master (and releases including #12356) is affected: the `2026.01.xx` based branches do not include #12356 and work correctly in 2D.

offtherailz:
## Description

This PR fixes the elevation value shown by the MousePosition plugin (`showElevation: true`) in 2D, broken by a regression of #12356.

The mousemove/click handlers passed lon/lat degrees to the elevation lookup, while the elevation layer resolves the tile in tile grid units, using the map zoom as tile level. The handlers now pass the native map coordinate, and the layer transforms it to the tile grid projection and resolves the value from the tile levels actually requested by the renderer, instead of relying on the map zoom (this also makes the lookup work when the tile grid projection differs from the map one).

 - [x] Bugfix

## Issue

#12589

**What is the new behavior?**

The altitude is shown by the MousePosition plugin while moving the mouse in 2D, for maps with a configured `elevation` layer.

## Breaking change
 - [x] No

## Other useful information

No backport needed: the `2026.01.xx` based branches do not include #12356 and are not affected.

### DIFF
--- a/web/client/components/map/openlayers/Map.jsx
+++ b/web/client/components/map/openlayers/Map.jsx
@@ -179,7 +179,7 @@ class OpenlayersMap extends React.Component {
             if (this.props.onClick && !this.map.disabledListeners.singleclick) {
                 const latLng = this.coordinateToLatLng(event.coordinate);
                 if (!latLng) return;
-                const { pos, lat, lng } = latLng;
+                const { lat, lng } = latLng;
                 /*
                  * Handle special case for vector features with handleClickOnLayer=true
                  * Modifies the clicked point coordinates to center the marker and sets the layerInfo for
@@ -213,7 +213,7 @@ class OpenlayersMap extends React.Component {
                     latlng: {
                         lat: finalLat,
                         lng: finalLng,
-                        z: this.getElevation(pos, event.pixel)
+                        z: this.getElevation(event.coordinate, event.pixel)
                     },
                     rawPos: event.coordinate.slice(),
                     modifiers: {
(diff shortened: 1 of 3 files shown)


dsuren1 (approved):

dsuren1:
@ElenaGallo Kindly test it in DEV when available. Thanks!

offtherailz:
- #12601

PR #12591, opened 2026-07-07
Fix #12583 Elevation in MousePosition not shown in 3D

### LINKED ISSUE #12583: Elevation in mouseposition not shown in 3D for some special cases
## Description

## How to reproduce

*Expected Result*

*Current Result*

- [x] Not browser related

## Other useful information

offtherailz:
## Description

This PR fixes the elevation value shown by the MousePosition plugin (`showElevation: true`) in 3D, when an `elevation` layer is configured.

The Cesium elevation layer was created but never added to the map imagery layers: the async proxy detection in `Layer.jsx` always delegated to the update flow, and the `elevation` layer type does not recreate the layer when nothing changed, so it was never added and its tiles were never requested. Now, after the detection, the already created layer is directly added when its proxy setup is correct, and it is recreated through the update flow only when the detection requires a different one (the `elevation` type now also recreates on `forceProxy` change).

An explicit `forceProxy: true` in the layer options was also overridden by the detection result: the explicit option now takes precedence, so the proxy can be forced by configuration for servers that reject browser requests with an HTTP error while exposing valid CORS headers.

 - [x] Bugfix

## Issue

#12583

**What is the new behavior?**

The altitude is shown by the MousePosition plugin while moving the mouse in 3D, for maps with a configured `elevation` layer.

## Breaking change
 - [x] No

## Other useful information

The affected code is also present in the `2026.01.xx` based branches: the fix applies cleanly there and should be backported. Verified with a WMS `application/bil16` DEM (`sf:sfdem` on gs-stable) used both as elevation layer and as WMS terrain; a reproduction context is attached to the issue.

### DIFF
--- a/web/client/components/map/cesium/Layer.jsx
+++ b/web/client/components/map/cesium/Layer.jsx
@@ -341,14 +341,14 @@ class CesiumLayer extends React.Component {
             {
                 ...newProps.options,
                 securityToken: newProps.securityToken,
-                forceProxy: this._isProxy,
+                forceProxy: newProps.options?.forceProxy || this._isProxy,
                 imageryLayersTreeUpdatedCount: newProps.imageryLayersTreeUpdatedCount,
                 position: newProps.position
             },
             {
                 ...oldProps.options,
                 securityToken: oldProps.securityToken,
-                forceProxy: this._prevIsProxy,
+                forceProxy: oldProps.options?.forceProxy || this._prevIsProxy,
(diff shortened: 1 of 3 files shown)


dsuren1 (approved):

dsuren1:
@offtherailz Some tests are failing in node v20 gha. Kindly check. Thanks

dsuren1:
@ElenaGallo Kindly test it in DEV when available. Thanks!

offtherailz:
- #12603

PR #12581, opened 2026-07-03
Fixed #12579 - Fixed terrain switch and black on load issues

offtherailz:
## Description

Fixed #12579 - Fixed terrain switch and black on load issues

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix
 - [ ] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [ ] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**

Fix #12579

**What is the new behavior?**

Tilt the map to see better the result 

image

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #12592, opened 2026-07-07
Backport c040 issues with 3D terrain switch and elevation in 3D mode

### LINKED ISSUE #12579: 3D map renders completely black when the map contains a terrain layer
## Description

When a map contains a background layer of `"type": "terrain"` (any provider: `wms`, `cesium`, `cesium-ion`), the Cesium 3D view can render completely black: no globe, no imagery, no terrain. No errors in the browser console and no tile requests are issued, so the failure is completely silent. The 2D view of the same map works correctly.

A second symptom of the same defect: switching the terrain from the background selector never applies the selected terrain — the scene stays on the flat Ellipsoid terrain.

The issue happens more frequently from 2026.01.xx, when the terrain layer is re-created for some reason (e.g. authenticationRules etc...); before 2026.01.xx it looks more rare.

<details><summary>Root cause</summary>
<p>

`createLayer` in `web/client/components/map/cesium/plugins/TerrainLayer.js` executes `map.terrainProvider = undefined;`. The Cesium `Scene.terrainProvider` setter cancels the pending ready-listener registered by a previous `scene.setTerrain()` call, so any `createLayer` execution not followed by a new `add()` kills the terrain currently applied (or pending) and the globe never renders. Extra `createLayer` executions always happen because:

1. `BackgroundSelector` automatically injects a hidden "Ellipsoid" terrain layer at runtime, and `CesiumLayer.componentDidMount` calls `createLayer` regardless of visibility;
2. `updateLayer` re-creates the layer on any watched prop change (`securityToken`, `security`, `requestRuleRefreshHash`, ...).

The final outcome depends on the order of these events relative to the last `setTerrain`, which is why the same map can work on one instance/flow and be black on another.

The terrain-switch symptom has the same root: `remove()` of the deselected terrain layer unconditionally resets the scene terrain and runs after the `add()` of the newly selected one, cancelling its pending application.

Regression introduced by the Cesium 1.131 upgrade (#11103/#11331, first in 2025.02.00), where the provider became a promise applied via `scene.setTerrain()`; the `map.terrainProvider = undefined;` line predates it and was harmless with the old synchronous assignment. #11648 and #12346 (2026.01.xx) added new `updateLayer` recreation triggers, making the bug much more frequent.

</p>
</details>

### 1. Black 3D scene on map load

1. Create a new map
2. Import (Import plugin, drag & drop) the file `terrain-black-3d-repro.json`

[terrain-black-3d-repro.json](https://github.com/user-attachments/files/29630502/terrain-black-3d-repro.json)

3. The map opens directly in 3D mode → the whole map area is black, no console errors, no terrain/imagery tile requests
4. Counter-check: open the background selector and switch the terrain to "Ellipsoid" → the map renders again

### 2. Terrain switch never applies the selected terrain

1. Create a new map
2. Import the file `terrain-switch-repro.json` (Ellipsoid is the active terrain, a quantized-mesh terrain is configured but not active — the map renders correctly on load)

[terrain-switch-repro.json](https://github.com/user-attachments/files/29639866/terrain-switch-repro.json)

3. Open the background selector → "Terrains" → select "swiss terrain"
4. The radio button updates but the globe stays flat: `layer.json` is requested, no `.terrain` tile request is ever issued. Switching back and forth never applies the terrain

Reproduced on https://dev-mapstore.geosolutionsgroup.com/mapstore/ (master).

*Expected Result*

The 3D globe renders with the configured imagery and terrain (or degrades gracefully to a flat globe if the terrain endpoint is unreachable). Switching terrains from the background selector always applies the selected terrain.

*Current Result*

The whole 3D scene is black. Terrain switching gets stuck on Ellipsoid. A terrain provider load failure (unreachable server, CORS) also leaves the scene black instead of falling back.

- [x] Not browser related

## Other useful information

Proposed fixes (verified locally, each covered by a regression test failing on the current code):

1. Remove the `map.terrainProvider = undefined;` line in `createLayer`: the terrain lifecycle is already fully handled by `scene.setTerrain()` in `add()`/`remove()`.
2. Make `remove()` reset the scene terrain only when the removed layer is the currently applied one.
3. Listen to `Terrain.errorEvent` in `add()` and fall back to the ellipsoid terrain with an explicit console error when the provider fails to load.
4. `proxy.properties`: add `(.*layer\.json.*)|(.*\.terrain.*)` to `reqtypeWhitelist.generic` and `application/vnd.quantized-mesh` to `mimetypeWhitelist`, so `forceProxy: true` becomes usable on quantized-mesh terrain layers (today the proxy replies 403 on `.terrain` tiles). Note: the current `reqtypeWhitelist.generic` also misses a `|` between `(.*\.ifc.*)` and `(.*3dtiles.*)`.

Related findings: `OSMLayer.js` still points to the deprecated `//a.tile.openstreetmap.org/` subdomain (CORS failure in Cesium, OSM imagery not displayed in 3D); the proxy-detection request in `CesiumLayer.addLayer` has no rejection handler (unhandled `AxiosError` on CORS-protected services).

### LINKED ISSUE #12583: Elevation in mouseposition not shown in 3D for some special cases

offtherailz:
In draft, waiting for approval

## Description

Backport to `c040-2026.01.xx` of the following fixes:

- #12581 
    - **Fix #12579 Terrain switch and black on load issues**: keeps the active terrain when a terrain layer is recreated on update or a deselected terrain layer is removed, and falls back to the ellipsoid terrain when the terrain provider fails to load (e.g. unreachable server or CORS error), avoiding a completely black globe.
- https://github.com/geosolutions-it/MapStore2/pull/12591 
    - **Fix #12583 Elevation in MousePosition not shown in 3D**: the elevation layer is now added to the Cesium imagery layers after the proxy detection, it is recreated when the detection requires a different proxy setup, and an explicit `forceProxy` option in the layer configuration takes precedence over the detection.

The 2D elevation fix (#12589) is not included on purpose: this branch does not contain #12356, so it is not affected by that regression.

 - [x] Bugfix

## Issue

#12579, #12583

**What is the new behavior?**

Terrain switch works correctly and the globe no longer turns black when the terrain provider fails to load; the elevation is shown by the MousePosition plugin in 3D.

## Breaking change
 - [x] No

### DIFF
--- a/web/client/components/map/cesium/Layer.jsx
+++ b/web/client/components/map/cesium/Layer.jsx
@@ -448,6 +448,13 @@ class CesiumLayer extends React.Component {
         if (this._isProxy === undefined && newProps?.options?.url) {
             const urls = castArray(newProps.options.url);
             return axios(urls[0], { noProxy: true })
+                // this request is used only to detect if the layer needs the proxy,
+                // a failure (e.g. CORS error) is expected and must not produce an unhandled rejection
+                // preventing the error avoids an unhandled rejection in console.
+                .catch(() => {
+                    console.warn(`Failed to detect proxy for ${urls[0]}. This is expected if the server does not support CORS.`);
+                    return null;
+                })
                 .finally(() => {
                     this._isProxy = !!getProxyCacheByUrl(urls[0]);
                     this.updateLayer(newProps, this.props);
--- a/web/client/components/map/cesium/Layer.jsx
+++ b/web/client/components/map/cesium/Layer.jsx
@@ -341,14 +341,14 @@ class CesiumLayer extends React.Component {
             {
                 ...newProps.options,
                 securityToken: newProps.securityToken,
-                forceProxy: this._isProxy,
+                forceProxy: newProps.options?.forceProxy || this._isProxy,
                 imageryLayersTreeUpdatedCount: newProps.imageryLayersTreeUpdatedCount,
                 position: newProps.position
             },
             {
                 ...oldProps.options,
                 securityToken: oldProps.securityToken,
-                forceProxy: this._prevIsProxy,
+                forceProxy: oldProps.options?.forceProxy || this._prevIsProxy,
                 imageryLayersTreeUpdatedCount: oldProps.imageryLayersTreeUpdatedCount,
                 position: oldProps.position
             },
@@ -457,7 +457,17 @@ class CesiumLayer extends React.Component {
                 })
                 .finally(() => {
                     this._isProxy = !!getProxyCacheByUrl(urls[0]);
-                    this.updateLayer(newProps, this.props);
+                    // the layer was created with the forceProxy value coming from its options,
+                    // recreate it through updateLayer only when the detection requires a different proxy setup,
+                    // otherwise the already created layer must simply be added to the map
+                    const createdWithProxy = !!newProps?.options?.forceProxy;
+                    const needsProxy = createdWithProxy || this._isProxy;
+                    if (createdWithProxy === needsProxy) {
+                        this._addLayer(newProps);
+                        this.updateZIndex(newProps.position);
+                    } else {
+                        this.updateLayer(newProps, this.props);
+                    }
                     this._prevIsProxy = this._isProxy;
                 });
         }
--- a/web/client/components/map/cesium/__tests__/Layer-test.jsx
+++ b/web/client/components/map/cesium/__tests__/Layer-test.jsx
@@ -1896,6 +1896,51 @@ describe('Cesium layer', () => {
         expect(cmp.layer).toBeTruthy();
         expect(cmp.layer.getElevation).toBeTruthy();
     });
+    it('should add the elevation layer to the map imagery layers after the proxy detection', (done) => {
+        const options = {
+            type: 'elevation',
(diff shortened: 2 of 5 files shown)


tdipisa:
@offtherailz as far I understood from the above description you should now backport only the following to the [c040-2026.01.xx](https://github.com/geosolutions-it/MapStore2/tree/c040-2026.01.xx) branch:

https://github.com/geosolutions-it/MapStore2/pull/12591

offtherailz:
> @offtherailz as far I understood from the above description you should now backport only the following to the [c040-2026.01.xx](https://github.com/geosolutions-it/MapStore2/tree/c040-2026.01.xx) branch:
> 
> #12591
@tdipisa :  There are 3 bugs on master: 
- https://github.com/geosolutions-it/MapStore2/issues/12579
- https://github.com/geosolutions-it/MapStore2/issues/12583
- https://github.com/geosolutions-it/MapStore2/issues/12589

Backporting only 2 of them because the 3rd (#12589) is due to a recent regression, not present on this branch

dsuren1 (approved):

PR #12603, opened 2026-07-08
[Backport 2026.02.xx] Fix #12583 Elevation in MousePosition not shown in 3D

### LINKED ISSUE #12583: Elevation in mouseposition not shown in 3D for some special cases

offtherailz:
# Description
Backport of #12591 to `2026.02.xx`.

Fixes #12583


tdipisa:
@offtherailz there is a failing check here. Can you please verify and fix? Thank you.

dsuren1 (approved):

PR #12669, opened 2026-07-16
CRS change crashes with grid-based projections when layer resolution limits are configured  #12627

### LINKED ISSUE #12627: CRS change crashes with grid-based projections when layer resolution limits are configured
## Description

Changing the map CRS can crash when the selected CRS uses a datum grid and a layer has `minResolution` or `maxResolution`.

The problem is caused by `convertResolution(`web/client/utils/MapUtils.js`)`, which transforms a random point to the target CRS and then passes it to `getPointResolution` together with the source projection. The point is therefore interpreted in the wrong CRS, and the unnecessary transformation can fail when it falls outside the datum grid coverage.

Some errors observerd:
```
mapstore2.js?c97a1ffaf62e2a5c:2 Uncaught TypeError: coordinates must be finite numbers
    at Ce (
```
```
mapstore2.js?c97a1ffaf62e2a5c:2 Uncaught TypeError: cannot destructure property 'zoom'
```

## How to reproduce

- Configure a CRS that uses a datum grid.
- Add a layer with `minResolution` or `maxResolution`.
- Change the map to the configured CRS.

*Expected Result*

The map CRS changes successfully and the layer resolution limits are updated.

*Current Result*

The CRS change fails with a transformation error.

- [x] Not browser related

rowheat02:
## Description

This PR fixes the CRS change when the target projection uses a datum grid and some layers have `minResolution` or `maxResolution` configured.

The resolution conversion was transforming a point to the target CRS and then using it with the source projection. This could generate invalid coordinates and crash the CRS change when the target projection requires a datum-grid correction.

While fixing this, we also found an issue when switching the CRS back and forth. For example:

`EPSG:25830 → EPSG:4326 → EPSG:25830`

The converted resolution limits were snapped to the resolution list, so they did not return to their original values. Because of this, layers that were visible before the CRS change could become invisible after switching back.

The limits are now converted using the meters-per-unit ratio without snapping, making the conversion reversible.

 - [x] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes

## Issue

#12627
The CRS change can fail with datum-grid projections when layer resolution limits are configured. Switching the CRS back and forth can also change the limits and make visible layers invisible.

**What is the new behavior?**

The CRS changes without crashing, and layer resolution limits remain stable after multiple CRS changes.

## Breaking change
 - [x] No


tdipisa:

web/client/utils/MapUtils.js
```diff
@@ -24,7 +24,7 @@ import {
     omit,
     isObject
 } from 'lodash';
-import { get as getProjectionOL, getPointResolution, transform } from 'ol/proj';
+import { get as getProjectionOL, getPointResolution } from 'ol/proj';
```
offtherailz: Needed by the suggestion below.

```suggestion
import { get as getProjectionOL, getPointResolution, transform } from 'ol/proj';
```

web/client/utils/MapUtils.js
```diff
@@ -532,28 +532,29 @@ export function getRandomPointInCRS(crs) {
...
 
-    // Transform the resolution
+    // `point` must stay in sourceProjection's own space
     const transformedResolution = getPointResolution(
         sourceProjection,
         sourceResolution,
-        transform(randomPoint, sourceCRS, targetCRS),
+        point,
         targetProjection.getUnits()
     );
```
offtherailz: `convertResolution` does half the conversion.

A resolution is map units per pixel, and one unit is not the same physical length in every CRS. Converting one takes two steps:

1. source resolution to real ground metres per pixel, which is what `getPointResolution(sourceProjection, ...)` returns;
2. ground metres per pixel to target CRS units, dividing by how many real metres one target unit is worth at that location.

Step 2 is missing, so the ground value comes back as if it already were a target resolution. Its divisor is close to 1 wherever the target projection is barely distorted, which is the case for EPSG:3857 to UTM, and that is why the reported bug is fixed. Two cases where it is not, both readable straight from the OpenLayers source.

EPSG:3857 as target. `ol/proj/epsg3857.js` defines the point resolution as `resolution / cosh(point[1] / RADIUS)`, which for Mercator is `resolution * cos(lat)`. So one 3857 unit is worth `cos(lat)` real metres, and skipping the division stores `cos(lat)` times the correct value: 79% of it at latitude 38, 71% at 45, half of it at 60. The value is then snapped to the nearest zoom level, so it can land on the wrong one.

EPSG:4326 as target. `getPointResolution` returns early when the requested unit is `degrees` and the source projection has no `getPointResolutionFunc`, which is the case for every CRS registered through proj4:

```js
if ((projUnits == 'degrees' && !units) || units == 'degrees') {
    pointResolution = resolution;
```

The resolution comes back untouched, so a map moving from a metric national grid to EPSG:4326 stores around `1003` degrees per pixel where `0.0106` is correct. After the snap, the layer visibility limits are gone. This one is not a precision problem.

Both cases are covered by the tests in the suggestion further down, which fail on this branch with `Expected 0.29358461905046174 to be less than 0.01` and `Expected 2000 to be less than 0.03`.

The suggestion below adds step 2. It brings `transform` back, but on the map center rather than a point forced to the equator, and if proj4 cannot resolve the datum grid it returns `[NaN, NaN]`, so that case falls back to the current expression and your regression tests still pass.

```suggestion
    const point = anchorPoint || getRandomPointInCRS(sourceCRS);

    // getPointResolution expects the point in the space of the projection passed as first argument,
    // so `point` stays in sourceCRS: measuring the source resolution needs no target transform
    const groundResolution = getPointResolution(sourceProjection, sourceResolution, point, 'm');

    // ground meters covered by one targetCRS unit at the same location
    let metersPerTargetUnit;
    try {
        const targetPoint = transform(point, sourceCRS, targetCRS);
        if (targetPoint.every((value) => Number.isFinite(value))) {
            metersPerTargetUnit = getPointResolution(targetProjection, 1, targetPoint, 'm');
        }
    } catch (e) {
        metersPerTargetUnit = undefined;
    }

    const transformedResolution = metersPerTargetUnit > 0
        ? groundResolution / metersPerTargetUnit
        // targetCRS not reachable from this point: convert by units only, ignoring its local distortion
        : getPointResolution(sourceProjection, sourceResolution, point, targetProjection.getUnits());
```

web/client/actions/map.js
```diff
@@ -91,42 +92,59 @@ export const changeCRS = (crs) => ({
...
+            const anchor = anchorPoint && [anchorPoint.x, anchorPoint.y];
+            layersWithLimits.forEach(layer => {
+                const options = {};
+                if (layer.minResolution) {
+                    const { transformedResolution } = convertResolution(sourceCRS, crs, layer.minResolution, anchor);
+                    options.minResolution = snapToResolution(transformedResolution, newResolutions);
+                }
+                if (layer.maxResolution) {
+                    const { transformedResolution } = convertResolution(sourceCRS, crs, layer.maxResolution, anchor);
+                    options.maxResolution = snapToResolution(transformedResolution, newResolutions);
```
offtherailz: `snapToResolution` returns `undefined` when the conversion is not finite, and that `undefined` still gets written into the layer. The reducer does `{...layer, ...options}` (`LayersUtils.js:1155`), so the visibility limit disappears.

With the `convertResolution` suggestion applied this branch is hard to reach, since the fallback always returns a finite number, so it is a safety net rather than a live bug. Still worth having: a deleted limit looks exactly like a layer that never had one.

Keeping the stale value is a compromise too, since it is a number in the previous CRS units, but it stays visible and can be fixed from the TOC.

```suggestion
                if (layer.minResolution) {
                    const { transformedResolution } = convertResolution(sourceCRS, crs, layer.minResolution, anchor);
                    const snapped = snapToResolution(transformedResolution, newResolutions);
                    if (snapped === undefined) {
                        console.warn(`Cannot convert minResolution of layer ${layer.id} to ${crs}, keeping the current value`);
                    } else {
                        options.minResolution = snapped;
                    }
                }
                if (layer.maxResolution) {
                    const { transformedResolution } = convertResolution(sourceCRS, crs, layer.maxResolution, anchor);
                    const snapped = snapToResolution(transformedResolution, newResolutions);
                    if (snapped === undefined) {
                        console.warn(`Cannot convert maxResolution of layer ${layer.id} to ${crs}, keeping the current value`);
                    } else {
                        options.maxResolution = snapped;
                    }
```

web/client/utils/__tests__/MapUtils-test.js
```diff
@@ -2437,6 +2439,16 @@ describe('Test the MapUtils', () => {
     it('convertResolution', () => {
         expect(convertResolution('EPSG:3857', 'EPSG:4326', 2000).transformedResolution).toBe(0.017986440587896155);
```
offtherailz: If you take the suggestion above, this hard-coded expected value changes in the 7th decimal and the test fails, so it needs updating in the same commit.

The number itself carries no meaning: it is the same conversion, only the meters-per-degree now comes from an OpenLayers great-circle measurement (111195.08) instead of its `METERS_PER_UNIT.degrees` constant (111194.87), which differ by ~2 parts per million.

```suggestion
        expect(convertResolution('EPSG:3857', 'EPSG:4326', 2000).transformedResolution).toBe(0.01798640727449076);
```

web/client/utils/__tests__/MapUtils-test.js
```diff
@@ -2437,6 +2439,16 @@ describe('Test the MapUtils', () => {
...
+    it('convertResolution does not fail when the target CRS relies on an unavailable datum grid', () => {
+        // simulates a datum grid (e.g. NTv2/gsb) that can't be resolved for the given point
+        proj4.defs('TEST:6265-GRID-CRS', '+proj=utm +zone=30 +ellps=GRS80 +units=m +no_defs +nadgrids=unavailable-grid-6265');
+        register(proj4);
+
+        const result = convertResolution('EPSG:3857', 'TEST:6265-GRID-CRS', 2000);
+        expect(result).toExist();
+        expect(isNaN(result.transformedResolution)).toBe(false);
+        expect(isFinite(result.transformedResolution)).toBe(true);
+    });
```
offtherailz, line 2486: Two tests for the conversion itself. The first pins the ground size across a metric CRS and Mercator in both directions, the second checks the target unit is honoured. On this branch they fail with `Expected 0.29358461905046174 to be less than 0.01` and `Expected 2000 to be less than 0.03`; with the `convertResolution` suggestion applied they pass.

The expected values are derived in the test rather than hard-coded, so they stay readable.

```suggestion
    });
    it('convertResolution keeps the same size on the ground when converting to a Mercator CRS', () => {
        proj4.defs('TEST:UTM30N', '+proj=utm +zone=30 +ellps=GRS80 +units=m +no_defs');
        register(proj4);
        const lat = 45;
        const utmPoint = proj4('EPSG:4326', 'TEST:UTM30N', [-1.13, lat]);
        const mercatorPoint = proj4('EPSG:4326', 'EPSG:3857', [-1.13, lat]);
        // one UTM metre is a real metre, one EPSG:3857 metre is cos(lat) real metres,
        // so the same pixel size on the ground needs a 1 / cos(lat) larger resolution in EPSG:3857
        const expected = 2000 / Math.cos(lat * Math.PI / 180);

        const mercatorResolution = convertResolution('TEST:UTM30N', 'EPSG:3857', 2000, utmPoint).transformedResolution;
        expect(Math.abs(mercatorResolution / expected - 1)).toBeLessThan(0.01);

        const back = convertResolution('EPSG:3857', 'TEST:UTM30N', mercatorResolution, mercatorPoint).transformedResolution;
        expect(Math.abs(back / 2000 - 1)).toBeLessThan(0.01);
    });
    it('convertResolution returns degrees when the target CRS is geographic', () => {
        proj4.defs('TEST:UTM30N', '+proj=utm +zone=30 +ellps=GRS80 +units=m +no_defs');
        register(proj4);
        const utmPoint = proj4('EPSG:4326', 'TEST:UTM30N', [-1.13, 45]);
        // 2000 metres is roughly 0.02 degrees at this latitude
        const result = convertResolution('TEST:UTM30N', 'EPSG:4326', 2000, utmPoint).transformedResolution;
        expect(result).toBeGreaterThan(0.01);
        expect(result).toBeLessThan(0.03);
    });
```

offtherailz (changes requested):
Good,
Two things before merge in inline suggestions you can apply:
1) guard + warning if conversion fails
2) fix to the unit coneversion  (if passing, for instance, from meeters to degrees etc). I attached a test. Please verify that makes sense to you too.

Could you please verify that the changes I suggested still fits our use case and they work well also in different latitude with inverse conversion @rowheat02 ? thank you.

offtherailz:
Pushed a commit on top of yours with the fix for the problem you found.

The conversion itself was fine after the suggestions; what broke the round trip was the snap.
`snapToResolution` used `getResolutions(targetCRS)`, and that returns whatever is in
`mapOptions.view.resolutions` for *any* SRS because
`updateMapResolutionsOnCrsChangeEpic` writes that array with `view.projection` stripped. Since the
thunk runs before that epic, on the way back the limits in metres were snapped onto the previous
CRS ladder in degrees: `maxResolution` became ~0.703125 metres, i.e. the layer only showed up below
1:2657. Snapping between two ladders is also a ratchet, so the limits shrank on every switch.

So the limits are now converted through the meters-per-unit ratio of the two CRS - the same ratio
`closestMatchedZoom` uses to keep the map scale, which is what keeps the layer's relation with the
view resolution - and they are no longer snapped, so the conversion is reversible. No point
transform is involved anymore, which also removes the original #12627 crash at the root.

What to test:
- a map in EPSG:25830 with a layer having min/maxResolution, switch to EPSG:4326 and back: the
  layer stays visible and the limits return to the original values (repeat a few times, they must
  not drift);
- the Visibility limits panel in the layer settings must keep showing the same scale denominators
  in every CRS (the limit is a scale, only its unit changes);
- after the round trip, zooming out past the max scale must still hide the layer, so the threshold
  is still where it was.

Two pre-existing problems came up while testing this, unrelated to the PR; I will open them
separately: the map scale drifts one zoom level per CRS switch, and `mapOptions.view.resolutions`
is saved without `view.projectis lying for any other caller.

offtherailz (approved):

offtherailz:
Backport failed for `2026.02.xx`: couldn't find remote ref `2026.02.xx`.
Please ensure that this GitHub repo has a branch named `2026.02.xx`.

allyoucanmap:
@rowheat02 the automatic backport failed, please provide a new backport

rowheat02:
Backport for 2026.02.xx has been created manually: https://github.com/geosolutions-it/MapStore2/pull/12762

PR #12691, opened 2026-07-20
Fix #12689 - Remove unnecessary src=null on WMS tile load error

### LINKED ISSUE #12689: WMS tile load error handler sets img.src = null, causing a spurious network request to a literal "null" URL
## Description
In the custom authenticated tile `loadFunction` (`WMSLayer.js`), when a tile fetch fails, the error handler sets `image.getImage().src = null` before calling `image.setState(3)`. Assigning `null` to an `` element's `src` property does not clear it — the DOM coerces it to the string `"null"`, so the browser issues a real GET request to a relative URL literally named `null` (e.g. `https://yourhost/null`), which always 404s. This happens once per failed tile load, adding console/network noise with no benefit.

## How to reproduce
- Configure a WMS layer using the authenticated request path with custom headers (`requestsConfigurationRules` / `security` matching a rule with `headers`), so tile loading goes through the `axios.get(..., {headers, responseType:'blob'})` branch of `loadFunction`.
- Make a tile request fail (e.g. an invalid/expired token, or a layer with an OGC exception response).
- Watch the Network tab: alongside the failed WMS request, a second request to a URL ending in `/null` appears and 404s.

*Expected Result*
No additional request should be made when a tile fails to load; only the original failing request should appear in the network log.

*Current Result*
An extra request to `/null` fires for every failed tile, purely as a side effect of setting `img.src = null`.

- [x] Not browser related

|---|---|
|Chrome| 150 (standard DOM behavior, not browser-specific) |
</details>

## Other useful information
Relevant code: `web/client/components/map/openlayers/plugins/WMSLayer.js`:
```js
}).catch(errorMessage => {
    image.getImage().src = null;   // needed to trigger the MS imageloaderror event in Map.onLayerError
    image.setState(3);
    failTiles.add(src);
    console.error(errorMessage);
});
```
The comment states this line is needed to trigger the tile's error signal to the rest of the app (`Map.onLayerError`). This is not actually necessary: `image.setState(3)` (the very next line) already sets OpenLayers' internal tile/image state to `ERROR`, and OpenLayers' own source classes (`ol/source/UrlTile.js` `handleTileChange`, `ol/source/Image.js` `handleImageChange`) dispatch their `tileloaderror`/`imageloaderror` events purely based on that internal state, synchronously — with no dependency on the `` element's native `error` DOM event. Verified locally: removing the `src = null` line eliminates the `/null` requests while `onLayerError` still fires correctly.

Suggested fix: delete the `image.getImage().src = null;` line; nothing else needs to change. PR to follow.

offtherailz:
Closes #12689

Removes `image.getImage().src = null;` from the tile load error handler in `WMSLayer.js`. Assigning `null` to an ``'s `src` coerces to the literal string `"null"`, causing a real (always-404) request to a relative URL named `null`.

The line immediately after it, `image.setState(3)`, already sets OpenLayers' internal tile/image state to `ERROR`. Traced through `ol/source/UrlTile.js` (`handleTileChange`) and `ol/source/Image.js` (`handleImageChange`): both dispatch `tileloaderror`/`imageloaderror` purely based on that internal state, synchronously, with no dependency on the `` element's native `error` DOM event. So the `src = null` line isn't needed for the error signal MapStore listens for (`Map.onLayerError`) — it's redundant, and its only effect is the spurious network request.

Verified locally: removing the line eliminates the `/null` requests while the layer's error state and `onLayerError` still propagate correctly.

### DIFF
--- a/web/client/components/map/openlayers/__tests__/Layer-test.jsx
+++ b/web/client/components/map/openlayers/__tests__/Layer-test.jsx
@@ -357,6 +357,40 @@ describe('Openlayers layer', () => {
             }, 200);
         });
     });
+    it('render wms singleTile layer with error does not request a spurious "null" url', (done) => {
+        mockAxios.onGet().reply(r => {
+            expect(r.url.indexOf('SAMPLE_URL') >= 0 ).toBeTruthy();
+            return [200, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
+            "<ows:ExceptionReport xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">\n" +
+            "  <ows:Exception exceptionCode=\"InvalidParameterValue\" locator=\"srsname\">\n" +
+            "    <ows:ExceptionText>msWFSGetFeature(): WFS server error. Invalid GetFeature Request</ows:ExceptionText>\n" +
+            "  </ows:Exception>\n" +
+            "</ows:ExceptionReport>"];
+        });
+        const options = {
+            type: 'wms',
+            visibility: true,
+            singleTile: true,
+            url: 'SAMPLE_URL',
+            name: 'osm:vector_tile'
+        };
+        const layer = ReactDOM.render(<OpenlayersLayer
+            type="wms"
+            options={{
+                ...options
+            }}
+            map={map} />, document.getElementById("container"));
+        expect(layer.layer.getSource()).toBeTruthy();
+        layer.layer.getSource().on('imageloaderror', (e) => {
+            setTimeout(() => {
+                // the error handler must not set img.src = null, which the DOM coerces
+                // to the literal string "null" and resolves into a spurious request
+                const img = e.image.getImage();
+                expect(img.getAttribute('src') === 'null' || (img.src || '').endsWith('/null')).toBe(false);
+                done();
+            }, 200);
+        });
+    });
     it('creates a tiled wms layer for openlayers map with long url', (done) => {
         let options = {
             "type": "wms",
--- a/web/client/components/map/openlayers/__tests__/Layer-test.jsx
+++ b/web/client/components/map/openlayers/__tests__/Layer-test.jsx
@@ -40,6 +40,8 @@ import { addElevationTile, getElevationKey } from '../../../../utils/ElevationUt
 
 
 import { Map, View } from 'ol';
+import ImageState from 'ol/ImageState';
+import TileState from 'ol/TileState';
 import VectorSource from 'ol/source/Vector';
 import { defaults as defaultControls } from 'ol/control';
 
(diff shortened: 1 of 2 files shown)


tdipisa:

web/client/components/map/openlayers/plugins/WMSLayer.js
```diff
@@ -96,8 +96,7 @@ const loadFunction = (options, headers) => function(image, src) {
                         throw new Error(response.dataText);
                     }
                 }).catch(errorMessage => {
-                    image.getImage().src = null;   // needed to trigger the MS imageloaderror event in Map.onLayerError
```
allyoucanmap, line 99: The src equal to `null` was there to cover the single tile case that uses a different image class (`ol/Image`) that does not include the setState function as the tile image class (`ol/ImageTile`). Anyway the null should be removed and a proposal is to include a separate function to handle both cases:

```js
import ImageState from 'ol/ImageState';

// ...

const setErrorState = (image) => {
    if (image.setState) {
        image.setState(ImageState.ERROR);
        return;
    }
    image.state = ImageState.ERROR;
    image.changed();
};
```

then replace image.setState in the load function:

```diff
- image.setState(3);
+ setErrorState(image);
```

there are multiple image.setState in the custom loadFunction and I think we need to update them all

web/client/components/map/openlayers/__tests__/Layer-test.jsx
```diff
@@ -357,6 +357,40 @@ describe('Openlayers layer', () => {
...
+        layer.layer.getSource().on('imageloaderror', (e) => {
+            setTimeout(() => {
+                // the error handler must not set img.src = null, which the DOM coerces
+                // to the literal string "null" and resolves into a spurious request
+                const img = e.image.getImage();
+                expect(img.getAttribute('src') === 'null' || (img.src || '').endsWith('/null')).toBe(false);
+                done();
+            }, 200);
+        });
+    });
```
allyoucanmap, line 466: This test was not reaching the `if (headers) {` branch inside the custom loadFunction because the `requestsConfigurationRules` weren't set. It was reaching the `img.src = newSrc` triggering the default image error. We need two test one for singleTile and one for the tiled layer both setting the request rules to ensure it reach the correct section of the loadFunction, below an update version of the two tests:

```js
it('render wms singleTile layer with error does not request a spurious "null" url', (done) => {
    ConfigUtils.setConfigProp('requestsConfigurationRules', [{
        urlPattern: '.*\\/geoserver.*',
        headers: {
            Authorization: 'Bearer ${securityToken}'
        }
    }]);
    setStore({
        getState: () => ({
            security: {
                token: "########-####-####-####-###########"
            }
        })
    });
    let request;
    mockAxios.onGet().reply((r) => {
        request = r;
        return [200, new Blob(["<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
        "<ows:ExceptionReport xmlns:ows=\"http://www.opengis.net/ows\">\n" +
        "  <ows:Exception exceptionCode=\"InvalidParameterValue\" locator=\"srsname\">\n" +
        "    <ows:ExceptionText>WMS server error. Invalid GetMap Request</ows:ExceptionText>\n" +
        "  </ows:Exception>\n" +
        "</ows:ExceptionReport>"], {type: 'text/xml'})];
    });
    const options = {
        type: 'wms',
        visibility: true,
        singleTile: true,
        url: '/geoserver/wms',
        name: 'ws:layer'
    };
    const layer = ReactDOM.render(<OpenlayersLayer
        type="wms"
        options={{
            ...options
        }}
        map={map}
        securityToken="########-####-####-####-###########" />, document.getElementById("container"));
    ConfigUtils.setConfigProp('requestsConfigurationRules', undefined);
    expect(layer.layer.getSource()).toBeTruthy();
    layer.layer.getSource().once('imageloaderror', (e) => {
        expect(request).toBeTruthy();
        expect(request.headers.Authorization).toBe('Bearer ########-####-####-####-###########');
        // the error handler must not set img.src = null, which the DOM coerces
        // to the literal string "null" and resolves into a spurious request
        const img = e.image.getImage();
        expect(img.getAttribute('src') === 'null' || (img.src || '').endsWith('/null')).toBe(false);
        done();
    });
});
it('render wms tiled layer with error does not request a spurious "null" url', (done) => {
    ConfigUtils.setConfigProp('requestsConfigurationRules', [{
        urlPattern: '.*\\/geoserver.*',
        headers: {
            Authorization: 'Bearer ${securityToken}'
        }
    }]);
    setStore({
        getState: () => ({
            security: {
                token: "########-####-####-####-###########"
            }
        })
    });
    let request;
    mockAxios.onGet().reply((r) => {
        request = r;
        return [200, new Blob(["<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
        "<ows:ExceptionReport xmlns:ows=\"http://www.opengis.net/ows\">\n" +
        "  <ows:Exception exceptionCode=\"InvalidParameterValue\" locator=\"srsname\">\n" +
        "    <ows:ExceptionText>WMS server error. Invalid GetMap Request</ows:ExceptionText>\n" +
        "  </ows:Exception>\n" +
        "</ows:ExceptionReport>"], {type: 'text/xml'})];
    });
    const options = {
        type: 'wms',
        visibility: true,
        singleTile: false,
        url: '/geoserver/wms',
        name: 'ws:tiled_layer'
    };
    const layer = ReactDOM.render(<OpenlayersLayer
        type="wms"
        options={{
            ...options
        }}
        map={map}
        securityToken="########-####-####-####-###########" />, document.getElementById("container"));
    ConfigUtils.setConfigProp('requestsConfigurationRules', undefined);
    expect(layer.layer.getSource()).toBeTruthy();
    layer.layer.getSource().once('tileloaderror', (e) => {
        expect(request).toBeTruthy();
        expect(request.headers.Authorization).toBe('Bearer ########-####-####-####-###########');
        // the error handler must not set img.src = null, which the DOM coerces
        // to the literal string "null" and resolves into a spurious request
        const img = e.tile.getImage();
        expect(img.getAttribute('src') === 'null' || (img.src || '').endsWith('/null')).toBe(false);
        done();
    });
});
```

allyoucanmap (changes requested):
A final note( that we should probably track in a different issue), other layer type implementations in OpenLayers have custom tile load function settings image to null that may cause a similar problem (to be investigated):

- WMTSLayer
- ArcGISLayer
- TileProviderLayer
-  TMSLayer

offtherailz:
PR updated @allyoucanmap

allyoucanmap (approved):

offtherailz:
- #12749

PR #12701, opened 2026-07-21
Support enable/disable rules on BrandNavbar menu items

offtherailz:
## Summary
- `leftMenuItems`/`rightMenuItems` entries in the `BrandNavbar` plugin can now declare an `enabled` property, either a plain boolean or a plugin expression string (same syntax as `cfg.disablePluginIf`), e.g. `"{state('usergroups').includes('editor')}"`.
- Items without `enabled` keep the current behaviour (always shown).
- Reuses the existing `PluginsUtils.handleExpression`/`getMonitoredState` machinery, no new expression syntax introduced.

## Test plan
- [x] Added `BrandNavbar-test.jsx` covering: item without `enabled`, `enabled: false`, expression resolving false, expression resolving true against `state('usergroups')`.
- [x] Ran the new suite locally via karma (chrome headless) — 4/4 passing.

### DIFF
--- a/web/client/plugins/ResourcesCatalog/BrandNavbar.jsx
+++ b/web/client/plugins/ResourcesCatalog/BrandNavbar.jsx
@@ -8,8 +8,10 @@
 
 import React from 'react';
 import { Glyphicon } from 'react-bootstrap';
+import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


offtherailz:
Sent in error

PR #12702, opened 2026-07-21
Support enable/disable rules on request configuration rules

### LINKED ISSUE #12704: Support enable/disable rules on request configuration rules
## Description

`requestsConfigurationRules` (and legacy `authenticationRules`) currently apply unconditionally whenever a rule's `urlPattern` matches. There is no way to activate/deactivate a rule depending on runtime conditions such as the current user's groups or role. This makes it hard to attach different auth headers/params to the same URL pattern depending on who is logged in (e.g. only send a Bearer token for users belonging to a specific group).

Add support for an `enabled` property on each rule: either a plain boolean, or a plugin expression string using the same syntax already supported for `cfg.disablePluginIf` (e.g. `state('usergroups')`, `includes(...)`, `hasAtLeastOne(...)`). Rules without `enabled` keep applying unconditionally, so the change is backward compatible.

## Acceptance criteria

- [x] Each entry in `requestsConfigurationRules` can declare an `enabled` property (boolean or expression string)
- [x] Rules with `enabled: false` are skipped
- [x] Rules with an expression string are evaluated against the monitored state (same mechanism as `cfg.disablePluginIf`), e.g. `"{includes(state('usergroups'), 'editor')}"`
- [x] Rules without `enabled` keep the current behaviour (always applied)
- [x] Legacy `authenticationRules` converted via `convertAuthenticationRulesToRequestConfiguration` also respect `enabled`
- [x] Documentation updated in `local-config.md`
- [x] Unit tests covering: no `enabled` property, `enabled: false`, expression resolving to true, expression resolving to false

## Other useful information

Implemented in #12702.

offtherailz:
Fixes #12704

## Summary
- Entries in `requestsConfigurationRules` (and legacy `authenticationRules`) can now declare an `enabled` property: a plain boolean or a plugin expression string (same syntax as `cfg.disablePluginIf`), e.g. `"{includes(state('usergroups'), 'editor')}"`.
- Rules without `enabled` keep the current behaviour (always applied).
- Reuses the existing `PluginsUtils.handleExpression`/`getMonitoredState` machinery, no new expression syntax introduced.
- Docs updated in `local-config.md`.

## Test plan
- [x] Added tests in `SecurityUtils-test.js`: no `enabled` property, `enabled: false`, expression resolving false, expression resolving true against `state('usergroups')`.
- [x] Ran the full `SecurityUtils-test.js` suite locally via karma (chrome headless) — 59/59 passing, no regressions.

### DIFF
--- a/web/client/utils/SecurityUtils.js
+++ b/web/client/utils/SecurityUtils.js
@@ -20,6 +20,7 @@ import castArray from "lodash/castArray";
 
 import {setStore as stateSetStore, getState} from "./StateUtils";
 import { parseUrl, WMS_GET_CAPABILITIES_VERSION } from '../api/WMS';
+import { getMonitoredState, handleExpression } from './PluginsUtils';
 
 export const USER_GROUP_ALL = 'everyone';
 
@@ -249,6 +250,20 @@ export const convertAuthenticationRulesToRequestConfiguration = (authRules = [])
 };
 
 /**
+ * Checks a request configuration rule's `enabled` property.
+ * `enabled` can be a plain boolean or a plugin expression string (e.g. `"{state('usergroups').includes('editor')}"`),
+ * evaluated with the same syntax used for `cfg.disablePluginIf` (see `PluginsUtils.handleExpression`).
+ * Rules without an `enabled` property are always applied.
+ * @param {object} rule the request configuration rule
+ * @returns {boolean} true if the rule's `enabled` resolves to a truthy value
+ */
+const isRuleEnabled = (rule) => {
+    const enabled = rule?.enabled ?? true;
+    const monitoredState = getMonitoredState(getState());
+    return !!handleExpression((path) => get(monitoredState, path), undefined, enabled);
+};
+
+/**
  * Gets all request configuration rules from Redux state or config
  * Automatically converts authenticationRules to new format if requestsConfigurationRules is missing
  * @returns {Array} Array of request configuration rules
@@ -257,19 +272,19 @@ export const getRequestConfigurationRules = () => {
     // First try to get from Redux state (if available)
     const stateRules = get(getState(), 'security.rules', []);
     if (!isEmpty(stateRules)) {
-        return stateRules;
+        return stateRules.filter(isRuleEnabled);
     }
 
     // Try to get new format from config
     const configRules = ConfigUtils.getConfigProp('requestsConfigurationRules');
     if (!isEmpty(configRules)) {
-        return configRules;
+        return configRules.filter(isRuleEnabled);
     }
 
     // If new format is missing, convert old authenticationRules format
     const authRules = ConfigUtils.getConfigProp('authenticationRules');
     if (!isEmpty(authRules)) {
-        return convertAuthenticationRulesToRequestConfiguration(authRules);
+        return convertAuthenticationRulesToRequestConfiguration(authRules).filter(isRuleEnabled);
     }
 
     // No rules found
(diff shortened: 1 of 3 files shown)


tdipisa:

web/client/utils/SecurityUtils.js
```diff
@@ -248,6 +249,20 @@ export const convertAuthenticationRulesToRequestConfiguration = (authRules = [])
...
+ * Checks a request configuration rule's `enabled` property.
+ * `enabled` can be a plain boolean or a plugin expression string (e.g. `"{state('usergroups').includes('editor')}"`),
+ * evaluated with the same syntax used for `cfg.disablePluginIf` (see `PluginsUtils.handleExpression`).
+ * Rules without an `enabled` property are always applied.
+ * @param {object} rule the request configuration rule
+ * @returns {boolean} true if the rule's `enabled` resolves to a truthy value
+ */
+const isRuleEnabled = (rule) => {
+    const enabled = rule?.enabled ?? true;
+    const monitoredState = getMonitoredState(getState());
```
allyoucanmap: Other getMonitoredState are including also the ConfigUtils.getConfigProp('monitorState') as second argument to include additional monitorState configured in a custom localConfig.json

`const monitoredState = getMonitoredState(getState(), ConfigUtils.getConfigProp('monitorState'));`

web/client/utils/SecurityUtils.js
```diff
@@ -248,6 +249,20 @@ export const convertAuthenticationRulesToRequestConfiguration = (authRules = [])
...
+    return !!handleExpression((path) => get(monitoredState, path), undefined, enabled);
+};
```
allyoucanmap: This is just a proposal, if we think `isRuleEnabled` is enough we can keep it for the moment.

At the moment we run `isRuleEnabled` for each rule, which means we call `getMonitoredState` and its internal memoize resolver (filterState) once per rule on every request. This could be costly because the resolver builds the cache key with `JSON.stringify` over every monitored entry, and the entry list is configurable via monitorState, so we cannot predict what ends up in there (it could technically be a large object).

The proposal is a function that takes `rules` and returns filtered rules:
```js
/**
 * Filters out rules whose `enabled` property resolves to falsy.
 * `enabled` is a boolean or a plugin expression string (same syntax as `cfg.disablePluginIf`).
 * Rules without an `enabled` property are always applied.
 * @param {object[]} rules the request configuration rules
 * @returns {object[]} the enabled rules
 */
const filterEnabledRules = (rules = []) => {
    const needsMonitoredState = rules.some(rule => isString(rule?.enabled));
    const monitoredState = needsMonitoredState ? getMonitoredState(getState(), ConfigUtils.getConfigProp('monitorState')) : {};
    const getMonitored = (path) => get(monitoredState, path);
    return rules.filter(rule => !!handleExpression(getMonitored, undefined, rule?.enabled ?? true));
};
```

applied like this:

```js
return filterEnabledRules(stateRules);
```

```js
return filterEnabledRules(configRules);
```

```js
return filterEnabledRules(convertAuthenticationRulesToRequestConfiguration(authRules));
```

allyoucanmap (changes requested):

web/client/utils/SecurityUtils.js
offtherailz: done

web/client/utils/SecurityUtils.js
offtherailz: done

allyoucanmap (approved):

offtherailz:
- #12748

allyoucanmap:

PR #12707, opened 2026-07-21
#12650: Zoom to filtered layer from the filter widget

### LINKED ISSUE #12650: [Filter Widget] Zoom to filtered layer from the filter widget
## Description

Add a zoom-to action to the filter widget. Zoom the map to the filtered geometry of the connected layers. Support a manual button and an automatic zoom on filter change.

**Model**
- Zoom-to is an interaction with a target, so it fits the interaction model. Add a new target type `zoomTo` whose target is a map, the main map or a dashboard map widget.
- Transient effect. No persisted `appliedData` and no cleanup. It dispatches `zoomToExtent` only.
- Bound to the filter through the shared `source.nodePath` (filter id). Siblings of the filter `applyFilter` interactions.
- Availability: a map is a valid `zoomTo` target only when the filter has an `applyFilter` connection to a layer inside it.

**Enable and mode (per connection)**
- Connect (plug) enables zoom-to for the target map. Unplugged means no zoom.
- When connected, a single boolean `autoZoom` in the interaction `configuration`: off is manual (button in the widget), on is automatic zoom on selection change.
- Single map context: the selector implicitly targets the main map. Dashboard: one target per map.

**Mock**
- **Widget**
  
Image

- **Widget wizard**
 
Image

**Extent computation**
- WPS bounds (GeoServer) as primary, WFS plus turf bbox as fallback. Consistent with the widget WPS and WFS value fetching.
- Union of the connected layers, reprojected to a common CRS.
- New action `zoomToFilterExtent(widgetId, filterId)` shared by manual and auto. Reuses `zoomToExtent` (`actions/map.js`).

## Acceptance criteria

- [x] A user connects (plugs) a `zoomTo` target (a map) to enable zoom-to
- [x] An `autoZoom` config toggles automatic vs manual when connected
- [x] Manual (`autoZoom` off): a button in the widget zooms to the filtered extent
- [x] Auto (`autoZoom` on): the map zooms on selection change
- [x] The extent is the filtered target layer geometry, not the filter source
- [x] Multiple connected layers zoom to the union extent
- [x] Works with a single map and with dashboard maps
- [x] Zoom to action should move the map on the visible map area taking also into account to widgets location. An heuristics is needed to find best visible map area falling back to full map zoom if the widgets are covering the full map
- [x] Unit tests for extent computation and mode handling
- [x] i18n updated (en, it, fr, de, es)

## Other useful information

- The first implementation may limit zoom-to to a single connected layer, the model already supports multiple.

dsuren1:
## Description
This PR enhances the dynamic filter widget by adding a new interaction `Zoom To` which let's the user perform `autoZoom` or manual zoom on the target map. Dashboard: one target per map

 - [x] Feature

## Issue

- #12650 

**What is the new behavior?**
- Zoom the map to the filtered geometry of connected layers. Supports both manual zoom via a button and automatic zoom when filters change.

  #### Map
  
  https://github.com/user-attachments/assets/d9ef2b62-e4cf-48ae-866b-bbaa662cd57b

  #### Dashboard
  
  https://github.com/user-attachments/assets/8d1c1ddb-45dd-4191-980d-a742c2b47e7c

## Breaking change
 - [x] No


web/client/components/map/openlayers/Map.jsx
```diff
@@ -714,13 +715,16 @@ class OpenlayersMap extends React.Component {
...
+        const {top = 0, right = 0, bottom = 0, left = 0} = resolveZoomToExtentPadding(mapEl, padding);
+        const paddingValues = [top, right, bottom, left];
         this.map.getView().fit(bounds, {
             size: this.map.getSize(),
             // mapPaddingSelector returns null while the layout epic is mid-computation
             // (e.g. during a setView swap on projection change). Pass undefined in that
             // case so OL applies its own [0,0,0,0] default - passing null reaches
             // View.fitInternal where padding[1] dereferences and throws.
-            ...(padding ? { padding: [padding.top || 0, padding.right || 0, padding.bottom || 0, padding.left || 0] } : {}),
+            ...(paddingValues.some(value => value > 0) ? { padding: paddingValues } : {}),
```
allyoucanmap: It's better we don't include Widgets logic inside the Map component, so please restore this file. We should avoid to import this utils and the padding should be resolved at epic level when we are using `buildZoomToExtentAction` (note buildZoomToExtentAction has a padding argument but never used). If we externalize this to the epic also the Cesium map will get the correct padding without changes on the map engine components

web/client/components/widgets/enhancers/interactionZoomToExtent.js
```diff
@@ -0,0 +1,36 @@
...
+        }
+        const hook = hookRegister?.getHook(ZOOM_TO_EXTENT_HOOK);
+        if (hook) {
+            const { extent, crs, maxZoom, padding } = targetMap.zoomToRequest;
+            hook(extent, { crs, maxZoom, ...(padding ? { padding } : {}) });
+        }
+        updateProperty(id, 'maps', { mapId: targetMap.mapId, zoomToRequest: undefined }, 'merge');
+        return {};
+    }
+);
```
allyoucanmap: This component is using recompose, we should avoid to use it in new files, also it's triggering a side effect inside the withPropsOnChange. This should become an HOC with a react function that uses an useEffect

web/client/epics/interactions.js
```diff
@@ -953,6 +962,149 @@ const applyInteractionEffectForApplyStyle = (interaction, state, targetContainer
...
+            const seen = new Set();
+            resolvedLayers.forEach(rl => {
+                const key = `${rl.mapWidgetId || 'main'}-${rl.mapId || 'main'}`;
+                if (!seen.has(key)) {
+                    seen.add(key);
+                    uniqueTargets.push(rl);
+                }
+            });
+
+            const actions = uniqueTargets.map(target => buildZoomToExtentAction(unionExtent, target, targetContainer));
```
allyoucanmap: we should pass the padding here

web/client/components/widgets/builder/wizard/common/interactions/interactionHelpers.js
```diff
@@ -20,6 +24,31 @@ const DEFAULT_NODE_DISABLED = {
...
+    if (isMapZoomToTarget(zoomTargetNodePath)) {
+        return applyFilterConnections.some(connection => isMapLayerPath(connection?.target?.nodePath));
+    }
+
+    const zoomMapId = extractMapIdFromNodePath(zoomTargetNodePath);
+    return applyFilterConnections.some(connection => {
+        const layerNodePath = connection?.target?.nodePath;
+        return !isMapLayerPath(layerNodePath) && extractMapIdFromNodePath(layerNodePath) === zoomMapId;
+    });
+}
```
allyoucanmap: This function has a similar structure of the findAllApplyFiltersForZoomTo, couldn't this be directly replaced by `findAllApplyFiltersForZoomTo(...).length > 0`

web/client/utils/InteractionUtils.js
allyoucanmap, line 40: Do we need to add `APPLY_ZOOM_TO` event?

web/client/components/widgets/builder/wizard/common/interactions/interactionConstants.js
```diff
@@ -42,11 +44,23 @@ export const CONFIGURATION_METADATA = {
...
+    autoZoom: {
+        label: "widgets.filterWidget.autoZoomLabel",
+        visibleWhen: {
+            targetType: {
+                isEqual: TARGET_TYPES.APPLY_ZOOM_TO
+            }
+        },
+        infoMsgByTargetType: {
+            "default": "widgets.filterWidget.autoZoomInfo"
+        }
```
allyoucanmap, line 64: Now that we have inline button is this Metadata really used?

If we have something like this

```
autoZoom: {
    label: "widgets.filterWidget.autoZoomLabel",
    renderAs: 'inline-button',
    visibleWhen: {
        targetType: { isEqual: TARGET_TYPES.APPLY_ZOOM_TO },
        plugged: { isEqual: true }
    },
    infoMsgByTargetType: { "default": "widgets.filterWidget.autoZoomInfo" }
}
```

We could add `renderAs: 'inline-button'` all the other will be rendered as checkbox when renderAs is undefined, also adding `plugged: { isEqual: true }` we can remove the hardcoded `showAutoZoom && plugged`

We can add the renderAs types in constants.

web/client/components/widgets/builder/wizard/common/interactions/InteractionButtons.jsx
```diff
@@ -27,10 +27,20 @@ const TButton = tooltip(Button);
...
     return (
         <FlexBox gap="xs" className={`ms-interaction-buttons${isPlugConstrained ? ' is-disabled' : ''}`}>
+            {showAutoZoom && plugged && <TButton
+                onClick={() => onAutoZoomChange(!autoZoom)}
+                borderTransparent
+                tooltip={autoZoomForAllMaps ? <Message msgId="widgets.filterWidget.autoZoomForAllMapsTooltip" /> : <Message msgId="widgets.filterWidget.autoZoomInfo" />}
+                variant={autoZoom ? "primary" : undefined}
+            >
+                <Message msgId="widgets.filterWidget.autoZoomLabel" />
+            </TButton>}
```
allyoucanmap: **Note, all of this is pseudo code needs to be validated**

Based on previous comment this could become, so we can remove hardcoded inline props and use the Metadata:

```js
{getVisibleConfigurationKeys(configuration, context, 'inline-button').map(key => {
    const metadata = CONFIGURATION_METADATA[key];
    return (<TButton key={key} borderTransparent
        variant={configuration[key] ? "primary" : undefined}
        tooltip={<Message msgId={metadata.infoMsgByTargetType?.[context.targetType] ?? metadata.infoMsgByTargetType?.default} />}
        onClick={() => setConfiguration({ ...configuration, [key]: !configuration[key] })}>
        <Message msgId={metadata.label} />
    </TButton>);
})}
```

We could create the new getVisibleConfigurationKeys function looking to how the InteractionConfiguration.jsx implements const visibleConfigurationKeys then we can externalize a function able to check configuration keys, eg.

```js
const getVisibleConfigurationKeys = (configuration = {}, context = {}, renderAs = CONFIGURATION_RENDER_MODES.PANEL_CHECKBOX) =>
    Object.keys(configuration).filter((key) =>
        isConfigurationVisible(CONFIGURATION_METADATA[key], context)
        && (CONFIGURATION_METADATA[key]?.renderAs ?? CONFIGURATION_RENDER_MODES.PANEL_CHECKBOX) === renderAs);

```

We could re-use this function in the following files

```js
// InteractionConfiguration.jsx
const visibleConfigurationKeys = getVisibleConfigurationKeys(configuration, context);

// InteractionButtons.jsx
const inlineKeys = getVisibleConfigurationKeys(configuration, context, CONFIGURATION_RENDER_MODES.INLINE_BUTTON);

// InteractionTargetsList.jsx
const isConfigurable = getVisibleConfigurationKeys(configuration, context).length > 0;
```

web/client/components/widgets/builder/wizard/common/interactions/InteractionTargetsList.jsx
```diff
@@ -9,7 +9,7 @@ import React from 'react';
 import FlexBox from '../../../../../layout/FlexBox';
 // import { filterTreeWithTarget } from '../../../../../../utils/InteractionUtils';
 import Message from '../../../../../I18N/Message';
-import { findNodeById, getItemPluggableStatus, isMapTimeTarget } from '../../../../../../utils/InteractionUtils';
+import { findNodeById, getItemPluggableStatus, isMapTimeTarget, isAnyZoomToTarget, getGlobalAutoZoom, updateGlobalZoomInteractionsAutoZoom, containsMultipleZoomToNodes } from '../../../../../../utils/InteractionUtils';
```
allyoucanmap: We should remove all the global auto zoom sync, each interaction should have it's own configuration, to make this clear to the final user we can add in the tooltip of the zoom to button when visible the names of map that the zoom to will trigger (if it's just one map the list of map title is not needed)

allyoucanmap (changes requested):
No cleanup path handles APPLY_ZOOM_TO, so a plugged zoom-to interaction can outlive the applyFilter sibling it depends on, how to reproduce:

- open a map and add a wms layer
- add a filter widget
- connect the filter widget to the layer
- connect the zoom to action to the layer
- save the widget
- verify the zoom to is working
- edit the widget again
- disconnect only the filter interaction
- save the widget
- **Expected** the zoom to button is not visible because there is not event filter applied
- **Current** the zoom to button is visible even if there is not event filter applied

web/client/components/map/openlayers/Map.jsx
dsuren1: This would require updating the Cesium map component as well, since it currently **doesn't** account for padding. I initially thought this was intentional and focused only on the OpenLayers map, but it looks like the padding support may have been missed in the original Cesium implementation. I will update it to handle the padding consistently

web/client/utils/InteractionUtils.js
dsuren1: I don't think so. `EVENT_TARGET_MAP` isn't used anywhere yet

allyoucanmap (approved):

allyoucanmap:

PR #12719, opened 2026-07-22
#11772 Save dashboard with maximized widget cause problem with resize/move

### LINKED ISSUE #11772: Save dashboard with maximized widget cause problem with resize/move
## Description

This problem is present also on 2025.01.xx, it's not a regression. At the moment the maximize logic for widgets is modifying the layout on the fly, it would be nice if we are able to simplify the way we are managing the fullscreen widget

https://github.com/user-attachments/assets/222418f9-e3e4-41dc-8bd8-3925c2b6dff3

## How to reproduce

- login
- create a new dashbaord
- add a widget
- maximize the created widget
- save the dashboard

*Expected Result*

- on save it's still possible to move and resize the widget card

*Current Result*

- on save it's not possible to move and resize the widget card

- [x] Not browser related

|---|---|
|Internet Explorer| |
|Edge| |
|Chrome| |
|Firefox| |
|Safari| |
</details>

Binabh:
## Description
Fix state management to track resizable widgets when they are maximized.

 - [x] Bugfix

## Issue

#11772 

**What is the new behavior?**
Resize and move is working properly even after maximizing widgets

## Breaking change
 - [x] No


offtherailz (changes requested):
Sorry I still see the same issue

[Screencast from 2026-08-04 12-10-29.webm](https://github.com/user-attachments/assets/19864198-cba1-4926-993b-1ac6a4b8d3c8)

Binabh:
Sorry from my side. I checked maximize->minimize->save but did not check maximize->save during my manual verification. I have pushed fix for that case also. Please re-review and provide feedback if any improvement still needed. @offtherailz

offtherailz (changes requested):
Thanks @Binabh, the `maximize -> save` case is fixed now: I checked the save path (`widgetsConfig` -> `filterLinkedWidgets`/`getUpdatedLayout` -> `getWidgetsForSave`/`getLayoutsForSave`) and the dashboard reload after save, and the persisted `dataGrid`/`layouts` are the pre-maximize ones, as expected.

While testing I hit another broken case in the same area, so I would like to fix it here, since this PR is about the state management of the maximized widgets.

### How to reproduce

- create a new dashboard
- add a widget
- maximize it
- delete it (from the widget toolbar, while maximized)
- add another widget

Result: the new widget is not shown at all, and the dashboard stays empty until a page reload.

### Root cause

`case DELETE` in `web/client/reducers/widgets.js` removes the widget from `containers[target].widgets` but leaves `containers[target].maximized` untouched, so the snapshot still contains the deleted widget (plus the pre-maximize `layout`/`layouts`).

Then in `web/client/plugins/Dashboard.jsx`:

```js
widgets: !isEmpty(maximized) && Array.isArray(maximized.widget) && maximized.widget.some(w => w.layoutId === selectedLayoutId)
    ? widgets.filter(w => maximized.widget.some(mw => mw.id === w.id))
    : widgets,
```

`maximized` is not empty and the deleted widget still matches `selectedLayoutId`, so only widgets listed in `maximized.widget` are kept: the deleted one is gone and every widget added later is filtered out. `layouts` also stays in the maximized shape (`md: []`, `xxs` with a single 1x1 cell).

The same happens in the map viewer through `getVisibleFloatingWidgets` (`web/client/selectors/widgets.js`), which filters on `id === maximized.widget.id`: deleting the maximized widget hides all the widgets.

Note this is not a regression of this PR (`master` behaves the same), but it is the same class of problem, so it makes sense to fix it together.

### Suggested fix

In `case DELETE`, if the widget is part of the maximized state, restore it first by reusing the reducer itself (same pattern already used by `TOGGLE_COLLAPSE_ALL`), then delete it:

```js
case DELETE: {
    const path = `containers[${action.target}].widgets`;
    const maximizedWidget = castArray(state?.containers?.[action.target]?.maximized?.widget || [])
        .find(w => w?.id === action.widget.id);
    // restore the layout captured on maximize before removing the widget
    const restoredState = maximizedWidget
        ? widgetsReducer(state, toggleMaximize(maximizedWidget, action.target))
        : state;
    const updatedState = arrayDelete(`containers[${action.target}].widgets`, {
        id: action.widget.id
    }, restoredState);
    ...
```

This covers both the single maximized widget and the multi-view case with more than one maximized widget, since `TOGGLE_MAXIMIZE` already handles both. `toggleMaximize` has to be imported from `../actions/widgets` (like `toggleCollapse` already is), and the widget must be taken from the snapshot because the multi-view branch needs its `layoutId`.

One more thing to check while you are there: the final `set(path, allWidgets.map(...), state)` of `case DELETE` uses `state` as base instead of `updatedState`. It works today only because the mapped list comes from `updatedState`, but with the change above it would drop the restored layout, so it should use `updatedState` (or `restoredState`).

Please also add a reducer test for `DELETE` of a maximized widget, checking that `maximized` is cleaned up and `layout`/`layouts` are restored.

### Minor (optional)

- `getWidgetsForSave` (`web/client/selectors/widgets.js`): when the snapshot has no `dataGrid` the widget is returned untouched, so it is still saved with `isDraggable: false`/`isResizable: false`. It should not happen with widgets created by `INSERT`, but it can with older saved configurations. A guard would be safer.
- `getLayoutsForSave` returns the whole pre-maximize `layouts` snapshot, so layout changes done while a widget is maximized (a widget added or deleted in the meantime) are lost or leave an orphan entry on save. Low impact, but worth keeping in mind.

offtherailz (changes requested):
I noticed this other bug. Please fix it contextually and test well the maximized functionality, finding out possible problems

https://github.com/user-attachments/assets/589e4a50-a62b-464c-9828-4924ef5b1610

Binabh:
Thank you for your review @offtherailz . I will check these and other actions for maximized widgets causing any problems. I will also try to add more tests to cover these cases.

Binabh:
@offtherailz I applied your suggestion and fixed `maximize->delete->new widget` case. I also tested for any other error case like `maximize->edit`,`reposition->maximize->minimize`,`reposition->maximize->minimize` but could not find any error at the moment.
I am re-requesting your review. Let me know your feedback on it.

offtherailz (changes requested):
The bug is still present on the map. @Binabh please test widely before to submit PRs . 

https://github.com/user-attachments/assets/ba0de9c4-a988-4509-aaf4-fd6dd67d8e2d

Clicking maximize and then minimize does unblock the widget, because the
un-maximize branch forces isDraggable: true / isResizable: true. That is
an accidental escape hatch rather than a fix: the layouts restored at that
point are the already-degraded ones, so the widget's original position and
size are lost for good.

## Suggested approach 

The current fix strips the transient state at save time. That works, but it
has to be repeated at every save path, and there are four of them today:
mapsave.js, ResourcesCatalog/selectors/save.js, epics/permalink.js and
DashboardExport.jsx. One has already been missed.

Here I attached a patch file with an incomplete prototype that follows a more stable solution. Apply it to your branch to see what are the effective changes.

[pr12719-prototype-b.patch](https://github.com/user-attachments/files/31074702/pr12719-prototype-b.patch)

Binabh:
@offtherailz Thank You for your feedback. I have pushed commits.

I took your patch as starting point and then fixed tests, multi-view widget id filtering and removed unnecessary functions.

I manually tested in following:
- Dashboard: Maximize->Save, Maximize->Delete, Maximize->Minimize->Save, Multiple-views
- Map: Maximize->Save, Maximize->Delete, Maximize->Minimize->Save
- Read-only Mode: User is able to maximize/minimize but not change the position

Thank you for your review and sorry for too much back-and-forth in this PR.

offtherailz (approved):

offtherailz:
 - Pull request Pull request is a draft

PR #12428, opened 2026-05-26
Upgrade to Spring 7

axl8713:
## Description

These changes cover the Spring 7 upgrade for MapStore.

Key changes:

  - Spring Framework 5.x → 7.0.8
  - Spring Security 5.x → 7.1.0
  - Jakarta Servlet API 6.0.0 (Jakarta EE 10, aligned with GeoServer ecosystem)
  - Tomcat 10.1.55 (Tomcat 10.1 is the target for EE 10)
  - Minimum supported JDK raised to 17 (required by Spring 7; aligned with GeoServer)
  - Maven compiler source/target updated from 8 to 17 in java/services/pom.xml
  - All CI/CD workflows updated to use JDK 17 as the build JDK; CI matrix still tests on [17, 21]

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [ ] Bugfix
 - [x] Feature
 - [ ] Code style update (formatting, local variables)
 - [ ] Refactoring (no functional changes, no api changes)
 - [x] Build related changes
 - [ ] CI related changes
 - [ ] Other... Please describe:

## Issue

**What is the current behavior?**

#<issue>

**What is the new behavior?**

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [X] Yes, and I documented them in migration notes
 - [] No

There will be a new set of requirements to have the GeoStore with Spring 7 running.

Issue #12025, opened 2026-02-27
Update of Spring dependency


tdipisa:
Move to Sprint 7. The task includes the following involved parts:
- GeoStore
- MapStore java part [here](https://github.com/geosolutions-it/MapStore2/tree/master/java)
- Mapfish print: work already done with https://github.com/mapfish/mapfish-print-v2/commit/bca81620badf1735840622ef012727e332377e93

PR #12731, opened 2026-07-24
Align standard pom template mapstore-services.version to 1.13-SNAPSHOT

offtherailz:
## Description

Realigns the standard project pom template to the current master snapshot version.

`project/standard/templates/pom.xml` had `mapstore-services.version` set to `1.12-SNAPSHOT`. This was regressed by the Spring 7 merge (#12428), which was branched from a long-lived branch and overwrote the alignment previously done in #12708. Master builds and deploys `1.13-SNAPSHOT`, so a project generated from the master template pulled the wrong (older) `mapstore-services` artifact.

All other template properties (spring, spring-security, jackson, jakarta servlet-api, tomcat, geostore-webapp, http_proxy, print-lib) were verified aligned with the root pom; only `mapstore-services.version` was off.

## Issue
#12025

## What kind of change does this PR introduce?
- [x] Bugfix
- [x] Build related changes

## Follow-up
No automated guard currently catches template/root version drift (this is how it regressed). A CI check comparing `mapstore-services.version` in the template against the root pom version is recommended as a follow-up.

### DIFF
--- a/project/standard/templates/pom.xml
+++ b/project/standard/templates/pom.xml
@@ -30,7 +30,7 @@
         <cargo.version>1.10.27</cargo.version>
 
         <!-- MapStore‑specific -->
-        <mapstore-services.version>1.12-SNAPSHOT</mapstore-services.version>
+        <mapstore-services.version>1.13-SNAPSHOT</mapstore-services.version>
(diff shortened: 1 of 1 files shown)


axl8713 (approved):
LGTM.

PR #12741, opened 2026-07-27
WMS Coalesce feature option for layers and maps

### LINKED ISSUE #12662: Coalesce adjacent WMS layers into a single GetMap request
## Overview

Adjacent WMS layers pointing at the same endpoint are currently rendered as separate `GetMap` requests. When they share a service they can be served by one request using comma-separated params:

```
[{ ...wmsA }, { ...wmsB }]  ->  one request:  ...&LAYERS=a,b&STYLES=sA,sB
```

This is a rendering-path optimization only. The redux layer model stays one-entry-per-layer, so identify, legend, TOC and settings are unaffected (they read the redux model, not engine layers).

### Proposed By

Stefano Bovio @allyoucanmap 

### State
* [x] Under Discussion
* [ ] In Progress
* [ ] Completed
* [ ] Rejected
* [ ] Deferred

### Motivation

Fewer HTTP requests/tiles and server-side compositing for the common case of themed layers from the same GeoServer differing only in name/style. No change to the redux model or to identify/legend/TOC/settings.

## Proposal

A shared pure `groupWMSLayers(layers)` transform turns the ordered layer list into render units (single layers, or a WMS group with a merged `name`/`style`). Called from both render paths: `plugins/Map.jsx` (main map) and `components/map/BaseMap.jsx` (dashboards, widgets, geostories).

- Grouping is gated by a map-configuration flag (`coalesce` at map config level), disabled by default for backward compatibility. When enabled it is automatic and derived (recomputed on every layer-list change); redux layer entries are untouched.
- Each engine's WMS `create`/`update` builds the comma `LAYERS`/`STYLES` on one source. Visibility/style changes keep the group key stable, so they go through `updateParams` (the existing `update` already diffs `LAYERS`/`STYLES`). Membership changes (add/remove/reorder) change the group key, so the group is recreated. Acceptable because one composite source refetches all tiles on any `LAYERS`/`STYLES` change.

### Merge rules

When coalescing is enabled at map config level, only maximal contiguous runs of eligible WMS layers coalesce. Any other layer (non-WMS, or a non-mergeable WMS) is a hard boundary. Two adjacent layers merge only if all hold:

- not opted out (`coalesce !== false`); both `type: wms`, non-vector, same `singleTile`;
- same URL, same opacity, same security/auth;
- every request-wide param identical: `format`, `transparent`, `tiled`, `version`, `srs`, `tileSize`, tile grids, `baseParams`, `ENV`, and all dimensions (`TIME`/`ELEVATION`);
- no per-layer param that v1 does not merge: `CQL_FILTER`, `FILTER`, `SLD`/`SLD_BODY`, `VIEWPARAMS`, `SORTBY`.

A run also stops once it reaches the maximum group size (`coalesceWMSLayersMaxGroupSize`, default 10). Further eligible layers start a new group, so a long contiguous run is split into several capped groups.

Only `LAYERS` (`name`) and `STYLES` (`style`) differ between members. `STYLES` is positional (empty allowed: `sA,,sC`).

### Key behaviours and decisions

- Coalescing remains inactive by default, as in the current proposal.
- Background layers are excluded
- **Enablement:** single `coalesce` flag at map configuration level, disabled by default (strict `=== true` to enable, so absent maps behave exactly as today, no migration). ~No per-layer flag in v1; a per-layer or TOC override is deferred.~  an override will be possible for each layer via the Layer settings -> Display -> coalesce.
- **Opacity** differences block the merge (one opacity per source); per-sublayer opacity out of scope.
- **Visibility** handled by membership: a hidden member is dropped from `LAYERS`/`STYLES`; all hidden hides the group.
- **Reorder:** moving a foreign layer between members splits the group (contiguity required).
- **`_v_`** (cache-bust) is not a merge blocker; the group value is `Math.max` of members.
- **Long URLs / POST:** coalesced requests are longer. A GET->POST fallback exists in `WMSLayer.js` gated by `maxURLLength` (default off), keeping the endpoint plus first param in the URL (hybrid). Long groups may need `maxURLLength` configured and the POST path hardened; servers must accept POST `GetMap`.
- **Load/error feedback needs an adapter:** `onLayerLoading/onLayerLoad/onLayerError` are keyed by a single `layerId`; the group synthetic id is not a real layer, so the event stream must be fanned out to member ids. `onCreationError` needs member attribution; `onWarning` bbox check must consider all members.
- **Error attribution:** a combined `GetMap` fails all-or-nothing; accepted for v1.
- **Limit:** cap members per group at `coalesceWMSLayersMaxGroupSize` (default 10, configurable at map plugin config level). Runs longer than the cap split into multiple groups; guards `LAYERS`/`STYLES` URL length and all-or-nothing failure blast radius.

- Data model / precedence
  - Layer settings `layer.coalesce = undefined` → layer inherits the map default
  - Layer settings `layer.coalesce = true | false` → always wins over the map default
  - Map settings → `mapOptions.coalesceWMSLayers` (`undefined` means `false`)
  - localConfig → `miscSettings.coalesceWMSLayers` (used if map is `undefined`)

### Acceptance criteria

- [ ] Adjacent same-endpoint WMS layers differing only in name/style render as one `GetMap`.
- [ ] Non-WMS or non-mergeable layers between them correctly split the group (z-order preserved).
- [ ] Opacity/visibility/reorder behave per the rules above.
- [ ] Per-layer loading/error state still works (adapter fan-out).
- [ ] Coalescing is off by default; the map-config `coalesce` flag enables it. Existing saved maps load unchanged (no migration).
- [ ] With `maxURLLength` configured, a group whose combined URL exceeds it still renders via GET->POST.
- [ ] A run longer than `maxCoalesce` (default 10/15) splits into multiple capped groups, z-order preserved.
- [ ] Works and is tested on OpenLayers and Cesium.
- [ ] layer settings an option default, true, false, default is inherits the value from the map

### Backwards Compatibility

Fully backwards compatible. Coalescing is disabled by default and only enabled by the map-config `coalesce` flag, so existing maps render exactly as today. The redux model is unchanged and no migration is needed; grouping is derived at render time.

## Future evolution

- Merging of currently blocked per-layer params.

### UI

- **Layer Display tab**: 3-state selector (`default`, `true`, `false`) editing `layer.coalesce`. `default` = inherit from map.
- **Map settings**: this is where I wanted to put the checkbox to enable the map-level default, but Giovanni points out that in GeoNode that settings panel doesn't exist (and honestly it's barely used in MapStore too). So for GeoNode it's still open **where** to expose this control.

Layer tab : 
```
Combine requests: [ Use map default ▾ ]
                      Use map default
                      Enabled
                      Disabled
```
Map setting: 
```
☐ Combine WMS requests by default ⓘ
```
With Tooltip which explains the functionality

## Voting
Project Steering Committee:

* @tdipisa: +1
* @offtherailz: +1
* @allyoucanmap: +1
* @MV88: +1
* @giohappy: +1
* @jna4k:

## Links

* [Email Discussion]()
* [Pull Request](https://github.com/geosolutions-it/MapStore2/pull/12741)
* [Linked Issue]()

stefanocudini:
## Description

Issues contains the details, add new feature for Map and Layers, Coalesce adjacent WMS layers into a single GetMap request.
image

 - [x] Feature

## Issue

https://github.com/geosolutions-it/MapStore2/issues/12662

Each WMS layer is always loaded with a single getMap per layer

**What is the new behavior?**
Adjacent WMS layers pointing at the same endpoint are currently rendered as separate GetMap requests.

## Other useful information

This map conatins more wms layers useful for testing:
https://dev-mapstore.geosolutionsgroup.com/mapstore/#/viewer/67842

### DIFF
--- a/docs/developer-guide/local-config.md
+++ b/docs/developer-guide/local-config.md
@@ -208,6 +208,20 @@ For configuring plugins, see the [Configuring Plugins Section](plugins-documenta
     }
     ```
 
+  - `enabled` - Optional, a plain boolean or a plugin expression string (same syntax as a plugin's `disablePluginIf`) to
+    conditionally apply the rule, e.g. based on the current user's groups. Rules without `enabled` are always applied.
+    Example:
+
+    ```json
+    {
+      "urlPattern": ".*geoserver.*",
+      "params": {
+        "authkey": "${securityToken}"
+      },
+      "enabled": "{includes(state('usergroups'), 'editor')}"
+    }
+    ```
+
 !!! note "Backward Compatibility"
     The old `useAuthenticationRules` and `authenticationRules` configuration still works and will be automatically converted to the new format. However, the new format is recommended for better flexibility and features like expiration support.
 
--- a/docs/developer-guide/local-config.md
+++ b/docs/developer-guide/local-config.md
@@ -78,6 +78,8 @@ This is the main structure:
   "miscSettings": {
       // Use POST requests for each WMS length URL highter than this value.
       "maxURLLength": 5000,
+      // Combine adjacent compatible WMS layers into a single GetMap request
+      "coalesceWMSLayers": false,
       // Custom path to home page
       "homePath": '/home'
   },
--- a/product/config/ldap/geostore-spring-security-ldap.xml
+++ b/product/config/ldap/geostore-spring-security-ldap.xml
@@ -29,6 +29,13 @@
 		<security:anonymous />
         <security:intercept-url pattern="/**" access="permitAll" />
 	</security:http>
+
+    <bean id="entityManager"
+          class="org.springframework.orm.jpa.support.SharedEntityManagerBean">
+        <property name="entityManagerFactory" ref="geostoreEntityManagerFactory"/>
+        <property name="persistenceUnitName" value="geostore"/>
+    </bean>
+
 	<!-- Entry point -->
 	<bean id="restAuthenticationEntryPoint" class="it.geosolutions.geostore.services.rest.security.RestAuthenticationEntryPoint" >
 		<property name="realmName" value="GeoStore"></property>
--- a/web/client/components/TOC/fragments/settings/Display.jsx
+++ b/web/client/components/TOC/fragments/settings/Display.jsx
@@ -51,7 +51,8 @@ export default class extends React.Component {
         mapBbox: PropTypes.object,
         resolutions: PropTypes.array,
         zoom: PropTypes.number,
-        hideInteractiveLegendOption: PropTypes.bool
+        hideInteractiveLegendOption: PropTypes.bool,
+        coalesceWMSLayers: PropTypes.bool
     };
 
     static defaultProps = {
@@ -296,6 +297,7 @@ export default class extends React.Component {
                                         projection={this.props.projection}
                                         onChange={this.props.onChange}
                                         disableTileGrids={!!this.props.isCesiumActive}
+                                        coalesceWMSLayers={this.props.coalesceWMSLayers}
                                     />
                                 </>
                             ))}
--- a/web/client/components/TOC/fragments/settings/Display.jsx
+++ b/web/client/components/TOC/fragments/settings/Display.jsx
@@ -51,8 +51,7 @@ export default class extends React.Component {
         mapBbox: PropTypes.object,
         resolutions: PropTypes.array,
         zoom: PropTypes.number,
-        hideInteractiveLegendOption: PropTypes.bool,
-        coalesceWMSLayers: PropTypes.bool
+        hideInteractiveLegendOption: PropTypes.bool
     };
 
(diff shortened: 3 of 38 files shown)


allyoucanmap (approved):
@stefanocudini I approved and reviewed the code adding some commits, please double check them. If everything is fine we can merge it

allyoucanmap:
we should squash with the `Squash and merge` button cleaning the commit description because the commit history is not clean, thanks

stefanocudini (commented):
ok

allyoucanmap:

stefanocudini:
@dsuren1 @allyoucanmap merging this PR seems to have broken Geonode-mapstore-client's dependency on the `resourceToLayers` method which is now not exposed, we restore it in a [new PR](https://github.com/geosolutions-it/MapStore2/pull/12781)

Issue #12597, opened 2026-07-07
Identify: media rendering for EXTERNAL DATA attributes


allyoucanmap:
## Description

Upgrade `EXTERNAL DATA` attributes from plain text (generated from GeoJSON) to configurable media. Adds the display-type selector to the Attributes config and the renderers that drive it.

**Config** (Attributes section, extending the EXTERNAL DATA issue)
- `Feature info display type` selector per attribute: `Image`, `Video`, `Audio`, `Panorama Image`, `PDF`, `URL`, `Iframe`, `Media`
- `Media type attribute` selector, visible only when type = `Media`: picks another attribute holding the media type, with fallback to string / guess-by-extension

Attribute configuration inputs related to media type

<img width="1919" height="951" alt="Image" src="https://github.com/user-attachments/assets/520b59ba-ecf2-41d9-a5e3-100c309e5f1b" />

**Renderers** (new components under `components/`)
- Image / Video / Audio: reuse/extract the GeoStory Media component into a store-agnostic renderer
- URL: `<a>`
- Iframe: `<iframe>`
- PDF: port the GeoNode iframe viewer (`geonode-mapstore-client .../MediaViewer/PdfViewer.jsx`)
- Panorama Image: new component embedding the approved existing panorama lib; lazy-loaded (dynamic import on tab open) to keep it out of the main bundle. During the mockup evaluation we tested https://github.com/mpetroff/pannellum because it's a simple panorama viewer giving also script to generate compatible multi resolution panorama images (before starting the development we should review also what kind of panorama views are supported in the latest version of [CesiumJS](https://cesium.com/learn/cesiumjs-learn/display-panoramic-images/) and evaluate if this could be a possible alternative)
- `Media`: dispatch by the media-type attribute value, defined fallback precedence

Note: all the content included inside these component needs to be checked and in case sanified (e.g. url, src, ...)

Result rendered inside a Panorama image media in the configured tab

<img width="1919" height="951" alt="Image" src="https://github.com/user-attachments/assets/e046fe41-088e-42ee-bc19-bb8531c2ab88" />

## Acceptance criteria

- [ ] Per attribute, an editor user can pick a display type and see the value rendered as that media in the tab
- [ ] `Media` type dispatches by the chosen attribute; fallback path is defined and tested
- [ ] Panorama and PDF render their content
- [ ] Image / Video / Audio / URL / Iframe render correctly
- [ ] Extracted media component is documented as a reusable core component
[...]

PR #12753, opened 2026-07-30
Identify: EXTERNAL DATA view type  #12596

### LINKED ISSUE #12596: Identify: EXTERNAL DATA view type
## Description

New Identify view type `EXTERNAL DATA` that drills into a separate spatial/non-spatial WFS table using a CQL filter template built from the clicked feature's properties, then shows the returned attributes. Values render as GeoJSON here (similar to properties); media rendering is a separate issue.

Initially the `EXTERNAL DATA` will support only WFS but in the future may be extended to support other sources.

**Config panel** (expanded via the view edit button)
- WFS url input (mandatory)
- layer name input (mandatory; enabled once a valid WFS url is set, populated/validated via `DescribeFeatureType` / GetCapabilities)
- CQL filter template input (mandatory): plain text with `${properties['x']}` placeholders
- Attributes section: per-column show/hide and alias only

Configuration panel for `EXTERNAL DATA` type showing needed configuration url, layer name and CQL

Image

Configuration panel for `EXTERNAL DATA` type showing the connected WFS layer. Note this mockup does not include the validate button that needs to be included during the development

Image

**Runtime**
- the tab lists all features returned for the layer
- one CQL per feature: interpolate the template with each feature's properties (reuse the existing `${}` template engine) with CQL value escaping, fire the `GetFeature` requests in parallel, render each feature's block as its result arrives. Let's evaluate possible alternative strategies during the development (e.g.: optimization: batch into a single CQL (`(F1) OR (F2) ...`) + demultiply rows back to features; requires an explicit join-key config, out of scope here TBD)
- render returned attributes (respecting show/hide + alias) as plain values
- lazy: fire only on tab open; cache per (feature, view)
- ensure requests pass into axios and apply the MapStore proxy when needed for route external WFS to handle CORS / mixed-content

## Acceptance criteria

- [ ] Configuring url + layer + CQL template produces a working tab
- [ ] Opening the tab fetches and lists the configured attributes for every returned feature
- [ ] One CQL per feature, fired in parallel and cached per feature
- [ ] An EXTERNAL DATA view with missing mandatory fields (url / layer / CQL) is flagged invalid and not rendered as a tab until completed.
- [ ] CQL values are escaped
- [ ] A test action validates the template (syntax + placeholders) and optionally runs it against a sample feature fetched from the source layer, showing the resulting request/response.
- [ ] External WFS requests go through the proxy when needed
- [ ] Docs and i18n (en/it/fr/de/es) updated

## Other useful information

- Out of scope: display-type selector and media renderers (separate issue)

### LINKED ISSUE #12598: Identify: media rendering for PROPERTIES attributes
## Description

Give the existing `PROPERTIES` Identify view type the same attribute editor built for `EXTERNAL DATA`: clicking the view's edit button opens the full Attributes editor (per-field show/hide, alias, and the `Feature info display type` selector with its media renderers), reusing the same components. Today `PROPERTIES` renders each feature attribute as plain text (HTML strings excepted) via `components/data/identify/viewers/row/PropertiesViewer.jsx`, with only per-field `alias` and show/hide (`include`/`exclude`) and no dedicated editor UI. This issue wires the shared Attributes editor to the `PROPERTIES` view and drives the rendered values through the shared renderers.

This is purely about reusing those components on `PROPERTIES`; no new editor UI or renderers are introduced here.

**Config** (the view's edit button opens the same Attributes editor as `EXTERNAL DATA`)
- Reuse the full per-attribute Attributes editor from `EXTERNAL DATA`, with one row per attribute exposing:
  - show/hide checkbox
  - alias
  - `Feature info display type` selector (`Image`, `Video`, `Audio`, `Panorama Image`, `PDF`, `URL`, `Iframe`, `Media`)
- `Media type attribute` selector, visible only when type = `Media`: picks another attribute holding the media type, with fallback to string / guess-by-extension
- Attribute list is populated from the layer's known attributes (e.g. `DescribeFeatureType` / existing layer fields), not from a remote WFS source
- Default display type stays the current plain-text/GeoJSON rendering so unconfigured attributes are unchanged
- Alias precedence: today the attribute alias comes from the global layer fields settings (`layer.fields`, configured via `components/TOC/fragments/LayerFields/`). The per-view Attributes config inside the Identify viewer, when an alias is set, takes precedence over the global one; when left empty it falls back to the global `layer.fields` alias (and then to the raw attribute name)

**Runtime** (`components/data/identify/viewers/row/PropertiesViewer.jsx`)
- For each attribute, dispatch to the shared media renderer by its configured display type (same components used by `EXTERNAL DATA`)
- Attributes with no display type configured keep the current behavior (plain value, HTML string handling preserved)
- Respect show/hide + alias, applying the per-view alias precedence over the global `layer.fields` alias
- Content (url, src, ...) is sanitized by the shared renderers

## Acceptance criteria

- [ ] The `PROPERTIES` view exposes an edit button that opens the same Attributes editor as `EXTERNAL DATA` (show/hide, alias, display type per attribute)
- [ ] Per attribute, an editor user can pick a display type and see the value rendered as that media in the Identify tab
- [ ] Attributes without a configured display type render exactly as before (plain value + existing HTML string handling)
- [ ] `Media` type dispatches by the chosen attribute; fallback path matches the `EXTERNAL DATA` behavior
- [ ] The Attributes editor and renderer components are the same ones built for `EXTERNAL DATA` (no duplication)
- [ ] Existing saved maps with a plain `PROPERTIES`/`featureInfo` load and behave as before
- [ ] Unit tests cover: configured display type renders the right media, unconfigured attribute falls back to plain value, alias/show-hide preserved
- [ ] Docs and i18n (en/it/fr/de/es) updated

rowheat02:
## Description

This PR introduces a new Identify view type called External Data. It allows identified features to be linked to a separate WFS-enabled table through a configurable CQL filter and displays selected attributes from the related records.

https://github.com/user-attachments/assets/2b8e6d36-17ee-457e-bc37-bbeb6eeedabd

Also, added configurable attributes to Properties views (from [issue](https://github.com/geosolutions-it/MapStore2/issues/12598)) , custom media rending will be handled in #12597
image

 - [x] Feature

## Issue

#12596
#12598

**What is the new behavior?**

A new External Data  Identify view is available. It allows users to configure an external WFS layer and build a CQL filter using properties from the identified feature.
Also, for Properties views, users can configure which field to hide and change the alias of some fields.
## Breaking change
 - [x] No

## Other useful information
- External Data currently supports WFS sources only.
- CQL templates can reference identified feature properties using ${properties.name} or ${properties['name']}.
- One external request is executed per identified feature, with results limited to 10 features per request.
- Requests use the existing MapStore Axios and proxy handling.


allyoucanmap (approved):

allyoucanmap:
@ElenaGallo please test this new feature on dev, thanks

stefanocudini:
hi @allyoucanmap @rowheat02 I seen that sometimes this timeout delay introduced in this PR, causes the CI tests to block(random in node V20 or v22), it probably happens at random times depending on the load at that moment:

https://github.com/geosolutions-it/MapStore2/pull/12753/changes#diff-96d8fa449107875236178604a05f560b2078797e8ececdbbc03d4cd17dfc3879R153

The CI action logs
```text
FAILED TESTS:                                                                                                        
    ExternalDataEditor                                                                                               
      ✖ loads feature types for an initial WFS URL                                                                   
        Chrome Headless 150.0.0.0 (Linux 0.0.0)                                                                      
      Error: Expected true to be false                                                                               
          at assert (webpack://mapstore2/./node_modules/expect/lib/assert.js?:29:9)                                  
          at Expectation.toBe (webpack://mapstore2/./node_modules/expect/lib/Expectation.js?:70:28)                  
          at eval                                                                                                    
  (webpack://mapstore2/./web/client/components/TOC/fragments/settings/__tests__/ExternalDataEditor-test.jsx?:180:97) 
```

allyoucanmap:
Hi @stefanocudini , thanks for reporting, we are going to fix this in a related PR

PR #12779, opened 2026-08-07
Added a filter for framable resources.

axl8713:
## Description

Not all the resources exposed from MapStore should be framable, so this filter limits framing to same-origin requests where appropriate.

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [X] Other... Please describe: Application container deployment change (web.xml)

## Issue

#<issue>

**What is the new behavior?**

## Breaking change
 - [x] No


java/web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -76,6 +76,39 @@
         <url-pattern>*.txt</url-pattern>
     </filter-mapping>
 
+    <filter>
+        <filter-name>viewFramingFilter</filter-name>
+        <filter-class>org.apache.catalina.filters.HttpHeaderSecurityFilter</filter-class>
```
offtherailz, line 81: This is a tomcat only filter. "org.apache.catalina.filters." 
Not sure it should be used ... We are using only tomcat right now. So not sure if we can skip, maybe is not blocking

java/web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -76,6 +76,39 @@
...
+        <dispatcher>REQUEST</dispatcher>
+        <dispatcher>FORWARD</dispatcher>
+    </filter-mapping>
+    <filter-mapping>
+        <filter-name>viewFramingFilter</filter-name>
+        <url-pattern>/index.html</url-pattern>
+    </filter-mapping>
+    <filter-mapping>
+        <filter-name>viewFramingFilter</filter-name>
+        <url-pattern>/configs/*</url-pattern>
```
offtherailz: What does this really covers? JSON are not "embeddable". I think it will never be applied so it doesn't protect from any issues. I think this was a false interception of something, by AI

java/web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -76,6 +76,39 @@
...
+            <param-value>false</param-value>
+        </init-param>
+    </filter>
+    <filter-mapping>
+        <filter-name>viewFramingFilter</filter-name>
+        <!-- FORWARD is required: / is resolved to index.html through an internal forward -->
+        <url-pattern>/</url-pattern>
+        <dispatcher>REQUEST</dispatcher>
+        <dispatcher>FORWARD</dispatcher>
+    </filter-mapping>
```
offtherailz: This mapping never matches: the filter is applied to / by the /index.html mapping below, not by this one, and the FORWARD dispatcher is not involved either.

offtherailz (commented):
Apply the same suggested in the main web.xml to the template one of course.

java/web/src/main/webapp/WEB-INF/web.xml
axl8713: To confirm, MapStore is already tied to Tomcat as the reference deployment target:
- **docs/developer-guide/requirements.md**: Tomcat 10.1 declared as the reference container (Jakarta EE 10 upgrade)
- **Dockerfile**: official image built FROM tomcat:10.1-jdk17
- **binary/pom.xml**: runtime dependencies on tomcat-catalina, tomcat-jasper, tomcat-jdbc
- **binary/bin-war/pom.xml**: war packaging for the official binary distribution

Using the built-in HttpHeaderSecurityFilter doesn't add a new constraint on top of that.

java/web/src/main/webapp/WEB-INF/web.xml
axl8713: This mapping was added because configs/localConfig.json was the specific endpoint named in the original report. If your knowledge confirms that `/configs/*` only ever serves JSON (no HTML/interactive content), I am fine dropping this mapping.

java/web/src/main/webapp/WEB-INF/web.xml
axl8713: You're right. Good catch! Removed that filter on the root.

offtherailz (dismissed):

offtherailz:
Sorry @axl8713 forgot to mention that this change need to be added also to migration guidelines. Could you please add it?

offtherailz (approved):

offtherailz:
It works, denying the embedding for all the application except embedded pages.

PR #12787, opened 2026-08-11
#12774 Title and Name should be available in General Tab for all layer types that support it

Binabh:
## Description

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Feature

## Issue

**What is the current behavior?**

#12774 

**What is the new behavior?**
Name is editable for `wfs`,`arcgis` and `arcgis-feature`

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [ ] Yes, and I documented them in migration notes
 - [x] No

PR #12802, opened 2026-08-14
#12571 ,  #12573 Support of distinct search-typename in WMS layers and way to edit urls of layers 

### LINKED ISSUE #12571: A way to edit urls of layers
## Description
The only way currently to edit a layer of an url is to add it anew. 
This makes adapting to changing servers and hostnames a very tedious task.

Worse still is the lack of control over the search url. If no csw service exist there is no way to set the search url.

## Acceptance criteria
- a way to edit the url or a layer

## Other useful information
heres a mockup
Image

### LINKED ISSUE #12573: support of distinct search-typename in WMS layers
## Description
The search object is a reference to a corresponding WFS-service on a WMS layer. 
It is used in various places like filters and attribute table.
At the moment the typename is assumed to be identical to the layer-name of the WMS.
this assumption is not always correct and so for some layers, having a corresponding wfs is impossible.
We lack a seperate typename in the searchObject.

Example
```json
{
    "name": "pdk:theme_pdk_lokalplandelomraade_vedtaget",
    "type": "wms",
    "url": "https://geoserver1.plandata.dk/geoserver/wms",
    "search": {
        "url": "https://geoserver2.plandata.dk/geoserver/wfs",
        "type": "wfs",
        "typeName": "pdk:lokalplandelomraade"
    },
    ...
}
```
## Acceptance criteria
- separate typename in layer.search.typeName
- used in every place the layers wfs is used
- gui to edit the url and typename

mockup

Image

Binabh:
## Description
This PR covers two functionalities of letting user change layer URL and WFS linked service for WMS Layers

 - [x] Feature

## Issue

#12571 #12573 

**What is the new behavior?**
User is able to change URLs and WFS linked service and it properly reflects on network calls.

## Breaking change
 - [x] No


offtherailz (changes requested):
Hi, some feedback from the UI point of view:

- [x] Validation is missing on the layer URL and on the WFS URL/typeName fields (currently only the       layer name is validated). The layer URL should go through the same validation as the layer name, and the WFS fields as well — the typeName could be validated with a DescribeFeatureType call, which would also give us the field refresh (see below). Note that with no validation an empty value is accepted: since `getSearchUrl` falls back to the layer URL, an empty WFS URL silently sends WFS requests to the WMS endpoint. GeoServer answers those on `/wms` too, so nothing looks broken, but it breaks as soon as the WFS lives on a different endpoint or host.
- [x] Fields stay saved once they have been downloaded, even if the layer changed completely: they are not reloaded when the typeName (or the layer name) changes, and the stale list is persisted on the layer. This is a pre-existing behaviour that may affect bbox and other items too, so it may be worth a separate issue. On reload I would keep the merge as it is: fields with the same name keep their customizations, fields that no longer exist disappear, and the dedicated button is there for whoever wants to drop the customizations — otherwise renaming a layer or changing an attribute means rewriting everything.
- [x] When you turn the WFS panel off and on again, the typeName is recreated with the layer name while the URL is left empty. Both are actually rewritten from defaults, so a custom typeName is silently replaced by the layer name. Could we use the same DescribeLayer request to retrieve the WFS URL (and typeName) when re-enabling the panel? Please note that the current merge keeps the existing `search` on top of the detected one, so it has to be reversed for an explicit user action, and we should handle servers that do not support DescribeLayer by simply leaving the field empty.

Binabh:
@offtherailz Thank you for your feedback. As Per your suggestion I have added validation for url, searchUrl and typename and have also appropriately populated values from DescribeLayer. I have also added tests and also verified manually. Let me know if any further improvements are needed.
Thank You!

offtherailz (changes requested):
### WMS 

- [x] I noticed this regression of layer name for WFS layers

https://github.com/user-attachments/assets/955d1e62-b4be-482c-997b-a0080f6c4ad7

WFS link service editing looks to work fine

### WFS

- [x] Type name is not editable, only URL... this is a little inconsistent. IT was indicated in #12571  [comment](https://github.com/geosolutions-it/MapStore2/issues/12571#issuecomment-4926127418):

> Same capability for native WFS layers (type: wfs): editing layer.url and typeName on the WFS layer's General tab

Image

Binabh:
@offtherailz Thank You for your review. The issue of `Type Name not editable` is addressed by another PR(already merged) and also available here as this has already synced with `master`. Instead of Type Name Another PR uses `Name` as other layer types also use that.

 Also, the regression when editing WMS name is now solved. (video attached below)

https://github.com/user-attachments/assets/22a71596-871a-4665-a024-3f75c4716c21



I have re-tested that other regression and feed backs reported on https://github.com/geosolutions-it/MapStore2/pull/12787 are not re-introduced while solving merge conflict.
Please let me know if any further changes are required. Thank You!

offtherailz (approved):
No time to test for me, but code changes look good. I'd suggest to widly test this features on dev.

tdipisa:
@offtherailz thanks. I'm going to merge, @ElenaGallo will verify in DEV shortly.

PR #12850, opened 2026-09-02
#12849 Unit Conversion is not properly happening in case of measuring area in 3D view 

Binabh:
## Description
Fix unit conversion issue in area measurement in 3D map and add relavent tests

**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/geosolutions-it/MapStore2/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x", remove the others)
 - [x] Bugfix

## Issue

**What is the current behavior?**
#12849 

**What is the new behavior?**
Value is being properly converted when converting units

## Breaking change
**Does this PR introduce a breaking change?** (check one with "x", remove the other)
 - [x] No

PR #12859, opened 2026-09-03
[Backport 2026.02.xx] #12849 Unit Conversion is not properly happening in case of measuring area in 3D view 

### LINKED ISSUE #12849: Unit Conversion is not properly happening in case of measuring area in 3D view
## Description
Issue reported when changing UoM after creating an Area measurement in 3D. If the unit of measurement is changed, the value remains the same and the conversion to the new unit of measurement is not performed. For example, if 118.47 m2 is changed to ft2, the area label will read 118.47 ft2, which is an error.

## How to reproduce
- Create a new map
- Switch to 3D view
- Use measure tool to measure area
- Switch UoM and see that only unit changes but not the value

https://github.com/user-attachments/assets/6e109afb-8055-4c46-822e-31d0d68245d4

*Expected Result*
The area value is properly converted

*Current Result*
Area value is not properly converted just unit is changed

- [x] Not browser related

offtherailz:
# Description
Backport of #12850 to `2026.02.xx`.

Fixes #12849


tdipisa (approved):

tdipisa:
@ElenaGallo this will be soon ready to be tested on QA.

